Issue #167, opened 2018-02-14
Parameters of the lambda expressions inside TriangleMesh::GetMinBound are not const references


csukuangfj:
See the code
https://github.com/IntelVCL/Open3D/blob/5382f39abf7ad625fb6bb80b247332703665a87e/src/Core/Geometry/TriangleMesh.cpp#L49-L61

The parameter types should be `const Eigen::Vector3d&` .

PR #177, opened 2018-02-14
Resolves #176.

### LINKED ISSUE #176: Misplaced const for LineSet operator+
See the code
https://github.com/IntelVCL/Open3D/blob/84d5681c3996958320bab41478a28431b2f7dc44/src/Core/Geometry/LineSet.h#L56

It should be

```.cpp
virtual LineSet operator+(const LineSet &lineset) const;
```

In addition, I find that `LineSet` is not inherited in the project and I don't see the point
of making the following functions `virtual`
https://github.com/IntelVCL/Open3D/blob/84d5681c3996958320bab41478a28431b2f7dc44/src/Core/Geometry/LineSet.h#L55-L56

csukuangfj:
Resolves #176 and similar problems like #167.

### DIFF
--- a/src/Core/Geometry/LineSet.h
+++ b/src/Core/Geometry/LineSet.h
@@ -52,8 +52,8 @@ public:
 	void Transform(const Eigen::Matrix4d &transformation) override;
 
 public:
-	virtual LineSet &operator+=(const LineSet &lineset);
-	virtual const LineSet operator+(const LineSet &lineset);
+	LineSet &operator+=(const LineSet &lineset);
+	LineSet operator+(const LineSet &lineset) const;
 
 public:
 	bool HasPoints() const {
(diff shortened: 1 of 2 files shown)


src/Core/Geometry/LineSet.h
```diff
@@ -52,8 +52,8 @@ class LineSet : public Geometry3D
 	void Transform(const Eigen::Matrix4d &transformation) override;
 
 public:
-	virtual LineSet &operator+=(const LineSet &lineset);
-	virtual const LineSet operator+(const LineSet &lineset);
+	LineSet &operator+=(const LineSet &lineset);
+	const LineSet operator+(const LineSet &lineset) const;
```
qianyizh: Should be ``LineSet operator+(const LineSet &lineset) const;``

csukuangfj: @qianyizh 
yes, you are right.

PR #184, opened 2018-02-14
Minor improvement.

csukuangfj:
Minor improvement.


src/Core/Geometry/RGBDImage.cpp
```diff
@@ -57,7 +57,7 @@ RGBImagePyramid CreateRGBDImagePyramid(const RGBDImage& rgbd_image,
 			num_of_levels, with_gaussian_filter_for_depth);
 	RGBImagePyramid rgbd_image_pyramid;
 	rgbd_image_pyramid.clear();
-	for (int level = 0; level < num_of_levels; level++) {
+	for (size_t level = 0; level < num_of_levels; level++) {
```
syncle: Good point. I think this lines
https://github.com/IntelVCL/Open3D/blob/9411b84e92d0994985632d564fbd4fac3e4af10a/src/Core/Odometry/Odometry.cpp#L450

need to be 
```
size_t num_levels = iter_counts.size();
```

src/Core/Geometry/TriangleMeshFactory.cpp
```diff
@@ -49,10 +49,11 @@ std::shared_ptr<TriangleMesh> CreateMeshSphere(double radius/* = 1.0*/,
 	mesh_ptr->vertices_[1] = Eigen::Vector3d(0.0, 0.0, -radius);
 	double step = M_PI / (double)resolution;
 	for (int i = 1; i < resolution; i++) {
+	    double alpha = step * i;
```
syncle: Inconsistent indent.

src/IO/ClassIO/ImageIO.h
```diff
@@ -33,14 +33,14 @@ namespace three {
...
-/// \return If the read function is successful.
+/// \return True if the read function is successful, false otherwise.
 bool ReadImage(const std::string &filename, Image &image);
 
 /// The general entrance for writing an Image to a file
 /// The function calls write functions based on the extension name of filename.
 /// If the write function supports quality, the parameter will be used.
 /// Otherwise it will be ignored.
-/// \return If the write function is successful.
+/// \return True if the write function is successful, false otherwise.
```
syncle: `return true if the write function is successful, false otherwise. `. The same thing applies for the other edits.

src/IO/FileFormat/FileJPG.cpp
```diff
@@ -39,7 +39,7 @@ bool ReadImageFromJPG(const std::string &filename, Image &image)
 	JSAMPARRAY buffer;
 
 	if ((file_in = fopen(filename.c_str(), "rb")) == NULL) {
-		PrintWarning("Read JPG failed: unable to open file.\n");
+		PrintWarning("Read JPG failed: unable to open file: '%s'.\n", filename.c_str());
```
syncle: do we need `'%s'` instead of `%s`?
and lets remove period as it will display like `Read JPG failed: unable to open file: image.name.`

src/Core/Geometry/RGBDImage.cpp
csukuangfj: @syncle thank you for the review.
I think this line is fine, since it is converted to `int` explicitly.

Changing it to `size_t` requires changing the following line at the same time.
https://github.com/IntelVCL/Open3D/blob/9411b84e92d0994985632d564fbd4fac3e4af10a/src/Core/Odometry/Odometry.cpp#L466

But if `level` in the above line is changed to `size_t`, then it is always larger than or
equal to 0, since it is unsigned. Then the for loop never ends.

src/Core/Geometry/TriangleMeshFactory.cpp
csukuangfj: Fixed.

I prefer using **spaces** instead of **tabs**, though Open3D recommends using **tabs**.

src/IO/ClassIO/ImageIO.h
csukuangfj: fixed.

src/IO/FileFormat/FileJPG.cpp
csukuangfj: quotes and dots are removed.

PR #236, opened 2018-03-19
expose lambda_geometric in registration_colored_icp

### LINKED ISSUE #235: Photo-metric weight in Colored ICP
Hi Authors,

Now it seems that Photo-metric weight is a fixed one (with geometric weighting 0.968) in `registration_colored_icp()`, would you be possible to open an interface in this function to tuning the weight? 

I wonder where is this weighting ratio coming from and why is it this number?

Thanks!

syncle:
This fixes #235.

### DIFF
--- a/src/Core/Registration/ColoredICP.cpp
+++ b/src/Core/Registration/ColoredICP.cpp
@@ -39,9 +39,6 @@ namespace three{
 
 namespace {
 
-const double default_lambda_geometric = 0.968;
-const int max_neighbors_for_gradient_approximation = 30;
-
 class PointCloudForColoredICP : public PointCloud {
 public:
 	std::vector<Eigen::Vector3d> color_gradient_;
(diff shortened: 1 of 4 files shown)


src/Core/Registration/ColoredICP.cpp
```diff
@@ -233,13 +230,13 @@ double TransformationEstimationForColoredICP::ComputeRMSE(
 RegistrationResult RegistrationColoredICP(const PointCloud &source,
 		const PointCloud &target, double max_distance,
 		const Eigen::Matrix4d &init/* = Eigen::Matrix4d::Identity()*/,
-		const ICPConvergenceCriteria &criteria/* = ICPConvergenceCriteria()*/)
+		const ICPConvergenceCriteria &criteria/* = ICPConvergenceCriteria()*/,
+		double lambda_geometric/* = 0.968*/)
```
syncle: I am not quite sure whether this extension is acceptable. I put ``lambda_geometric`` as the last entry of the function call in order to make users don't have to care about this parameter.

PR #261, opened 2018-03-28
.gitignore updates

takanokage:
Added more entries for:
- Linux/Windows/MacOS
- C/C++, Python, CMake
- Visual Studio/Visual Studio Code/etc.

Based on https://github.com/github/gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,4 +14,412 @@ src/Sandbox/
 src/Visualization/Shader/Shader.h
 _build/
 doxygen/
+
+################################################################################################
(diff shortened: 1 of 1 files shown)


syncle:
I am not sure whether this is really necessary. We recommend to make a **build** folder for users. As git simply ignores the build folder, the job is done. I think this is clear way instead of specifying all the possible file formats for all the platforms.

syncle:
After the discussion, we decide to close as it is not necessary for now. It is also important to keep .gitignore simple.

Issue #278, opened 2018-04-06
include guards


takanokage:
I noticed we're using the non-standard `pragma once` in place of include guards.
I propose we replace `pragma once` with the following or similar in order to avoid unforeseen bugs later on:

```
// for Core.h
#ifndef OPEN3D_CORE_H
#define OPEN3D_CORE_H
...
#endif /* OPEN3D_CORE_H */
```

While `pragma once` is supported on most compilers there are corner cases where its behavior is  undefined aka compiler dependent.

takanokage:
Any chance we could discuss removing the other exceptions to the Google coding style? I'd be willing to make the fix.

takanokage:
I wrote a script in order to update all the include guards.
Found a couple of headers with no guards at all.
I will initiate a PR momentarily. Please take a look.
 I won't include the script in the PR but you can find it in the history of commits.

takanokage:
Well, I've made my case and offered the implementation.

PR #283, opened 2018-04-07
use standard include guards

takanokage:
Issue #278
Replace `#pragma once` with the standard `#ifndef ... ` include guards.
For this purpose I wrote a script that did the modifications.
Found a couple of headers with no include guards at all.


syncle:
Let's finish the discussion in #278, and decide to merge.

Issue #277, opened 2018-04-05
install py3d from build


takanokage:
This is about installing py3d module when building from sources.
This is different from #217.

Currently the Open3D based python code has to start with something like this:

```
import sys, os
sys.path.append("../..")
from py3d import *
```
where the location to py3d is explicitly specified.

Instead we could simply use the following:

```
import py3d
```
if the py3d module gets installed.

takanokage:
True for the case you're describing.
However I'm talking about what we can do when building from sources. I believe I have a solution, will initiate a PR hopefully later today.

takanokage:
I have tested my solution for this on Ubuntu and Mac and it works. I can import py3d from anywhere on the disk.
I'll hold onto the PR until I figure out how to do the same on Windows.

syncle:
Addressed in #286. Thanks @takanokage!

takanokage:
thank you!

PR #286, opened 2018-04-09
Install the python module from sources

takanokage:
#277

Created installer from sources. (make install after building the sources)
Added scripts for install/uninstall.
Updated the getting started documentation regarding the import of the python module.
Removed `sys.path.append("../..")` from the python examples.


syncle:
Oh wait, based on the discussion we had, I thought `make install` does not work on windows. However, this PR may break running python tutorials on windows. Is that true?

src/Visualization/CMakeLists.txt
```diff
@@ -53,10 +53,10 @@ else ()
...
 endif ()
 
-install(TARGETS Visualization
-	RUNTIME DESTINATION ${CMAKE_INSTALL_PREFIX}/bin
-	LIBRARY DESTINATION ${CMAKE_INSTALL_PREFIX}/lib
-	ARCHIVE DESTINATION ${CMAKE_INSTALL_PREFIX}/lib
-)
-file(GLOB headers "${CMAKE_CURRENT_SOURCE_DIR}/*.h")
-install(FILES ${headers} DESTINATION ${CMAKE_INSTALL_PREFIX}/include)
+# install(TARGETS Visualization
```
syncle: please remove commented lines. Applied to other cases too.

docs/getting_started.rst
```diff
@@ -71,15 +71,27 @@ Running Open3D tutorials
...
+    # uninstall the Python module
+    ./uninstall.sh
+
+For quick sanity check, try importing the library from the python interactive shell:
+
+.. code-block:: sh
+
+    python
+
+    >>> import py3d
```
syncle: Please add more description for windows. Please check my comments in this PR.

util/scripts/install.sh
```diff
@@ -0,0 +1,9 @@
+echo
+echo installing...
```
syncle: Can you echo a path where the libraries are installed?

util/scripts/uninstall.sh
```diff
@@ -0,0 +1,8 @@
+echo
+echo uninstalling...
```
syncle: Can you echo the uninstallation path?

takanokage:
ok, so it works on Windows as well? I didn't try it because my Windows environment is still not working after half a day of attempts. I'm pretty mad right now at how much trouble this is behind the corp network/proxies.

util/scripts/install.sh
takanokage: that is echoed a the end of the install:

```
Install the project...
-- Install configuration: "Release"
-- Installing: /home/dpetre/.local/lib/python3.5/site-packages/py3d.cpython-35m-x86_64-linux-gnu.so
```

util/scripts/uninstall.sh
takanokage: I just made a commit about it.

src/Visualization/CMakeLists.txt
takanokage: fine with me. that was the "install" script for the c/c++ side of things. it's not working so I decided to hide it entirely. removing the code it's also fine.

takanokage: DONE

docs/getting_started.rst
takanokage: OK, I see, it works by default on Windows too. Awesome :).

takanokage: DONE

util/scripts/uninstall.sh
takanokage: DONE

syncle:
and please remove 
```
import sys
sys.path.append("../..")
```
in documentations and tutorial scripts if necessary. 
Some scripts will not require `sys`.

takanokage:
I don't know if it overwrites or not. I see the install process detects the existing installation and says:
`-- Up-to-date: /home/dpetre/.local/lib/python3.5/site-packages/py3d.cpython-35m-x86_64-linux-gnu.so`.

PR #284, opened 2018-04-07
Docker for Open3D

takanokage:
Docker solution for Open3D.

PR #304, opened 2018-04-12
Revert "Docker for Open3D"

syncle:
Found some issue this with #284. Revert for now.


syncle:
Yes. There are couple of things that can be improved.

- In documentation, optional steps can be moved back of the document. It breaks the flow of the tutorial document. 
- `Add user to “docker” group` can be part of docker installation.
- `Usage notes` should start with `VNC` and `Running in terminal`. The explanation about the `tools` can be moved in the next place. This is more user friendly and for the better document flow.
- In VNC, it is hard to follow the instruction. Add some example script that runs out of the box. It is ok to take an example with famous and free VNC viewer and include corresponding screen shots.
- In running in the terminal, the path are not correct. For example, not `utilities` but `util`. 
- In my case, the shell `./docker_sample.sh` are not copied to `Advanced` folder. 
- I think the callee of `./docker_sample.sh` does not have to be `headless_rendering.py` as this docker solution is known to have a issue with headless rendering.

This is a great contribution, but I would like to make it more matured. This PR can be considered again once there is a request from users for this.

PR #315, opened 2018-04-16
Tab indent -> 4 space indent

### LINKED ISSUE #308: Tab indent vs space indent
Open3D uses tab for the indent. 

From http://open3d.org/docs/contribute.html. 
> Use tab for indent. In IDE it should be 4 spaces wide. Use two indents for a forced line break (usually due to the 80 character length limit).

We got several requests regarding the space based indent (4 spaces) instead. Which one would be better? This issue have been discussed within Open3D admins, but we would like to listen public opinions. We are open for the either suggestions.

syncle:
This resolves #308. Only tabs in source code (exclude external folder) is modified.


takanokage:
Looks good to me.

Issue #166, opened 2018-02-14
Request to use some test framework.


csukuangfj:
The current project is not extensively tested.

Some kind of framework, say gtest, may be introduced into the project.

qianyizh:
Agree.

I thought about adding gtest for a long time. But I realized it is non-trivial, especially it needs to be integrated with CI and tested under three different compiler chains.

It would be really great if you can contribute. I am happy to adopt it.

qianyizh:
A rough plan in resolving this issue:

- [ ] Setup test framework for Python code
    - [ ] Choose one of the testing packages (unittest? pytest?)
    - [ ] Write a couple of unit tests in ``src/Python/Unittest``. The folder should have the same folder structure as ``src/Python`` and files with ``filename_test.py``, e.g., ``src/Python/Unittest/Core/py3d_camera_test.py`` is the test file for ``src/Python/Core/py3d_camera.cpp``
    - [ ] Write CMake file and link the unittest with Travis CI. There should be a CMake flag to turn on/off unittest (default is ON).
    - [ ] Pull request 1
- [ ] Setup test framework for C++ code
    - [ ] Choose one of the testing libraries (gtest?)
    - [ ] Write a couple of unit tests in ``src/Unittest``. The folder should have the same folder structure as ``src`` (except ``src/Python``) and files are named ``filename_test.cpp``
    - [ ] Write CMake file and link the unittest with Travis CI
    - [ ] Pull Request 2
- [ ] Write tutorials for unit test, update documents for developers, PR 3

takanokage:
I'm starting work on this. I'll continue to provide assistance on #130 and #298 as needed.

takanokage:
I have implemented a proof of concept for using googletest as our test framework.
For now I have basic unit test proof of concept for the Geometry/Image class.

Next I'm going to write some documentation. Any preferences regarding where we should place the unit testing in the doc tree?

syncle:
> For now I have basic unit test proof of concept for the Geometry/Image class.

That's amazing. You got to manage gtest. Thanks! 

I would like to suggest making documents later. First, this will reduce your work. [Check out this example](https://github.com/pybind/pybind11/tree/master/tests). They just provide readable codes and scripts without documentation. Second, test applications and functions will be the most frequently changing part. It may not easy to maintain documentations in that frequency as well. :)

takanokage:
Ok, this is what I'm going to do:
- polish the code just a tiny bit.
[...]

Issue #317, opened 2018-04-16
AppVeyor failure


syncle:
AppVeyor configuration in the current master branch keeps failing.
The error comes from some `pip` commands:
https://github.com/IntelVCL/Open3D/blob/5bb9fb0bbed6b5dcc243476cf7535caf01c0f304/.appveyor.yml#L35-L36

These lines can be omitted.

Sahloul:
@syncle
> I would like to mention couple of things for you. We have built the library for two years from the scratch. During this procedure, we had to set strict principle and rules to make the library to be light-weight but powerful. Therefore, **we regard contribution page very important. One of the core principle is the consensus. If there is an issue, we submit it on the issue board, after the discussion and decide to work for it, then switch to work and submit pull request.**

It is your word against you!

syncle:
You're right. It was my bad. I didn't expect this will break your PR. I personally thought this is rather independent issue. However, as your spending priceless time, I had to be more careful and shouldn't merge before I got your comment. My apologies.

PR #323, opened 2018-04-17
Unit Testing

takanokage:
#166 

This is a skeleton implementation of how Google Test can be used to implement unit testing.
A proof of concept is provided through `src/UnitTest/Image_unittest.cpp`. This example is not meant to fully test the `Core\Geometry\Image` class, only to show how it's done.

Brief documentation is provided on the `Compilation Options` page.

Please take a look and let me know if you have improvement suggestions.


takanokage:
There's something wrong with the master branch and/or the AppVeyor script.
By default the unit tests are disabled.

syncle:
Yes. The cure can be found from #317. I halted merging yesterday but you can make a PR again.

takanokage:
Ok, no problem.

takanokage:
This can be a squash merge once we're ready.

src/UnitTest/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
...
+cmake_minimum_required(VERSION 3.0)
+
+# Locate GTest
+find_package(GTest REQUIRED)
+include_directories(${GTEST_INCLUDE_DIRS})
+
+# Specify the unit tests
+add_executable(unitTests
+               main.cpp
+               Image_unittest.cpp)
```
syncle: I think `image.cpp` would be fine, as it is under UnitTest folder.

syncle:
Can you open PR for AppVeyor issue too? Thanks!

src/UnitTest/CMakeLists.txt
takanokage: yeah, we could do that.

takanokage:
Actually I haven't worked on the AppVeyor at all so far. I don't know what that issue is about.
Expect a ramp up period (and subsequently no progress on other things) if you really want me to help with that.

takanokage:
I just added skeleton unit test code for most of the features in the three libraries: Core, IO & Visualization.
By default each case displays a "Not implemented." message in red. I could use the FAIL() macro to actually cause a fail by default in each test case however the output of the unit test becomes quite inelegant (output becomes very verbose which seems overkill for the purpose of just signaling that the test case hasn't been implemented yet.)

The output will look like this:

```
[----------] 1 test from PinholeCameraTrajectoryIO
[ RUN      ] PinholeCameraTrajectoryIO.Default
[          ] Not implemented.
[       OK ] PinholeCameraTrajectoryIO.Default (0 ms)
[----------] 1 test from PinholeCameraTrajectoryIO (0 ms total)
```

There might be cases where it doesn't make sense to ever add tests. If that is the case we can just delete the offending unit test file.

Oh, and by the way I maintained the same directory structure as was found under Core, IO and Visualization.

takanokage:
I've also been investigating how to get the unit tests running on Windows.
So far I figured out the google test installation part. However `find_package(GTest REQUIRED)` fails on Windows and I'm not sure there's anything to do about it. If we really want unit tests on Windows we might have to provide a manual path for including and linking against the gtest.

syncle:
@takanokage: checkout this https://github.com/iat-cener/tonatiuh/wiki/Installing-Google-Test-For-Windows

takanokage:
Yeah, I saw that link. As it happens that didn't help. I've gone past the gtest installation, it's the cmake automation for finding gtest libs and headers that fails.

takanokage:
Question: do we want to add Windows support as a separate PR or as part of this PR?

takanokage:
I propose we make it a separate PR because the Windows side has some issues that will take time to address. We should also be fine with the current solution which is supported on both macOS and Linux.

syncle:
Can you test with CI with unit test flag on? I can merge this if it passes build. @qianyizh can you briefly review this PR too?

takanokage:
Great, we agree on all points (including `Win32`).
I'm already on it.

takanokage:
gtest required in order to build the unit tests.
I'm working on a script to install that as part of the CI tests.

takanokage:
Let's disconnect this pull request. The new requirement to test unit test building in Travis CI has complicated things.

PR #339, opened 2018-05-04
Color map optimization #230

### LINKED ISSUE #230: Color map optimization
We need an implementation of my SIGGRAPH 2014 paper:
Color Map Optimization for 3D Reconstruction with Consumer Depth Cameras
It is actually very handy in producing high quality textures for 3D models, and it is not very complicated.

syncle:
This resolves #230 

Example:
initial_zoomnon_rigid_zoom


src/Core/Geometry/Image.cpp
```diff
@@ -61,6 +61,13 @@ Eigen::Vector2d Image::GetMaxBound() const
     return Eigen::Vector2d(width_, height_);
 }
 
+bool Image::TestImageBoundary(double u, double v,
+        double inner_margin = 0.0) const
+{
+    return (u >= inner_margin && u < width_ - inner_margin &&
+        v >= inner_margin && v < height_ - inner_margin);
+}
+
```
syncle: I add this as it is frequently used in this PR. However, this function also can be in Geometry2D.h. Not quite sure this is the right place.

src/Core/Utility/Eigen.cpp
```diff
@@ -33,20 +33,25 @@ namespace three{
...
-    if (fabs(det) < 1e-6 || std::isnan(det) || std::isinf(det))
-        solution_exist = false;
-
-    if (solution_exist) {
-        // Robust Cholesky decomposition of a matrix with pivoting.
-        Eigen::MatrixXd x = A.ldlt().solve(b);
-        return std::make_tuple(solution_exist, std::move(x));
+    if (check_det) {
+        bool solution_exist = true;
+        double det = A.determinant();
```
syncle: This is quite bad idea for large scale matrix. I add ``check_det`` option here.

src/Core/Utility/Eigen.h
```diff
@@ -53,7 +53,7 @@ Eigen::Vector6d TransformMatrix4dToVector6d(const Eigen::Matrix4d &input);
 
 /// Function to solve Ax=b
 std::tuple<bool, Eigen::VectorXd> SolveLinearSystem(
-    const Eigen::MatrixXd &A, const Eigen::VectorXd &b);
+    const Eigen::MatrixXd &A, const Eigen::VectorXd &b, bool check_det = true);
```
syncle: it checkes determinant by default. Hence, it should not affect original pipeline.

src/Python/Core/py3d_registration.cpp
```diff
@@ -326,4 +364,9 @@ void pybind_registration_methods(py::module &m)
             "Function for computing information matrix from RegistrationResult",
             "source"_a, "target"_a, "max_correspondence_distance"_a,
             "transformation_result"_a);
+    m.def("texture_map_optimization",
+			&TextureMapOptimization,
+			"Function for optimizing texture mapping for reconstructed scenes",
+			"mesh"_a, "imgs_rgbd"_a, "camera"_a,
+            "option"_a = TextureMapOptmizationOption());
```
syncle: I put texture_map_optimization in registration module. Not quite sure this is the right place.

syncle:
I found a minor bug on this PR. Working on this now.

src/Python/Core/py3d_registration.cpp
```diff
@@ -35,6 +35,12 @@
 #include <Core/Registration/FastGlobalRegistration.h>
 #include <Core/Registration/ColoredICP.h>
 
+// #include <Core/Geometry/Image.h>
+#include <Core/Geometry/RGBDImage.h>
+#include <Core/Geometry/TriangleMesh.h>
+#include <Core/Registration/ColorMapOptimization.h>
+#include <Core/Camera/PinholeCameraTrajectory.h>
```
syncle: Can we place ColorMapOptimization under py3d_registration? Can add these headers? Maybe this would not match the category.

src/Python/Tutorial/Advanced/color_map_optimization.py
```diff
@@ -0,0 +1,58 @@
...
+# The MIT License (MIT)
+# See license file or visit www.open3d.org for details
+
+from py3d import *
+from trajectory_io import *
+import os, sys
+sys.path.append("../Utility")
+from common import *
+
+path = "[set_this_path_to_fountain_dataset]"
```
syncle: I didn't include dataset for ColorMapOptimization. Instead, I had added google drive link to download the dataset. The file size for this tutorial is about 30MB (including mesh, 33 color and depth image pairs), which is not need to be maintained in Github repo.

qianyizh:
Wow, this is awesome! Let me review it during the weekend.
I have been dreaming to have it in Open3D for years!!

qianyizh:
I have some non-trivial comments on this. So the review process is slow.
I will try to finish it asap.

src/Core/Registration/ColorMapOptimization.cpp
```diff
@@ -0,0 +1,683 @@
...
+    float v = (Vt(1) * f.second) / Vt(2) + p.second;
+    float z = Vt(2);
+    return std::make_tuple(u, v, z);
+}
+
+std::tuple<std::vector<std::vector<int>>, std::vector<std::vector<int>>>
+        MakeVertexAndImageVisibility(const TriangleMesh& mesh,
+        const std::vector<RGBDImage>& images_rgbd,
+        const PinholeCameraTrajectory& camera,
+        const ColorMapOptmizationOption& option)
```
qianyizh: I think back when I made visibility test, I mask all vertices around depth discontinuity as invisible. This is because at depth discontinuity, the depth image becomes noisy, the measurement is not accurate in terms of both depth and x-y value. The projection also gets unstable. This happens frequently when you do a 360 scan of an object. So I chose to mask visible vertices only when:
1.  projected depth is close to the depth in depth image
2. the point on depth image should have a normal close to the viewing ray

Can you try a sequence which is a 360 scan and see if you see any boundary ghosting? If so, the pixels around depth discontinuity need to be eroded.

qianyizh: This also applies when averaging pixel values. Only points with a good normal orientation (not too far away from the viewing ray) gets averaged.

qianyizh:
@syncle The build is broken. Can you check?

syncle:
Got it. I am working on this now. This PR is also needed to be tested on the suggested case.

src/Core/Geometry/Image.cpp
```diff
@@ -296,4 +303,37 @@ std::shared_ptr<Image> FlipImage(const Image &input)
...
+                        if (*pi == 255) {
+                            *PointerAt<unsigned char>(*output, x, y, 0) = 255;
+                            xx = half_kernel_size;
+                            yy = half_kernel_size;
+                        }
+                    }
+                }
+            }
+        }
+    }
```
syncle: I am not quite like this function. All the filtering functions in Open3D are implemented using 1D domain to save computational burden, but for this function, I had to use 4 nested for-loops. Maybe there would be better implementation for this.

src/Python/Core/open3d_colormap.cpp
```diff
@@ -0,0 +1,80 @@
...
+}
+
+void pybind_colormap_optimization_methods(py::module &m)
+{
+    m.def("color_map_optimization",
+            &ColorMapOptimization,
+            "Function for color mapping of reconstructed scenes via optimization",
+            "mesh"_a, "imgs_rgbd"_a, "camera"_a,
+            "option"_a = ColorMapOptmizationOption());
+}
```
syncle: I am not sure color map is good name for this module. We already uses color map name for vertex/mesh coloring. However, I had to use this name again because texture_map_optimization may lead users to think some 2D texture map representations.

src/Python/Tutorial/Misc/sampling.py
```diff
@@ -0,0 +1,57 @@
...
+                    pose_graph_global.nodes[fragment_id].pose,
+                    pose_graph_fragments[fragment_id].nodes[local_frame_id].pose)))
+            copyfile(depth_image_path[i], out_path + "depth/" + \
+                    os.path.basename(depth_image_path[i]))
+            copyfile(color_image_path[i], out_path + "image/" + \
+                    os.path.basename(color_image_path[i]))
+            cnt += 1
+    copyfile(path + "/scene/cropped.ply",
+            out_path + "/scene/integrated.ply")
+    write_trajectory(traj, out_path + "scene/key.log")
```
syncle: This python script just samples every ``sampling_rate`` frames from the RGBD sequence. Actually it would be good this script can select only sharp frames.

syncle:
Few changes:
- Implement a function to generate depth discontinuity mask
- Image dilation function in Image class
- Some misc scripts that can be used with Reconstruction system
- Tested with newly captured images

### Color image
color

### Depth discontinuity mask
mask

### Before optimization
screen shot 2018-06-28 at 6 52 05 pm

### After optimization
screen shot 2018-06-28 at 6 51 31 pm

src/Python/Tutorial/Misc/sampling.py
qianyizh: Yes. This is a good point. A simple fix is to use ``cv2.Sobel`` to generate the magnitude of image gradient, then sum it to estimate the sharpness. Then use a rolling window to pick the sharpest frames once in a while.

src/Python/Core/open3d_colormap.cpp
qianyizh: Yeah it is a bit confusing. But I can't come up with better names. Let's just use this one for now.

src/Core/Geometry/Image.cpp
qianyizh: I think it is fine for now. In the long run, we should seriously consider how to use OpenCV in Open3D. Actually I think [vcpkg](https://github.com/Microsoft/vcpkg) might be our savior. All that is missing is a package management system on Windows. Thus we had to include all the source code and are afraid of linking to big packages like OpenCV. But if Windows have a package management system, lots of our problems will be solved. We should invest time in this. @syncle @takanokage

qianyizh:
Looks good. Merging.

lrxjason:
Is it normal for the Residual error increase in some iteration and decrease later?
[Iteration 0001] Residual error : 17140.453393, reg : 0.000000
[Iteration 0002] Residual error : 15428.754199, reg : 37.112254
[Iteration 0003] Residual error : 14066.208086, reg : 92.925209
[Iteration 0004] Residual error : 12799.874035, reg : 159.233673
[Iteration 0005] Residual error : 11881.530189, reg : 224.982045
[Iteration 0006] Residual error : 10978.509944, reg : 291.869497
[Iteration 0007] Residual error : 10164.817606, reg : 357.346255
[Iteration 0008] Residual error : 9418.470241, reg : 419.045432
[Iteration 0009] Residual error : 8753.664132, reg : 471.310913
[Iteration 0010] Residual error : 8271.234603, reg : 523.733466
[Iteration 0011] Residual error : 7810.671108, reg : 574.601559

> [Iteration 0012] Residual error : 7503.250006, reg : 623.425672
> [Iteration 0013] Residual error : 7182.370144, reg : 668.310040
> [Iteration 0014] Residual error : 7944.772004, reg : 3436.285845
> [Iteration 0015] Residual error : 8398.769080, reg : 1366.687143
> [Iteration 0016] Residual error : 8075.550004, reg : 1086.460814
> [Iteration 0017] Residual error : 7834.747559, reg : 1121.454900
> [Iteration 0018] Residual error : 7645.106373, reg : 1087.304552
> [Iteration 0019] Residual error : 7586.346190, reg : 1110.170485
> [Iteration 0020] Residual error : 7510.479685, reg : 1142.089592
> [Iteration 0021] Residual error : 7360.104927, reg : 1175.957038

[Iteration 0022] Residual error : 7256.626958, reg : 1210.944181
[Iteration 0023] Residual error : 7166.587797, reg : 1246.043558
[Iteration 0024] Residual error : 7113.245672, reg : 1324.791503
[Iteration 0025] Residual error : 7013.289640, reg : 1359.360750
[Iteration 0026] Residual error : 6900.794533, reg : 1392.757199
[Iteration 0027] Residual error : 6926.551166, reg : 1339.472581
[Iteration 0028] Residual error : 6869.239273, reg : 1354.990521
[Iteration 0029] Residual error : 6785.210600, reg : 1379.802116
[Iteration 0030] Residual error : 6785.631435, reg : 1411.250541
[Iteration 0031] Residual error : 6757.596825, reg : 1582.067014
[Iteration 0032] Residual error : 6721.654318, reg : 1448.798946
[Iteration 0033] Residual error : 6709.315793, reg : 1454.303839
[Iteration 0034] Residual error : 6604.872329, reg : 1470.278581

syncle:
That may happen when images are being bended too much. You can notice `reg` increases suddenly for this reason. In that case, you can make iteration more conservative way - making anchor point more stubborn. Consider to increase ``option.non_rigid_anchor_point_weight``.

lrxjason:
Thank you very much. After the increase, the error decreased more stable. @syncle 
After color optimization, Some parts of the surface is black which has color in the non-optimized mesh. Is there any parameter to deal with that?

syncle:
Consider turn off ``non_rigid_camera_coordinate``. In case of non-rigid image deformation, there is a the case if a image is warped inwards.

chunduri11:
Do you have any plans to consider implementing **alternating optimisation** for Color map optimisation?
This could give significant speed up with out compromising on the quality of the results.

theNded:
> Do you have any plans to consider implementing **alternating optimisation** for Color map optimisation?
> This could give significant speed up with out compromising on the quality of the results.

I have commented that in #2686.

chunduri11:
Thanks, this information Is very useful.

Issue #225, opened 2018-03-05
Update Point Cloud in visualizer without closing it 


zjudmd1015:
Hi Authors,

I want to show registration process in an online way (update it step by step, gradually). Is there a way to update Point Cloud in the visualizer without closing it?

THX!

syncle:
Related to #145. Can you try with the solution @felixendres suggested?

syncle:
Addressed in #343.

PR #343, opened 2018-05-09
Non blocking visualization

### LINKED ISSUE #145: Non-blocking rendering
Hi,
I am wondering if there is a way to render an image and interact with it while the code keep running.

I am implementing a non-rigid registration pipeline based on many optimization cycles, and I'd like to see how the optimization work in every loop. 
Now running "draw_geometries" method, it waits until I close the windows to goes on.

Thanks!
R

syncle:
This resolves #145, #225, #343.

Example (live visualization of ICP registration)

visualize_icp_iteration


src/Python/Visualization/py3d_visualizer.cpp
```diff
@@ -56,6 +56,10 @@ void pybind_visualizer(py::module &m)
         .def("run", &Visualizer::Run, "Function to activate the window")
         .def("close", &Visualizer::Close,
                 "Function to notify the window to be closed")
+        .def("reset_view_point", &Visualizer::ResetViewPoint, "Function to reset view point")
+        .def("update_geometry", &Visualizer::UpdateGeometry, "Function to update geometry")
+        .def("update_renderer", &Visualizer::UpdateRender, "Function to inform render needed to be updated")
```
syncle: I am not very sure this function needed to be exposed.

qianyizh:
Could have a document (maybe append to this one) to discuss the rendering loop of Open3D.

The rendering loop is as follows (see ``Visualizer::Run()``):
1. Check if geometry has changed, if yes, re-bind geometry to shaders
2. Check if view parameters have changed, if yes, re-render the scene
3. PollEvent() to check any user mouse/keyboard input, and respond to it
4. Go back to 1

Note that both binding geometry and rendering are costly operations, thus they are executed in a lazy way. There are two flags control them individually. Function ``update_geometry()`` and ``update_renderer()`` set these flags to dirty individually. After rebinding/rendering, these flags are cleared once again.

It's the programmers' responsibility to explicitly call ``update_geometry()`` and ``update_renderer()``.

syncle:
OK, working on it.

Issue #217, opened 2018-02-28
install Open3D as python package


EthanZhangYi:
HI,
Following the instruction document, `py3d` is compiled in `build/lib`. How to install it as a python package and manage it by `pip` or `conda` ? or I just need to copy the py3d.so to the default python package path?

qianyizh:
Yes. So far copying py3d.so to your working directory is the best solution.

This is also related to #130

syncle:
This is good suggestion. I actually have been thinking about this idea for a long time. This will make accessibility of Open3D even easier as ``conda`` or ``pip`` are widely used python version control tool now days. Just typing ``conda install py3d`` or ``pip install py3d`` would be great. How about officially submit py3d to anaconda and pip once we have stable release?

qianyizh:
I think the right timing is when we have v1.

takanokage:
Just a note this used to be milestone 0.2 but it's not showing up anymore in the list.

qianyizh:
Yeah I just had a discussion with @syncle and we think v0.2 might be too early to do this. Thanks for the note!

syncle:
The solution that copies py3d module to default python path is addressed in #286.

takanokage:
well, yes and no.
Yes, #286 is a good solution for installing from sources.
Would be nice if we place this in a public repository so that people just say something like:
`pip install open3d`.

qianyizh:
I think this is lower priority for now. I also think Open3D is not mature enough to be adopted by pip or conda. 😸 

Let's focus on other high priority things first then we revisit this.

harpone:
Copying works fine, but then the module is not recognized by e.g. PyCharm's package manager (can import it fine though), which is a nuisance...

takanokage:
addressed as part of milestone [0.2](https://github.com/IntelVCL/Open3D/tree/v0.2).

syncle:
Useful link for conda packaging: https://conda.io/docs/user-guide/tutorials/build-pkgs-skeleton.html

PR #350, opened 2018-05-10
PyPi support

takanokage:
I've reached a point where I need somebody else to check my work.
I've uploaded packages to PyPi at https://pypi.org/project/open3d-python/.
Installation PASSED on:

1. macOS 10.13.4 with Python 2.7
2. Windows 10 with Python 2.7 and 3.5
3. Ubuntu 16.04 with Python 2.7 and 3.5

For now the .so (mac and linux) and .pyd (win32) binaries need to be manually built and copied to the right locations before building the .whl package. We may need to change this in the future and maybe build the binaries on the target directly - I don't know how to do that today.

Edit 180510.1232
#217


util/pip_package/open3d/linux/readme.txt
```diff
@@ -0,0 +1 @@
+place linux binaries here
```
syncle: Question: is this just place holder?

LICENSE
```diff
@@ -1,5 +1,6 @@
-MIT License
+The MIT License (MIT)
 
+Open3D: www.open3d.org
```
syncle: Just question: do we need to change license file?

syncle:
> For now the .so (mac and linux) and .pyd (win32) binaries need to be manually built and copied to the right locations before building the .whl package.

Two options (maybe I am wrong):
1) Give warning message saying binaries need to be built prior to run this script, and the packaging script immediately quits.
2) Adding cmake build command into packaging script.

util/pip_package/open3d/linux/readme.txt
takanokage: it's a reminder for whomever is building the pip package.

LICENSE
takanokage: sorry about the confusion.

We used to have two license files, one in open3d/ and one in open3d/src.
This is my proposal to keep just one. It's the license we used to have in open3d/src/ because it had one additional line: `Open3D: www.open3d.org`

Any other changes are just wraparound differences.

takanokage:
I'd leave this as is for now unless we want to push the PR even further. We
need to research this a bit more.

This is how we build the wheel:

$ python setup.py bdist_wheel

There's no way that I know of to check if the right binaries have been
included in the build. We could edit MANIFEST.in and specifically name each
binary that needs to be copied however the binary names could change
depending on the python and compiler version used so I'd recommend against
that.

We can't take the cmake approach either because we need the binaries for
linux, windows and macos before building the pip package. However those
binaries cannot be all built by one cmake build command.
Which brings us to another note:
1. the binaries are built on their respective OSs
2. the pip package is built on Ubuntu after the binaries are copied over.

takanokage:
Do we have a disclaimer anywhere to address the current situation of releasing actual binaries?
We're doing our due diligence however it's a binary, you know, no guarantees and all...

util/pip_package/setup.py
```diff
@@ -0,0 +1,58 @@
...
+        "Topic :: Multimedia :: Graphics :: Graphics Conversion",
+        "Topic :: Multimedia :: Graphics :: Viewers",
+        "Topic :: Scientific/Engineering",
+        "Topic :: Scientific/Engineering :: Mathematics",
+        "Topic :: Scientific/Engineering :: Visualization",
+        "Topic :: Software Development :: Libraries :: Python Modules",
+        "Topic :: Utilities",
+    ],
+    description = ("Open3D is an open-source library that supports rapid development of software that deals with 3D data.."),
+    install_requires=['numpy', 'matplotlib', 'opencv-python', ],
```
qianyizh: is ``opencv-python`` required?
I would rather not having it since it is a big library and not many people have it.

util/pip_package/setup.py
```diff
@@ -0,0 +1,58 @@
...
+    include_package_data=True,
+    keywords = "3D reconstruction pointcloud",
```
qianyizh: keywords = "3D reconstruction, point cloud, mesh, RGB-D, visualization"

qianyizh:
Only a few minor comments from me.

@takanokage can you do a project-wise check to make sure there is no dangling ``py3d`` keyword around?

@syncle everything looks good to m. you can merge any time when you think it is ready.

takanokage:
Again, good call. I see I have actually missed one or two `py3d` references in the documentation. I'll fix right away.

syncle:
Sorry I didn't aware this PR was on my hand. @takanokage, can you resolve conflicts? I will merge once the conflict is resolved.

PR #379, opened 2018-06-01
add make install/uninstall rules and pkg-config support

BlazingForests:
add make install rule for .so file
add make uninstall rule
add pkg-config support

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -138,3 +138,31 @@ endif(EXISTS "${Open3D_SOURCE_DIR}/Experimental/CMakeLists.txt")
 if (Open3D_BUILD_PYTHON_BINDING)
     add_subdirectory(Python)
 endif(Open3D_BUILD_PYTHON_BINDING)
+
+
+# config
+CONFIGURE_FILE(open3d.pc.in "${PROJECT_BINARY_DIR}/open3d.pc" @ONLY)
+
+
+# install
+install(DIRECTORY ./
+  DESTINATION include/open3d
+  FILES_MATCHING PATTERN "*.h"
+  PATTERN "*.txt" EXCLUDE
+  PATTERN "Core/*"
+  PATTERN "IO/*"
+  PATTERN "Visualization/*"
+  PATTERN "CMake/*" EXCLUDE
+  PATTERN "Experimental/*" EXCLUDE
+  PATTERN "External/*" EXCLUDE
+  PATTERN "Python/*" EXCLUDE
+  PATTERN "Test/*" EXCLUDE
+  PATTERN "Tools/*" EXCLUDE
+  PATTERN "UnitTest/*" EXCLUDE
+)
+
+INSTALL(FILES "${PROJECT_BINARY_DIR}/open3d.pc" DESTINATION lib/pkgconfig)
+
+# uninstall
+configure_file("${CMAKE_CURRENT_SOURCE_DIR}/cmake_uninstall.cmake" "${CMAKE_CURRENT_BINARY_DIR}/cmake_uninstall.cmake" IMMEDIATE @ONLY)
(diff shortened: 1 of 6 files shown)


takanokage:
@BlazingForests Thank your for your submission.
I have already implemented install/uninstall rules and other cmake fixes. Please check my [issue_130](https://github.com/takanokage/Open3D/tree/issue_%23130) branch and let me know if that is satisfactory to you. I am in the process of polishing and validating my contribution which includes over 110 commits before submitting a PR in the next couple of days.

Some of this features that will soon be added to Open3D through my [issue_130](https://github.com/takanokage/Open3D/tree/issue_%23130) branch:
- make install/uninstall Open3D and each of the external dependencies that are custom built
- only one library output (open3d.lib) instead of three (core.lib, io.lib, visualization.lib)
- enable the creation of a new C++ project that works with only installed Open3D headers and libraries
- find_package(Open3D) functionality allowing the creation of a C++ project that has minimal information about the Open3D headers & libs names and location. 
- Open3D version; allows accessing/displaying the Open3D version from C++; allows CMake to search for a specific Open3D version
- pip package version and other information based on CMakeLists.txt settings; version specified in one central location only
- substantially simplified CMakeLists.txt files in order to increase applicability, readability and maintainability
- etc.
- new documentation describing those changes

BlazingForests:
@takanokage Thank you. Awesome!

I have some suggestions.

* I think if the functionality is separate, can work alone. Don't need to merge them to an single library file. That is good for modularization and keep lightweight. More dependency means more heavy and more hard to manage.

* All the so files have prefix (eg `libopen3d_`), Core, IO, that kind of name are too common, can not unique.

* Please add .pc.in file for pkg-config support, each sub library need a separate .pc.in file.

* In `src/Python/CMakeLists.txt` file,  `${PYTHON_EXECUTABLE} -m site --user-site`  doesn't work for python2, I change to `${PYTHON_EXECUTABLE} -c "import site; print(site.getusersitepackages())"` is works for both python2 and python3.

* **libtinyfiledialogs.a** don't need to install

takanokage:
@BlazingForests Thank you for your suggestions! I will consider them for the upcoming PR.

takanokage:
@BlazingForests 

- `I think if the functionality is separate, can work alone.`
   A: The IO lib depends on the Core lib and the Visualization lib depends on the IO lib and the Core lib - they are not independent. The Visualization lib is basically the same as my proposed Open3D lib. Static linking against Open3D automatically prunes the features that are not used - no need to do it manually.

- `All the so files have prefix (eg libopen3d_)`
   A: composing everything together under Open3D solves the uniqueness part.

- `Please add .pc.in file for pkg-config support`
   I agree we should add this.

- `In src/Python/CMakeLists.txt file, ${PYTHON_EXECUTABLE} -m site --user-site doesn't work for python2`
   A: works on my machines. Tested on Ubuntu 16.04, macOS 10.13.4, Windows 10 with Python 2.7. I do however acknowledge this is an issue as more than one user has reported this.
  TODO: I'll check again, maybe some additional circumstances cause this to fail in the way you described. I'd appreciate if you have any additional information that might help shine a light on this issue.

- `libtinyfiledialogs.a don't need to install`
   A: not true. _libtinyfiledialogs.a_ is not included in the Core/IO/Visualization or Open3D libraries if static building. I'm talking about the case when a C++ project is created independently of the Open3D CMake configuration/generation. In such a case the C++ project should work with the Open3D installed headers and libs and _libtinyfiledialogs.a_ is required.

takanokage:
@BlazingForests 
- `libtinyfiledialogs.a don't need to install`
   You just gave me an idea: I could make _libtinyfiledialogs_ an object library and thus force it's inclusion in the Open3D lib the way I did with Core/IO/Visualization. I think it should work in which case there wouldn't be a need to install _libtinyfiledialogs.a_.

qianyizh:
I will close this PR since there is a more thorough solution tracked in https://github.com/IntelVCL/Open3D/pull/381

Issue #130, opened 2018-02-02
Open3D cpp interface and "make install"


amiltonwong:
Hi, authors,

If I want to use Open3D as a substitute for PCL to deal with 3D data manipulation and visualization (my system uses CMake for project building).  What cmake instructives should I append in my CMakeLists.txt  file? It'll be very useful to have an example of CMakeLists.txt and also a simple example on how to call Open3D C++ interface.

THX!

qianyizh:
We don't have detailed documentation for the cpp interface.
But there are examples in:
https://github.com/IntelVCL/Open3D/tree/master/src/Test
and
https://github.com/IntelVCL/Open3D/tree/master/src/Tools
Hope they are helpful.

syncle:
That's good point. As mentioned above, we have several cpp examples. The cpp examples should be self-explanatory. They are hidden gem of Open3D that are used in our past projects, and some of nice applications (like cropping point cloud) are not addressed in Python Tutorial.  Our intention was to guide Open3D users to be familiar with Python environment first.

The best answer for your question for now is to add your custom cpp file in test folder and add entry of your application in CMakeList.txt file. The Open3D build will generate standard-alone executable. 

I began to consider adding a getting started guide for cpp examples. Please stay tuned!

ma-bru:
It would also be really nice if "make install" could install a Open3D-config.cmake. Ideally with the new cmake style of exporting targets instead of the old OPEN3D_LIBRARIES, OPEN3D_INCLUDE_DIRS style.
That would allow people to just use the standard cmake FindPackage() function.

qianyizh:
It would be nice if someone is interested in building the "make install" function for Open3D. I started it in 2015 (that's why it is very old style!) and have not got time to finish it.

Can anyone contribute a PR?

takanokage:
Qianyi, you mentioned you might have a few ideas on how to best approach this.
Just a note for you to send me the resources when you have time.

qianyizh:
There are two parts:
 * Setup make install
 * Setup C++ interface
This is also related to #234 

To set up make install, we only need to edit the cmake files. Just see what glfw has done:
https://github.com/glfw/glfw/blob/0d4534733b7a748a21a1b1177bb37a9b224e3582/CMakeLists.txt
https://github.com/glfw/glfw/blob/0d4534733b7a748a21a1b1177bb37a9b224e3582/src/CMakeLists.txt
We probably need an uninstall recipe too, example:
[...]

PR #381, opened 2018-06-05
make install

takanokage:
Issue #130 

cmake install overhaull


takanokage:
A more detailed list of changes:
[381.docx](https://github.com/IntelVCL/Open3D/files/2082278/381.docx)

syncle:
This is a HUGE and GREAT contribution for the project. Thanks @takanokage! I just have read your summary document. Among many changes, I mostly liked clear cmake structure, version control system, and so on!

I just have some questions:
- I tried building from source using this branch. In my case (MacOS Sierra) I saw ``jpeg dependency not met`` and got failed. I removed build folder, changed ``BUILD_JPEG`` flag in ``src/CMakeLists.txt`` as ``ON``, and got success. Could it be possible that any external libraries  does not met dependency requirement can be automatically build from source? 
- I would like to test find_package too. Could you guide me how can I begin with?

Thanks again!

takanokage:
@syncle Thank you!
- `Could it be possible ... automatically build from source?`
  I have thought about it and understand the value. OpenCV seems to favor the explicit way - the user has to turn build options to ON, the OpenCV CMake won't do it automatically. So I did it the way OpenCV does it.
- `I would like to test find_package too.`
  Let me work on this, the topic is fresh in my mind. There might be a way to do it from command line. I'll make a change to this PR if it works.

takanokage:
The following command find Open3D installed in the `~/open3d_install` user folder:

`$ cmake --find-package -DNAME=Open3D -DCOMPILER_ID=GNU -DLANGUAGE=C -DMODE=EXIST -DCMAKE_PREFIX_PATH="~/open3d_install/lib/CMake"`

The return of the command is `Open3D found`. I just need to integrate this in the CI scripts.

takanokage:
here's a [reference](https://stackoverflow.com/questions/28863366/command-line-equivalent-of-cmakes-find-package) to the command line example of find_package.

takanokage:
@syncle 

- `Could it be possible ... automatically build from source?`
  Another thing I noticed when turning a build option to ON if the corresponding dependency was missing: Windows CMake GUI still shows the build option as OFF. This is clearly an issue in the CMake GUI however until this is improved users may get confused: Open3D would build fine with the build option either ON or OFF.

syncle:
Thanks for the comments @takanokage.

> Windows CMake GUI still shows the build option as OFF

Maybe this is Cmake GUI does not get refreshed. Is that happens even if build folder is made from scratch?

The cmake command also works in my Mac without problem actually. Can I ask why 'find_package works on Ubuntu' in your documentation? I also would like to know if there is a example script/code that can build open3d application using cmake find_package.

.gitignore
```diff
@@ -15,3 +15,11 @@ src/Visualization/Shader/Shader.h
 _build/
 doxygen/
 .vs/
+src/open3d_config.h
+util/pip_package/setup.py
```
syncle: Why do we need to add these two files in .gitignore? I understand src/open3d_config.h is generated from source, but setup.py is not.

docs/_static/C++/CMakeLists.txt
```diff
@@ -0,0 +1,129 @@
+cmake_minimum_required(VERSION 3.0)
```
syncle: Aha. You are trying to include this example program into docs/_static folder. How about moving this into Open3D/examples in the future?

docs/compilation.rst
```diff
@@ -1,50 +1,78 @@
...
+          -DBUILD_GLEW=ON    \
+          -DBUILD_GLFW=ON    \
+          -DBUILD_JPEG=ON    \
+          -DBUILD_JSONCPP=ON \
+          -DBUILD_PNG=ON     \
+          ../src
+
+.. tip:: This can save a lot of time on Windows where it can be particularly difficult to install the Open3D dependencies.
+
+.. note:: Enabling these build options may increase the compilation time substantially.
```
syncle: I don't think it will substantially increase the time :) Let's just say 'increase the compilation time'

docs/getting_started.rst
```diff
@@ -28,17 +28,29 @@ Run the following commands to install the dependencies and use CMake to build th
...
-Install Python binding module by running the following script.
+The ``CMAKE_INSTALL_PREFIX`` argument is optional.
+
+.. tip:: In order to build the Open3D Python module with a particular Python version in mind set the either the ``PYBIND11_PYTHON_VERSION`` or the ``PYTHON_EXECUTABLE`` cmake option.
+
+.. code-block:: bash
+
+    cmake ../src -DPYBIND11_PYTHON_VERSION=2.7
+    # or
+    cmake ../src -DPYTHON_EXECUTABLE=/usr/bin/python
```
syncle: The python issue is addressed in docs/compilation.rst. How about providing a link?

takanokage:
@syncle 

- `Maybe this is Cmake GUI does not get refreshed`
  The GUI was unable to display the true value of the build options if the build options are changed in the _src/CMakeLists.txt_ as part of some logic that flips the build options on. Seems that the issue is unrelated to wether the GUI is refreshed or not.

- `The cmake command also works in my Mac without problem actually.`
  At the time I wrote the documentation I didn't test creating a custom C++ project with _find_package_ on mac yet. `make install` should work on mac with no problems. However I expect issues when creating the C++ project due to the fact that fewer External dependencies can be installed on mac than on Ubuntu (on Ubuntu this only works with all the dependencies installed).

- `example script/code that can build open3d application using cmake find_package`
  Docs » Tutorial » C++ interface shows how to create a custom open3d application in C++ an provides links to download a _CMakeLists.txt_ and a _.cpp_ for this purpose. I have also update the documentation, there was an error concerning which lines of _CMakeLists.txt_ were displayed.

src/CMakeLists.txt
```diff
@@ -112,29 +185,103 @@ include_directories(${GLFW_INCLUDE_DIRS})
...
+endif ()
+
+target_link_libraries(${CMAKE_PROJECT_NAME}
+                      ${GLFW_LIBRARIES}
+                      ${GLEW_LIBRARIES}
+                      ${JPEG_LIBRARIES}
+                      ${JSONCPP_LIBRARIES}
+                      ${OMP_LIBRARIES}
+                      ${PNG_LIBRARIES}
+                      tinyfiledialogs)
```
syncle: Missing {}?

src/Tools/ManuallyAlignPointCloud/CMakeLists.txt
```diff
@@ -1,6 +1,9 @@
-file(GLOB HEADER_FILES "*.h")
 file(GLOB SOURCE_FILES "*.cpp")
-add_executable(ManuallyAlignPointCloud ${HEADER_FILES} ${SOURCE_FILES})
-target_link_libraries(ManuallyAlignPointCloud Visualization IO Core)
-set_target_properties(ManuallyAlignPointCloud PROPERTIES
-    FOLDER "Tools")
+
```
syncle: redundant lines?

.gitignore
takanokage: actually _setup.py_ is generated as well from _setup.py.in_.
I just double checked and _setup.py_ is not committed to the repository.

docs/_static/C++/CMakeLists.txt
takanokage: I think that's potentially a great idea, I didn't know where to put it. We need to discuss this more offline.

And you just gave me an idea: why not configure cmake install to install the documentation files as well? We can do this as a separate PR. What do you think?

docs/tutorial/C++/cplusplus_interface.rst
```diff
@@ -0,0 +1,51 @@
...
+.. warning:: The following is supported at this time only on Ubuntu with all dependencies installed.
+
+Let's create a basic C++ project based on CMake and Open3D installed libraries and headers.
+
+1. Get the code from :download:`TestVisualizer.cpp <../../_static/C++/TestVisualizer.cpp>`
+2. Get the CMake config file from :download:`CMakeLists.txt <../../_static/C++/CMakeLists.txt>`
+3. mkdir -p build
+4. cd build
+5. cmake ..
+6. make -j
```
syncle: How about make it like ``make -j``? These are shell commands.

.gitignore
syncle: Aha got it.

docs/compilation.rst
takanokage: got it.

docs/_static/C++/CMakeLists.txt
syncle: Sounds good idea. Is the proposal intended for CI?

src/CMakeLists.txt
takanokage: There's no tinyfiledialogs_LIBRARIES variable to use here, tinyfiledialogs always gets built from source. I guess I can manually add it - would you prefer that?

syncle: Yes, if it can be addressed quickly. This will make cmake file more consistent.

src/Tools/ManuallyAlignPointCloud/CMakeLists.txt
takanokage: _ManuallyAlignPointCloud_ was treated differently than the other _src/Tools_ projects. Because of that I can't replace the CMake logic for it with a one liner the way I did with the other _src/Tools_ projects.

docs/tutorial/C++/cplusplus_interface.rst
takanokage: got it.

src/Tools/ManuallyAlignPointCloud/CMakeLists.txt
syncle: I see. I actually meant to say that let's remove empty lines.

src/CMakeLists.txt
takanokage: done.

docs/getting_started.rst
takanokage: done

docs/_static/C++/CMakeLists.txt
takanokage: True, they wouldn't need to. In light of that my idea is incomplete.

I was thinking more about how we would go about packaging Open3D in the future - we might want to put some documentation inside the package. However I agree, the form that documentation might take might be different from this one.

takanokage:
I don't know what's going on, the same passed all the Travis CI tests on my branch.

syncle:
> I don't know what's going on, the same passed all the Travis CI tests on my branch.

Yes. I don't think this is not the problem from this PR. I've just re-executed the build again. Let's wait again.

I am actually tried ``tutorial/C++/cplusplus_interface.html`` using my Ubuntu 16.04 machine and Mac but both of them didn't worked out with basic library warnings. 

From my Ubuntu
```
-- The C compiler identification is GNU 5.4.0
-- The CXX compiler identification is GNU 5.4.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found Open3D 0.1.5.2
-- Found eigen3 3.2.92
-- Found glew 1.13.0
-- Found glfw 3.2.1
-- Found libjpeg 
-- Found jsoncpp 1.7.2
-- Found OpenGL
-- Found OpenMP 
CMake Error at CMakeLists.txt:128 (message):
  libpng not found


-- Configuring incomplete, errors occurred!
```
At the time of Open3D compilation, Cmake couldn't find libpng, and I had to turn on the flag regarding libpng build. I installed Open3D using the provided script (``build.sh`` in Open3D root folder), and checked the python tutorial runs well without hassle.

takanokage:
Yes, I have observed similar behavior. At this time I don't know how to make this work if the dependencies are not installed, all of them.

takanokage:
I have experimented quite a lot with GLFW on Ubuntu 16.04.
1. I removed libglfw3-dev from my system.
2. Built Open3D with GLFW from source.
3. Looked for ways to convince src/External/GLFW to install the library.
3.1. GLFW_INSTALL=ON doesn't do anything.
3.2. Updating the GLFW CMakeLists.txt with my own install makes the custom C++ project complain about missing library dependencies.

I haven't found a way to fix this yet. Ideally it should work out of the box without the programmer having to add more and more find_package(...) statements in order to address all the library dependencies of the dependencies of Open3D. The best option at the moment seems to be to ask the programmer to install all the dependencies rather than having Open3D try to install them from source.  

For now find_package(Open3D) is supported on Ubuntu only with all of the dependencies installed. More work is required to make it work on Ubuntu with missing dependencies and on Mac and Windows.

syncle:
Thanks for the clarification. Now I understand what is happening better. By the way, I have no objection that this PR is more than enough for 'make install' task. Regarding the ongoing issue, we can keep addressing this for future PR. @qianyizh Can you also review and comment on this PR?

src/External/GLFW/CMakeLists.txt
```diff
@@ -375,7 +375,7 @@ endif()
 #--------------------------------------------------------------------
 # MODIFICATION BY Open3D: overwrite some GLFW cmake variables
 #--------------------------------------------------------------------
-set(GLFW_INCLUDE_DIRS "${Open3D_SOURCE_DIR}/External/GLFW/include")
+set(GLFW_INCLUDE_DIRS "${CMAKE_CURRENT_SOURCE_DIR}/include")
 set(GLFW_LIBRARIES "glfw")
 set(GLFW_INCLUDE_DIRS ${GLFW_INCLUDE_DIRS} PARENT_SCOPE)
 set(GLFW_LIBRARIES ${GLFW_LIBRARIES} PARENT_SCOPE)
```
qianyizh: At the end of this CMakeLists file, there should be install rules as in other CMakeLists files, something like:
```
+# install
+install(TARGETS ${GLEW_LIBRARY}
+        RUNTIME DESTINATION ${CMAKE_INSTALL_PREFIX}/bin
+        LIBRARY DESTINATION ${CMAKE_INSTALL_PREFIX}/lib
+        ARCHIVE DESTINATION ${CMAKE_INSTALL_PREFIX}/lib)
```

In its current status, if ``BUILD_GLFW`` is on, and make install, GLFW won't be install into the Open3D install directory.

src/External/GLFW/CMakeLists.txt
```diff
@@ -32,7 +32,7 @@ set(GLFW_INSTALL OFF)
 option(GLFW_VULKAN_STATIC "Use the Vulkan loader statically linked into application" OFF)
```
qianyizh: The current install behavior of GLFW is different than other libraries when ``BUILD_GLFW`` switch is ``ON``. I have done some initial testing, I think the solution here is to change:
```
set(GLFW_INSTALL OFF)
```
to
```
set(GLFW_INSTALL ON)
```
Then its install behavior is consistent with other libraries (i.e., glfw3.lib is copied to the install directory of Open3D).

docs/_static/C++/CMakeLists.txt
```diff
@@ -0,0 +1,129 @@
...
+if (PKGCONFIG_FOUND)
+    pkg_search_module(EIGEN3    eigen3>=3.2.7   QUIET)
+    pkg_search_module(JPEG      libjpeg         QUIET)
+    pkg_search_module(GLFW      glfw3           QUIET)
+    pkg_search_module(GLEW      glew            QUIET)
+    pkg_search_module(JSONCPP   jsoncpp>=1.7.0  QUIET)
+    pkg_search_module(PNG       libpng>=1.6.0   QUIET)
+endif (PKGCONFIG_FOUND)
+
+option(Open3D_USE_NATIVE_DEPENDENCY_BUILD "Prioritize Found dependency build" ON)
```
qianyizh: Remove ``Open3D_USE_NATIVE_DEPENDENCY_BUILD``, it is not used any more.

docs/_static/C++/CMakeLists.txt
```diff
@@ -0,0 +1,129 @@
...
+find_package(Open3D    QUIET HINTS "~/open3d_install/lib/CMake")
+
+if (PKGCONFIG_FOUND)
+    pkg_search_module(EIGEN3    eigen3>=3.2.7   QUIET)
+    pkg_search_module(JPEG      libjpeg         QUIET)
+    pkg_search_module(GLFW      glfw3           QUIET)
+    pkg_search_module(GLEW      glew            QUIET)
+    pkg_search_module(JSONCPP   jsoncpp>=1.7.0  QUIET)
+    pkg_search_module(PNG       libpng>=1.6.0   QUIET)
+endif (PKGCONFIG_FOUND)
```
qianyizh: These lines force CMake to find the packages installed in the system.
However, I just made a quick test, and there is an easy workaround to build everything from source code included in Open3D. In this way, you don't even need to install the libraries, 😁 

First, you need to apply the change I mentioned above for GLFW CMakeLists.txt file. Then once you install Open3D, all header files will be copied to Open3D include folder and all libraries will be copied to Open3D lib folder.

Then create a CMakeLists.txt like this:
```
cmake_minimum_required(VERSION 3.0)

project(TestVisualizer)

find_package(PkgConfig QUIET)
find_package(OpenGL    QUIET)
find_package(pybind11  QUIET)
find_package(OpenMP    QUIET)
find_package(Open3D    QUIET HINTS "~/open3d_install/lib/CMake")

# Open3D
if (Open3D_FOUND)
    message(STATUS "Found Open3D ${Open3D_VERSION}")

    add_executable(TestVisualizer TestVisualizer.cpp)

    include_directories(${Open3D_INCLUDE_DIRS})

    link_directories(${Open3D_LIBRARY_DIRS})

    set_target_properties(TestVisualizer PROPERTIES
            RUNTIME_OUTPUT_DIRECTORY "${CMAKE_RUNTIME_OUTPUT_DIRECTORY}")

    target_link_libraries(TestVisualizer ${Open3D_LIBRARIES} jsoncpp jpeg glew png zlib glfw3 opengl32 glu32)
    target_link_libraries(TestVisualizer tinyfiledialogs)

else ()
    message(SEND_ERROR "Open3D not found")
endif ()
```

I have tested it on Windows 10 + VS 2017. It works. Actually, the right way to do it is: when we compile Open3D and install it, we write a customized FindOpen3D.cmake file. It remembers which dependent libraries are compiled from source included in Open3D, and include these dependent libraries in ``${Open3D_LIBRARIES}``. In this case, when an external C++ program is linked to ``${Open3D_LIBRARIES}``, it links to the Open3D library and all dependent libraries that were compiled with it.

qianyizh:
Overall, this PR is great! I think it is almost ready.
I have a change proposal (see above comments) that can support **build from source** and **link with external c++ programs** simultaneously. @syncle @takanokage please take a look.

But since this PR is already very big, I am leaning towards merging it first, then addressing my proposal in the next PR.

Thoughts?

syncle:
I agree with @qianyizh. I will merge this PR, once additional comments are addressed.

src/External/GLFW/CMakeLists.txt
takanokage: I have also tried this option and it didn't work for some reason.
I will double check.

docs/_static/C++/CMakeLists.txt
takanokage: done.

takanokage:
I am double checking QianYi's findings.
I'll let you know how it goes.

src/External/GLFW/CMakeLists.txt
takanokage: I found out why this wasn't working for me: I was setting GLFW_INSTALL in src/CMakeLists.txt and the value was overridden in src/External/GLFW/CMakeLists.txt.

docs/_static/C++/CMakeLists.txt
takanokage: At this time I recommend we move fwd with this PR as is (with only minor modifications) and address the build dependencies issue in a separate PR.

There actually multiple problems, I'll mention below just two of the major ones.
I can fix these however this will take more than just half a day.
1. GLFW_LIBRARIES is set to glfw but the output of the build GLFW is actually glfw3. Somehow this works fine when building Open3D but when building a custom C++ project only based on installed Open3D libs & headers (and dependencies) the linker fails to find the lib. Setting the GLFW_LIBRARIES to glfw3 breaks the build of Open3D. Setting the Open3D_LIBRARIES to "...; glfw3;..." still doesn't work as the custom build project expects more libraries (dependencies of GLFW). I tested this on Ubuntu after uninstalling _libglfw3_ and _libglfw3-dev_.

```
[100%] Linking CXX executable TestVisualizer
/usr/bin/ld: /home/dpetre/open3d_install/lib/libglfw3.a(vulkan.c.o): undefined reference to symbol 'dlclose@@GLIBC_2.2.5'
//lib/x86_64-linux-gnu/libdl.so.2: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
CMakeFiles/TestVisualizer.dir/build.make:94: recipe for target 'TestVisualizer' failed
make[2]: *** [TestVisualizer] Error 1
CMakeFiles/Makefile2:67: recipe for target 'CMakeFiles/TestVisualizer.dir/all' failed
make[1]: *** [CMakeFiles/TestVisualizer.dir/all] Error 2
Makefile:83: recipe for target 'all' failed
make: *** [all] Error 2
```

2. set(GLFW_INSTALL ON) breaks the `make uninstall`

```
CMake Error at cmake_uninstall.cmake:3 (message):
  Cannot find install manifest:
  "/home/dpetre/open3d/issue_130/build/External/GLFW/install_manifest.txt"

External/GLFW/CMakeFiles/uninstall.dir/build.make:57: recipe for target 'External/GLFW/CMakeFiles/uninstall' failed
make[3]: *** [External/GLFW/CMakeFiles/uninstall] Error 1
CMakeFiles/Makefile2:226: recipe for target 'External/GLFW/CMakeFiles/uninstall.dir/all' failed
make[2]: *** [External/GLFW/CMakeFiles/uninstall.dir/all] Error 2
CMakeFiles/Makefile2:233: recipe for target 'External/GLFW/CMakeFiles/uninstall.dir/rule' failed
make[1]: *** [External/GLFW/CMakeFiles/uninstall.dir/rule] Error 2
Makefile:188: recipe for target 'uninstall' failed
make: *** [uninstall] Error 2
```

takanokage: otherwise all the other mechanisms (like custom creating Open3D include dirs and libraries based on build options) mentioned by QianYi are already in place.

takanokage:
Ok, as I mentioned above, after investigating QianYi's suggestions I recommend we keep going as is with this PR (with only the small changes proposed) and address the install of dependencies and custom C++ project based on that in a separate PR.

qianyizh:
Yeah, that's also my guess. I will merge this PR.
@takanokage can you open a separate issue for the changes you mentioned in the comments?

takanokage:
#393

Issue #393, opened 2018-06-13
improve make install/uninstall and find_package(Open3D) to handle building dependencies from source


takanokage:
When the dependencies are not installed and have to be built from source features like make install/uninstall and find_package(Open3D) don't work.

Some of the issues encountered so far and other things that need to be done:

1. GLFW_LIBRARIES is set to glfw but the output of the build GLFW is actually glfw3. Somehow this works fine when building Open3D but when building a custom C++ project only based on installed Open3D libs & headers (and dependencies) the linker fails to find the lib. Setting the GLFW_LIBRARIES to glfw3 breaks the build of Open3D. Setting the Open3D_LIBRARIES to "...; glfw3;..." still doesn't work as the custom build project expects more libraries (dependencies of GLFW). I tested this on Ubuntu after uninstalling libglfw3 and libglfw3-dev.

```
[100%] Linking CXX executable TestVisualizer
/usr/bin/ld: /home/dpetre/open3d_install/lib/libglfw3.a(vulkan.c.o): undefined reference to symbol 'dlclose@@GLIBC_2.2.5'
//lib/x86_64-linux-gnu/libdl.so.2: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
CMakeFiles/TestVisualizer.dir/build.make:94: recipe for target 'TestVisualizer' failed
make[2]: *** [TestVisualizer] Error 1
CMakeFiles/Makefile2:67: recipe for target 'CMakeFiles/TestVisualizer.dir/all' failed
make[1]: *** [CMakeFiles/TestVisualizer.dir/all] Error 2
Makefile:83: recipe for target 'all' failed
make: *** [all] Error 2
```

2. set(GLFW_INSTALL ON) breaks the make uninstall
CMake Error at cmake_uninstall.cmake:3 (message):
  Cannot find install manifest:
  "/home/dpetre/open3d/issue_130/build/External/GLFW/install_manifest.txt"

```
External/GLFW/CMakeFiles/uninstall.dir/build.make:57: recipe for target 'External/GLFW/CMakeFiles/uninstall' failed
make[3]: *** [External/GLFW/CMakeFiles/uninstall] Error 1
CMakeFiles/Makefile2:226: recipe for target 'External/GLFW/CMakeFiles/uninstall.dir/all' failed
make[2]: *** [External/GLFW/CMakeFiles/uninstall.dir/all] Error 2
[...]

PR #397, opened 2018-06-18
simplify the creation of a custom C++ project

takanokage:
Polished the CMake system in order to simplify the creation of a custom C++ project.
Now the CMakeLists.txt of a custom C++ project only needs to find Open3D and the required headers, libraries and library directories are automatically provided for both Open3D and its external dependencies.


syncle:
Can you also check xcode build? I got some issue with this branch.
```
clang: error: no such file or directory: ~/Open3D/build-xcode/lib/Debug/libOpen3D.a
```
When building `MergeMesh` application.

qianyizh:
This is pretty awesome. I like short CMakeLists files for external C++ programs. I will test this PR tonight.

syncle:
I tried to compile this PR with XCode. For example, 
```
cmake -G Xcode ../src -DCMAKE_INSTALL_PREFIX=~/open3d_install
```
and I got a compile error aforementioned.

takanokage:
This seems to be a CMake bug manifested when using object libraries together with Xcode.
I propose we open a separate issue for that and move with the current PR as is. The solution might not be as simple as not using object libraries because this is the mechanism that I was hoping to use to fix #393.

- [supporting link1](https://cmake.org/pipermail/cmake/2016-April/063169.html)
- [supporting link2](https://cmake.org/Bug/view.php?id=14768)
- [supporting link3](https://github.com/tensorflow/tensorflow/issues/15369)
- [supporting link4](https://gitlab.kitware.com/cmake/cmake/issues/17500)

syncle:
I got it. I approve merge.

Issue #377, opened 2018-05-31
pip packaging: add support for 64bit Python on Windows


takanokage:
At this time [open3d-python](https://pypi.org/project/open3d-python/) does not include binaries compiled for 64bit Python on Window 10.

This feature request was first mentioned in #369.
#369 has been addressed and should be closed.

syncle:
addressed in #406

qm2k:
Thank you for your work, is this change already included in 0.3.0.0? I still see `ImportError: DLL load failed` on 64-bit Python 3. 32-bit Python 2.7 works fine.

yxlao:
@qm2k for "ImportError: DLL load failed on 64-bit Python 3", which python 3 version is that? Currently we only have 2.7, 3.5 and 3.6 supported

qm2k:
Python 3.7 indeed. Is there some trick to make Open3D work with 3.7, or would you recommend downgrading to 3.6 for now?

yxlao:
Compiling from source should work: http://www.open3d.org/docs/getting_started.html#windows. You'll need to specify `PYTHON_EXECUTABLE` to point to the python 3.7 executable in the cmake flags, e.g.
```
-DPYTHON_EXECUTABLE=C:\your-path-to-python3.7\python.exe 
```
Since there are a few request already to support python 3.7, we'll consider adding support for that in the next release.

qm2k:
Thank you for your help!

PR #406, opened 2018-06-21
add support for 64bit Python on Windows

takanokage:
#377 

Added support for 64bit Python 2.7+ and 3.5+ on Windows.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -5,7 +5,7 @@ mkdir -p build
 cd build
 
 # you can specify a custom install location and the python version
-cmake ../src -DCMAKE_INSTALL_PREFIX=~/open3d_install -DPYTHON_EXECUTABLE=/usr/bin/python3.5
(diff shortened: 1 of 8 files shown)


build.sh
```diff
@@ -5,7 +5,7 @@ mkdir -p build
 cd build
 
 # you can specify a custom install location and the python version
-cmake ../src -DCMAKE_INSTALL_PREFIX=~/open3d_install #-DPYTHON_EXECUTABLE=/usr/bin/python
+cmake ../src -DCMAKE_INSTALL_PREFIX=~/open3d_install -DPYTHON_EXECUTABLE=/usr/bin/python3.5
```
syncle: Is it necessary to specify python version?

takanokage: good catch!
I needed that in order to generate a binary for each python version on Ubuntu.
Fixed.

PR #418, opened 2018-06-27
 Supporting interactive cropping of meshes

syncle:
VisualizerWithEditing class supports interactive cropping of meshes. 
- Adding SelectDownSample function for Triangular mesh
Keep a triangle only if all the three corresponding vertices are selected by user.
Preserves vertex ordering
- Generalize Tools/ManuallyCropPointCloud to Tools/ManuallyCropGeometry
Either point cloud and mesh is supported.

**Example of mesh cropping**:
```
./ManuallyCropGeometry --mesh [path_to_mesh_file] 
```
mesh
mesh_crop

**Example of point cloud cropping**:
```
./ManuallyCropGeometry --pointcloud [path_to_mesh_or_pointcloud_file] 
```
pointcloud
pointcloud_crop

### DIFF
--- a/src/Core/Geometry/DownSample.cpp
+++ b/src/Core/Geometry/DownSample.cpp
@@ -114,9 +114,8 @@ std::shared_ptr<TriangleMesh> SelectDownSample(const TriangleMesh &input,
     std::vector<std::vector<int>> vertex_to_face_temp(input.vertices_.size());
     int triangle_id = 0;
     for (auto trangle : input.triangles_) {
-        vertex_to_face_temp[trangle[0]].push_back(triangle_id);
-        vertex_to_face_temp[trangle[1]].push_back(triangle_id);
-        vertex_to_face_temp[trangle[2]].push_back(triangle_id);
+        for (int i=0; i<3; i++)
+            vertex_to_face_temp[trangle(i)].push_back(triangle_id);
         triangle_id++;
     }
     // for each vertex, remove list if it is non-selected vertices
@@ -134,7 +133,7 @@ std::shared_ptr<TriangleMesh> SelectDownSample(const TriangleMesh &input,
     }
     std::vector<bool> mask_observed_vertex(input.vertices_.size());
     for (auto vertex_ids : triangle_idx) {
-        if (vertex_ids.size() == 3)
+        if ((int)vertex_ids.size() == 3)
             for (int i=0; i<3; i++)
                 mask_observed_vertex[vertex_ids[i]] = true;
     }
@@ -145,21 +144,25 @@ std::shared_ptr<TriangleMesh> SelectDownSample(const TriangleMesh &input,
             new_vertex_id[i] = cnt++;
     }
     // push face that has 3 elements.
+    triangle_id = 0;
     for (auto vertex_ids : triangle_idx) {
-        if (vertex_ids.size() == 3) {
+        if ((int)vertex_ids.size() == 3) {
             Eigen::Vector3i new_face;
             for (int i=0; i<3; i++)
-                new_face[i] = new_vertex_id[vertex_ids[i]];
+                new_face(i) = new_vertex_id[input.triangles_[triangle_id][i]];
             output->triangles_.push_back(new_face);
         }
+        triangle_id++;
     }
     // push marked vertex.
-    for (size_t i : indices) {
-        output->vertices_.push_back(input.vertices_[i]);
-        if (has_normals) output->vertex_normals_.push_back(
-                input.vertex_normals_[i]);
-        if (has_colors) output->vertex_colors_.push_back(
-                input.vertex_colors_[i]);
+    for (int i=0; i<mask_observed_vertex.size(); i++) {
+        if (mask_observed_vertex[i]) {
+            output->vertices_.push_back(input.vertices_[i]);
+            if (has_normals) output->vertex_normals_.push_back(
+                    input.vertex_normals_[i]);
+            if (has_colors) output->vertex_colors_.push_back(
+                    input.vertex_colors_[i]);
+        }
     }
     PrintDebug("Triangle mesh sampled from %d vertices and %d triangles to %d vertices and %d triangles.\n",
             (int)input.vertices_.size(), (int)input.triangles_.size(),
--- a/src/Core/Geometry/DownSample.cpp
+++ b/src/Core/Geometry/DownSample.cpp
@@ -110,42 +110,44 @@ std::shared_ptr<TriangleMesh> SelectDownSample(const TriangleMesh &input,
     auto output = std::make_shared<TriangleMesh>();
     bool has_normals = input.HasVertexNormals();
     bool has_colors = input.HasVertexColors();
-    // For each vertex, list what faces are observed. use vector
-    std::vector<std::vector<int>> vertex_to_face_temp(input.vertices_.size());
+    // For each vertex, list face indices.
+    std::vector<std::vector<int>> vertex_to_triangle_temp(input.vertices_.size());
     int triangle_id = 0;
     for (auto trangle : input.triangles_) {
         for (int i=0; i<3; i++)
-            vertex_to_face_temp[trangle(i)].push_back(triangle_id);
+            vertex_to_triangle_temp[trangle(i)].push_back(triangle_id);
         triangle_id++;
     }
-    // for each vertex, remove list if it is non-selected vertices
-    std::vector<std::vector<int>> vertex_to_face(input.vertices_.size());
+    // Remove face indices of vertex_to_triangle_temp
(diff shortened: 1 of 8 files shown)


src/Core/Geometry/DownSample.cpp
```diff
@@ -103,6 +104,74 @@ std::shared_ptr<PointCloud> SelectDownSample(const PointCloud &input,
...
+                    input.vertex_normals_[i]);
+            if (has_colors) output->vertex_colors_.push_back(
+                    input.vertex_colors_[i]);
+        }
+    }
+    PrintDebug("Triangle mesh sampled from %d vertices and %d triangles to %d vertices and %d triangles.\n",
+            (int)input.vertices_.size(), (int)input.triangles_.size(),
+            (int)output->vertices_.size(), (int)output->triangles_.size());
+    return output;
+}
```
syncle: This function is bit long. I have tried to figure out the most efficient way to implement an algorithm that picks faces having three selected points. Maybe the names of variables are not very concise.

syncle: I am not very sure this function can be placed here. In DownSample.cpp, most of the functions are point cloud operations.

src/Tools/CMakeLists.txt
```diff
@@ -13,6 +13,6 @@ endmacro(TOOL)
 
 TOOL(ConvertPointCloud      ${CMAKE_PROJECT_NAME})
 TOOL(EncodeShader)
-TOOL(ManuallyCropPointCloud ${CMAKE_PROJECT_NAME})
+TOOL(ManuallyCropGeometry   ${CMAKE_PROJECT_NAME})
```
syncle: I changed the application name. Do we have any dependent projects regarding this program names?

src/Visualization/Utility/SelectionPolygon.cpp
```diff
@@ -182,6 +200,18 @@ std::shared_ptr<PointCloud> SelectionPolygon::CropPointCloudInPolygon(
...
+        const TriangleMesh &input, const ViewControl &view)
+{
+    return SelectDownSample(input, CropInRectangle(input.vertices_, view));
+}
+
+std::shared_ptr<TriangleMesh> SelectionPolygon::CropTriangleMeshInPolygon(
+        const TriangleMesh &input, const ViewControl &view)
+{
+    return SelectDownSample(input, CropInPolygon(input.vertices_, view));
+}
```
syncle: Can we use template functions?

src/Visualization/Utility/SelectionPolygon.h
```diff
@@ -71,6 +74,10 @@ class SelectionPolygon : public Geometry2D
             const PointCloud &input, const ViewControl &view);
     std::shared_ptr<PointCloud> CropPointCloudInPolygon(
             const PointCloud &input, const ViewControl &view);
+    std::shared_ptr<TriangleMesh> CropTriangleMeshInRectangle(
+            const TriangleMesh &input, const ViewControl &view);
+    std::shared_ptr<TriangleMesh> CropTriangleMeshInPolygon(
+            const TriangleMesh &input, const ViewControl &view);
```
syncle: Same as previous comment, maybe template functions.

src/Visualization/Visualizer/VisualizerWithEditing.cpp
```diff
@@ -392,6 +393,33 @@ void VisualizerWithEditing::KeyPressCallback(GLFWwindow *window,
...
+                    filename = default_filename.c_str();
+                }
+                if (filename == NULL) {
+                    PrintInfo("No filename is given. Abort saving.\n");
+                } else {
+                    SaveCroppingResult(filename);
+                }
+                view_control.ToggleLocking();
+                InvalidateSelectionPolygon();
+                InvalidatePicking();
```
syncle: I want to reduce these lines. This is too redundant and quite similar to previous lines in if statement. Maybe define template functions that does this job.

src/Tools/CMakeLists.txt
qianyizh: I think this is fine.

src/Visualization/Utility/SelectionPolygon.cpp
qianyizh: Yes, as long as the implementation gets declared in .cpp files and does not leak into .h.

src/Visualization/Utility/SelectionPolygon.h
qianyizh: I think it is fine. Let's leave it for now.
Plus, I can't think of any other use cases. These two are all that we need.

src/Visualization/Visualizer/VisualizerWithEditing.cpp
qianyizh: Up to you.

src/Core/Geometry/DownSample.cpp
```diff
@@ -103,6 +104,74 @@ std::shared_ptr<PointCloud> SelectDownSample(const PointCloud &input,
...
+    // Only a face with three selected points contributes to mark mask_observed_vertex.
+    std::vector<bool> mask_observed_vertex(input.vertices_.size());
+    for (auto vertex_ids : triangle_to_vertex) {
+        if ((int)vertex_ids.size() == 3)
+            for (int i=0; i<3; i++)
+                mask_observed_vertex[vertex_ids[i]] = true;
+    }
+    // Rename vertex id based on selected points
+    std::vector<int> new_vertex_id(input.vertices_.size());
+    for (int i=0, cnt=0; i<mask_observed_vertex.size(); i++) {
```
qianyizh: VS does not throw out warnings?
I think you should use ``i < (int)mask_observed_vertex.size()``, or ``auto i``

src/Core/Geometry/DownSample.cpp
```diff
@@ -103,6 +104,74 @@ std::shared_ptr<PointCloud> SelectDownSample(const PointCloud &input,
...
+        if ((int)vertex_ids.size() == 3) {
+            Eigen::Vector3i new_face;
+            for (int i=0; i<3; i++)
+                new_face(i) = new_vertex_id[input.triangles_[triangle_id][i]];
+            output->triangles_.push_back(new_face);
+        }
+        triangle_id++;
+    }
+    // Push marked vertex.
+    for (int i=0; i<mask_observed_vertex.size(); i++) {
```
qianyizh: Same comment as above.

src/Core/Geometry/DownSample.cpp
```diff
@@ -103,6 +104,74 @@ std::shared_ptr<PointCloud> SelectDownSample(const PointCloud &input,
...
+    }
+    // Push a triangle that has 3 selected vertices.
+    triangle_id = 0;
+    for (auto vertex_ids : triangle_to_vertex) {
+        if ((int)vertex_ids.size() == 3) {
+            Eigen::Vector3i new_face;
+            for (int i=0; i<3; i++)
+                new_face(i) = new_vertex_id[input.triangles_[triangle_id][i]];
+            output->triangles_.push_back(new_face);
+        }
```
qianyizh: triangle_normals_?

src/Visualization/Utility/SelectionPolygon.cpp
```diff
@@ -127,6 +128,23 @@ std::shared_ptr<PointCloud> SelectionPolygon::CropPointCloud(
     }
 }
 
+std::shared_ptr<TriangleMesh> SelectionPolygon::CropTriangleMesh(
```
qianyizh: Should these functions be exposed in Python too?

src/Core/Geometry/DownSample.cpp
```diff
@@ -103,6 +104,74 @@ std::shared_ptr<PointCloud> SelectDownSample(const PointCloud &input,
     return output;
 }
 
+std::shared_ptr<TriangleMesh> SelectDownSample(const TriangleMesh &input,
```
qianyizh: should this be exposed in python too?

syncle:
Thanks @qianyizh. I will revisit this once the other PR is addressed.

src/Visualization/Utility/SelectionPolygon.cpp
syncle: done.

src/Core/Geometry/DownSample.cpp
syncle: done.

qianyizh:
This is great.
I think this document should be edited too.
Add a note or something saying that meshes are also supported.
http://www.open3d.org/docs/tutorial/Basic/pointcloud.html#crop-point-cloud
But this is a minor fix and can be done later.
I will merge it.

Issue #411, opened 2018-06-26
point-to-plane ICP registration


daegeumer:
Some dataset(for example Microsoft 7-scenes dataset) doesn't perform any ICP iteration using point-to-plane option. Strangely, point-to-point option works well for the same data set and the tutorial dataset also works well using point-to-plane option. I don't understand what is the problem. When the ICP algorithm doesn't converge, the result transformation of point-to-plane algorithm is always the Identity regardless of option(threshold, trans_init, relative_fitness and so on). Microsoft 7-scenes dataset consists of two png files - depth.png and rgb.png. I tried to combine two pngs to rgbd image like tutorial and I also tried to use only depth png.
I installed open3d from source a few weeks ago in Ubuntu 16.04

syncle:
HI @daegeumer. Can you check you had computed point cloud normal before applying point to plane ICP?

qianyizh:
@syncle we have seen similar questions recently. Maybe we should add a warning to point-to-plane ICP when there is no normal attach. (The current behavior is to fail quietly?)

Also, I think I am seeing more and more use cases directly using depth images as input, maybe we should support normal computation / bilateral filtering / sampling directly for depth images too.

syncle:
Good point. I would need to follow-up with @daegeumer, if it is occurred by such problem. Regarding the depth image processing, let's open a new issue.

rukie:
Personal experience: Orienting normals towards camera location significantly improves normals results (after estimating), thus improving point-to-plane registration. I typically review normals in Meshlab, as it easily lets you scale the size of normals for visualization.

qianyizh:
FYI, you can review normals in Open3D visualizer too.
Just press ``N`` and the normals will be shown. The length of the normals can be changed with ``[`` and ``]``.

syncle:
Hi @daegeumer, can you confirm the answers helped to resolve the issue?

daegeumer:
oh, I'm sorry. Do I have to close the issue? Thank you for your comments. These comments help me to solve the problem. I'll close the issue right now. Thanks!

syncle:
No you don't have to. (I sometimes close inactive threads to focus on opening issues). I just wanted to ask whether the issue is resolved.

daegeumer:
Ok! Thank you for your concern!👍

PR #419, opened 2018-06-28
Point to plane ICP error message

syncle:
This PR related to #411.

Key changes:
- Define TransformationEstimationType
- Adding TransformationEstimationType variable in TransformationEstimation class
- RegistrationICP terminates early with error message like below if there is no normal direction is given
```
Error: TransformationEstimationPointToPlane requires pre-computed normal vectors.
```

### DIFF
--- a/src/Core/Registration/ColoredICP.cpp
+++ b/src/Core/Registration/ColoredICP.cpp
@@ -46,6 +46,8 @@ public:
 
 class TransformationEstimationForColoredICP : public TransformationEstimation {
 public:
+    TransformationEstimationType GetTransformationEstimationType()
+            const override { return type_; };
     TransformationEstimationForColoredICP(
             double lambda_geometric = 0.968) :
             lambda_geometric_(lambda_geometric) {
@@ -63,6 +65,8 @@ public:
 
 public:
     double lambda_geometric_;
+    const TransformationEstimationType type_ =
+            TransformationEstimationType::ColoredICP;
 };
 
 std::shared_ptr<PointCloudForColoredICP>
--- a/src/Core/Registration/ColoredICP.cpp
+++ b/src/Core/Registration/ColoredICP.cpp
@@ -65,6 +65,8 @@ public:
 
 public:
     double lambda_geometric_;
+
+private:
     const TransformationEstimationType type_ =
             TransformationEstimationType::ColoredICP;
 };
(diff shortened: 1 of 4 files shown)


src/Core/Registration/ColoredICP.cpp
```diff
@@ -63,6 +65,10 @@ class TransformationEstimationForColoredICP : public TransformationEstimation {
 
 public:
     double lambda_geometric_;
+
+private:
+    const TransformationEstimationType type_ =
+            TransformationEstimationType::ColoredICP;
```
syncle: I am not very sure whether this private variable is necessary for every child class of TransformationEstimation. The same concern applies to TransformationEstimationPointToPoint and TransformationEstimationPointToPlane.

src/Core/Registration/Registration.cpp
```diff
@@ -146,8 +146,16 @@ RegistrationResult RegistrationICP(const PointCloud &source,
         const ICPConvergenceCriteria &criteria/* = ICPConvergenceCriteria()*/)
 {
     if (max_correspondence_distance <= 0.0) {
+        PrintError("Error: Invalid max_correspondence_distance.\n");
```
syncle: Do we have some standard form of error message? In this example, it is just ``error: [message]``

qianyizh: I usually use "[function_name] some_message".

src/Core/Registration/ColoredICP.cpp
qianyizh: I think this is fine. Usually, it is not recommended to let the subclasses keep the knowledge of which class it is. But here we don't have other options (or we have to let ComputeEstimation() yell every time it gets called).

And we do this to Geometry class too. Every subclass has a type_ which indicates which type it is. So I think it is fine.

The downside though, is that it will be hard to extend - say, someone wants to write a customized TransformationEstimation using Python. Then there might be other problems. Anyways, we will worry about it later.

qianyizh:
Looks good. I will merge.

src/Core/Registration/ColoredICP.cpp
yangece: In fact, we do want to write a customized Python subclass of the TransformationEstimation class. For example, 
"class TransformationEsimationTranslation(TransformationEstimation):
      def ...
"
I wonder if we can do that in Python, or we also have to change the pybind11 interfaces? Thanks!

qianyizh: I did it before in C++. It is doable.

However I don't think it is doable in Python in its current way (or at least I have not tested).

There are performance consideration though. These functions are called extensively and you definitely want them to be written in native C++ in production code.

yangece: It makes sense to do it in C++. I wonder if it is straightforward for developers to change Open3D  C++ code to do translation-only ICP. Or for this particular case, we have to change some of the Eigen functions, like the Umeyama implementation. Thanks!

Issue #422, opened 2018-06-29
Visualizer::CreateWindow confliction problem


dotchang:
On windows MSVC environment, the name of CreateWindow is defined by windows.h which is basic library by microsoft. (I think it's terrible, but...)
When I tried to use Open3D and other windows software, it makes name confliction problem.

Could you please consider changing the name?

syncle:
I may misunderstood the point, but Visualizer::CreateWindow() is the function of Visualizer class. Is this naming conflict making some technical issue (such as compilation error)?

dotchang:
WinUser.h which is a part of WindowsSDK10 of the microsoft gives a global definition as follows, 
(line 4283)
#ifdef UNICODE
#define CreateWindow  CreateWindowW
#else
#define CreateWindow  CreateWindowA
#endif // !UNICODE

Then, it makes problems. Do you have good solutions?

qianyizh:
Did this happen when you are compiling Open3D or when you are compiling an external C++ program that is linked to Open3D?

dotchang:
It happened when I tried to use external C++ program.
I want to use Open3D with openFrameworks.

qianyizh:
Umm, I think I understand this now.

I remember a long time ago ``three::filesystem::MakeDirectory()`` was named ``three::filesystem::CreateDirectory()``, then it conflicts with Windows' ``CreateDirectory`` which is a macro mask of ``CreateDirectoryA`` or ``CreateDirectoryW`` (yes, terrible design, but...). I looked for workarounds and eventually decided to change the function name.
https://github.com/IntelVCL/Open3D/blob/dd511a0d3f973735d15ae9b90e292077886f1e1b/src/Core/Utility/FileSystem.h#L52

I think ``three::Visualizer::CreateWindow`` is the same case. Let me hot fix it.

qianyizh:
@syncle to explain it a bit more, the problem is with the macro definition. C++ macro overrides everything. So once a windows header file is included the macro overrides all the valid function names. In a cpp file that uses ``three::filesystem::CreateDirectory()``, it is translated into ``three::filesystem::CreateDirectoryA()`` or ``three::filesystem::CreateDirectoryW()``, which breaks the build.

syncle:
Addressed in #433.

PR #433, opened 2018-07-05
fix name conflict of CreateWindow()

qianyizh:
This addresses #422 

I also changed the name of ``DestroyWindow()`` so that they are consistent.

### DIFF
--- a/src/Python/Visualization/open3d_visualizer.cpp
+++ b/src/Python/Visualization/open3d_visualizer.cpp
@@ -43,11 +43,11 @@ void pybind_visualizer(py::module &m)
         .def("__repr__", [](const Visualizer &vis) {
             return std::string("Visualizer with name ") + vis.GetWindowName();
         })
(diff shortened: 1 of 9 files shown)


src/Python/Visualization/open3d_visualizer.cpp
```diff
@@ -43,11 +43,11 @@ void pybind_visualizer(py::module &m)
         .def("__repr__", [](const Visualizer &vis) {
             return std::string("Visualizer with name ") + vis.GetWindowName();
         })
-        .def("create_window", &Visualizer::CreateWindow,
+        .def("create_window", &Visualizer::CreateVisualizerWindow,
```
syncle: Shouldn't we also need to change ``create_windows`` as well? Every Python function names are lower case + under bar style of original c++ function name.

qianyizh: I thought about it but decided against it. This would look like a weird name in Python interface.

PR #434, opened 2018-07-05
Adding centered version file

### LINKED ISSUE #428: Automatically update version in docs
When we generate the project files with CMake the Open3D version in `src/CMakeLists.txt` gets updated in the C++ code and `setup.py` script used to package Open3D for pypi.

Let's update the Open3D version throughout the documentation using the same mechanism.

In summary, what I'm proposing is that we specify the Open3D version (and maybe a few other things) in one central location and have it updated throughout the project automatically. The current mechanism I've used is CMake.

syncle:
- adding version.txt
- src/CMakeList.txt and docs/conf.py automatically refers version.txt

Resolves #428.


takanokage:
IMHO the proper way to do it was through CMake generate without adding the two new files.

syncle:
It just adds single file (version.txt) BTW. Feel free to submit PR if it is necessary :)

PR #450, opened 2018-07-10
Extend tsdf input

### LINKED ISSUE #410: [Question] denoising point cloud and mesh reconstruction
Hi guys,

After solving the issue with the linking in the mex script. I am able to load and visualize my point using the following script:

```
#include "mex.h"

// cpp system headers
#define _USE_MATH_DEFINES
#include <cmath>
#include <iostream>
#include <vector>
#include <fstream>
#include <memory>

// open3D headers
#include <Eigen/Dense>
#include <Core/Core.h>
#include <IO/IO.h>
#include <Visualization/Visualization.h>

using namespace Eigen;
using namespace std;
using namespace three;

template<typename... Args>       
void vprintf(const char *format, Args... args) {
    mexPrintf(format, args...);
}
template<typename... Args>
void PrintInfo(const char *format, Args... args) {
    vprintf(format, args...);
}
template<typename... Args>
void PrintDebug(const char *format, Args... args) {
    vprintf(format, args...);
}

void PrintPointCloud(const three::PointCloud &pointcloud)
{
// 	using namespace three;
	bool pointcloud_has_normal = pointcloud.HasNormals();
	PrintInfo("Pointcloud has %d points.\n",
			(int)pointcloud.points_.size());

	Eigen::Vector3d min_bound = pointcloud.GetMinBound();
	Eigen::Vector3d max_bound = pointcloud.GetMaxBound();
	PrintInfo("Bounding box is: (%.4f, %.4f, %.4f) - (%.4f, %.4f, %.4f)\n",
			min_bound(0), min_bound(1), min_bound(2),
			max_bound(0), max_bound(1), max_bound(2));

// 	for (size_t i = 0; i < pointcloud.points_.size(); i++) {
// 		if (pointcloud_has_normal) {
// 			const Eigen::Vector3d &point = pointcloud.points_[i];
// 			const Eigen::Vector3d &normal = pointcloud.normals_[i];
// 			PrintDebug("%.6f %.6f %.6f %.6f %.6f %.6f\n",
// 					point(0), point(1), point(2),
// 					normal(0), normal(1), normal(2));
// 		} else {
// 			const Eigen::Vector3d &point = pointcloud.points_[i];
// 			PrintDebug("%.6f %.6f %.6f\n", point(0), point(1), point(2));
// 		}
// 	}
	mexPrintf("End of the list.\n\n");
}


/* The gateway function */
void mexFunction(
     int          nlhs,
     mxArray      *plhs[],
     int          nrhs,
     const mxArray *prhs[]
     )
{

      /* Check for proper number of arguments */
      
      if (nrhs != 3) 
      {
        mexErrMsgIdAndTxt("MATLAB:mexcpp:nargin",
            "pointCloud2Mesh requires 3 input arguments, the pointCloud points, the corresponding color and the albedo map!!!");
      }
  
      /* check that number of columns in first and second input argument is 3 */
      if (mxGetN(prhs[0]) != 3) {
          mexErrMsgIdAndTxt("MyToolbox:arrayProduct:not3Columns", "Point cloud input must have 3 columns.");
      }
      if (mxGetN(prhs[1]) != 3) {
          mexErrMsgIdAndTxt("MyToolbox:arrayProduct:not3Columns", "Normals must have 3 columns.");
      }
      
      /* check that number of columns in the third input argument is 1 */
      if (mxGetN(prhs[2]) != 1) {
          mexErrMsgIdAndTxt("MyToolbox:arrayProduct:not3Columns", "Albedo map must have 1 column.");
      }
      
      SetVerbosityLevel(VerbosityLevel::VerboseAlways);
      
      /* Get sizes */
      mwSize nRows = mxGetM(prhs[0]);
      mwSize nCols = mxGetN(prhs[0]);
      
      /* Read input data  */
      MatrixXd pnts = Map<MatrixXd>( mxGetPr(prhs[0]), nRows, nCols );
      MatrixXd pntsColor = Map<MatrixXd>( mxGetPr(prhs[1]), nRows, nCols );
      double* pntsAlbedo = mxGetPr(prhs[2]); 
      
      PointCloud pointcloud;
      for(int i = 0; i < nRows; i++){
          pointcloud.points_.push_back(Eigen::Vector3d(pnts(i,0), pnts(i,1), pnts(i,2)));
          pointcloud.colors_.push_back(Eigen::Vector3d(pntsColor(i,0), pntsColor(i,1), pntsColor(i,2))); // values need to be between 0-1
      }
      
      PrintPointCloud(pointcloud);
      
      std::shared_ptr<PointCloud> pointcloud_ptr(new PointCloud);
	  *pointcloud_ptr = pointcloud;
      
      // normal estimation
      EstimateNormals(*pointcloud_ptr, three::KDTreeSearchParamRadius(0.05));
      mexPrintf("Normals estimated!!!\n\n");
      
      OrientNormalsToAlignWithDirection(*pointcloud_ptr, Eigen::Vector3d(0.0, 0.0, 1.0));
      
      DrawGeometries({pointcloud_ptr}, "Open3D PointCloud", 1920, 1080);
      
//       Visualizer visualizer;
// 	  std::shared_ptr<PointCloud> pointcloud_ptr(new PointCloud);
// 	  *pointcloud_ptr = pointcloud;
// 	  pointcloud_ptr->NormalizeNormals();
// 	  BoundingBox bounding_box;
// 	  bounding_box.FitInGeometry(*pointcloud_ptr);
//       
// //       // test normal estimation
// //       EstimateNormals(*pointcloud_ptr, three::KDTreeSearchParamRadius(0.05));
// 
// 	  visualizer.CreateWindow("Open3D", 1600, 900);
// 	  visualizer.AddGeometry(pointcloud_ptr);
// 	  visualizer.Run();
// 	  visualizer.DestroyWindow(); 
      mexPrintf("Exiting.\n");
}
```

where I am passing the points and the corresponding color and albedo values for each point (extracted with kinect v2). Then I am assigning the points and the color to the Open3D PointCloud structure and I am able to show the point cloud in the visualizer after I have estimated and oriented the normals. Now I would like to create the mesh from the point cloud but before that, if you notice the point cloud there are some noisy points that I would like to get rid of, see the image below:

image

Thus, I would like to ask if there is a corresponding function that I can call for this. I couldn't find something relevant at least to the files that I've looked in, though I might have missed it. Another idea that came to my mind is whether down sampling could be used for this purpose or not?

Second question is, how I can reconstruct the mesh out of the point cloud, is there any C++ tutorial that I can follow? I've seen this discussion here [https://github.com/IntelVCL/Open3D/issues/270](url) however I am not sure if the answer there still stands. Also is it possible that I use TSDF for this since I do have the depth and rgb images as well as the intrinsic information of the camera.

and last question as I said I am also passing the abledo value of each point thus is it possible after I create the mesh to assign to each created face a corresponding albedo value as an interpolation of the albedo values from the vertices that create the triangle or any other way?

Thanks.

You can find an example point cloud here if that helps: [ptCloud1.zip](https://github.com/IntelVCL/Open3D/files/2133455/ptCloud1.zip)

syncle:
Resolves #410.

- Float type intensity image + depth image can integrate TSDF volume. 
- Before this PR, only 8bit color image + float type depth image were only able to be used for TSDF integration. This is irrational since we can make point cloud or RGBD image using float type intensity image + depth image.

Tested with several test sets.


src/Core/Integration/TSDFVolume.h
```diff
@@ -33,6 +33,12 @@
 
 namespace three {
 
+enum class TSDFVolumeColorType {
+    None = 0,
+    RGB8Bit = 1,
+    Gray32Bit = 2,
```
syncle: I am not very confident about this names. You can suggest any.

TheodoreT: @syncle maybe you could adapt something like similar to the opencv's `Mat` types e.g. `CV_64FC1`, `CV_8UC3`, etc... here is a list http://ninghang.blogspot.com/2012/11/list-of-mat-type-in-opencv.html . Thus in the current case you could have something like `3D_8UC3` (or `3D_8U` without the channel info) and `3D_32FC1` (`3D_32F`)  or if this is getting too specific considering that in your case you have only two cases to worry about, you could go for something like `RGB8` (or with small `rgb8`) and equivalently `RGB32` (or `rgb32`) or `GRAY32` (or `gray32`).

syncle: Good suggestion. How do you think? @qianyizh?

qianyizh: I think it is a good suggestion. Can you do it in this PR?

syncle: Sure. Will do.

qianyizh:
I actually think the naming can be more general and should not be limited to ``TSDFVolumeColorType``. Maybe make it an enum of ``Image::ColorType``? We will definitely have more good uses of it in the future.

syncle:
I agree. For generalization, ``Image::ColorType``need to include depth images as well. In that sense, ``RGB8`` or ``GRAY32`` may not a good terminology as the float type depth image is not classified in the both.  

This change will involve touching many of functions in Image class and RGBDImage class that simply utilizes ``num_of_channels_`` and ``bytes_per_channel_``. These variables are in public domain. Having ``num_of_channels_``, ``bytes_per_channel_``, and ``Image::ColorType`` together in public domain would be redundant. 

With considering these, how about changing it this way?
- add enum for ``Image::Type`` instead of ``Image::ColorType``.
- ``Image::Type`` will have ``8UC3`` and ``32FC1`` instead of ``RGB8`` or ``GRAY32``, in order to correctly express depth images. We may define other types too.
- ``Image::Type`` also will have ``Unspecified`` type likewise how we did in ``Geometry`` class.
- make ``num_of_channels_`` and ``bytes_per_channel_`` is in private domain. Once ``Image::Type`` is specified, It is just used in internally, and add interface function to retrieve the value.
- Remove ``TSDFVolumeColorType``. ``TSDFVolume`` does not need to be initialized with specific image format. ``TSDFVolume`` just takes either Images of will have ``8UC3`` and ``32FC1`` for integration. Other type of images are rejected.
- Maybe we need to open another PR for this? (don't merge this PR, but new PR will be build from this PR to avoid confusion for Open3D users)

There must be some weakness or exceptional cases. @qianyizh: please share your thoughts on this suggestion.

qianyizh:
All sounds good to me. Except I think we should merge this PR for now and open a new ticket for the next change as you described. Is that fine?

syncle:
OK I am fine with that.

PR #453, opened 2018-07-12
Update TestRealSense.cpp

TheodoreT:
fixing the recent change of the CreateWindow to CreateVisualizerWindow related to IntelVCL#422 in the TestRealSense.cpp file which leads to an error in the compilation if build tests is enabled


syncle:
Good catch. Thanks @TheodoreT! Can you also change this line as well? 
https://github.com/IntelVCL/Open3D/blob/bd6e7170134f1ed974ddcedd20fb7761408ef10f/src/Visualization/Visualizer/Visualizer.h#L104
I will merge if you fix this too. 👍

TheodoreT:
done ;-)

PR #464, opened 2018-07-18
Optimisation of colored point cloud registration code and doc.

baptiste-mnh:
The Step 2 had no effect on step 3.


syncle:
Hi @baptiste-mnh, thanks for this pull request, but the step 2 and step 3 should be independent. As mentioned in the document, step 2 is presented as a baseline, just to show that failure case of conventional ICP. Step 3 independently runs without result of Step 2.

argosvr:
We made different tests aligning several cameras and without linking step 2 and 3 it was impossible to get a correct result. Linking 2 and 3 gave the result expected.

baptiste-mnh:
Ok, I''ll publish an issue with a dataset as soon as possible.

PR #466, opened 2018-07-22
Orient normals based on MST

ppinies:
@germanros1987 I have added an algorithm and its unit test to orient normals using a MST as requested in https://github.com/IntelVCL/Open3D/issues/270.

Note that given an intial point to create a MST that connects the point cloud passed as input, KNN can  just connect a subset of the original points invalidating the assumption in section 3.3 in http://hhoppe.com/recon.pdf, i.e., "the surface is assumed to consist of a single connected component".


syncle:
Hi @germanros1987, Could I ask why this PR is was closed without merge or review? This feature is good to me though.

ytyfeng:
This feature looks great. How do I use it in Python?

PR #499, opened 2018-08-09
Fix bug for Xcode IDE

### LINKED ISSUE #401: Fix Xcode build.
There seems to be a CMake bug manifested when using object libraries together with Xcode.
The solution might not be as simple as not using object libraries because this is the mechanism that I was hoping to use to fix #393.

[supporting link1](https://cmake.org/pipermail/cmake/2016-April/063169.html)
[supporting link2](https://cmake.org/Bug/view.php?id=14768)
[supporting link3](https://github.com/tensorflow/tensorflow/issues/15369)
[supporting link4](https://gitlab.kitware.com/cmake/cmake/issues/17500)

syncle:
This resolves #401.

---

**Root cause:**
- ``libOpen3D.a`` is made from several cmake ``OBJECT`` libraries. Some IDE (such as Xcode) does not produce ``libOpen3D.a`` if there is no real file in ``add_library``.
- Reference: https://cmake.org/cmake/help/v3.12/command/add_library.html#object-libraries

---

**Change summary:**
- Adding ``Open3DConfig.cpp`` and include this file into ``add_library(Open3D)`` to add a real file.
- ``Open3DConfig.cpp`` has ``PrintOpen3DVersion()`` in ``three`` namespace. It just prints open3d version.
- replacing manual print functions with ``PrintOpen3DVersion()``.
- replacing ``printf`` with ``PrintInfo``.
- removing ``#include "../open3d_config.h"`` in ``IO/IO.h`` and ``Visualization/Visualization.h`` as they are redundant.
- Changing name ``open3d_config.h`` -> ``Open3DConfig.h`` to comply with file naming rule.

---

**Notes:**
- This PR modifies several CMakeList.txt files. Careful review is required. (Especially pip package build) @takanokage @qianyizh, please review.

---
This change is [Reviewable](https://reviewable.io/reviews/intelvcl/open3d/499)


qianyizh:
I mean we should include them right?
For example if a Visualizer class wants to access the version of Open3D, the config file needs to be included in Visualization.h. Right?

syncle:
You're right. Just fixed.

PR #501, opened 2018-08-10
Fix test data path in scripts under src/Python/Tutorial/Advanced.

bing-jian:
When running those Python scripts under src/Python/Tutorial/Advanced at their own directory,  the top-level test data path should not be "../../TestData/",  instead it should be changed to "../../../Test/TestData"

### DIFF
--- a/src/Python/Tutorial/Advanced/camera_trajectory.py
+++ b/src/Python/Tutorial/Advanced/camera_trajectory.py
@@ -2,6 +2,7 @@
 # The MIT License (MIT)
 # See license file or visit www.open3d.org for details
 
+import os
 import numpy as np
 from open3d import *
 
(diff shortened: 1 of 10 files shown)


syncle:
Hi @bing-jian. Thanks for the contribution!

Actually, the paths in Tutorial scripts are intended to be used from build folder. This is described in [getting started page](http://open3d.org/docs/getting_started.html) (please notify if the document was not clear). If this PR is accepted, running python tutorials from build folder will not work. 

However, you have a good point. We may need to reorganize folder structure to avoid confusion.

qianyizh:
@syncle I assume you will follow up on this.

bing-jian:
@syncle ,   Thanks for the quick response and pointing me to the Getting Started page which I did not read carefully.   My understanding now is that these tutorial scripts are supposed to run from build/lib/Tutorial. If that's the case, then as you said, the paths in current scripts are correct. However, I still feel this assumption is not intuitive, at least for me.  I usually think the build folder are mostly for temporary files and can be safely deleted after successful installation. On the other hand, the source code will be kept as that's the place to dig into the implementation.  So for me, it is more natural to run the script under the src/...  instead of finding and running code under the build/lib/....   Just my two cents.

BTW,  Open3D is such a great library for working with RGB-D and point cloud data.  Thank you and @qianyizhi so much for creating Open3D,  truly appreciate all the efforts your guys put into this.

PR #515, opened 2018-08-23
Image & ImageFactory unit tests

takanokage:
Added unit tests for the following test cases:
- Image
- ImageFactory

In total I have added 32 unit tests.
Of these 2 are currently failing and 30 are passing.

I have also added all the remaining test cases and templates for their corresponding unit tests.
In total there are 87 test cases but I have grouped some of them and we're left with about 80 test cases (Image and ImageFactory are just one test case for example).
Total unit tests: 792 disabled and 32 active.
I have marked the incomplete unit tests with the prefix DISABLED_ in order to have them skipped when running the unit tests. Please remove the DISABLED_ prefix when adding code to a unit test. This is a good (but not perfect) way to keep track of our progress in implementing the unit tests.

NOTE:
Please note that in order to add these tests in a timely manner I have taken the "black box" approach for testing. That is, rather then implementing test for any possible inputs to the units in question, I have created specific inputs and their expected outputs for each unit. These tests are not proof. When optimizing the code or implementing for specific accelerators these tests may fail to capture flaws in the code.

---
This change is [Reviewable](https://reviewable.io/reviews/intelvcl/open3d/515)


takanokage:
please note

takanokage:
I need to perform a few more checks.

PR #529, opened 2018-08-31
fix PinholeCameraIntrinsic initialization

cfwen:
in python color_map_optimization example

---
This change is [Reviewable](https://reviewable.io/reviews/intelvcl/open3d/529)


syncle:
Great catch! Thanks @cfwen!

PR #539, opened 2018-09-05
PointCloud

takanokage:
NOTE: this should be merged after the Image test case. It is essentially a branch to the work on the Image test case.

Main contribution: the PointCloud test case.

Additional contributions:
- fix some bugs in the Image test case that are only visible when another test case, namely PointCloud, is run before the Image test case 
- added more utilities to UnitTest.h/.cpp like Rand(...) and Print(...) that help with the implementation of the unit tests
- added the UnitTest namespace to UnitTest.h/.cpp
- removed duplicate unit tests

---
This change is [Reviewable](https://reviewable.io/reviews/intelvcl/open3d/539)


qianyizh:
Sorry for hopping into the discussion late, been v/ busy...

Re black box testing: my overall feeling is that we need **both** explicit reasoning test and black box testing. Actually we need three types of tests. They have different purposes:

 * Explicit reasoning tests: check the logic of the code (e.g., @yxlao 's example), cover corner cases (e.g., dump in an invalid input, or all 0 input, or possibly-out-of-memory input, etc). It has better readability. Future engineers should be able to understand it. It has more restrict coding style and documentation requirements too.
 * Black-box tests: prevent future code from breaking the current functionalities. The basic assumption is that the current code is correct, and we build test upon current code. The tests serve as a guard so that that future code change won't break current functionalities.
 * Bug-fixing tests: every bug should have a failing unit-test. Future changes should pass these tests - so that the old bugs won't come out again.

IMO, all these tests are useful. But since we are in the v/ early stage of adding tests. I am fine with adding them bit by bit. I think the main type of tests being added here (and in previous PR) are black-box testing. @takanokage also wrote some explicit reasoning tests.

Overall this is good progress. And I can't wait to see #556 happens...

qianyizh:
I am also a big hater of random numbers since they are compiler related and execution order related. But I think we have a fix of it in #551 
So this PR is good to go.

takanokage:
Agreed on all the above.

Thank you!

takanokage:
I'll rebas the following PRs as soon as I get to a computer.

PR #551, opened 2018-09-12
Custom init to replace rand() init.

takanokage:
This PR addresses the concern that different compilers might generate different random number sequences.

---
This change is [Reviewable](https://reviewable.io/reviews/intelvcl/open3d/551)


takanokage:
Thanks!

PR #556, opened 2018-09-13
UnitTest integration into TravisCI

takanokage:
Integrated UnitTest into TravisCI.

Some modifications made to the unit tests in order to address some issues due to platform differences. Such an example is the fact that the vectors were ordered differently in TravisCI. To address this I sorted the results before comparing. std::sort doesn't give correct results in TravisCI/macos so I had to implement a very basic sort instead.

NOTE: if the order issues comes up in other tests we'll have to sort those data sets too. So far the tests are passing so maybe we won't need to sort everything.

---
This change is [Reviewable](https://reviewable.io/reviews/intelvcl/open3d/556)


qianyizh:
This is awesome...

syncle:
There are some warning messages arise from UnitTest. @takanokage: Please check https://travis-ci.org/IntelVCL/Open3D/jobs/437199232

takanokage:
Yes, I've noticed that sometimes the compiler can complain about this addressing mode:

v[0, 1] which is equivalent with v(1, 0).

I have already changed the addressing to use the v(1, 0) pattern in a later PR.

qianyizh:
Exciting moment!

PR #576, opened 2018-09-19
disable 3rd party static library installation

yxlao:
Fixes https://github.com/IntelVCL/Open3D/issues/569. There are 7 of such libraries.

Previously, with `-DBUILD_SHARED_LIBS=ON`, the following libs were installed:

```
-- Installing: /usr/local/lib/libglew.a
-- Installing: /usr/local/lib/libglfw3.a
-- Installing: /usr/local/lib/libjpeg.a
-- Installing: /usr/local/lib/libjsoncpp.a
-- Installing: /usr/local/lib/libzlib.a
-- Installing: /usr/local/lib/libpng.a
-- Installing: /usr/local/lib/libtinyfiledialogs.a
-- Installing: /usr/local/lib/libOpen3D.a
```

After the fix, they are no longer installed, but still in the `build` directory
```
➜  ~/repo/Open3D/build (disable-3rd-party-install) tree lib
lib
├── Python
│   └── open3d.so
├── libOpen3D.dylib
├── libglew.a
├── libglfw3.a
├── libjpeg.a
├── libjsoncpp.a
├── libpng.a
├── libtinyfiledialogs.a
└── libzlib.a

1 directory, 9 files
```

---
This change is [Reviewable](https://reviewable.io/reviews/intelvcl/open3d/576)


syncle:
@yxlao: Please address CI issues.

yxlao:
@syncle yup, the CI presents a more complicated issue than i originally anticipated. 

```
                                        +-----+
                                 +----->+dep.a|
                   +--------+    |      +-----+
             +---->+open3d.a+----+Both
             |     +--------+    |      +------+
 +-------+   |                   +----->+dep.so|
 |app.exe+---+One-of                    +------+
 +-------+   |                           +-----+
             |                    +----->+dep.a|
             |     +---------+    |      +-----+
             +---->+open3d.so+----+Both
                   +---------+    |      +------+
                                  +----->+dep.so|
                                         +------+
```

1. By default, `app.exe` have linker flag `-lglfw3 -lzlib -ltinyfiledialogs -lglfw3 -lzlib -ltinyfiledialogs`. If this is the case, we must install 3rd party static libraries. In order to avoid installation of 3rd party static libraries, we need to change the linker flag to only link `Open3D` (shall the linker flag also include all pre-installed share libs?). 
2. If we change `app.exe`'s linker flag to only link `Open3D`. Then we have two cases:
    - 2.1. If we build shared library `libOpen3D.so`. 
        - `libdep.a`: All `libdep.a` is embedded inside `libopen3d.so`. I got this case fully working if all the dependencies are compiled from source.
        - `libdep.so`: `libOpen3D.so` also links `libdep.so`. We need to make sure that `app.exe` can also resolve symbols in `libdep.so` even without specifying `-ldep` in the linker flag.
    - 2.2 If we build shared library `libOpen3D.a`. 
        - `libdep.a`: Then all `libdep.a` is **not** embedded inside `libopen3d.a`. We need to ask cmake to [create an archive](https://stackoverflow.com/a/14200728/1255535) to include all `libdep.a` into `libopen3d.a`. 
        - `libdep.so`: We also make sure that `app.exe` resolve symbols in `libdep.so` even without specifying `-ldep` in `app.exe`'s linker flag.

In summary, the complexity is due to 1) open3d can be shared or static 2) 3rd party dependencies can be shared or static. If we always build 3rd party libraries from source as static libraries, it would greatly reduce the complexity. 

I'll experiments on those combinations, and if you have some recommendations, please let me know as well.

yxlao:
Thanks @takanokage for the comments. Agreed on the suggestion to redesigning the linking system of our library.

My overall impression is that: we shall reduce combination of ways of how Open3D is built. Currently, 1) open3d can be shared or static 2) 3rd party dependencies can be shared or static and they can be pre-installed or built from source.

Adding a discussion point of my current solution in this PR:
```
If Open3D is a shared library
    - Built-from-source 3rd-party libraries is embedded in Open3D
    - Built-from-source 3rd-party libraries will NOT be installed
    - A user app needs to link
          1) Open3D
          2) Pre-installed 3rd-party libraries
If Open3D is a static library
    - Built-from-source 3rd-party libraries is NOT embedded in Open3D
    - Built-from-source 3rd-party libraries will be installed
    - A user app needs to link
          1) Open3D
          2) Pre-installed 3rd-party libraries
          3) Built-from-source 3rd-party libraries
```

qianyizh:
👍 the logic is much clearer and more readable.

PR #585, opened 2018-09-25
Adding CropTriangleMesh to SelectionPolygonVolume and pybindings

dbadrian:
AFAIK there is no dedicated way to crop a TriangleMesh using a SelectionPolygonVolume (as created with the visualizer) via the python bindings. 

This PR would resolves this by adding the equivalent of the crop_point_cloud method for a TriangleMesh to SelectionPolygonVolume and the pybindings.

It seems to do exactly what I want, but I can't tell if this might cause a problem somewhere else. So feel free to see this as a feature request with reference implementation.

---
This change is [Reviewable](https://reviewable.io/reviews/intelvcl/open3d/585)

### DIFF
--- a/src/Python/Visualization/open3d_utility.cpp
+++ b/src/Python/Visualization/open3d_utility.cpp
@@ -28,6 +28,7 @@
 
 #include <Core/Utility/FileSystem.h>
 #include <Core/Geometry/PointCloud.h>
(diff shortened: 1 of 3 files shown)


syncle:
This PR looks good to me. Thanks for the contribution @dbadrian!

PR #597, opened 2018-10-04
Loading & saving viewpoint

syncle:
Fixes #387 and #567.
- Currently saving single view point is not straightforward. For example, to save a current view as a PinholeCamera model, it requires to make a new instance of PinholeCameraTrajectory and needs manual assignment to ``PinholeCameraTrajectory.intrinsic`` and ``PinholeCameraTrajectory.extrinsic``. For example:
```
def save_view_point(pcd, filename):
    vis = Visualizer()
    vis.create_window()
    ctr = vis.get_view_control()
    vis.add_geometry(pcd)
    vis.run() # user changes the view and press "q" to terminate
    param = ctr.convert_to_pinhole_camera_parameters()
    trajectory = PinholeCameraTrajectory()
    trajectory.intrinsic = param[0]
    trajectory.extrinsic = Matrix4dVector([param[1]])
    write_pinhole_camera_trajectory(filename, trajectory)
    vis.destroy_window()

def load_view_point(pcd, filename):
    vis = Visualizer()
    vis.create_window()
    ctr = vis.get_view_control()
    trajectory = read_pinhole_camera_trajectory(filename)
    vis.add_geometry(pcd)
    ctr.convert_from_pinhole_camera_parameters(
            trajectory.intrinsic, trajectory.extrinsic[0])
    vis.run()
    vis.destroy_window()
```
- For this issue, I defined new class PinholeCameraParameters that can fully describe single view point. PinholeCameraTrajectory will have a array of PinholeCameraParameters so that the camera trajectory can have varying camera intrinsic parameters, such as zoom-in or zoom-out while changing the view point.
- With PinholeCameraParameters, saving and loading camera parameter is easy and straightforward. For example:
```
def save_view_point(pcd, filename):
    vis = Visualizer()
    vis.create_window()
    vis.add_geometry(pcd)
    vis.run() # user changes the view and press "q" to terminate
    param = vis.get_view_control().convert_to_pinhole_camera_parameters()
    write_pinhole_camera_parameters(filename, param)
    vis.destroy_window()

 def load_view_point(pcd, filename):
    vis = Visualizer()
    vis.create_window()
    ctr = vis.get_view_control()
    param = read_pinhole_camera_parameters(filename)
    vis.add_geometry(pcd)
    ctr.convert_from_pinhole_camera_parameters(param)
    vis.run()
    vis.destroy_window()
```
- I've tested some codes in example folder, but this PR may introduce some conflicts I don't aware. CAREFUL review is required.

---
[...]

PR #606, opened 2018-10-10
PinholeCameraIntrinsic

takanokage:
This change is [Reviewable](https://reviewable.io/reviews/intelvcl/open3d/606)


qianyizh:
@syncle can you take a look at this too? This PR seems to have a dependency on #597

PR #618, opened 2018-10-17
Automated make conda-package

yxlao:
## Changes
1. Added `make conda-package`. Tested on Ubuntu, Mac and Windows. 
2. Renamed pip package creation targets for consistency. 

## Summary of usage
In summary, we now have:

```bash
# C++ make
make                     # our regular make, creates build/lib/Python/open3d.so
make install             # does not install any python module

# python
make python-package      # creates python package in build/lib/python_package

# pip
make pip-package         # creates wheel in build/lib/python_package/pip_pacakge
make install-pip-package # installs wheel in virtualenv
make all-pip-packages    # creates all wheels in PyPI: "open3d-python, py3d, ..."

# conda
# Note 1: only use it when you're inside a conda virtualenv. You might need to 
# re-run cmake to pick up the conda version of python. 
# Note 2: we don't provide `make install-conda-package` similar to pip. This
# is becuase:
#     1) `make install-pip-whell` works in conda environment for local build
#     2) `make conda-package` is mainly for internal use to distribute conda
make conda-pacakge       # creates wheel in build/lib/python_package/conda_pacakge
```

## Motivation
This is a follow up to the automated pip wheel creation PR: https://github.com/IntelVCL/Open3D/pull/604

The motivation is to automate pip and conda package creation. This is necessary since now the python build process is more complicated (including compiled module, non-compiled module, and later Jupyter plugin with compiled JavaScripts).

---
This change is [Reviewable](https://reviewable.io/reviews/intelvcl/open3d/618)


qianyizh:
This is great... Need to be highlighted in next release. @germanros1987

PR #627, opened 2018-10-23
Odometry

takanokage:
Odometry unit tests.
The last three unit tests in this test case, (DoSingleIteration, ComputeMultiscale & ComputeRGBDOdometry) will be addressed in a subsequent PR as they require more complicated initialization. Also, CreateInformationMatrix unit test fails validation in TravisCI although it passes locally and it will be fixed in a subsequent PR.

---
This change is [Reviewable](https://reviewable.io/reviews/intelvcl/open3d/627)


syncle:
I've reviewed this PR in #631.

PR #628, opened 2018-10-23
Fix Rand(float,...)

takanokage:
Fixed Rand(float) to return values between 0 and 1.
Updated all the dependencies.
In a few cases updated name variables (like rgbdImage to rgbd_image) in order to meet guideline requirements.

---
This change is [Reviewable](https://reviewable.io/reviews/intelvcl/open3d/628)


syncle:
I've reviewed this PR in #631.

PR #630, opened 2018-10-24
Expose non-default constructor of VisualizerWithEdit in Python and add counter to automatic filename

dbadrian:
This PR aims to e xpose the VisualizerWithEdit constructor with arguments through its python bindings. Without one can not define the default filepath which very inconvenient. 

Additionally, this PR adds a counter to the "cropped.ply" file, since most likely one will apply more than one cropping (and requires all to obtain the same final results when loading and applying the crop selections). The counter is only increased if a file is actually saved (although by that point, the point cloud already changed, and the user might start over anyway).

---
This change is [Reviewable](https://reviewable.io/reviews/intelvcl/open3d/630)

### DIFF
--- a/src/Python/Visualization/open3d_visualizer.cpp
+++ b/src/Python/Visualization/open3d_visualizer.cpp
@@ -106,6 +106,7 @@ void pybind_visualizer(py::module &m)
     py::detail::bind_default_constructor<VisualizerWithEditing>(
             visualizer_edit);
     visualizer_edit
+        .def(py::init<double, bool, const std::string &>())
         .def("__repr__", [](const VisualizerWithEditing &vis) {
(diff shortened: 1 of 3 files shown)


dbadrian:
___
*[src/Visualization/Visualizer/VisualizerWithEditing.cpp, line 420 at r1](https://reviewable.io/reviews/intelvcl/open3d/630#-LPd_6pE3vsMDz60jqE_:-LPeYKUd7b-H1epMO3IP:b-896fix) ([raw file](https://github.com/intelvcl/open3d/blob/46d07ce9145b3f0b6c4b29ac410841e63daa9c4d/src/Visualization/Visualizer/VisualizerWithEditing.cpp#L420)):*

> ```
> ····················
> ```
redundant space
</blockquote></details>

Done.

PR #650, opened 2018-10-30
Fix Conda and Pip packaging issues

yxlao:
Several fixes related to Conda and Pip packaging:
- Build platform-specific targets (instead of `any` `none`) for wheel
- Fixed python 2.7 import `JVisulizer`
- Fixed conda dependency conflicts (resulting in forced downgrade)
- Disabled conda executable check (`conda` command could be a bash function instead of an executable, CMake may complain)


qianyizh:
One of the CI is broken. I have restarted it.
Otherwise this PR looks ready to me.

PR #655, opened 2018-11-01
Add option to disable jupyter build

yxlao:
Before: all based on whether `npm` is detected
Now: only when `ENABLE_JUPYTER=ON`, cmake will try to detect `npm`

---
This change is [Reviewable](https://reviewable.io/reviews/intelvcl/open3d/655)


yxlao:
thanks @syncle !

PR #666, opened 2018-11-06
moved back AccumulatedPoint to a hidden namespace.

takanokage:
This change is [Reviewable](https://reviewable.io/reviews/intelvcl/open3d/666)


takanokage:
This was based on Open3D master and can/should be merged first.

src/UnitTest/Core/Geometry/AccumulatedPoint.cpp
```diff
@@ -26,41 +26,41 @@
...
+// #include "Core/Geometry/AccumulatedPoint.h"
 #include "Core/Geometry/PointCloud.h"
 
 // ----------------------------------------------------------------------------
 //
 // ----------------------------------------------------------------------------
 TEST(AccumulatedPoint, Default)
 {
-    int size = 100;
+    // int size = 100;
```
syncle: Let's remove this test for now.

takanokage: done.

PR #681, opened 2018-11-12
Print utilities

takanokage:
added more print utilities to assist in the implementation of the black box unit tests.

---
This change is [Reviewable](https://reviewable.io/reviews/intelvcl/open3d/681)


syncle:
As a note 
```
    void Print(const Eigen::Matrix<T, M, N> &matrix, const int& tabs = 1, const char& terminator = ';')
```
Function can be used for general purpose so I am recommending put this function into ``Open3D/src/Core/Utility/Console.cpp``

takanokage:
I'm in favor of that.
Let's make another task for that in Asana. We need a discussion. (I'll add the task)
The Print utilities were meant for UnitTest only and QianYi has had in the past a pretty strong opposition to changing/extending the existing utilities in Open3D.

syncle:
Thanks @takanokage - I just opened a feature request too.

takanokage:
oh, great, thanks!

PR #682, opened 2018-11-12
Rand utilities

takanokage:
added more utilities for generating input data for the unit tests.
fixed bug: non-scaled float/double values must fall between 0.0 & 1.0 => updates to the RGBDImage tests for this reason.

---
This change is [Reviewable](https://reviewable.io/reviews/intelvcl/open3d/682)


syncle:
> These utilities don't actually generate random numbers.

Umm I don't get it. Why do you name this module "Rand" then?

PR #684, opened 2018-11-12
Constructor initialization

takanokage:
fixed bug due to PinholeCameraIntrinsic constructor not initializing member data.

---
This change is [Reviewable](https://reviewable.io/reviews/intelvcl/open3d/684)

### DIFF
--- a/src/Core/Camera/PinholeCameraIntrinsic.cpp
+++ b/src/Core/Camera/PinholeCameraIntrinsic.cpp
@@ -32,7 +32,8 @@
 
 namespace open3d{
 
-PinholeCameraIntrinsic::PinholeCameraIntrinsic()
+PinholeCameraIntrinsic::PinholeCameraIntrinsic() : width_(-1), height_(-1),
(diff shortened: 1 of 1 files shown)


syncle:
One quick comment. Can you check if there are other cases in current master that initializes intrinsic matrix? If so, can you check that they initializes the intrinsic like ``width_(-1), height_(-1), intrinsic_matrix_(Eigen::Matrix3d::Zero()``? Let's make code consistent.

syncle:
@takanokage - please check my question. This PR is straightforward, so I can merge once the question is answered.

takanokage:
I'll check right away.

takanokage:
I can find no other case where PinholeCameraIntrinsic gets initialized.
Still, the question remains what would be good default values for width and height? I chose -1 so that the code will fail if uninitialized width & height are used anywhere in the code. Maybe you have another idea?

syncle:
I think it is good - I also considered 0, but couldn't see clear difference. We can comeback this issue later if necessary.

PR #687, opened 2018-11-12
FileSystem

takanokage:
implemented the Core/Utility/FileSystem test case.

---
This change is [Reviewable](https://reviewable.io/reviews/intelvcl/open3d/687)


takanokage:
___
*[src/UnitTest/Core/Utility/FileSystem.cpp, line 43 at r1](https://reviewable.io/reviews/intelvcl/open3d/687#-LR8cMXv-EBfgRqHbNzD:-LRU4yWc1_BV0-5Exk7z:b-896fix) ([raw file](https://github.com/intelvcl/open3d/blob/35c0b13f7c8fd1d2c3a19fdcc56b5b9646c52433/src/UnitTest/Core/Utility/FileSystem.cpp#L43)):*
<details><summary><i>Previously, takanokage (Dan Petre) wrote…</i></summary><blockquote>

This was a great suggestion after all.
I found & fixed a small bug in GetFileNameWithoutExtension.
The bug was causing a missing character at the end when there was no extension.
</blockquote></details>

Done.

PR #688, opened 2018-11-12
PinholeCameraTrajectory

takanokage:
implemented PinholeCameraTrajectory test case.  …
fixed bug in PinholeCameraTrajectory ConvertToJsonValue(...) and ConvertFromJsonValue(...).

---
This change is [Reviewable](https://reviewable.io/reviews/intelvcl/open3d/688)


takanokage:
___
*[src/Core/Camera/PinholeCameraTrajectory.cpp, line 75 at r2](https://reviewable.io/reviews/intelvcl/open3d/688#-LRDYZnxBn3Kd4wLzyR8:-LREObXP3vTUBSFuEF7r:b-8twnyc) ([raw file](https://github.com/intelvcl/open3d/blob/8e24cc10ccaf899c20f1fbffc6275a3e738eaeb8/src/Core/Camera/PinholeCameraTrajectory.cpp#L75)):*

Thanks. Please go ahead. BTW, I think we should keep ``PinholeCameraParameters`` because ``PinholeCameraTrajectory`` is vector of ``PinholeCameraParameters``, right?
</blockquote></details>

Sounds right.
The way I implemented the fix to PinholeCameraTrajectory was by calling the PinholeCameraParameters ConvertToJsonValue(...) for each element in the vector.

Below is an example of what PinholeCameraTrajectory with 2 PinholeCameraParameters looks like when converted to Json::Value.
Does this look acceptable to you or do you have other ideas?

```
{
	"class_name" : "PinholeCameraTrajectory",
	"parameters" : 
	[
		{
			"class_name" : "PinholeCameraParameters",
			"extrinsic" : 
			[
				0.10793991159086103,
				-0.045205896275679502,
				0.25774184952384882,
				-0.27043105441631332,
				0.026801820391231024,
				0.90445945034942565,
				0.8323901360074013,
				0.27142345591980188,
				0.43459385886536617,
				-0.7167948892883933,
				0.21393775251411729,
				-0.96739885675134085,
				-0.51422645874052608,
				-0.72553684642796257,
				0.60835350845398084,
				-0.68664182149183084
			],
			"intrinsic" : 
			{
				"height" : 480,
				"intrinsic_matrix" : 
				[
					0.68037543430941905,
					-0.21123414636181392,
					0.56619844751721171,
					0.59688006695214657,
					0.82329471587356862,
					-0.60489726141323208,
					-0.32955448857022196,
					0.53645918962380801,
					-0.44445057839362445
				],
				"width" : 640
			},
			"version_major" : 1,
			"version_minor" : 0
		},
		{
			"class_name" : "PinholeCameraParameters",
			"extrinsic" : 
			[
				0.27510453540603841,
				0.048574380133568518,
				-0.012834026018545996,
				0.94555004776713902,
				-0.41496643117394594,
				0.54271539558782966,
				0.053489958426677697,
				0.53982767255037456,
				-0.1995427558196442,
				0.78305890401036438,
				-0.43337050798971699,
				-0.29508330547021855,
				0.61544904001776546,
				0.83805294793008489,
				-0.86048944753617485,
				0.89865415072937216
			],
			"intrinsic" : 
			{
				"height" : 480,
				"intrinsic_matrix" : 
				[
					-0.198111211507633,
					-0.74041910643708853,
					-0.78238239594846148,
					0.99784903600711794,
					-0.56348618937818618,
					0.025864788808796835,
					0.67822446938521441,
					0.22527966519132248,
					-0.40793676460531392
				],
				"width" : 640
			},
			"version_major" : 1,
			"version_minor" : 0
		}
	],
	"version_major" : 1,
	"version_minor" : 0
}
```

takanokage:
___
*[src/Core/Camera/PinholeCameraTrajectory.cpp, line 75 at r2](https://reviewable.io/reviews/intelvcl/open3d/688#-LRDYZnxBn3Kd4wLzyR8:-LRERFvk4XPIVElIkQpQ:b-m94n9y) ([raw file](https://github.com/intelvcl/open3d/blob/8e24cc10ccaf899c20f1fbffc6275a3e738eaeb8/src/Core/Camera/PinholeCameraTrajectory.cpp#L75)):*

Yes. It looks good to me as it conveys the right encapsulation. Please update the code and test data accordingly. Thanks!
</blockquote></details>

I have tested reading the json file from Python and works great.
After updating the json file several Python examples need to be updated as well - I'm working on it.

takanokage:
___
*[src/Core/Camera/PinholeCameraTrajectory.cpp, line 54 at r2](https://reviewable.io/reviews/intelvcl/open3d/688#-LRDXyVH86AQCqDYlHu7:-LRERZiuEVOTef_wNmjG:b-f9ou34) ([raw file](https://github.com/intelvcl/open3d/blob/8e24cc10ccaf899c20f1fbffc6275a3e738eaeb8/src/Core/Camera/PinholeCameraTrajectory.cpp#L54)):*

This changes core of Open3D, so we need to be careful. Could you double-check that we have full test functions and it passes?
</blockquote></details>

work in progress.

takanokage:
@syncle I see that some python examples read the trajectory from a `.log`.
Given that the structure of PinholeCameraTrajectory has changed, is it accurate to assume that the methods for reading/writing `.log` and the format of the `.log` should change as well?

syncle:
Nope. Log file is only supported for backward compatibility. Ideally we want to use only .json format. If there is a case need to make ``PinholeCameraTrajectory`` using log file, it need to take single intrinsic matrix and log file having several extrinsics, and IO function need to make a class by copying intrinsic matrix to make ``PinholeCameraParameter``s. I remember I have modified the I/O for .log file in that way.

takanokage:
I would recommend we remove support for `.log` files completely. I imagine the move to `.json` happened a long time ago so everyone is using that instead of `.log`. It will simplify the scripts and help eliminate any confusions moving forward.

1. we need to let people know ahead of time that we're going to remove support for `.log` files)
2. remove support for all `.log` not this sprint but the next.

Any thoughts?

syncle:
I don't think so. Log file support is made for supporting other datasets like supporting I/O for velodyne dataset. We cannot just drop it.

qianyizh:
Actually I think the datasets that use .log format are datasets that we maintain. So it **is possible** to deprecate the support for .log format. But, in order to do this, we need to clean up our datasets to transfer the .log files to .json. I think this has long term value and we should budget time to do it.

The datasets that currently use .log format are:
 * Stanford 3D scene (http://qianyi.info/scenedata.html)
 * Redwood SUN3D (http://redwood-data.org/indoor/models.html)
 * Redwood synthetic (http://redwood-data.org/indoor/dataset.html)
 * ~~Redwood CPCP (http://redwood-data.org/indoor_lidar_rgbd/download.html), it already has .json format~~
 * Redwood registration benchmark (http://redwood-data.org/indoor/regbasic.html)
 * TanksAndTemples (https://www.tanksandtemples.org/download/)

@syncle have I missed anything?

PR #692, opened 2018-11-13
Image

takanokage:
+ using namespace directive in order to reduce code line length.
removed global variables.
shortened local variables names.
simplified comparisons using unit_test::ExpectEQ(...).

---
This change is [Reviewable](https://reviewable.io/reviews/intelvcl/open3d/692)


syncle:
Let me merge once @yxlao's comment is addressed.

takanokage:
___
*[src/UnitTest/UnitTest.cpp, line 176 at r1](https://reviewable.io/reviews/intelvcl/open3d/692#-LRDYA7f9bBCD7tszZti:-LRgqxwkEcAdeTLLHqSt:b-896fix) ([raw file](https://github.com/intelvcl/open3d/blob/7744bdd68a801d1b1628469f1a17d33b7442133d/src/UnitTest/UnitTest.cpp#L176)):*

> ```
> ExpectEQ(&v0[0], &v1[0], v0.size());
> ```

@Dan Wondering if it would be possible to change this to:
```cpp
ExpectEQ(v0.data(), v1.data(), v0.size());
```

From [this](https://stackoverflow.com/a/25420131/1255535), seems like even if vector is empty, `v0.data()` is still safe to call. 
</blockquote></details>

Done.

PR #694, opened 2018-11-13
RGBDOdometryJacobian

takanokage:
implemented the RGBDOdometryJacobianFromHybridTerm and RGBDOdometryJacobianFromColorTerm test cases.
added new reference data for RGBDImage based on fixes to Rand float/double.

---
This change is [Reviewable](https://reviewable.io/reviews/intelvcl/open3d/694)


syncle:
This PR includes some code change on Rand utilities. Should this PR be merged after #682 is merged?

takanokage:
The order of merging doesn't matter for any of the new PRs.
In order to make this happen some code sections appear in several PRs.

yxlao:
Yeah, it shall be fine to have duplicated changes in different PR. The overlapping part will be automatically resolved when we merge.

takanokage:
even if there are conflicts let me know and I'll rebase/solve any issue.

syncle:
Thanks @yxlao! https://github.com/IntelVCL/Open3D/pull/694#pullrequestreview-174584341 is a great comment and I feel that I am not 100% sure the code in Open3D strictly follows the recommendation. I just opened a issue #701 regarding this for the sanity check.

takanokage:
___
*[src/UnitTest/Core/Odometry/OdometryTools.cpp, line 54 at r1](https://reviewable.io/reviews/intelvcl/open3d/694#-LRDqpEE6AASxgsf8JTV:-LRJ8hKw9QEEwy1JXIb2:bgblqim) ([raw file](https://github.com/intelvcl/open3d/blob/4499f0a93946c377da97b9e99ddc4115fc848046/src/UnitTest/Core/Odometry/OdometryTools.cpp#L54)):*

This is not the recommended way to use `shared_ptr`. This line works only because the copy constructor is called, that is, this line creates another copy of the Image.
To illustrate the point, here's a snippet that i created:

```cpp
#include <iostream>
#include <memory>
#include <string>

using namespace std;

class Dog {
public:
    Dog(const string& name) {
        name_ = name;
    }
    Dog(const Dog& src) {
        cout  << "Copy constructor called" << endl;
        name_ = src.name_;
    }
private:
    string name_;
};

int main()
{
    // Method 1: not recommended
    Dog corgi("Corgi");
    shared_ptr<Dog> shared_corgi = make_shared<Dog>(corgi);
    
    // Method 2: recommended
    shared_ptr<Dog> proper_shared_corgi = make_shared<Dog>("Corgi");
    return 0;
}
```
From the output of the code, we can see that the copy constructor is called. 

In general in most cases it's not a good idea to assign existing object to a shared_ptr. More discussions at: https://stackoverflow.com/questions/20195611/assigning-existing-values-to-smart-ptrs and https://stackoverflow.com/questions/8466459/create-a-boostshared-ptr-to-an-existing-variable/8466476.

I would change it to 
```cpp
shared_ptr<Image> image = make_shared<Image>();
image->PrepareImage(width,
                       height,
                       num_of_channels,
                       bytes_per_channel);
...
```
</blockquote></details>

Good points.
I would argue that the use of make_shared is fine and that we need to properly add copy constructors to all our classes (or explicitly delete them if the design requires such a behavior).

So my answer is that that we actually need to explicitly add a copy constructor to Image **if** the default one is not doing the right thing. It would seem that in this case however it is doing the right thing, is that correct?

Any thoughts?

takanokage:
___
*[src/UnitTest/Core/Odometry/OdometryTools.cpp, line 54 at r1](https://reviewable.io/reviews/intelvcl/open3d/694#-LRDqpEE6AASxgsf8JTV:-LRJGcWx2lRpG62qrr5R:b-ehm9q2) ([raw file](https://github.com/intelvcl/open3d/blob/4499f0a93946c377da97b9e99ddc4115fc848046/src/UnitTest/Core/Odometry/OdometryTools.cpp#L54)):*

I still believe we should use `make_shared` in the recommended way. The main purpose of the smart pointer is to manage the lifetime of objects. If we do a copy every time we create a smart pointer, it defeats the original purpose.  So even when the proper constructor is defined, to make the code clean, we shouldn't invoke it.

Invoking the copy constructor is kind of like doing this:
```cpp
int add(int a, int b) {
    int sum = a + b;
    int sum2 = sum;
    return sum2;
}
```
</blockquote></details>

I believe we all are aware of the costs/benefits of copy constructors.
I'm suggesting that the code is functionally correct.
We can discuss performance however I'm suggesting that we should instead discuss performance/(cost of making the change). Yes we can eliminate a copy here and in doing that my estimate is that the ratio of performance/(cost of making the change) is close to zero. That's the only reason why I wouldn't make the change - not worth the time.

takanokage:
___
*[src/UnitTest/Core/Odometry/OdometryTools.cpp, line 54 at r1](https://reviewable.io/reviews/intelvcl/open3d/694#-LRDqpEE6AASxgsf8JTV:-LROPaUtAS8nsgqx2BJN:b-jzpvje) ([raw file](https://github.com/intelvcl/open3d/blob/4499f0a93946c377da97b9e99ddc4115fc848046/src/UnitTest/Core/Odometry/OdometryTools.cpp#L54)):*

Umm I don't understand why you insist here. Why don't you just try? :)
</blockquote></details>

I will make the change. Sorry for giving you trouble.

I will say this. The two links provided as arguments in favor of making a shared_ptr object that I initialize afterwards have been misinterpreted. That's not what the links say. The links say that you shouldn't make_shared of objects that are **not dynamically allocated** aka that are residing on the stack. This is clearly not applicable to our discussion.

Also, given that:
1. the code is functionally correct
2. there's no performance bottleneck that needs to be fixed
perhaps the value of the change to the the project is zero. Maybe I'm wrong, in which case I welcome correction.

takanokage:
___
*[src/UnitTest/Core/Odometry/OdometryTools.cpp, line 54 at r1](https://reviewable.io/reviews/intelvcl/open3d/694#-LRDqpEE6AASxgsf8JTV:-LRShOVxBtKg94BpJvzI:bp6hi42) ([raw file](https://github.com/intelvcl/open3d/blob/4499f0a93946c377da97b9e99ddc4115fc848046/src/UnitTest/Core/Odometry/OdometryTools.cpp#L54)):*

> The links say that you shouldn't make_shared of objects that are not dynamically allocated aka that are residing on the stack.

So in the previous code, `image` is indeed residing on the stack.

```cpp
Image image; // on the stack
...
return make_shared<Image>(image);
```


</blockquote></details>

Oh well, technically parts of `image` do reside on the stack and parts on the heap.
`Open3D::Image` is not a POD (see pure virtual methods in base class) which pretty much guarantees `image` is not entirely the stack. 

so, the following member variables may in fact be on the stack:
- `width_`
- `height_`
- `num_of_channels_`
- `bytes_per_channel_`

`data_` on the other hand is a dynamic type, whatever/most memory it uses is on the heap.

But all that may not matter in our discussion given the fact that `make_shared` creates another object on the heap using it's copy constructor. So the code is fine.

takanokage:
thanks for reviewing/approving!

PR #711, opened 2018-11-20
Relative paths in CMake package config

pezcode:
This makes the entire install directory relocatable. Currently

- paths to include and library directories are hardcoded
- 3rd-party include directories incorrectly point to the source directory

This could for example pave the way for an installer with a prebuilt library for Windows.

---
This change is [Reviewable](https://reviewable.io/reviews/intelvcl/open3d/711)


syncle:
Thanks @pezcode and @yxlao. I have some quick questions for @yxlao before merge.
- Can I ask this PR have tested on different platforms?
- Will it be necessary to update our documentation page with this change?

yxlao:
I've tested it on Ubuntu, and I believe @pezcode have tested it on Windows, right? 

Also, our CI build the `TestVisualize.cpp` on [ubuntu, mac](https://github.com/IntelVCL/Open3D/blob/master/.travis.yml#L166) and [windows](https://github.com/IntelVCL/Open3D/blob/master/.appveyor.yml#L61) with the new CMake. 

The documentation shall be fine, as we still use the same cmake commands to link with Open3D. The only difference is that the installed library can now be moved around.

pezcode:
I tested it on Win 10 with VS 2017

pezcode:
I just tested another change on Ubuntu and noticed that the install target doesn't copy the include files for flann, liblzf and rply. Is that intentional?
Previously the CMake config file would populate Open3D_INCLUDE_DIRS with references to the 3rd party include folders inside the source tree, so this kinda worked.
Now that all paths are relative to the install directory, these folders don't exist.

Would outside code ever use these includes (or indirectly through Open3D's headers)? If so, the install step should install those folder as well.

Also, it says "2 discussions left (pezcode)". Do I need to do something here?

qianyizh:
Looks good to me.
CI builds Open3D, installs it, then builds Visualzer.cpp, then runs it.
https://github.com/IntelVCL/Open3D/blob/master/examples/Cpp/Visualizer.cpp

As long as this passes the installation should be good.

PR #717, opened 2018-11-24
Factorize internal functions for ColormapOptimization module

syncle:
This PR reorganizes internal functions in ColorMapOptimization.

Key changes
- A new function for building Jacobian matrices that follows RGBDOdometry structure
- Non-rigid optimization has more than 6 variables (6D camera pose + anchor points). I added a new Eigen helper function that can take n-elements and multiplication pattern. This could be used for general purpose, but non-rigid colormap optimization is the only use case, so I put it inside colormapoptimization.
- Took out ImageWarpingField - it is not general purpose, so I just place it inside colormapoptimization folder.
- Some useful function for mesh colorization using color images are placed inside TriangleMeshAndImageUtilities.cpp. I can polish a tutorial. However, there are many non-general purpose functions as well. We need to discuss for this.
- A new general purpose image processing function is placed in Image.cpp such as CreateDepthBoundaryMask.

I think this PR need some more discussion before it is merged.

### DIFF
--- a/src/Core/ColorMap/ColorMapOptimization.cpp
+++ b/src/Core/ColorMap/ColorMapOptimization.cpp
@@ -35,6 +35,8 @@
 #include <Core/Utility/Console.h>
 #include <Core/Utility/Eigen.h>
 
+#include <iostream>
+
 namespace open3d {
 
 namespace {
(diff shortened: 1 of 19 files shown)


syncle:
I fixed some issues. Please review.

syncle:
You're right @yxlao. I just fixed it.

PR #725, opened 2018-12-04
Move 3rdparty

syncle:
This PR moves 3rdparty folder to Open3D-3rdparty repo. 
@yxlao please check if this PR will not make any compilation issue.


syncle:
BTW, is there a way to squeeze the commit history regarding the 3rdparty folder? I am concerned about git will keep maintain change log forever.

yxlao:
Although there is a way to squeeze, that would require all users do force update the git history, so it might be too much overhead.

PR #743, opened 2018-12-18
Imported CMake target

pezcode:
Using `find_package(Open3D)` now defines a target _Open3D::Open3D_ that you can link to with `target_link_libraries`. Include dirs, libraries, compiler definitions are transitive so you don't need to do anything else.

You can check out the updated [C++ sample project](https://github.com/pezcode/Open3D/blob/cmake-target/docs/_static/C++/CMakeLists.txt). It also showcases a nice way of copying over the .dll on Windows by using the `TARGET_FILE` property.

This would close https://github.com/IntelVCL/Open3D/issues/710 and https://github.com/IntelVCL/Open3D/issues/716.

There is something that might need fixing and I'm not entirely sure how to go about it:
- GCC requires the OpenMP flag on the compiler AND linker line, otherwise the proper libraries won't be included (gomp and pthread in this case)
- Setting CMAKE_CXX_FLAGS deals with this by passing the flag on to the linker (this doesn't happen with MSVC from my testing, some googling makes me assume CMake keeps a list of options it passes on)
- There is no way to set a property on the imported target that works the same way CMAKE_CXX_FLAGS does

Solution 1 (how it is currently):

Setting the INTERFACE_COMPILE_OPTIONS property does NOT pass flags on to the linker, so we can pass it manually. This however causes problems with compilers that don't want the OpenMP flags passed to the linker (e.g. MSVC): 
> LINK : warning LNK4044: unrecognized option '/openmp'; ignored
Would have to check for specific compilers and when to set both, and this is where it would get messy.

Solution 2:

Explicitly link to the needed OpenMP libraries at build time. CMake 3.9 and higher define a target OpenMP::OpenMP_CXX which has all the needed libraries. This probably also means there won't be any need to require anyone linking to Open3D to add any compiler flags. I don't know if setting the minimum version to 3.9 is acceptable, it got released in July 2017. Could also do this optionally.

Solution 3:

Manually define the libraries at build time, like [here](https://github.com/loriab/psi4/blob/master/external/common/lapack/FindTargetOpenMP.cmake).

Solution 4:

Require the user to still set CMAKE_CXX_FLAGS but that kind of defeats the entire point of this change.

---
This change is [Reviewable](https://reviewable.io/reviews/intelvcl/open3d/743)


pezcode:
Any ideas on what's going wrong with the automated build? I did some testing on Win10 and Ubuntu 18.04 and didn't run into problems

Specifically:
> CMake Error: Problem processing arguments. Aborting.

yxlao:
For the [semaphore CI](https://semaphoreci.com/syncle/open3d-2/branches/pull-request-743/builds/1)
`CMake Error: Problem processing arguments. Aborting.` is due to an error in the previous step:
```
CMakeFiles/OpenMP.dir/OpenMP.cpp.o: In function `TestMatrixMultiplication(int, char**)':
OpenMP.cpp:(.text+0x3cfa): undefined reference to `pthread_create'
collect2: error: ld returned 1 exit status
examples/Cpp/CMakeFiles/OpenMP.dir/build.make:104: recipe for target 'bin/examples/OpenMP' failed
make[2]: *** [bin/examples/OpenMP] Error 1
CMakeFiles/Makefile2:2092: recipe for target 'examples/Cpp/CMakeFiles/OpenMP.dir/all' failed
make[1]: *** [examples/Cpp/CMakeFiles/OpenMP.dir/all] Error 2
Makefile:127: recipe for target 'all' failed
make: *** [all] Error 2
```

For the [Travis CI](https://travis-ci.org/IntelVCL/Open3D/builds/469318759?utm_source=github_status&utm_medium=notification) looks like the build of open3d is successful (since there are no change to the build itself), but the TestVisualizer linking failed (e.g. [this](https://travis-ci.org/IntelVCL/Open3D/jobs/469318763#L4233)).

pezcode:
Ok...
Travis fails on Linux when building static libs, apparently X11 isn't getting linked into the visualizer test.
Semaphore fails on Linux when not building 3rd party modules from source, threads library isn't getting linked into samples.

I can't reproduce this on my Ubuntu 18.04 VM, both make install and building the C++ sample succeed. Afraid I don't have the time to troubleshoot this much more, maybe I can revisit it after New Year's. Feel free to ignore or scrap this for now I guess.

pezcode:
I wanted to take another shot at this by explicitly linking the OpenMP libraries. That requires a custom FindOpenMP.cmake script which goes into the 3rdparty repo's CMake folder.

That's going to be a problem with CI I assume. What's the best way to do this? Just put the script into the normal repository and run CI on that for now?

yxlao:
For the CMake version, we could consider adding a
```
message(STATUS "CMake version: ${CMAKE_VERSION}")
```
in the top level CMake. It seems that CMake is not documented on [Semaphore's docs](https://semaphoreci.com/docs/supported-stack.html).

pezcode:
Seeing how you [restructured folders and build scripts](https://github.com/IntelVCL/Open3D/commit/f82e73f73f0c3af724cf76e7864258a4814e422c) and considering my lack of freetime, I'd suggest to close this for now.
Would be a better idea to revisit this from scratch at some point.

yxlao:
Thanks @pezcode ! The imported CMake target still seems to be a good idea, we can revisit it in the future.

akashsharma02:
Is there any plan to incorporate these changes anytime soon? Linking Open3D using the C++ interface via `link_directories(${Open3D_LIBRARY_DIRS})` in CMakeLists.txt is quite messy, as it does not allow using dependencies which are common to Open3D and my project such as `fmt::format`.

Issue #744, opened 2018-12-18
Add rolling support to visualizer


syncle:
**Is your feature request related to a problem? Please describe.**
Interactive visualization need to add support rolling. It is useful as the current rotation is based on x and y directional rotations. To rotate geometry along z axis, it is not straightforward.

**Describe the solution you'd like**
Adding a new key + drag option to support z-axis rotation.

qianyizh:
I think rotation along z-axis only makes sense when it is under orthogonal view. And I think in Open3D there is such a function (in EditingVisualizer? at least I had it in ManuallyCropGeometry). To do it, you can first press X or Y or Z to lock to one axis.

syncle:
That's true, but the motivation here is that supporting z-axis rotation in basic visualizer with free view mode. Remember with the basic visualizer, there were cases we are struggled to orient a geometry. We did this by zig-zag + drag which was not very fancy. I actually have submitted #745, so that the only OpenGL up vector and right vector to be rotated. This is feasible without orthogonal projection model. You may want to check #745. :)

germanros1987:
This is done in the new 0.10 release.

Issue #617, opened 2018-10-16
How to rotate a model along X axis? 


TaoHuUMD:
I used python. There is a simple code about rotation, but "ctr.rotate(0.0, 10.0)" only supports rotation along Y or Z.  How can I rotate a model along X? And translation also has the same problem. 

Another way is to set view matrix, but there is no related interface for python. How can I do?

syncle:
Hi, Please see my answer for setting view matrix: https://github.com/IntelVCL/Open3D/issues/615

``ctr.rotate`` and ``ctr.translate`` is intended for 2-dimensional mouse event; It computes view point change using mouse drag event. To freely rotate or translate the model, I recommend to use ``pcd.transform(T)`` with your transformation matrix ``T``.

tsyj0404:
thanks. pcd.transform(T)  seems work for rotation, but I cannot translate the model. Do I have to set camera?

    vis = Visualizer()
    vis.create_window(window_name = 'open3d',width = 600, height = 600) #false
    vis.add_geometry(pcd)

    trans_init = np.asarray(
                [[0.862, 0.011, -0.507,  **10**],
                [-0.139, 0.967, -0.215,  0],
                [0.487, 0.255,  0.835, 0],
                [0.0, 0.0, 0.0, 1.0]])

    pcd.transform(trans_init)

    draw_geometries_with_animation_callback([pcd], rotate_view)

syncle:
``draw_geometries*`` family function has ``reset_view_point`` inside that automatically adjusts camera depending on the geometry. In this manner, even though the geometry is translated, the camera follows too, and you are looking the same image as a result. 

In this case, you will need to mimic ``draw_geometries`` family, but without using ``reset_view_point``. A good example is [here](http://open3d.org/docs/tutorial/Advanced/non_blocking_visualization.html). You will see the script makes use of explicit call of ``Visualizer`` class, and using its helper function explicitly.

tsyj0404:
Thanks. It seems that the next transformation is based on the current state. There is no function in Visualizer() to delete the current geometry.  Can I transform the object always from its original state?  
And the transformation matrix is the same as the matrix of OpenGL, right?

tsyj0404:
thanks. I will make a try.

syncle:
Hi @tsyj0404. I submitted a PR that can do rolling the geometry (rotating along the viewing direction) hope this will be helpful.

taohucs:
thanks for your wonderful work!

[...]

PR #745, opened 2018-12-18
Panning by Mouse Wheel Drag & Rolling Option

### LINKED ISSUE #560: Pan Mouse Wheel Press
First of all, thanks for this amazing open source library!

I was wondering if it would it be possible to implement panning using the mouse wheel press. It's currently really hard to zoom in on some places due to how restricted the viewer is. It would be nice to introduce an option when we can "walk/translate" in a direction using the wheel.

Thanks!

syncle:
Resolves #560, #617, #744.

**Key changes**
- ``Shift + left mouse button + drag`` rotates geometry along z-axis (rolling)
- ``mouse wheel button + drag`` translates the geometry

**Need to check**
Check these additive modes are supported by other Visualization modes, such as ViewControlWithCustomAnimation or ViewControlWithEditing.

---
This change is [Reviewable](https://reviewable.io/reviews/intelvcl/open3d/745)

### DIFF
--- a/docs/tutorial/Basic/visualization.rst
+++ b/docs/tutorial/Basic/visualization.rst
@@ -25,9 +25,11 @@ Open3D provides a convenient visualization function ``draw_geometries`` which ta
 .. code-block:: sh
 
     -- Mouse view control --
-      Left button + drag        : Rotate.
-      Ctrl + left button + drag : Translate.
-      Wheel                     : Zoom in/out.
+      Left button + drag         : Rotate.
+      Ctrl + left button + drag  : Translate.
+      Wheel button + drag        : Translate.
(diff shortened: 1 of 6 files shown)


syncle:
Thanks for the comment @yxlao. I am revisiting this PR for the next release, but this implementation is in accordance with other previous functions such as ``Rotation`` that passes pixel values to the view point transform functions. I agree with that y-axis could help the user experience, but based on my test, this seems good for now. It could be addressed later

qianyizh:
I think both @yxlao and @syncle have a point. The current function is good enough for now so I will proceed to merge this PR. The more sophisticated behavior of ``Roll()`` can be added in the future.

PR #786, opened 2019-01-18
Adding STL file support

### LINKED ISSUE #785: STL file support
STL format is another very common format. It would be good if Open3D supports it.

Reading/Writing triangle mesh to stl file.

**Additional context**
STL format has quite different description for ascii and binary format. There are several stl extensions for the specific application. 
Reference to the stl format: https://en.wikipedia.org/wiki/STL_(file_format)
http://www.sgh1.net/posts/read-stl-file.md (Good C++ reference)
https://github.com/rsmith-nl/stltools (in Python)
https://github.com/hschendel/stl (MIT in GO language)

syncle:
Resolves #785.

- Supports read/write of STL file format
- Binary STL files are supported.

Example: Downloading and viewing geometry for 3D printing (in STL format).

screen shot 2019-01-17 at 5 34 54 pm

screen shot 2019-01-17 at 5 38 15 pm


qianyizh:
Two quick question:

1. According to https://en.wikipedia.org/wiki/STL_(file_format) , it seems that there is an ASCII STL format too? But we are not supporting it. Right?

2. Is there any lightweight mesh libraries so that we can use for IO, instead of reinventing the wheels? In this case, we will have STL and OBJ and maybe some other format almost out of the box. I briefly looked. I think PCL is using VTK (which I don't like since it is too heavy). MeshLab seems to be using something similar. OpenMesh might be another good candidate.

pezcode:
>  2. Is there any lightweight mesh libraries so that we can use for IO, instead of reinventing the wheels? In this case, we will have STL and OBJ and maybe some other format almost out of the box. I briefly looked. I think PCL is using VTK (which I don't like since it is too heavy). MeshLab seems to be using something similar. OpenMesh might be another good candidate.

You might want to look into the [Open Asset Import Library](https://github.com/assimp/assimp).

syncle:
@qianyizh: Yes there are ascii format, and I can add support for this (however, I am hesitating because of the answer below). My intension in this PR was to support binary format first because binary-STL is much popular than ascii-STL. 

Regarding general I/O for various geometry formats, the external libraries pointed by @qianyizh and @pezcode looks bit heavy, so I am not sure which method would be good option for us (please correct me if I am wrong). I am not working STL ascii format for now, because the issue can be resolved if we configure general I/O libraries.

qianyizh:
Sounds good to me.
@syncle can you resolve the conflict?
@yxlao can you check if your comments have been addressed?

syncle:
@qianyizh addressed the comments by @yxlao.

PR #802, opened 2019-01-28
Fix issues on examples/Cpp/RegistrationRANSAC.cpp

### LINKED ISSUE #800: RANSAC C++ example looping 50,000 times
The [RegistrationRANSAC C++ example](https://github.com/IntelVCL/Open3D/blob/master/examples/Cpp/RegistrationRANSAC.cpp) is doing some odd looping where it runs the registration 50,000 times.

I assume this was for benchmarking, but it would be nice to have a useful working example that also visualizes the result (maybe an extra commandline option for the visualizer, currently it's forced to false).

syncle:
Resolves #800.

---
This change is [Reviewable](https://reviewable.io/reviews/intelvcl/open3d/802)

### DIFF
--- a/examples/Cpp/RegistrationRANSAC.cpp
+++ b/examples/Cpp/RegistrationRANSAC.cpp
@@ -71,8 +71,7 @@ int main(int argc, char *argv[]) {
     }
 
     bool visualize = false;
-    if (ProgramOptionExists(argc, argv, "--visualize"))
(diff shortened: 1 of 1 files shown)


qianyizh:
@syncle build is failing due to code style check. Can you correct?

syncle:
Finished code-style check.

PR #804, opened 2019-01-29
UpdateGeometry() for single geometries

mthiede:
With many geometries loaded in the visualizer, UpdateGeometry() and AddGeometry() become slow. 

The reason is that in both cases, all geometries are invalidated.

With this pull request AddGeometry() only updates the new geometry and UpdateGeometry() gets an optional argument for updating only one specific geometry.

---
This change is [Reviewable](https://reviewable.io/reviews/intelvcl/open3d/804)

### DIFF
--- a/src/Visualization/Visualizer/Visualizer.cpp
+++ b/src/Visualization/Visualizer/Visualizer.cpp
@@ -356,7 +356,7 @@ bool Visualizer::AddGeometry(std::shared_ptr<const Geometry> geometry_ptr)
     ResetViewPoint();
     PrintDebug("Add geometry and update bounding box to %s\n",
             view_control_ptr_->GetBoundingBox().GetPrintInfo().c_str());
-    return UpdateGeometry();
+    return UpdateGeometry(geometry_ptr);
 }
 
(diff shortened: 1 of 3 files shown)


yxlao:
closing since #1392 is merged

src/Visualization/Visualizer/Visualizer.cpp
```diff
@@ -356,15 +356,17 @@ bool Visualizer::AddGeometry(std::shared_ptr<const Geometry> geometry_ptr)
...
 
-bool Visualizer::UpdateGeometry()
+bool Visualizer::UpdateGeometry(std::shared_ptr<const Geometry> geometry_ptr)
 {
     glfwMakeContextCurrent(window_);
     bool success = true;
     for (const auto &renderer_ptr : geometry_renderer_ptrs_) {
-        success = (success && renderer_ptr->UpdateGeometry());
+        if (geometry_ptr == nullptr || renderer_ptr->HasGeometry(geometry_ptr)) {
+            success = (success && renderer_ptr->UpdateGeometry());
```
nnop, line 368: I saw all `GeometryRender::UpdateGeometry()` return `true`, why checking if return success?

PR #808, opened 2019-01-30
Adding missing initializers for FGR class

### LINKED ISSUE #807: Missing initializers for FGR class
**Describe the bug**
FGR class has default parameter set, but they are not correctly assigned to the class.

https://github.com/IntelVCL/Open3D/blob/4e3b06026c2e51415400771254b308c97b0aab3a/src/Core/Registration/FastGlobalRegistration.h#L43-L55

The value ``decrease_mu`` and ``maximum_correspondence_distance `` will not be set.

syncle:
Fixes #807.

---
This change is [Reviewable](https://reviewable.io/reviews/intelvcl/open3d/808)


qianyizh:
Actually the test is broken by #801 
Should fix that first.

PR #809, opened 2019-01-30
 C++ project documentation for Windows

takanokage:
Updated the documentation on creating a C++ project based on installed files of Open3D on Windows.

This update was motivated by the discussion on Discord with Darth Blitz on 29/01/2019 regarding his trouble with this scenario.

---
This change is [Reviewable](https://reviewable.io/reviews/intelvcl/open3d/809)


takanokage:
___
*[docs/tutorial/C++/cplusplus_interface.rst, line 45 at r1](https://reviewable.io/reviews/intelvcl/open3d/809#-LXVAjD61ZuRzxoOxigi:-LXVoCzM3VFMe9Irjg3p:b-d1v8gm) ([raw file](https://github.com/intelvcl/open3d/blob/96733d38b6ae4c58f147f78257531d4ffa8f3d67/docs/tutorial/C++/cplusplus_interface.rst#L45)):*

> Most people on Windows don't even know they can use CMake from cmd line.  The CMake GUI is very convenient on Windows so most people are using that.
we could be consistent with: http://www.open3d.org/docs/compilation.html#windows, which contains both GUI and cmd for windows. 
</blockquote></details>

Well, I have the same comment for that section of the documentation.
IMHO that shouldn't be there because except maybe a few people, including you :P, no one is reading/using that.

PR #819, opened 2019-02-06
Fix Colomap segfault and divergence issues

yxlao:
Several changes:
1. When computing visibility map `images_mask` is now used, previously it was ignored due to a typo.
2. Index out-of-bound segfaults I: `baseidx < anchor_w_ * anchor_h_ * 2`, cannot be `=`
3. Index out-of-bound segfaults II: Given a pixel location, the base anchor point index `(ii, jj)` must satisfy
    ```cpp
    ii < anchor_w_ - 1;
    jj < anchor_h_ - 1;
    ```
    This is because if the base location is `(ii, jj)`, the four anchors are 
    ```cpp
    (ii, jj), (ii + 1, jj), (ii, jj + 1), (ii + 1, jj + 1)
    ```
    and we must have
    ```cpp
    ii + 1 < anchor_w_;
    jj + 1 < anchor_h_;
    ```
4. Divergence: `double anchor_step` shall be used instead of `int anchor_step`, this creates small misalignments when getting pixel values from the image, leading to divergence in the end.

Compared numerical results (`Residual error` and `reg`) for non-rigid optimization against code base before https://github.com/IntelVCL/Open3D/pull/717, and veified that they are identical.

---
This change is [Reviewable](https://reviewable.io/reviews/intelvcl/open3d/819)

### DIFF
--- a/src/Core/ColorMap/ColorMapOptimizationJacobian.cpp
+++ b/src/Core/ColorMap/ColorMapOptimizationJacobian.cpp
@@ -94,7 +94,7 @@ void ColorMapOptimizationJacobian::ComputeJacobianAndResidualNonRigid(
     pattern.setZero();
     r = 0;
     int anchor_w = warping_fields.anchor_w_;
-    int anchor_step = warping_fields.anchor_step_;
+    double anchor_step = warping_fields.anchor_step_;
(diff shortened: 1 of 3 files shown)


yxlao:
special thanks to @syncle for the discussions!

PR #821, opened 2019-02-07
Standardize API of SolveLinearSystemPSD

yxlao:
This PR contains only interface changes, numerical values of the algorithms stay the same.

- Rename to `SolveLinearSystemPSD`
- Optional `check_det` checks if the matrix is valid and full rank, otherwise it returns zero matrix
- Optional `check_psd` checks if the matrix is positive-semidefinite, otherwise the function SolveLinearSystemPSD cannot solve this matrix and return zero
- Without specification, the `SolveLinearSystemPSD` assumes the input matrix is PSD and bypass these two checks

### DIFF
--- a/src/Core/ColorMap/ColorMapOptimization.cpp
+++ b/src/Core/ColorMap/ColorMapOptimization.cpp
@@ -112,7 +112,8 @@ void OptimizeImageCoorNonrigid(
             bool success;
             Eigen::VectorXd result;
             std::tie(success, result) =
-                    SolveLinearSystem(JTJ, -JTr, false, true);
+                    SolveLinearSystem(JTJ, -JTr, /*prefer_sparse=*/true,
+                                      /*check_det=*/false, /*check_psd=*/false);
             Eigen::Vector6d result_pose;
             result_pose << result.block(0, 0, 6, 1);
             auto delta = TransformVector6dToMatrix4d(result_pose);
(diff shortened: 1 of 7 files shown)


takanokage:
___
*[src/UnitTest/Core/Utility/Eigen.cpp, line 74 at r1](https://reviewable.io/reviews/intelvcl/open3d/821#-LY8jxSh4spWn7Ouf0Jx:-LY8ogg06JE06H7lp4xl:b-bkzjle) ([raw file](https://github.com/intelvcl/open3d/blob/8df7d0a434ced68c97815a8a9cf48fbb64329580/src/UnitTest/Core/Utility/Eigen.cpp#L74)):*

thanks, done!
</blockquote></details>

this looks nice and compact.

takanokage:
___
*[src/Core/ColorMap/ColorMapOptimization.cpp, line 114 at r1](https://reviewable.io/reviews/intelvcl/open3d/821#-LY8nmIvA4-QtQmIqlhY:-LY8pqgF-6JF1i5p6k4z:bejkdws) ([raw file](https://github.com/intelvcl/open3d/blob/8df7d0a434ced68c97815a8a9cf48fbb64329580/src/Core/ColorMap/ColorMapOptimization.cpp#L114)):*

6 + 2 * num_anchor_in_width * num_anchor_in_height
</blockquote></details>

@syncle 
```
in non-rigid optimization, it become hundreds of sparse matrix
```
I apologize if my question is stupid.
SolveLinearSystemPSD seems to solve one equation at a time; can you explain more about how does it matter that there are hundreds of sparse matrices?

PR #830, opened 2019-02-15
Compute vertex adjacency map

yxlao:
Adjacency map is useful for implementing some mesh operations. There are some decision made with Jaesik. Please feel free to comment on the datastructure selection. 

1. map vs vector vs set
    ```cpp
    std::vector<std::unordered_set<int>> // selected, to fast check if two nodes are neighbors
    std::vector<int, std::vector<int>>
    std::unordered_map<int, std::unordered_set<int>>
    ```
2. size_t vs int
    ```cpp
    std::vector<std::unordered_set<int>> // selected
    std::vector<std::unordered_set<size_t>>
    ```

Tested with python as well
```python
type(mesh.adjacency_list[0])
Out[4]: set
type(mesh.adjacency_list)
Out[5]: list
```

---
This change is [Reviewable](https://reviewable.io/reviews/intelvcl/open3d/830)

### DIFF
--- a/src/Core/Geometry/TriangleMesh.cpp
+++ b/src/Core/Geometry/TriangleMesh.cpp
@@ -185,6 +185,18 @@ void TriangleMesh::ComputeVertexNormals(bool normalized /* = true*/) {
     }
 }
 
+void TriangleMesh::ComputeAdjacencyList() {
+    adjacency_list_.resize(vertices_.size());
+    for (const auto &triangle : triangles_) {
+        adjacency_list_[triangle(0)].insert(triangle(1));
+        adjacency_list_[triangle(0)].insert(triangle(2));
+        adjacency_list_[triangle(1)].insert(triangle(0));
+        adjacency_list_[triangle(1)].insert(triangle(2));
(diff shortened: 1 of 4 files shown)


syncle:
I like this design. As @yxlao stated, I would vote for ``std::vector<std::unordered_set<int>>`` as ``TriangleMesh`` has ``Eigen::Vector3i`` for vertices reference. This means ``size_t`` for vertex reference is redundant.

qianyizh:
This looks good. Just to copy paste my comments left on discord here for future reference.

Up vote for int instead of size_t, so it is consistent with vector3i for triangles. However I do have comments on unordered_set. In case the mesh is a manifold, you can build a half-edge structure of the mesh and track the neighborhood of a point in counterclockwise order. In this case, the neighboring points do have an order and using vector makes more sense.

The counter argument is that many meshes are not manifold. In this case there is no order in adjacent points and unordered_set is the right choice. Also, the very first step of half-edge algorithm is to build an unordered adjacent map, if my memory is correct. In this case we can say adjacent map is for general mesh and later on we can derive a subclass, say, halfedgemesh, which implements the order of adjacent points.

PR #833, opened 2019-02-16
alexs7 - minor fix to make the realsense D415 recording work

alexs7:
Encountered a problem when running ```python3 realsense_recorder.py --record_imgs``` on my machine (Macbook Pro). Which was:

`Traceback (most recent call last):
  File "realsense_recorder.py", line 107, in <module>
    profile = pipeline.start(config)
RuntimeError: Couldn't resolve requests`

Changing the resolution of the depth camera solves this.

Also added:

```
# Might need these lines aswell on macOS Mojave 10.14.3
# import sys
# sys.path.append('/usr/local/lib')
```
On top, as it might be needed for macOS users.

---
This change is [Reviewable](https://reviewable.io/reviews/intelvcl/open3d/833)

### DIFF
--- a/examples/Python/ReconstructionSystem/sensors/realsense_recorder.py
+++ b/examples/Python/ReconstructionSystem/sensors/realsense_recorder.py
@@ -1,3 +1,7 @@
+# Might need these lines aswell on macOS Mojave 10.14.3
+# import sys
+# sys.path.append('/usr/local/lib')
+
 # pyrealsense2 is required.
 # Please see instructions in https://github.com/IntelRealSense/librealsense/tree/master/wrappers/python
 import pyrealsense2 as rs
@@ -93,7 +97,7 @@ if __name__ == "__main__":
     if args.record_imgs or args.record_rosbag:
         # note: using 640 x 480 depth resolution produces smooth depth boundaries
         #       using rs.format.bgr8 for color image format for OpenCV based image visualization
-        config.enable_stream(rs.stream.depth, 1280, 720, rs.format.z16, 30)
+        config.enable_stream(rs.stream.depth, 640, 480, rs.format.z16, 30)
         config.enable_stream(rs.stream.color, 640, 480, rs.format.bgr8, 30)
         if args.record_rosbag:
             config.enable_record_to_file(path_bag)


syncle:
[D415 supports 1280x720](https://ark.intel.com/products/128256/Intel-RealSense-Depth-Camera-D415). This issue happens when the camera is connected with ordinal usb cable. Can you refer [this thread](https://github.com/IntelVCL/Open3D/issues/818)?

alexs7:
@syncle I am using a usb 3.1 cable and from what I checked my laptop has a usb 3.0 port.

syncle:
Can you that you can stream 1280x720 frames using [Intel RealSense SDK](https://github.com/IntelRealSense/librealsense)?

alexs7:
I followed this:

[](https://github.com/GitwellAnyohub/Real_Sensible_Moseq/blob/master/bash_and_install/install_realsense_sdk_mac_os.sh)

from this:

[](https://github.com/IntelRealSense/librealsense/issues/2764)

and when I try to run realsense-viewer in the command line, I get this:

```[1]    35764 segmentation fault  realsense-viewer```

syncle:
I think you are not using appropriate configuration for D415. Please check again. If the viewer can stream 1280x720, ``realsense_recorder.py`` should record the depth images.

alexs7:
@syncle What do you mean I am not using the appropriate configuration for D415? Is there a way to set it up or pre-configure it before usage? The way I used it is, I plugged it in and run the Open3D examples.

syncle:
The most common case is using low bandwidth USB 3.0 cable instead of using the cable provided with RealSense camera. In this case, the maximum resolution the camera can stream is getting limited.

alexs7:
@syncle Ok thanks. Even though I am using a USB 3.1 as mentioned I doubt that would be the problem. I will just have to stick with a lower res for now.

yxlao:
@alexs7, looks like I am getting the same behavior as @syncle suggests. I tested the scripts on both Ubuntu and macOS (with the link you send as installation guide), with both D415 and D435, and found that 1280x720 works on all cases. 

The conclusion seems to be that the cable matters a lot. I am using a MacBook with all type-c ports:
- If I use a 3rd-party type-c to type-c cable (potentially of bad quality), I can get only up to 640x480 
- If I use the original type-c to type-a cable that comes with RealSense plus a type-c dongle, then it works with 1280x720

enescanguven:
I just recognize that back USB port of my desktop is better. I recommend you to use back USB ports. even if front usb dock has 3.0 its not worked for my camera.

PR #864, opened 2019-03-07
Pybind docs parser and Google-style docstring generator

yxlao:
## Changes
- Parse and converts Pybind generated docstring to Google-style python docstring
- Allows injection of custom docstring to function arguments
- Converted `src/Python/io/io.cpp`  as an example demo
- The main motivation is to allow modification of Pybind generated docstrings. See the comments in the `open3d::docstring::FunctionDoc` class for discussions on the pros and cons of this approach.
- There are a lot of opportunities for future works, e.g.
    1. Better parser (fix numpy default argument format error)
    2. Extend to handle class member functions
    3. Convert all files to use `FunctionDoc` to generate docs 

### Auto-generated google-style docs
unknown
(Left: original, right: auto-generated)

### Custom docstring injection
unknown 1
(`Path to file` is added manually)

---
This change is [Reviewable](https://reviewable.io/reviews/intelvcl/open3d/864)


yxlao:
CI failing mainly due to the older version of gcc. Upgraded to gcc-4.9 for support of `regex`.

syncle:
LGTM - is this approach good for injecting some example codes into docstring as well?

yxlao:
@syncle yeah, there are several fields, we can put example codes in the `body_` field

```cpp
public:
    std::string name_ = "";
    std::vector<ArgumentDoc> argument_docs_;
    ArgumentDoc return_doc_;
    std::string summary_ = "";
    std::string body_ = "";
```

PR #865, opened 2019-03-08
Benchmark

takanokage:
This is a proposal for implementing performance tests.
The performance tests follow a similar pattern to examples/Cpp.
 - the code resides in <root_open3d>/benchmark
 - source files can use similar names to those in examples/Cpp
 - output binary names will be prefixed with `bench`; CMake requires different binary names, seemingly regardless of the fact that the output paths are different
 - build option BUILD_BENCHMARK, default OFF, can be enabled in order to build the tests

---
This change is [Reviewable](https://reviewable.io/reviews/intelvcl/open3d/865)


takanokage:
The PR and review are not necessary at this time.
I will redo the entire commit because there's an opportunity to use google benchmark for this task.
I'll create a new PR once I integrate google benchmark.

PR #876, opened 2019-03-21
Iteration number from RANSAC

syncle:
Current RANSAC family functions does not provide number of iterations after RANSAC finishes.
Extending RegistrationResult class a bit to convey iteration numbers.

---
This change is [Reviewable](https://reviewable.io/reviews/intelvcl/open3d/876)


syncle:
Found some bug with this PR. Will submit again.

PR #903, opened 2019-04-11
Initial Octree support

yxlao:
- `Octree` class
    - Constructor
    - Copy constructor
    - Node insertion
    - Node locate
    - Convert from PointCloud
    - Traverse function with labmda
- `OctreeNode` class: two sub types
- `OctreeNodeInfo` class.
    - Design decision: do not store the origin and size of a node in `OctreeNode`, store in `OctreeNodeInfo`
    - `OctreeNodeInfo` is computed then disgraded
    - Gain space efficiency with more computes
- Unit tests

Currently `OctreeLeafNode` only takes `color` property. The plan is to expand this to support flexible datatype.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/903)

### DIFF
--- a/src/Open3D/Geometry/Octree.cpp
+++ b/src/Open3D/Geometry/Octree.cpp
@@ -1,0 +1,415 @@
+// ----------------------------------------------------------------------------
+// -                        Open3D: www.open3d.org                            -
+// ----------------------------------------------------------------------------
+// The MIT License (MIT)
+//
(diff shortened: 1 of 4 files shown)


syncle:
This is great feature. Could we make some connections this with Voxel structure?

yxlao:
Yeah, the conversion between voxel gird and octree is planned

PR #914, opened 2019-04-16
Octree to VoxelGrid conversion

yxlao:
Input PointCloud (PointCloud to Octree)
Screenshot from 2019-04-16 16-58-44

VoxelGrid (Octree to VoxelGrid)
Screenshot from 2019-04-16 16-58-33

The visualization above depends on "VoxelGrid OpenGL 3.1+ visualizer fix" in #918 . Will polish and create another PR. Without the "VoxelGrid OpenGL 3.1+ visualizer fix", VoxelGrid can still be converted from Octree, but VoxelGrid cannot be optimized. 

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/914)

### DIFF
--- a/src/Open3D/Geometry/Octree.cpp
+++ b/src/Open3D/Geometry/Octree.cpp
@@ -31,6 +31,7 @@
 #include <unordered_map>
 
 #include "Open3D/Geometry/PointCloud.h"
+#include "Open3D/Geometry/VoxelGrid.h"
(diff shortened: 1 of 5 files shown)


qianyizh:
@syncle I think it is @germanros1987 's idea to unify all conversion functions to a form of "ToXXXX" and "FromXXXX", which I think is a good design :)

PR #929, opened 2019-04-24
Method to check if TriangleMesh is watertight

griegler:
The method checks if a triangle mesh is a closed 2-manifold.
A triangle mesh is a closed 2-manifold if V + F - E = 2.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/929)


yxlao:
As discussed offline, we could switch to detecting boundaries in half edge mesh.

griegler:
The latest commits added functions to TriangleMesh that can be used to check for edge manifold, vertex manifold, and self-intersections. In combination those can be utilized to perform a watertightness test.

syncle:
Awesome. @yxlao Could you check this PR again?

Issue #938, opened 2019-04-28
Remove dependencies on LGPL libraries in Eigen


jornb:
Currently, Open3D depends on Eigen's SimplicialCholesky implementation, [which is licensed under LGPL](http://eigen.tuxfamily.org/index.php?title=Main_Page#License). This is vastly more limiting than the licenses of all other third party dependencies, and I presume several people do not realize they need to comply to LGPL when using Open3D out of the box.

I found this when trying to build Open3D with the `EIGEN_MPL2_ONLY` compile definition, which disables Eigen components licensed under LGPL.

Suggested fixes in increasing severity (note that I don't know the internals of Open3d and cannot assess the feasibility of these suggestions):
* Clearly indicate the use of dependencies licensed under LGPL
* Detect the `EIGEN_MPL2_ONLY` cmake flag and disable compilation of all parts that require SimplicialCholesky
* Replace SimplicialCholesky with another solver

May be related to #405.

jornb:
For reference, the pull request #939 fixes my compilation errors when `EIGEN_MPL2_ONLY` is defined. However, I don't know enough of Open3D to know what effects this will have.

kuonangzhe:
Are there any updates? How will Open3D deal with LGPL problem? Thanks a lot! @qianyizh

kuonangzhe:
@yxlao @syncle @takanokage Could you mind sharing a bit of the design and plan for this part? It's been inactive for a while. Thank you so much!

AndyMender:
@yxlao @syncle @takanokage @kuonangzhe heads-up about licensing changes in Eigen3: https://gitlab.com/libeigen/eigen/-/commit/1dc1677d525b1df3cc32b3df5cd816e150b07185

germanros1987:
problem solved!

PR #939, opened 2019-04-28
Disable use of Eigen::SimplicialCholesky when EIGEN_MPL2_ONLY is defined

jornb:
Fixes compilation error when LGPL-licensed components of Eigen are disabled through the `EIGEN_MPL2_ONLY` flag.

See #938

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/939)

### DIFF
--- a/src/Open3D/Utility/Eigen.cpp
+++ b/src/Open3D/Utility/Eigen.cpp
@@ -68,6 +68,7 @@ std::tuple<bool, Eigen::VectorXd> SolveLinearSystemPSD(
 
     Eigen::VectorXd x(b.size());
 
+#ifndef EIGEN_MPL2_ONLY
     if (prefer_sparse) {
         Eigen::SparseMatrix<double> A_sparse = A.sparseView();
         // TODO: avoid deprecated API SimplicialCholesky
@@ -85,6 +86,7 @@ std::tuple<bool, Eigen::VectorXd> SolveLinearSystemPSD(
             PrintInfo("Cholesky decompose failed, switched to dense solver\n");
         }
     }
+#endif
 
     x = A.ldlt().solve(b);
     return std::make_tuple(true, std::move(x));


qianyizh:
@yxlao @syncle take a look at this?

syncle:
This is a good point. Thanks @jornb.

Just for the summary, ``Eigen::SimplicialCholesky`` was introduced in [this discussion](https://github.com/intel-isl/Open3D/issues/363#issuecomment-391784285) and merged via [this commit](https://github.com/intel-isl/Open3D/commit/baf7fa460655340ace8d4beafb9f43150e818c04). It was a contribution from the Open3D community. At that time, we didn't aware that ``Eigen::SimplicialCholesky`` is in LGPL domain.

Here are some options to consider:
1) Merge this PR. If users concern LGPL license, user should specify ``EIGEN_MPL2_ONLY`` flag. As Open3D is being advertised as MIT license now, we should revise documentation regarding this issue in a proper way. This part would be tricky.
2) Remove the use case of ``Eigen::SimplicialCholesky`` and close this PR. ``Eigen::SimplicialCholesky`` is only used in [PoseGraphOptimization](https://github.com/intel-isl/Open3D/blob/11a27bfaa83e57cec9c13afd4b7f4bcb60ff643a/src/Open3D/Registration/GlobalOptimization.cpp#L611-L613). This will make Open3D free from the license issue, and users don't need to worry at all. The downside of this option is that PoseGraphOptimization would be much slower because ``Eigen::SimplicialCholesky`` is hundreds of times faster than the alternative solvers in Eigen. 

What do you think @yxlao @qianyizh

jornb:
Thanks for the response @syncle!

I'd just like to add the following option for consideration.

    3. Remove use of `Eigen::SimplicialCholesky` unless some flag, e.g. `ALLOW_LGPL_LIBRARIES`, is set.

germanros1987:
@syncle I would like to give some closure to this PR. Should we consider the proposed option: "Remove use of Eigen::SimplicialCholesky unless some flag, e.g. ALLOW_LGPL_LIBRARIES, is set." or should we close the PR?

germanros1987:
@syncle do you have an opinion? I really want to make a decision soon.

germanros1987:
The PR is almost 1 year old and nobody has done anything to bring it in. Unfortunately, this means there is not enough interest in the feature. I am going to proceed to close it. If the author or any reviewer wants to recover this PR, please feel free to open it again.

PR #945, opened 2019-05-02
Mesh Simplification

griegler:
Implemented a vertex cluster mesh simplification (using vertex average and quadric minimization) and "Surface Simplification Using Quadric Error Metrics". For testing, I added also a simple mesh subdivision method (midpoint). The color and normal information is averaged in all implementations and not considered in the error minimization.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/945)


griegler:
I added 

    if(input.HasTriangleNormals()) {
        mesh->ComputeTriangleNormals();
    }

to the two simplification and the one subdivision method.

yxlao:
great, thanks @griegler , LGTM now after CI passing

yxlao:
@syncle once approved we can merge

syncle:
LGTM. Thanks @yxlao @griegler!

PR #947, opened 2019-05-03
add octree python bindings

yxlao:
- All properties and methods except lambdas are bound. 
- Docstrings
- Python unit tests

This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/947)


yxlao:
thanks @syncle !

PR #952, opened 2019-05-05
Fixes Eigen::floor compiler error issue #935

### LINKED ISSUE #935: VortexGrid and GlobalOptimization compilation problem on Ubuntu
**Describe the bug**
Seems the compilation can not succeed, any idea to fix it?
The problem arises when I compiled the VortexGrid.cpp and GlobalOptimization.cpp.

**To Reproduce**
1. git clone --recursive https://github.com/intel-isl/Open3D
2. cd Open3D |  . util/scripts/install-deps-ubuntu.sh
3. mkdir build | cd build
4. cmake ..
3. make -j8 and meets the problem.

**Expected behavior**
Seems the compilation can not succeed, any idea to fix it?

**Screenshots**

image

 - OS: Ubuntu 18.04
 - Python version:3.6
 - Open3D version: The latest from git.
 - How did you install Open3D?: gcc build from source

**Additional context**
It looks like that there are many compatibility bugs hidden in Open3D, but it seems to be an almost perfect open source package except for the bugs. I hope I could be a contributor to fix them.

yxlao:
Fixes #935
Thanks @WenyinWei

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/952)

### DIFF
--- a/src/Open3D/Geometry/VoxelGrid.cpp
+++ b/src/Open3D/Geometry/VoxelGrid.cpp
@@ -83,7 +83,7 @@ VoxelGrid VoxelGrid::operator+(const VoxelGrid &voxelgrid) const {
 
 Eigen::Vector3i VoxelGrid::GetVoxel(const Eigen::Vector3d &point) const {
     Eigen::Vector3d voxel_f = (point - origin_) / voxel_size_;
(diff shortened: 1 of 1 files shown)


yxlao:
Closing since actual reason is that the issue is due to using the system's previously installed Eigen

Issue #951, opened 2019-05-04
Reader for TUM trajectories


martinruenz:
Hey there!

I noticed that there is no reader for trajectories in the TUM format, which parametrizes a pose with 8 parameters (`<timestamp> <x> <y> <z> <qx> <qy> <qz> <qw>`). See: https://vision.in.tum.de/data/datasets/rgbd-dataset/file_formats

Implementing a reader for this format is super straight forward and I can add it to Open3D if you want to. I would add a `FileTUM.cpp` (or `FileTXT.cpp`) and map the extension `.txt` to the function `ReadPinholeCameraTrajectoryFromTUM`.

What do you think?

martinruenz:
Implemented in #953.

PR #953, opened 2019-05-06
Add TUM trajectory format

martinruenz:
Reader for TUM trajectory files, as described here #951

The clang-format was applied and coding style is as in `FileLOG.cpp`.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/953)


yxlao:
I see. You're right, it will detect the file extension and call the corresponding TUM reader.

PR #954, opened 2019-05-09
Loop Subdivision

griegler:
Implementation of triangle mesh subdivision by Loop to increase number of triangles and smooth surface. 
Additionally, the code adds additional Platonic solids to the TriangleMeshFactory.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/954)


syncle:
Like this example: [Open3D/ColorMap/ColorMapOptimization.cpp](https://github.com/intel-isl/Open3D/blob/master/src/Open3D/ColorMap/ColorMapOptimization.cpp) we are having unnamed namespace for the minimal functions that are only used in specific functions. I was trying to see if we can replace this with lambda function.

syncle:
Yeah. Yes, It is. @yxlao: can you review this PR as well?

PR #956, opened 2019-05-09
Added Basic Transformation Methods (Translate, Scale, Rotate) to Geometry3D

griegler:
Added the basic transformation methods to Geometry3D and all supported sub-classes. Rotation is supported in various Euler formats for now. 
In addition, the added methods, as well as the existing `Transform(.)` method return now a reference to the object itself to allow chaining of the methods, e.g. `mesh.translate((1,0,0)).scale(0.5)`. Maybe this should be considered standard for void methods.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/956)


src/Open3D/Utility/Eigen.cpp
```diff
@@ -278,5 +278,26 @@ template std::tuple<Eigen::Matrix6d, Eigen::Vector6d, double> ComputeJTJandJTr(
         int iteration_num, bool verbose);
 // clang-format on
 
+Eigen::Matrix3d RotationMatrixX(double radians) {
+    Eigen::Matrix3d rot;
```
pmoulon: Eigen already provide the code for this see https://github.com/openMVG/openMVG/blob/master/src/openMVG/numeric/numeric.cpp#L46

PR #963, opened 2019-05-14
Extended TriangleMesh Properties and Tests

griegler:
Implemented `IsOrientable` and `OrientTriangles` in TriangleMesh that can be used to infer if a triangle mesh is orientable, i.e., triangles can be ordered such that the triangle normals point all towards the same direction.

To test the orientation methods, I also added a factory method to create a Moebius strip with arbitrary number of twists to the TriangleMeshFactory.

Further, I added `IsIntersecting` to test if a triangle mesh is intersecting another triangle mesh.

The example script `mesh_properties.py` showcases all those TriangleMesh property methods.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/963)


griegler:
Superseded by #977

Issue #238, opened 2018-03-23
remove 3rdparty folder


takanokage:
I propose we remove the src/External folder.
I want to avoid any future confusion if other projects will have to use some of the same external dependencies. It is good practice to reuse those dependencies across many projects rather than duplicating the code for each project.

Some of these dependencies belong to existing git repositories so it should be easy to pull the necessary version when building Open3D.

For other dependencies we can instead create mirror repositories (fork) under our team's umbrella and use proper versioning to match the original source. In this case we can match master to the progress made by the original teams and create our own branches when we want to fix/extend the code ourselves.

qianyizh:
The main purpose of ``src/External`` is to make Open3D Windows-friendly. I still want to keep it so that Windows users can just pull the repo and compile Open3D out of the box.

I think your proposal is to use ``git submodule`` to replace the hard copies of the external libraries. This is a good suggestion. I thought about this. There are pros and cons.

Pros as mentioned in OP: it is cleaner, easier to use across projects, and easier to track the latest versions.

Cons: it is harder to track the exact version we need. ``submodule`` can track the exact version we need. But I had the experience that I had to edit some files (e.g., the CMake file) or fix some hot bugs (submitted to the official repo but could not wait until PR gets merged). In these cases, it is always better to have my own version of the library. Thus it is either a hard copy in ``src/External`` or a fork of the original repo. I chose the former because it reduces the maintenance cost from my side, :)

I dislike ``src/External`` too. I am open for discussions of a better solution of this.

syncle:
One suggestion is to maintain another repo just for Open3D dependencies. That repo is just tested for Open3D. This will make Open3D repo much compact.

syncle:
Just for update ``src/External`` folder is moved into ``3rdparty`` folder.

qianyizh:
The discussion is followed up on discord channel and offline discussions. Feel free to hop in and contribute your thoughts. https://discord.gg/qsm2fd

syncle:
Addressed in #725.

PR #967, opened 2019-05-15
Merged Open3D-3rdparty repository into main repository

griegler:
1) It seems to be much cleaner to have the code in one place.
2) It greatly simplifies PRs that involve changes in the 3rdparty folder (like adding an additional module).

No files have been changed. The only relevant diff is in `.gitmodules`, the other files have just moved from the `Open3D-3rdparty` repository.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/967)


syncle:
Thanks for your effort @griegler! This PR is opposite to previous PR #725. I was not able to fully understand the two reasons you mentioned. Could you give some more example why this PR is necessary?

qianyizh:
hahaha, finally one more vote for my original design...

@griegler to give you some context. My original design was exactly the same as you are doing in this PR. However, Dan strongly argued against it in this thread #238 . As a result of compromise, we end up with a solution in #725 . Hence the current structure.

There are indeed pros and cons of each structure. We can discuss in our meetings.

griegler:
@syncle @qianyizh I have seen (after the initial PR) the discussion and understand the reasoning to move the 3rdparty source out of the main repository, but I think the current setting is not satisfactory at all. It is actually the very same as having all the source in the main repository, but with the major disadvantage that when we have a PR that involves 3rdparty code (see convex hull PR, or the triangle triangle intersection) that with have to merge two PRs independently that actually belong together. 

Therefore, I think that the larger 3rdparty projects should be separate submodules (as I tried in the current form of this PR). Then, (most of) the 3rdparty source is hosted independently as needed, while at the same time it is easy to contribute PRs that involve 3rdparty code. 

I am happy to discuss this further in our meeting. Anyway, the current PR is failing the integration tests.

yxlao:
hi @griegler, I pushed your changes to a backup branch https://github.com/griegler/Open3D/tree/submodules-backup. In this PR, I reverted several changes such that only header-only library like Pybind11 and Eigen are submodules. Let's get this merged this first to unblock other PRs. 

Later, it should be possible to convert other 3rd party libraries using custom `ExternalProject_Add` function. We can convert them one-by-one.

PR #983, opened 2019-05-23
Voxel grid to octree

yxlao:
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/983)


yxlao:
@griegler ok, let's discuss on Monday's meeting

PR #994, opened 2019-05-29
Added center parameter to rotate method

griegler:
I added a `center` parameter to the `Rotate` methods of the `Geometry3D` classes. If it is set to `true`, then the vertices/points are rotated in the centered 3D space.
In addition, I center the points/vertices in the `Scale` methods, otherwise it would do something different as most users might expect.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/994)


griegler:
@qianyizh

PR #1003, opened 2019-06-05
Geometry functions as class methods

griegler:
Refactoring the `geometry` API such that no more functions are used. Instead, most methods are now class methods. Exceptions are `create_*` functions, which are now static class methods. Changes are applied to C++ and Python API. Therefore, I also retested the Python examples (and found also a few other minor bugs).

I tried to be as consistent with the naming of the methods as possible.

Note: `HalfEdgeTriangleMesh` is/was in some kind of limbo state. Should be revisited in another PR.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1003)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,12 @@ docs/python_api/
 _vimrc_local.vim
 examples/TestData/Armadillo.ply
 examples/TestData/Bunny.ply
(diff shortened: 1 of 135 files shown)


yxlao:
For issue 3, I'll try to find a solution.

Issue #1016, opened 2019-06-11
Open3D read_point_cloud skips NaNs 


renuka-apte:
Open3D seems to remove NaNs forcibly while reading a pcd, which doesn't make sense for organized pcds. 

https://github.com/intel-isl/Open3D/blob/master/src/Open3D/IO/FileFormat/FilePCD.cpp#L768

**To repro**
```
pcd = o3d.io.read_point_cloud('file.pcd')
```

**Expected behavior**
Perhaps pass a flag that allows for the reading of organized point clouds. IOW, put that line ^ behind a flag.

qianyizh:
I will add it as a feature request and let @germanros1987 to put it in the roadmap.

Also yes, in the current version, we assume the point cloud is an unorganized point cloud without NaNs. A flag to enable NaNs would be good.

griegler:
Addressed in #1022

PR #1022, opened 2019-06-14
Added option to remove NaN points in ReadPointCloud

griegler:
This addresses issue #1016. I added a parameter to control if NaNs should be removed after reading a `PointCloud`, or not. The default is set to `true`, such that the function is backwards compatible. 

Note that this fixes also a subtle bug in the `RemoveNaN` function that has so far been unnoticed.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1022)


griegler:
@yxlao

PR #1032, opened 2019-06-20
added a new function that allows to assign different colors to each po…

liubigli:
added a new function that allows to assign different colors to each point in the point cloud.
This could be useful to visualize a segmented point cloud where each label correspond to a color.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1032)


liubigli:
Yes you are totally right! I am sorry but I did not find this feature in the documentation. I cancel the pull request.

PR #1043, opened 2019-06-24
CMake, Registry, Tensor and Array for CUDA

yxlao:
- CMakes for enabling/disabling CUDA build
    - Enable with `BUILD_CUDA_MODULE=ON` flag
- Registry for static initialization
    - Eliminate conditional compilation of GPU code with `#ifdef`, instead, CPU/GPU classes/functions register themselves to the registry and available backends can be queried from the registry
    - Includes: Class registry, function registry, and singleton registry for globally unique objects
- Initial `Tensor` and `Array` class that uses `MemoryManager`, only the basic framework is implemented, details will be added later
- Unit tests
- Merging to `gpu` branch, not `master` branch

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1043)


yxlao:
more contexts on the registration pattern: https://docs.google.com/presentation/d/1nAu0ctKF04YGe71xJDVryvgqXsPuyeyVrJBOKF18LqE/edit?usp=sharing

PR #1046, opened 2019-06-27
Feature/asana tools

germanros1987:
Script to synchronize GitHub issues into ASANA tasks. For security reasons the json file to set up the environment will remain secret.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1046)


syncle:
@yxlao what do you think about this PR? Shall we merge?

PR #1060, opened 2019-07-07
Triangulation by ear clipping

kongsgard:
This PR adds support for loading meshes with non-convex faces.
The implementation is partly based on the ear clipping approach used in the 3rdparty header `tinyobjloader.h` used to load .obj files. The point inclusion in polygon test is acquired from [here](https://stackoverflow.com/a/43896965).

Currently, the triangulation by ear clipping algorithm is only added to the .off-loader. Additionally, it should be added to loading .ply files, but I'm not sure what would be the best approach to do this. .obj-files are handled by `tinyobjloader.h` and .stl-files are triangular representations, per the specifications.

Example triangulations are given in the illustration below:
Screenshot 2019-07-07 at 22 50 18

The chosen ear to base the triangulation on is the first one found in each step (assuming counterclockwise ordering of vertices). This, as seen in the right polygon above, results in narrow triangles for (almost) convex polygons. If desirable, the implementation could be slightly changed to search for an "optimal" ear (which would be the most extruded one). 

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1060)

### DIFF
--- a/src/Open3D/IO/ClassIO/TriangleMeshIO.cpp
+++ b/src/Open3D/IO/ClassIO/TriangleMeshIO.cpp
@@ -133,20 +133,30 @@ bool WriteTriangleMesh(const std::string &filename,
     return success;
 }
 
-int PointInclusionInPolygonTest(unsigned int nvert,
-                                double *vertx,
-                                double *verty,
-                                double testx,
-                                double testy) {
-    int i, j, c = 0;
-    for (i = 0, j = nvert - 1; i < nvert; j = i++) {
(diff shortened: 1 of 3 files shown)


kongsgard:
Additionally, should unit tests be added to [Open3D/src/UnitTest/IO/ClassIO/TriangleMeshIO.cpp](https://github.com/intel-isl/Open3D/blob/master/src/UnitTest/IO/ClassIO/TriangleMeshIO.cpp)?

kongsgard:
Thanks for the feedback!
I'll leave support for `ply` files to another PR.

PR #1069, opened 2019-07-11
Improved Logging/Print

griegler:
I did a rewrite and clean-up of the current Logging/Print functionalities of the library. The logging/printing is now based on a small, lightweight library [fmt](https://github.com/fmtlib/fmt) (included as submodule). This library would allow to also print custom types (e.g., Eigen, or future Tensor class). 

In addition I went over all uses of Print in the code base and replaced them with the most sensible if needed (Error, Warning, Info, Warning). They should be now consistent. Further, the print functions are now part of a `Logger` class, that is realized as a singleton (to avoid the global static variables).

I also did a rewrite of the progress bar. Again, I removed the gloabl static variables, and implemented a `ConsoleProgressBar` class that is now used instead. This class also has an `active` flag that is utilized in the `IO` classes to activate/deactivate the progress bar (default is, that the bar is not active anymore in those classes to not clutter the output of the user).

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1069)


yxlao:
Thanks @griegler , I do think that we should refactor our logging system. The `fmt` library looks nice, however, that would require us to use non C/C+ standard formatting syntax.

Another option is to use the C++ stream. In TF, there are two types: [`VLOG`](https://github.com/tensorflow/tensorflow/blob/7c4167188208ef9041b4938ac883d90970f61644/tensorflow/core/platform/default/logging.h#L119-L124) and [`LOG`](https://github.com/tensorflow/tensorflow/blob/7c4167188208ef9041b4938ac883d90970f61644/tensorflow/core/platform/default/logging.h#L87-L96).

```cpp
// VLOG for looging only
VLOG(0) << "foo";  // Automatic append endl
VLOG(1) << "bar";

// LOG may have control, e.g. Fatal will exit
LOG(INFO) << "small issue";
LOG(FATAL) << "big error";
```

Probably using the C++ stream syntax can be more friendly for the user. Additionally, it already supports Eigen, and we can add `ostream& operator<<` to our custom class to be printed.

griegler:
I am personally not really a fan of the stream approach, as it is quite clunky to format for example floats. The `fmt` library also supports standard C formatting, cf. [here](https://fmt.dev/latest/api.html#printf-formatting). So, I could add the functions also to the `Logger` class, if wanted.

griegler:
Added `Log*f` methods that accept `printf` like format strings. Further, I added a simple `Log.cpp` to the examples. How do you like this @yxlao?

griegler:
@yxlao CI is now passing after changes in the includes.

yxlao:
Thanks @griegler !

PR #1075, opened 2019-07-17
s/Debug/Warning

barnjamin:
Took me a while to figure out why I wasn't getting any points in my point cloud from a single byte depth image.  Might also make sense to add something mentioning that depth images need to be at least 2 bytes per pixel.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1075)


griegler:
Thanks @barnjamin for the PR. There is a PR (#1069) currently open that refactors the Logging/Printing framework. This also cleans up the Warning/Debug/Error messages and your changes are actually already included in this PR.

barnjamin:
Ah okay, thanks!

PR #1079, opened 2019-07-19
remove TSDFVoxelGrid class

### LINKED ISSUE #1078: UniformTSDFVolume::Reset does not clear voxels
**Describe the bug**
Calling `UniformTSDFVolume::Reset` won't actually reset the volume

**To Reproduce**
1. Create a UniformTSDFVolume and integrate some frames
2. Call `volume.Reset();`
3. Visualize somehow the volume, e.g. `open3d::visualization::DrawGeometries({ volume.ExtractPointCloud() });`

**Expected behavior**
I expect the volume to be empty.
In the previous example, I would expect an empty visualization window, but I still see the content up to the frames I integrated.

 - OS: Debian Testing (11/Bullseye)
 - Python version: 3.7.3 (but tested in C++)
 - Open3D version: last git commit
 - How did you install Open3D? build_from_source (GCC (Debian 8.3.0-6) 8.3.0)

**Additional context**
I think that overriding the `Clear` method also in `TSDFVoxel` (in `src/Open3D/Integration/UniformTSDFVolume.h`) would solve this. The problem is that the `voxels_` member is redefined, so the `Clear` of `Voxel` just clears its own voxels, but not the ones of the child class.
As a temporary solution I managed to solve the problem by calling `volume.voxel_grid_.voxels_.clear();` before `volume.Reset();`.

yxlao:
Fixes #1078 , since it is a dense voxel grid, we can simply use a `vector`.
As discussed, in the future, we will 1) split the voxel grid class to `DenseVoxelGrid` and `SparseVoxelGrid`, 2) use continuous buffers for each property separately, instead of a vector of voxels that contains different properties.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1079)

### DIFF
--- a/src/Open3D/Integration/ScalableTSDFVolume.cpp
+++ b/src/Open3D/Integration/ScalableTSDFVolume.cpp
@@ -119,13 +119,10 @@ std::shared_ptr<geometry::PointCloud> ScalableTSDFVolume::ExtractPointCloud() {
                 for (int y = 0; y < volume0.resolution_; y++) {
                     for (int z = 0; z < volume0.resolution_; z++) {
                         Eigen::Vector3i idx0(x, y, z);
-                        w0 = volume0.voxel_grid_.voxels_[volume0.IndexOf(idx0)]
-                                     .weight_;
-                        f0 = volume0.voxel_grid_.voxels_[volume0.IndexOf(idx0)]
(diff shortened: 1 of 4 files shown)


griegler:
LGTM after CI passes.

PR #1082, opened 2019-07-22
FileGLTF for TriangleMesh

kongsgard:
Added support for [glTF file format](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0).
Support for the alternative GLB file format, which is the binary version of glTF, is also added.

`tinygltf` is added as a submodule, in a similar manner as `tinyobjloader` for obj files.
Currently, only vertices, vertex normals, and vertex colors are supported.
Moreover, only the first mesh given in the glTF file will be read.

The following improvements can be added to a future PR:
- Read all meshes given in a glTF file by combining them into one mesh when reading the file
- Optimized writing of indices (store as `uint16_t` if `number_of_vertices <= 65536`)

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1082)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -22,3 +22,6 @@
 [submodule "3rdparty/fmt"]
 	path = 3rdparty/fmt
 	url = https://github.com/fmtlib/fmt.git
+[submodule "3rdparty/tinygltf"]
+	path = 3rdparty/tinygltf
+	url = https://github.com/syoyo/tinygltf.git
--- a/3rdparty/CMakeLists.txt
+++ b/3rdparty/CMakeLists.txt
@@ -197,6 +197,9 @@ else ()
     message(SEND_ERROR "TINYFILEDIALOGS dependency not met.")
 endif ()
 
+# tinygltf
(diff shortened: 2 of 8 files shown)


griegler:
> * Combining multiple meshes into one would result in a triangle soup, right? With this approach, I've kept a user warning stating that the meshes are combined into one.
>   (Preferably, the meshes would be loaded separately, but I'm not sure how we would do that in the current framework. Any other alternatives?)

I think there is no real alternative right now, without refactoring the IO framework in general. Combining them should be ok for the moment?

> * Is there a function to rotate a mesh using quaternions? For now, I've used the approach outlined [here](https://github.com/intel-isl/Open3D/blob/f4ee6189cd3b1980d6a58f438e41a43e94057ed7/src/Open3D/IO/FileFormat/FileTUM.cpp#L75).

This approach is good. Maybe we can add a helper function to `utility`, like we have for the rotation matrices (https://github.com/intel-isl/Open3D/blob/master/src/Open3D/Utility/Eigen.h#L118).

> * Is there a function to scale each axis x, y, and z of a mesh independently?

No, there is none. But it should be easy todo with transform, where `T` is a diagonal matrix with the scale factors on the diagonal.

PR #1094, opened 2019-07-30
Recorder

theNded:
Merge RGBDSensor core

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1094)


theNded:
Thanks for the review!
As this is a temporary branch, many implementations are still scratches, like the hard-coded paths and TODOs -- our current plan is to synchronize over this temp branch during our development, but maybe it is still better to polish the code before a PR.

We've noticed most of the problems:
1. We will remove the libjpeg and use libjpeg-turbo instead, as a submodule. Also, @yxlao has double checked AzureKinectSDK's build system, in which `jpegturbo` works reasonably efficient without the NASM compiler.
2. Regarding the naming, German's plan is to first implement the classes using AzureKinect's variable, and later adapt to others by refactoring. 
3. In [cmake](https://github.com/theNded/Open3D/blob/recorder/src/Open3D/IO/CMakeLists.txt#L9) we specify that it will only be built with Kinect library enabled.
4. Other minor problems will be fixed soon.

yxlao:
thanks @griegler for the review, we'll merge this to the recorder branch and fix the issues when it is ready for master

PR #1107, opened 2019-08-08
Enables visualization of RGBD images

theNded:
## Implemented
- Turns the `RGBDImage` class to a `Geometry2D` object to support side-by-side visualization 
- Image is rendered in two passes within one shader (Jet colormap is built-in glsl), may improve performance for the AzureKinect recorder
- Updated the visualizer example accordingly

## Problem
- Now depth-max value is specified by `option.image_max_depth_ = 3000`, while depth-scale is set to default 1000. It seems `option` cannot be easily set when calling `DrawGeometries`. Shall we consider maintaining `depth_scale_` and `depth_max_` properties in `RGBDImage`?

## Results
Intensity + uint16 depth
Intensity + uint16 depth

RGB + float 
RGB + float depth

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1107)


theNded:
@germanros1987 Yes, it's better if we finalize the design in this PR.

Issue #1096, opened 2019-07-30
Meshes are not visualized properly


kongsgard:
**Describe the bug**
Meshes are not visualized properly, as there is some clipping and missing triangles.

**To Reproduce**
```
# Load e.g. knot.ply in TestData
mesh = o3d.io.read_triangle_mesh("knot.ply")
o3d.visualization.draw_geometries([mesh])
```
Showing the wireframe also highlights the issues.

**Expected behavior**
Visualize the mesh as in Open3D 0.7.0.

**Screenshots**
<img width="1054" alt="Screenshot 2019-07-30 at 19 15 25" src="https://user-images.githubusercontent.com/13556755/62150493-69eba900-b2fe-11e9-8e50-3b606e739e0f.png">

**Environment:**
 - OS: macOS Mojave
 - Python version: 3.6
 - Open3D version: -
 - Is this remote workstation?:  no
 - How did you install Open3D?: build_from_source

**Additional context**
I tested previous commits, and it seems like the bug was introduced in #1080.

griegler:
Seems to be a regression of #1080. I will have a look at it.

griegler:
Fixed in #1122

PR #1080, opened 2019-07-22
Set z_near_ close to 0

griegler:
Set `z_near_` close to `0`.

Before:
ScreenCapture_2019-07-22-13-34-18

Afterwards:
ScreenCapture_2019-07-22-13-35-45

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1080)

PR #1122, opened 2019-08-19
Revert #1080 and added method for constant z planes

griegler:
#1080 introduced problems in the rendering of meshes, see for example #1096. This PR reverts those changes.
In addition, I added a methods to set/unset constant z planes of the visualizer, i.e., that are independent of the zoom and the bounding box size.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1122)

### DIFF
--- a/examples/Python/Basic/meshes.py
+++ b/examples/Python/Basic/meshes.py
@@ -96,20 +96,28 @@ def intersecting_boxes():
     return mesh
 
 
+def _relative_path(path):
+    script_path = os.path.realpath(__file__)
+    script_dir = os.path.dirname(script_path)
+    print(script_dir)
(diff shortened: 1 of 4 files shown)


examples/Python/Basic/meshes.py
```diff
@@ -96,20 +96,28 @@ def intersecting_boxes():
     return mesh
 
 
+def _relative_path(path):
+    script_path = os.path.realpath(__file__)
+    script_dir = os.path.dirname(script_path)
+    print(script_dir)
```
yxlao: remove print

griegler:
LGTM. @yxlao ready to merge?

Issue #1056, opened 2019-07-03
Is there an upsampling method or surface interpolation method? 


lvgeng:
Just wondering if there is a upsampling method or surface interpolation method.

I wish to interpolate a relatively smooth curved surface from several sampled points.

Convert a sparse point cloud to a dense point cloud and fix the holes at the same time.

Seems to be something like:

sparse point cloud -> reconstructed mesh -> dense point cloud ?

Any suggestions?

griegler:
I think for the suggested pipeline we are missing the middle part: reconstruction of the mesh. We are working on it, but for now you would need some other library for that.

lvgeng:
> I think for the suggested pipeline we are missing the middle part: reconstruction of the mesh. We are working on it, but for now you would need some other library for that.

Any suggestions for that? Like which python library might be suitable for it?
Or... is is possible to skip the mesh part, just do a sparse point cloud to dense point cloud interpolation?

lvgeng:
Or probably other methods? Does open3d have a smoothing method?

for point in sparse_point_cloud
      make a square planar point cloud p_sub at its position
      rough dense point cloud += p_sub
smoothed_dense_point_cloud = smooth(rough dense point cloud)
dense_point_cloud = downsample(smoothed_dense_point_cloud)

griegler:
#1130 implements the BPA.  You could now use it to reconstruct the mesh surface and then densely sample the mesh, i.e., with Poisson disk sampling.

Oberschnulli:
Does somebody have an implementation for that approach by chance?

Issue #270, opened 2018-04-02
reconstruct surface with triangle mesh in python


rarilurelo:
I'd like to reconstruct surfaces of point cloud `.pcd` file and convert it to mesh `.ply` file.

```
import argparse
from py3d import *
if __name__ == "__main__":
    parser = argparse.ArgumentParser()
    parser.add_argument('--file', type=str)
    args = parser.parse_args()
    pcd = read_point_cloud(args.file)
    draw_geometries([pcd])
    mesh = TriangleMesh()
    mesh.vertices = pcd.points
    mesh.triangles = triangle_mesh_surface(pcd.points)
    draw_geometries([mesh])
    write_triangle_mesh('out.ply', mesh)
```

Is there any function like `triangle_mesh_surface()` in Open3D?

syncle:
Nope. We don't have meshing algorithm from raw point cloud yet. I think good candidate is [poisson surface reconstruction](https://github.com/mkazhdan/PoissonRecon), or you can suggest any. We can consider as a next milestone.

As a side note, We do have an implementation for mesh from TSDF volume. [Check out this link.](http://open3d.org/docs/tutorial/Advanced/rgbd_integration.html)

griegler:
There is now a BPA implementation for surface reconstruction in Open3D, #1130.

PR #1130, opened 2019-08-20
Implementation of Ball Pivoting algorithm

griegler:
Includes:
* Implementation of BPA
* Example script on some synthetic point clouds
* New intersection tests in IntersectionTests
* Some unit tests

Addresses #1056 and #270,

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1130)


qianyizh:
This is awesome. I have been wanting it for a very very long time :D

ibrahimLearning:
@yxlao @benjaminum @griegler 
Anyone tell me please for point cloud data with x y z x_nromal y_normal z_normal, How can I generate the points for ball pivoting algorithm?
Because in this case I am confused about the loop?

```
 for x in range(4):
        for y in range(4):
            points.append((np.random.uniform(-2, 2), np.random.uniform(-2,
```
Thanks

griegler:
@ibrahimLearning sorry, but I do not get your question. You already have a point cloud with points and normals? Then it should be straight forward to use this algorithm? Where does this code snippet come from?
On another note, for questions it would be better to open an issue.

ibrahimLearning:
@griegler 
TriangleSurface

Thank you very much for your reply. I think you can solve this problem. Please see the image below after calling ball pivoting algorithm, why all points are not connected?:

griegler:
Maybe the normals do not agree (pointing to the same direction).

Issue #1068, opened 2019-07-10
Minimum bounding box


kongsgard:
**Is your feature request related to a problem? Please describe.**
Computation of the minimum bounding box for a mesh or point cloud.

**Describe the solution you'd like**
An API similar to the computation of convex hulls:
```
bbox = mesh.compute_minimum_bounding_box()
bbox_ls = o3d.geometry.LineSet.create_from_triangle_mesh(bbox)
```

**Describe alternatives you've considered**
Possibly, the following variations could be added:
- Computation of axis-aligned minimum bounding box, arbitrarily oriented minimum bounding box, and object-oriented minimum bounding box. 
- Add a function to compute the minimum bounding sphere.

PR #1131, opened 2019-08-21
Bounding Volume

griegler:
Implementation of `geometry` classes `OrientedBoundingBox` and `AxisAlignedBoundingBox`. The latter replaces also the utility class in `visualization`. Both can be computed from the existing `geometry` types, although for the `OrientedBoundingBox` I do not compute the exact minimum volume box.
Partly addresses #1068

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1131)


yxlao:
build restarted for macos

griegler:
thanks for the macos fix.

PR #1132, opened 2019-08-21
K4a with dlopen

yxlao:
- Complete `dlopen` approach
- Stub k4a interface built-in
- Use macros to quickly define k4a functions
- Do not need to link libk4a.so or libk4arecord.so, they will be loaded at runtime
- Python package no longer carry the shared libs

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1132)


theNded:
LGTM. In the next release we may split k4a_playback and k4a core as we will implement our own mkv parser.

PR #1140, opened 2019-08-23
color map knn interpolate for invisible point

yxlao:
before: 
Screenshot from 2019-08-23 16-34-49

after:
Screenshot from 2019-08-23 16-31-13

This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1140)


theNded:
LGTM. In the future we need to refactor this component into a class for easier configuration.

yxlao:
thanks @theNded , agreed, we shall refactor this to a class

Issue #1142, opened 2019-08-26
Apply transformation to PointCloud / Mesh


minihops123:
Hi,
I loaded a point cloud and tried to scale the point cloud by applying a transformation. 
For scaling the transformation matrix looks in the following:
[[1 0 0 0];
[0 1 0 0];
[0 0 1 0];
[0 0 0 s]];
where s is the scaling factor. 
Unfortunately this is ignored in applying the transformation to the pointcloud.
The error seems to be here:

https://github.com/intel-isl/Open3D/blob/af78edd6fa573aac72a16586f7b87f0521a98fcb/src/Open3D/Geometry/PointCloud.cpp#L84

Here, the transformed point is not normalized by the last entry of the homogenous vector but only the the first elements are returned.

        point = new_point.block<3, 1>(0, 0);

so one single line is missing:
new_point = new_point / new_point(3)

The same issue may arrise in the mesh.

griegler:
has been fixed in #1143

PR #1143, opened 2019-08-26
Transform Fix

griegler:
This PR fixes the problem observed in #1142, i.e., the point/vertex is not normalized by the 4th component after applying a transformation.


griegler:
@yxlao could you please restart travis.

PR #1144, opened 2019-08-26
Point Mapping in ConvexHull and Hidden Point Removal Operator

cosama:
This pull request will add two new features:

1) The call to `point_cloud.compute_convex_hull` can contain an optional argument of an IntVector, which will be filled with the mapping between the point in the mesh (returned) and the points in the original point cloud. One possible application of this is to retain additional information in the mesh:
```
pc = o3d.geometry.PointCloud()
pc.points = points  # a numpy array of  x, y, z values
pc.colors = colors  # a numpy array of r, g, b values

pt_map = o3d.utility.IntVector()
mesh = pc.compute_convex_hull(pt_map=pt_map)
mesh.vertex_colors = colors[np.asarray(pt_map)]
```

2) Implement the Hidden Point Removal operator, discussed in [Katz et. al. 'Direct Visibility of Point Sets', 2007](https://cgm.technion.ac.il/Computer-Graphics-Multimedia/Publications/Papers/2007/2007_Direct_Visibility_of_Point_Sets.pdf) as a member function of the point_cloud class. With the above change this could be achieved in pure python, but I think this is a very useful feature.

Todo (not sure what would be suitable):
- [ ] Add tests
- [x] Add examples
- [ ] Add optional argument to retain the `camera` vertex in the mesh

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1144)

### DIFF
--- a/examples/Python/Basic/convex_hull.py
+++ b/examples/Python/Basic/convex_hull.py
@@ -4,15 +4,7 @@
 
 # examples/Python/Basic/convex_hull.py
 
-import numpy as np
-import os
-import urllib.request
-import gzip
-import tarfile
-import shutil
-import time
 import open3d as o3d
-
 import meshes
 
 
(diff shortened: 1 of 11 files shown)


griegler:
Thanks! The changes look good to me. @yxlao can you please have another look.

yxlao:
sorry for the delay, looks good to me, thanks @cosama !

PR #1147, opened 2019-08-28
DBSCAN speed up

griegler:
This PR speeds up DBSCAN by pre-computing the neighbors using omp.

I further added a `GetCenter` method to `Geometry3D` and a `relative` parameter to `Translate` to allow to move the center to a specific position.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1147)

### DIFF
--- a/src/Open3D/Geometry/HalfEdgeTriangleMesh.cpp
+++ b/src/Open3D/Geometry/HalfEdgeTriangleMesh.cpp
@@ -282,6 +282,16 @@ Eigen::Vector3d HalfEdgeTriangleMesh::GetMaxBound() const {
     return Eigen::Vector3d((*itr_x)(0), (*itr_y)(1), (*itr_z)(2));
 }
 
+Eigen::Vector3d HalfEdgeTriangleMesh::GetCenter() const {
+    Eigen::Vector3d center(0, 0, 0);
+    if (!HasVertices()) {
+        return center;
+    }
+    center = std::accumulate(vertices_.begin(), vertices_.end(), center);
+    center /= double(vertices_.size());
+    return center;
+}
+
 AxisAlignedBoundingBox HalfEdgeTriangleMesh::GetAxisAlignedBoundingBox() const {
     return AxisAlignedBoundingBox::CreateFromPoints(vertices_);
 }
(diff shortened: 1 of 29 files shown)


src/Open3D/Geometry/HalfEdgeTriangleMesh.cpp
```diff
@@ -282,6 +282,16 @@ Eigen::Vector3d HalfEdgeTriangleMesh::GetMaxBound() const {
     return Eigen::Vector3d((*itr_x)(0), (*itr_y)(1), (*itr_z)(2));
 }
 
+Eigen::Vector3d HalfEdgeTriangleMesh::GetCenter() const {
+    Eigen::Vector3d center(0, 0, 0);
+    if (!HasVertices()) {
+        return center;
+    }
+    center = std::accumulate(vertices_.begin(), vertices_.end(), center);
```
pmoulon: Would not be faster to use Eigen vectorization with colwise operations?

src/Open3D/Geometry/HalfEdgeTriangleMesh.cpp
```diff
@@ -316,9 +326,13 @@ HalfEdgeTriangleMesh &HalfEdgeTriangleMesh::Transform(
 }
 
 HalfEdgeTriangleMesh &HalfEdgeTriangleMesh::Translate(
-        const Eigen::Vector3d &translation) {
+        const Eigen::Vector3d &translation, bool relative) {
+    Eigen::Vector3d transform = translation;
```
pmoulon: I would Eigen map the vector to a Mat3X array and apply the transform the the mapping.

src/Open3D/Geometry/LineSet.cpp
```diff
@@ -63,6 +63,16 @@ Eigen::Vector3d LineSet::GetMaxBound() const {
             });
 }
 
+Eigen::Vector3d LineSet::GetCenter() const {
+    Eigen::Vector3d center(0, 0, 0);
+    if (!HasPoints()) {
+        return center;
+    }
+    center = std::accumulate(points_.begin(), points_.end(), center);
```
pmoulon: Same as before, Map the memory to Eigen matrice and use Eigen optimized functions

src/Open3D/Geometry/HalfEdgeTriangleMesh.cpp
griegler: Could you provide an example on how to map `std::vector<Eigen::Vector3d>` to `Eigen::Map<Eigen::MatrixXd>` ?

griegler: I tried the code below, but not sure if the mapping is 100% correct.

```
Eigen::Map<Eigen::Matrix<double, Eigen::Dynamic, 3>> points((double*)points_.data(), points_.size(), 3);
center = points.colwise().mean()
```

However, given this code the mean of 1e8 points is computed in `0.1582[s]` (avg of 20 reps). In the current version it took `0.1636[s]`. The difference does not seem to be too big.

Issue #909, opened 2019-04-13
point.paint_uniform_color causes Segmentation fault and ipython crash


barlaensdoonn:
**IMPORTANT: Please use the following template to report the bug.**
_____

Using point_cloud.paint_uniform_color causes Segmentation fault and ipython to crash and exit

**To Reproduce**
```
import open3d

def load_o3d_and_viz(infile, gray=False):
    if type(infile) == str:
        vz_cloud = open3d.read_point_cloud(infile)
    if gray:
        vz_cloud = vz_cloud.paint_uniform_color([101, 101, 101])

    open3d.draw_geometries([vz_cloud], width=1600)

vz_out = 'tmp/table_scene_lms400.ply'

# behaves as expected
load_o3d_and_viz(vz_out)

# this causes "Segmentation fault (core dumped) message and ipython to exit
load_o3d_and_viz(vz_out, gray=True)
```

**Screenshots**
![paint_issue](https://user-images.githubusercontent.com/5253980/56086154-7f979a00-5e05-11e9-92ac-25ae6205712e.png)

**Environment (please complete the following information):**
 - OS: Ubuntu 16.04
 - Python version: 3.5.2
 - Open3D version: 0.5.0.0
 - Is this remote workstation?: no
 - How did you install Open3D?: pip

barlaensdoonn:
you can find the .pcd file referenced in [this PCL tutorial](http://pointclouds.org/documentation/tutorials/cluster_extraction.php#cluster-extraction)

qianyizh:
I think color is a double range from 0 to 1.
Can you try ``paint_uniform_color([0.5, 0.5, 0.5])``

qianyizh:
@yxlao is this documented?

singleye:
I had the same problem recently, which version of numpy do you use?

'paint_uniform_color' and assign 'colors' with 'Vector3dVector' will cause segmentation fault on numpy 2.0.1 (MacOS, python 3.9.6), after downgrade numpy to 1.26.4, the problem is gone.

PR #1150, opened 2019-08-30
Geometry Code Simplification and PaintUniformColor Warning

griegler:
I simplified (and removed code duplication) from the `Geometry3D` classes. 
Additionally, I addressed issue #909. If `color` is outside the valid range of `[0, 1]` the method prints a warning and clips the color value to the valid range.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1150)


yxlao:
looks great

PR #1155, opened 2019-09-03
suppress 3rd-party lib warnings

yxlao:
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1155)

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -33,6 +33,7 @@ def get_git_short_hash():
 
 
 def get_git_tag_short_hash(tag):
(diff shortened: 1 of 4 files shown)


docs/conf.py
```diff
@@ -33,6 +33,7 @@ def get_git_short_hash():
 
 
 def get_git_tag_short_hash(tag):
+    _ = subprocess.check_output(['git', 'fetch', '--tags'])
```
griegler: This line has no effect?

PR #1160, opened 2019-09-05
Azure Kinect build from source on Ubuntu 16.04

yxlao:
- Allow Ubuntu 16.04 to build azure kinect support when `open3d-azure-kinect-ubuntu1604` pip package is installed. This is useful since the released Open3D is built on Ubuntu 16.04 in order to be compatible with both 16.04 and 18.04. 
- Simplify `azure_kinect.cmake`: `k4a_INCLUDE_DIRS` is all we need at build time, no library or lib paths are needed
- Added the option for Windows users to specify `cmake -DK4A_INCLUDE_DIR=...` at at CMake config time
- Added the option for Windows user to specify `K4A_LIB_DIR` environment variable to point to K4a DLLs, e.g.
    ```
    set K4A_LIB_DIR=C:\Program Files\Azure Kinect SDK v1.2.0\sdk\windows-desktop\amd64\release\bin
    ```
- `std::map` instead dof `std::unorderd_map` for enum to string map, to be compatible with older compilers.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1160)

### DIFF
--- a/3rdparty/azure_kinect/azure_kinect.cmake
+++ b/3rdparty/azure_kinect/azure_kinect.cmake
@@ -1,3 +1,9 @@
+if (BUILD_AZURE_KINECT AND APPLE)
+    message(WARNING "Azure Kinect is not supported on macOS, setting BUILD_AZURE_KINECT to OFF")
+    set(BUILD_AZURE_KINECT OFF)
+    set(BUILD_AZURE_KINECT OFF PARENT_SCOPE)
+endif()
+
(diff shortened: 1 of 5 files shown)


yxlao:
Thanks @griegler 

I've pushed several updates that

- fixed the typo
- moved ``find_k4a_with_ubuntu_1604_pip_package`` as a function in cmake
- updated the `.rst` docs for this PR

Issue #1171, opened 2019-09-11
Wrong self intersecting triangles in triangle mesh


ShayGus:
I export a non intersecting triangulated mesh from blender as a ply file.
Then import it into open3d in python (open3d-python ver. 0.7.0) as a TriangleMesh.
Then run a get_self_intersecting_triangles() on this mesh: 

mesh. get_self_intersecting_triangles()

I get many intersecting triangles although, there are no intersections in the ply file we export.

Is there any thing else I should take into consideration?
Any additional info I can provide to find the issue?

griegler:
Can you provide a triangle mesh with the described problem to aid debugging?

ShayGus:
Sure here it is:
[patch_square.ply.zip](https://github.com/intel-isl/Open3D/files/3604046/patch_square.ply.zip)



And here's the text of it:

ply
format ascii 1.0
comment Created by Blender 2.79 (sub 0) - www.blender.org, source file: ''
element vertex 392
property float x
property float y
property float z
element face 670
property list uchar uint vertex_indices
end_header
0.353972 0.040076 1.373050
0.354228 0.043402 1.379656
0.357662 0.042552 1.376781
0.350225 0.037071 1.369384
0.350484 0.040982 1.375864
0.346675 0.038054 1.372114
0.346917 0.041968 1.378729
0.350727 0.044241 1.382615
0.339490 0.044284 1.384593
0.339641 0.047509 1.391732
0.343437 0.046323 1.388634
0.335503 0.041491 1.380698
0.335619 0.045560 1.387630
0.353694 0.036240 1.366655
0.357374 0.039373 1.370215
0.361012 0.041926 1.373882
0.373269 0.013790 1.448147
0.377677 0.004503 1.446140
0.376176 0.014255 1.444252
0.370163 0.013940 1.452037
0.374122 0.003626 1.449932
0.370509 0.003833 1.453747
0.373295 -0.007260 1.450395
0.377916 -0.007588 1.446012
0.343043 0.039137 1.374894
0.343250 0.043110 1.381614
0.339319 0.040303 1.377741
0.347130 0.045242 1.385594
0.344452 -0.036410 1.457586
0.351935 -0.030909 1.456583
0.348562 -0.029913 1.460394
0.355952 -0.023475 1.459091
0.352542 -0.022371 1.462828
0.359435 -0.014710 1.460805
0.355932 -0.013645 1.464452
0.325316 -0.049652 1.456557
[...]

PR #1177, opened 2019-09-13
Improve numeric stability of TriangleTriangle3d

griegler:
It seems that the current triangle triangle intersection test has numerical issues (#1171). Therefore, this PR adds a normalization step to this collision test.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1177)

### DIFF
--- a/src/Open3D/Geometry/IntersectionTest.cpp
+++ b/src/Open3D/Geometry/IntersectionTest.cpp
@@ -60,7 +60,8 @@ bool IntersectionTest::TriangleTriangle3d(const Eigen::Vector3d& p0,
               (p2 - mu).array().square() + (q0 - mu).array().square() +
               (q1 - mu).array().square() + (q2 - mu).array().square()) /
              5)
-                    .sqrt();
+                    .sqrt() +
+            1e-12;
     Eigen::Vector3d p0m = (p0 - mu).array() / sigma.array();
     Eigen::Vector3d p1m = (p1 - mu).array() / sigma.array();
     Eigen::Vector3d p2m = (p2 - mu).array() / sigma.array();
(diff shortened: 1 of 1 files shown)


src/Open3D/Geometry/IntersectionTest.cpp
```diff
@@ -54,12 +54,21 @@ bool IntersectionTest::TriangleTriangle3d(const Eigen::Vector3d& p0,
...
                                           const Eigen::Vector3d& q1,
                                           const Eigen::Vector3d& q2) {
-    return NoDivTriTriIsect(const_cast<double*>(p0.data()),
-                            const_cast<double*>(p1.data()),
-                            const_cast<double*>(p2.data()),
-                            const_cast<double*>(q0.data()),
-                            const_cast<double*>(q1.data()),
-                            const_cast<double*>(q2.data())) != 0;
+    const Eigen::Vector3d mu = (p0 + p1 + p2 + q0 + q1 + q2) / 6;
+    const Eigen::Vector3d sigma =
```
yxlao: add a small `epsilon` before sqrt?

src/Open3D/Geometry/IntersectionTest.cpp
```diff
@@ -54,12 +54,21 @@ bool IntersectionTest::TriangleTriangle3d(const Eigen::Vector3d& p0,
...
+            (((p0 - mu).array().square() + (p1 - mu).array().square() +
+              (p2 - mu).array().square() + (q0 - mu).array().square() +
+              (q1 - mu).array().square() + (q2 - mu).array().square()) /
+             5)
```
yxlao: just for the information, why did we choose the unbiased `5` instead of `6` here?

src/Open3D/Geometry/IntersectionTest.cpp
griegler: done.

src/Open3D/Geometry/IntersectionTest.cpp
griegler: No real reason, I guess for the purpose of normalization it does not matter too much?

ShayGus:
Guy's is there any estimation regarding when will it be merged?

griegler:
Pinging @yxlao.

PR #1180, opened 2019-09-14
std::unordered_map for voxelgrid structure

syncle:
### Key Change

previous version
```
std::vector<Voxel> voxels_
```
this version
```
std::unordered_map<Eigen::Vector3i, Voxel> voxels_
```

### Benefit
- O(1) access to any voxel with ``Eigen::Vector3i`` voxel index.
- concise C++/Python codes to retrieve/check existing voxels

### Additional Feature
- polish and reorganize mesh/voxel related Python examples
- check if a point belong to VoxelGrid
```
output = voxel_grid.check_if_included(o3d.utility.Vector3dVector(queries))
```

### ToDo
- Binding ``std::unordered_map`` to work with ``print`` function in Python
- Due to aforementioned reason, ``print(voxel_grid.voxels)`` won't work for now.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1180)


griegler:
Thanks. Once CI passes and merge conflicts are resolved, we should be good to go.
CI seems to fail because of a syntax error in a unit test:

```
/home/runner/Open3D/src/UnitTest/Geometry/VoxelGrid.cpp: In member function ‘virtual void VoxelGrid_Bounds_Test::TestBody()’:
/home/runner/Open3D/src/UnitTest/Geometry/VoxelGrid.cpp:40:25: error: no match for ‘operator=’ (operand types are ‘std::unordered_map<Eigen::Matrix<int, 3, 1>, open3d::geometry::Voxel, open3d::utility::hash_eigen::hash<Eigen::Matrix<int, 3, 1> > >’ and ‘<brace-enclosed initializer list>’)
     voxel_grid->voxels_ = {
```

griegler:
Thanks. LGTM. Just one more thing that I found while skimming the code again: Do we need `voxel_index_` within the class `Voxel` now? Seems a bit redundant, as this information is in the map anyway?

griegler:
@syncle I agree. Although I would prefer `VoxelData`, but that might be subjective :wink:

yxlao:
I ran the benchmark with `examples/Python/Benchmark/benchmark_tsdf.py`, which uses a `512 * 512 * 512` dense voxel grid. The performance difference is negligible.

yxlao:
@syncle i see, then it's the same implementation compared, LGTM for this PR

Issue #1179, opened 2019-09-14
HalfEdgeTriangleMesh errors


kongsgard:
**Describe the bug**
1) The example in `examples/Python/Basic/half_edge_mesh.py` does not work.
2) Converting the Stanford bunny given in `examples/TestData/Bunny.ply` from a TriangleMesh to a HalfEdgeTriangleMesh gives a segmentation fault.

**To Reproduce**
1) `python half_edge_mesh.py`
2)
```
mesh = o3d.io.read_triangle_mesh("Bunny.ply")
half_edge_mesh = o3d.geometry.HalfEdgeTriangleMesh.create_from_mesh(mesh)
```

**Expected behavior**
1) Find the single boundary, and display it in the visualizer.
2) Convert the TriangleMesh to a HalfEdgeTriangleMesh, which could be used for further processing such as e.g. boundary detection.

**Screenshots**
N/A

**Environment (please complete the following information):**
 - OS: OSX Mojave
 - Python version: 3.6.9
 - Open3D version: 0.8.0
 - Is this remote workstation?: no
 - How did you install Open3D?: conda

**Additional context**
1) I get the expected behavior if I change the name of the HalfEdgeTriangleMesh variable:
```
if __name__ == "__main__":
    mesh = o3d.io.read_triangle_mesh("../../TestData/sphere.ply")
    mesh = mesh.crop([-1, -1, -1], [1, 0.6, 1])
    half_edge_mesh = o3d.geometry.HalfEdgeTriangleMesh.create_from_mesh(mesh)
    mesh.compute_vertex_normals()
    num_vertices = len(mesh.vertices)
    draw_geometries_with_back_face([mesh])

    # Find a boundary vertex
    boundaries = half_edge_mesh.get_boundaries()
    assert len(boundaries) == 1
    boundary_vertices = boundaries[0]

    # Colorize boundary vertices
    vertex_colors = 0.75 * np.ones((num_vertices, 3))
    vertex_colors[boundary_vertices, :] = np.array([1, 0, 0])
    mesh.vertex_colors = o3d.utility.Vector3dVector(vertex_colors)
    draw_geometries_with_back_face([mesh])
```

2) If I add the same mesh cropping in the example above (although it doesn't seem to affect the mesh since the cropping is outside its boundaries), the conversion is successful. I can then also find the five holes in the mesh:
```
mesh = o3d.io.read_triangle_mesh('Bunny.ply')
[...]

PR #1184, opened 2019-09-16
New Mesh Class Structure

griegler:
Reorganizing the mesh geometries: There is a new `MeshBase` class that only contains `vertices_` and related attributes. `TriangleMesh` and `HalfEdgeTriangleMesh` are children of `MeshBase`.  Edit: `TetraMesh` is also a child of `MeshBase` now. ~~Further, this PR adds `TexturedTriangleMesh` which contains `uvs_` (3 per triangle) and a `texture_`.~~
`TriangleMesh` can now contain uv coordinates in `triangle_uvs_` (3 per triangle) and a `texture_` image.

The script `half_edge_mesh.py` does not segfault on execution anymore (addressing #1179), but the results still look weird. @yxlao could you please have a look, as the code mostly builds on your initial `HalfEdgeTriangleMesh` class.

~~`TexturedTriangleMesh` is bare bones, there are no IO, nor visualization capabilities. I think @theNded  wanted to look into this?~~

~~Right now the `TriangleMesh` operations, e.g. `Filter*`, `Simplification*` do not touch the `uvs_`, should we print a warning for now?~~
All `TriangleMesh` methods that change the number of `triangles_` print a warning if the mesh contains uv coordinates. In future PRs we can extend these methods to handle the uv coordinates correctly.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1184)


theNded:
Ah, actually I just did something similar. I cannot figure out a good way to load obj into varying meshtypes (`TriangleMesh` vs `TexturedTriangleMesh`, so my current solution is to hardcode uvs inside `TriangleMesh` class.
With this `MeshBase` class I think we need to change some of the function signatures to support polymorphism? 

For example, we can change
```cpp
bool ReadTriangleMeshFromOBJ(const std::string& filename,
                             geometry::TriangleMesh& mesh,
                             bool print_progress)
```
to something like
```cpp
bool ReadTriangleMeshFromOBJ(const std::string& filename,
                             std::shared_ptr<geometry::MeshBase>& mesh,
                             bool print_progress) 
```
and return the correct meshtype according to the file content parser.

griegler:
@theNded hmm, returning `std::shared_ptr<geometry::MeshBase>` does not seem to be ideal. Maybe `std::shared_ptr<geometry::TriangleMesh>`, but I still do not like that we would test for the real instance at runtime. 
Maybe we really can merge `TriangleMesh` and `TexturedTriangleMesh`? We have the `Has*` methods anyway.

theNded:
@griegler I have an initial implementation (haven't covered boundary cases though). If it makes sense maybe we can do this.
https://github.com/theNded/Open3D/blob/texture/src/Open3D/Geometry/TriangleMesh.h
https://github.com/theNded/Open3D/blob/texture/src/Open3D/IO/FileFormat/FileOBJ.cpp

griegler:
@theNded Yes, I am fine with having `uvs` and `texture` in `TriangleMesh`, but I think it shouldn't be `vertex_uvs`. `triangle_uvs` (3 uvs coordinates per triangle) seems to be a better choice and is also closer to the `obj` format.

theNded:
@griegler I agree. Should we temporarily remove `TexturedTriangleMesh` in this PR and let me modify `TriangleMesh` in another PR?

griegler:
@theNded I will merge `TexturedTriangleMesh` into `TriangleMesh` in this PR based on this discussion. Then, I can also add warnings to other methods (`Filter*`, ...) if `uvs` are set. Does this work for you?

theNded:
@griegler Sure, thanks!

yxlao:
@griegler HalfEdgeTriangleMesh fixes pushed

griegler:
@yxlao Thanks. LGTM. I just have one question, why does the `HalfTriangleMesh` needs the `triangles_` vector? All the information should be in the `half_edges_`?

yxlao:
@griegler that seems right, in that way we can keep the `HalfEdgeTriangleMesh` minimalistic. I will do that in a separate PR, since that will require a new `SimpleShaderForHalfEdgeTriangleMesh`. 

If that's okay for you we can merge this PR.

griegler:
@yxlao I think we are then ready to merge.

PR #1186, opened 2019-09-16
Implement image flip functions

theNded:
- Rename previous `Flip()` with `Transpose()` according to its semantics;
- Implement `FlipUD()` and `FlipLR()` to flip images. These are useful functions in loading textures.
- Python interface also supported.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1186)

### DIFF
--- a/src/Open3D/Geometry/Image.cpp
+++ b/src/Open3D/Geometry/Image.cpp
@@ -295,7 +295,7 @@ std::shared_ptr<Image> Image::Transpose() const {
     return output;
 }
 
-std::shared_ptr<Image> Image::FlipUD() const {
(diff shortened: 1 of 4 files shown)


theNded:
@syncle Thanks for the review! 
The initial naming comes from scipy, as suggested by @yxlao 
https://docs.scipy.org/doc/numpy/reference/generated/numpy.fliplr.html
But I do think `flip_horizontal` for python and `FlipHorizontal` for C++ are clearer. 
Any preferance?

theNded:
@syncle Done!

griegler:
LGTM.

Issue #1098, opened 2019-07-31
Files with Unicode path are not read


mwil:
**Describe the bug**
(This may only apply to Windows 10) I'm trying to read a PCD file using `open3d.io.read_point_cloud`. The operation yields the correct result if the path to the file is in ASCII, but for paths that include Unicode characters the read operation returns an empty object.

`os.path.isfile()` returns true for the file including Unicode, and I use the same path for other operations without issue. I think it must be related to how Open3D handles the path internally.

**To Reproduce**
- Create a Unicode folder ("出典データ" for me)
- Try reading the file using `read_point_cloud`

**Expected behavior**
Read the file from any valid location.

**Environment (please complete the following information):**
 - OS: Win 10 Pro
 - Python version: 3.7.3
 - Open3D version: 0.7.0
 - Is this remote workstation? no
 - How did you install Open3D? conda

griegler:
Yes, seems to be related to Windows and `fopen`. Cf. https://stackoverflow.com/questions/396567/is-there-a-standard-way-to-do-an-fopen-with-a-unicode-string-file-path
Not sure what the best way is to handle this. @yxlao @qianyizh any suggestions?

griegler:
Can you test PR #1190 if it fixes your problem?

griegler:
Fixed in #1190.

PR #1190, opened 2019-09-18
fopen wrapper to enable unicode paths in Windows

griegler:
Addresses #1098. 
Implemenation of an fopen wrapper in `utility::filesystem` that should handle unicode paths in Windows. Note that some file types/extensions might still not work with unicode paths on Windows (ply, png?), because fopen is used within the 3rdparty lib.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1190)


griegler:
LGTM.

PR #1193, opened 2019-09-19
Update type check for open3d PointCloud

msofka:
fixes https://github.com/intel-isl/Open3D/issues/949
(problem on Python 3.6)

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1193)


yxlao:
thanks!

PR #1194, opened 2019-09-20
Support rendering textured triangle mesh

theNded:
## IO
Supports `.obj` file's IO along with `.mtl` files, and possible `.png` texture files.
### Input
We can parse uv coordinates and materials (diffuse texture map to be specific). 
In theory, an obj file can hold multiple objects and multiple corresponding textures. At current, we only support one material and load the first encountered texture image. There are several reasons for this trade-off:
- In general, a file reader only returns one mesh object. Our previous code merges vertices and trianlges from all shapes into one mesh. Otherwise we have to significantly change the APIs.
- Most obj files, even holding multiple materials, have only one diffuse texture.

### Output
By default, we write an `.mtl` file along with an `.obj` file, even if there is no uv coordinates or textures. This is an convention of other softwares, e.g. _meshlab_. 
A general API change is an additional flag `write_triangle_uvs` in `WriteTriangleMesh` and corresponding writers for all formats. By default, it is `false`. It can be manually enabled when writing to `.obj` files.

## Rendering
The new shader is mostly copied from the `PhongShader`, but as the buffers, binded textures, and shader uniforms (texture) are changed, we separate everything in another class `TexturedPhongShader`.

We set a higher priority of rendering from texture than rendering from vertex colors. 
In `RenderOption::MeshColorOption::Color` mode, if we have `mesh.HasTriangleUvs() && mesh.HasTexture()`, we will disregard vertex colors.

## Tests
A basic texture `rock` is temporarily added to `TestData`. We may change this model if we have better options. 
Corresponding tests are added to `Basic/file_io.py` and `Basic/visualization.py`.

Other manual tests include 
- Read corrupted `.obj` file with missing `.mtl` and/or `.png` files, which correctly falls back to default vertex color rendering.
- Write obj file to relative / absolute paths, and load / render them in _Open3D, meshlab_, and _CloudCompare_. 

## Bug fixes
Fixed crashing bug where there is no normal in an `.obj` file. 
Location and reason: 
https://github.com/intel-isl/Open3D/blob/499d245d80fda050319727ea99e8ea33fecb57f4/src/Open3D/IO/FileFormat/FileOBJ.cpp#L100-L101
When `attrib.normals.size()` is 0, `idx.normal_index` will be -1. This will pass the test and trigger the out-of-bound problem.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1194)


yxlao:
@theNded @qianyizh just noticed the request, I have merged too soon -- let's make additional changes in another PR if necessary

theNded:
I guess @qianyizh doesn't have enough time to review, so I removed him for this PR as we've already discussed for a while. I guess the efficient way is to ask him to review when we have fundamental changes of APIs.

qianyizh:
Works for me.

PR #1232, opened 2019-10-10
Support relative paths in ReconstructionSystem

keunhong:
Allows `path_dataset` and `path_intrinsic` to be paths relative to the json file.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1232)

### DIFF
--- a/examples/Python/ReconstructionSystem/run_system.py
+++ b/examples/Python/ReconstructionSystem/run_system.py
@@ -6,11 +6,17 @@
 
 import json
 import argparse
-import time, datetime
+import time
+import datetime
+import os
 import sys
-sys.path.append("../Utility")
-from file import check_folder_structure
-sys.path.append(".")
+
+_this_path = os.path.realpath(__file__)
+_this_dir = os.path.dirname(_this_path)
+sys.path.append(os.path.join(_this_dir))
+sys.path.append(os.path.join(_this_dir, "../Utility"))
+
+from file import check_folder_structure, resolve_relative_path
 from initialize_config import initialize_config
 
 if __name__ == "__main__":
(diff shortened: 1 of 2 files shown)


examples/Python/ReconstructionSystem/run_system.py
```diff
@@ -6,11 +6,17 @@
...
+import datetime
+import os
 import sys
-sys.path.append("../Utility")
-from file import check_folder_structure
-sys.path.append(".")
+
+_this_path = os.path.realpath(__file__)
+_this_dir = os.path.dirname(_this_path)
+sys.path.append(os.path.join(_this_dir))
```
yxlao: put them in one line so that we don't have a temp variable? e.g.
```python
sys.path.append(os.path.join(os.path.realpath(__file__)))
```

examples/Python/Utility/file.py
```diff
@@ -30,9 +29,16 @@ def get_file_list(path, extension=None):
...
+    path = None
     for folder_name in folder_names:
-        if exists(join(path_dataset, folder_name)):
-            path = join(path_dataset, folder_name)
+        _path = join(path_dataset, folder_name)
+        if exists(_path):
+            path = _path
+
+    if path is None:
+        raise ValueError("Could not find any of {!r} in path {!r}".format(
```
yxlao: `FileNotFoundError`?

examples/Python/Utility/file.py
```diff
@@ -30,9 +29,16 @@ def get_file_list(path, extension=None):
 
 
 def add_if_exists(path_dataset, folder_names):
+    path = None
     for folder_name in folder_names:
```
yxlao: The original code looks pretty strange -- if there is more than one existing `folder_names`, only the last one will be picked up. We shall return the path if and only if there is exactly one `folder_name` inside `folder_names` exists.

So probably something like this
```python
valid_paths = [join(path_dataset, f) for f in folder_names if exists(join(path_dataset, f))]
if len(valid_paths) == 0:
    # exception not found
elif len(valid_paths) > 1:
     # exception more than one found
else:
    return valid_paths[0]
```

yxlao:
@keunhong could you please take a second look at the PR comments? Thanks.

germanros1987:
Given the lack of response from the author, I am closing this PR. Happy to re-open it in the future.

PR #1226, opened 2019-10-08
Slice support for Tensors

yxlao:
- Slice support for tensors, supports operations like
    - `t[1]`
    - `t[1:5, 0:10:2, 1]`
- Initilal print function for tensor
- `Clone` and `Copy` function, where `Clone` copies the whole buffer while `Copy` will compact the tensor
- Parameterized unit tests for available hardware backends
- Conditional compilation, shall work on machines without CUDA

Let's try to review and merge #1211 first, both of them shall be merged to `gpu` branch. While #1211 don't aim for passing CI, we'll enable (CPU) CI checks from this PR and onwards. Later we'll be adding the GPU CI checks as well.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1226)

PR #1255, opened 2019-10-18
changes to compile flags for Unix because of nvcc

benjaminum:
This PR addresses some problems with nvcc 10.0 and earlier for Unix builds.

- removed duplicate optimize flag
- removed global "-Wno-unused-result". Flag is now set for the specific files that need it and the examples only.
- Disabled LTO when using nvcc < 10.0 for python module

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1255)


benjaminum:
I think the CI issues are solved in #1226 
I will apply my changes on top of that after it has been merged and then open a new PR.

PR #1270, opened 2019-10-28
Improve docs for color map

RohanRathi:
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1270)

### DIFF
--- a/src/Open3D/ColorMap/ColorMapOptimization.h
+++ b/src/Open3D/ColorMap/ColorMapOptimization.h
@@ -80,25 +80,56 @@ public:
     ~ColorMapOptimizationOption() {}
 
 public:
-    /// Set to `true` to enable non-rigid optimization (optimizing camera extrinsic params and image wrapping field for color assignment), set to False to only enable rigid optimization (optimize camera extrinsic params).
+    /// Set to `true` to enable non-rigid optimization (optimizing camera
+    /// extrinsic params and image wrapping field for color assignment), set to
+    /// False to only enable rigid optimization (optimize camera extrinsic
+    /// params).
     bool non_rigid_camera_coordinate_;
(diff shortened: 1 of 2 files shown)


yxlao:
@RohanRathi also run style check:
http://www.open3d.org/docs/release/contribute.html#coding-style. I'd recommend using an editor plugin to make formatting automatic.

RohanRathi:
___
*[src/Python/open3d_pybind/color_map/color_map.cpp, line 180 at r1](https://reviewable.io/reviews/intel-isl/open3d/1270#-LsJkbs10gGHQKl8XuyF:-LsWdHte2PC6XLTcAcA8:b1sbvsr) ([raw file](https://github.com/intel-isl/open3d/blob/29fdbae686836856425b37e8b731a224328b357f/src/Python/open3d_pybind/color_map/color_map.cpp#L180)):*

This is not fixed. we need to add comma, not period. 
</blockquote></details>

Sure, I'm facing a weird issue though. Whenever I make changes to the python documentation, the changes don't reflect on the docs that are built. Even if in my local master I make some small change say in `color_map.cpp` and then build the docs. The original text is still there, instead of the change I made. Can you help with that?

RohanRathi:
___
*[src/Python/open3d_pybind/color_map/color_map.cpp, line 180 at r1](https://reviewable.io/reviews/intel-isl/open3d/1270#-LsJkbs10gGHQKl8XuyF:-LsavgNe5Mv-1HbTCa6i:b-yr7eea) ([raw file](https://github.com/intel-isl/open3d/blob/29fdbae686836856425b37e8b731a224328b357f/src/Python/open3d_pybind/color_map/color_map.cpp#L180)):*

@RohanRathi did you re-install the open3d python package after code update? i.e. run this command in the `build` directory:

```shell
```
</blockquote></details>

Shouldn't `python make_docs.py --sphinx --doxygen` update the docs?

yxlao:
@RohanRathi yes, but the `make_docs.py` imports the newly-built and installed `open3d` from the python path. When the actual code is changed, the `open3d` package shall be updated.

PR #1275, opened 2019-10-29
changes to compile flags for Unix because of nvcc

benjaminum:
This PR addresses some problems with nvcc 10.0 and earlier for Unix builds.

- removed duplicate optimize flag

- removed global "-Wno-unused-result". Flag is now set for the specific files that need it and the examples only.
- Disabled LTO when using nvcc < 10.0 for python module

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1275)


benjaminum:
Like #1255 but based on latest gpu branch 00bbddf213b0c12e4baf9cdbb1d64f9d139f8ba4 to pass CI checks

Issue #1278, opened 2019-10-29
Large PTS reading issue


YouYue123:
**Problem**

I am facing problem when loading a 4 billion **.PTS** point cloud file. Because the number of pts is exceeding the limit of signed 32 int (2,147,483,647) by referencing [this doc](https://en.cppreference.com/w/cpp/language/types)

**Solution**

I realised in Open3D you are reading header of PTS as integer, I made my change to unsigned long long for header reading. 

It works well from my side, dose this changing look like a good solution from you? I can make a pull request for this issue to support large PTS reading.

syncle:
Addressed via #1279. Thanks @YouYue123!

PR #1279, opened 2019-10-29
Large PTS File IO 

YouYue123:
Change num_of_pts from int to unsigned long long

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1279)


YouYue123:
Related to issue #1278

YouYue123:
👌unsigned int seems reasonably big enough for my problem as well. Will make the change soon

YouYue123:
Changed

YouYue123:
Do you think it will be a good idea to change other IO component from int to size_t in this way?

PR #1293, opened 2019-11-01
Assignment operator

yxlao:
Supports copy constructor and assignment of the following syntax:
```cpp
Tensor tensor_a(tensor_b);
Tensor tensor_a(tensor_b[0]);
tensor_a = tensor_b;
tensor_a = tensor_b[0];
tensor_a[0] = tensor_b;
tensor_a[0] = tensor_b[0];
tensor_a[0] = 100;  // with automatic casting
```

Notably, we now can assign to a non-contiguous tensor, e.g (in python equivalent).
```python
a = np.ones((10,))
b = np.ones((20,))
a[0:10:2] = b[0:20:4]
```

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1293)

### DIFF
--- a/src/Open3D/Container/Kernel/CPULauncher.h
+++ b/src/Open3D/Container/Kernel/CPULauncher.h
@@ -35,7 +35,7 @@ namespace kernel {
 class CPULauncher {
 public:
     /// Recover src tensor element offsets given dst tensor element offsets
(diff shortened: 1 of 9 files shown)


theNded:
Can we merge this? I would have to pull this in order to update my own branch now.

Issue #1267, opened 2019-10-25
Add concave hull function for point clouds


HowcanoeWang:
**Is your feature request related to a problem? Please describe.**
I am doing some point cloud analysis, I am benefited a lot from the build-in convex hull functions, I also need to calculate the concave hull to get another volume estimation.

**Describe the solution you'd like**
Will it possible to provide a function in `o3d.geometry.PointCloud` class, called `compute_concave_hull(self)` and Returns `open3d.geometry.TriangleMesh` of that?

**Describe alternatives you've considered**
If you can provide a function to calculate the perimeter and volume of the generated convex or concave hull, that would be very nice. Currently, I need to use `scipy.spatical` package to get these values.

Anyway, thanks a lot for such a good package!

griegler:
The concept of a concave hull is not very well defined in contrast to the convex hull. Do you have a specific algorithm in mind?

HowcanoeWang:
> The concept of a concave hull is not very well defined in contrast to the convex hull. Do you have a specific algorithm in mind?

https://www.mathworks.com/help/matlab/ref/alphashape.html
https://www.mathworks.com/help/matlab/ref/alphashape.boundaryfacets.html

Hoping something like these (though these materials are implemented in Matlab)

Or https://stackoverflow.com/questions/26303878/alpha-shapes-in-3d

BTW, it seems `compute_concave_hull(self, alpha=0.5)` need an alpha parameter.

griegler:
Implemented in #1320.

PR #1320, opened 2019-11-14
Alpha Shapes

griegler:
Implementation of Alpha Shapes.
Addressing #1267.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1320)


jiapei100:
How to use it in **Python**??

PR #1323, opened 2019-11-17
Added normal explanation 

atulim:
Added normal explanation and removed a documentation bug so that users can easily understand what each function stands for. Also there is required in change in documentation by your side. 
/// should be changes to /* or //! because multiline documentation written by people at Open3D is not understable by Doxygen when you use ///.

Thanks. Screenshots are attached. I will send another pull request by 22nd of November which will documentation of all the functions used in each 3D processing algorithm.
Screenshot from 2019-11-17 21-53-13
Screenshot from 2019-11-17 21-53-19
Screenshot from 2019-11-17 21-53-24
Screenshot from 2019-11-17 21-53-28

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1323)

### DIFF
--- a/src/Open3D/Camera/PinholeCameraIntrinsic.h
+++ b/src/Open3D/Camera/PinholeCameraIntrinsic.h
@@ -39,6 +39,7 @@ enum class PinholeCameraIntrinsicParameters {
     Kinect2ColorCameraDefault = 2,
 };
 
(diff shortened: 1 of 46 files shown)


yxlao:
Thanks @atulim , please re-submit the PR when it's ready, according to our offline discussions.

Issue #1285, opened 2019-10-30
Update examples CPP with RealSense2 to support latest devices (D415 camera).


JTrantow:
**Is your feature request related to a problem? Please describe.**
The problem is that examples/CPP/RealSense project is written for and uses thirdparty includes for the original LibRealSense SDK. This example needs an update to support LIBREALSENSE_RS2 and the latest cameras. The current implementation appears to support only the R200 and F200 cameras. Therefore the example code does not recognize the D415 camera.

**Describe the solution you'd like**
Update example/CPP/RealSense and thirdparty/librealsense to use RealSense SDK 2.0. Or possibly keep existing project to support for R200 and F200 and establish a new project and thirdparty/librealsense2 to support with the new devices. Maybe the equivalent to the python realsense_recorder.py?

**Describe alternatives you've considered**
Alternative could be switching from CPP to Python development which appears to support librealsense2, but would like to stay with CPP.

**Additional context**

argosvr:
very good idea ! it seems that you are not taking into account what the dr400 cameras can do.. see the last firmware, the last sdk and with proper light the kind of result you can get. We need a proper support for these cameras.

JTrantow:
I think the correct feature to add is new directories for 3rdParty/RealSense2 and examples/CPP/RealSense2.

Towards this end, I have created a new RealSense2 project that works with SDK2 and the new cameras(D415/D435i). It is essentially a port of examples/CPP/RealSense that works with RealSense SDK2. I don't think the example needs to be much more than this but a starting point CPP example in Open3D that links with SDK2 would be more relevant than the existing RealSense example.

I have examples/CPP/RealSense2 files ready for a pull request. However, I don't understand the bigger picture 3rdParty CMake ideas. I edit my VisualStudio project to find include files from a separate LibRealSense2 git directory and copy RealSense2.dll and RealSense2.lib into my build which isn't a portable solution. 

So I would need some help on the 3rd party CMake aspects Specifically:

1. Adding the 3rdParty/RealSense2 files or a CMake configs to download the files.
2. CMake configs to add the correct include and library paths to the project build.

argosvr:
still no answer here ? it is really a shame...

JTrantow:
Closed due to lack of interest.

PR #1329, opened 2019-11-20
example/CPP/RealSense2 working with RealSense SDK2 and D415 Camera.

JTrantow:
I added a RealSense2 example similar to the original RealSense example. 

I'm not sure of the design intent behind the 3rdparty directory but this example also requires a 3rdparty/librealsense2 directory to be populated from https://github.com/IntelRealSense/librealsense to build

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1329)


JTrantow:
This push is a follow up to new feature request #1285

JTrantow:
I don't understand the desired 3rdparty organization. The original librealsense is not a submodule, nor are many of the other 3rdparty, so I don't see a good pattern to follow.

I'm a newbie to Git and it's counter intuitive relative to my Subversion experience so please be patient. Looking at my pull request, I could have picked a better starting point and broken this into two pulls 1) for the 3rdparty/libRealSense2 and 2) for examples/CPP/RealSense2.  

I will delete everything from this pull except the RealSense2 example and open a new feature request for just the 3rdparty/libRealSense2. Ideally, someone who's good at Git submodules and CMake could handle the libRealSense2. If no one has time to add this, I will fumble through it.

PR #1337, opened 2019-11-22
Dimshuffle and transpose

yxlao:
Implements the following functions with PyTorch semantics.
```cpp
Tensor::Permute
Tensor::Transpose
Tensor::T
```

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1337)


yxlao:
will merge for now, we can discuss the exception handling later

PR #1339, opened 2019-11-22
Compilation with libcxx

Xardvor:
Merged master.
Brings compilation with libc++. Using it we can safely use any stl classes in our code

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1339)


Xardvor:
Pull request is obsolete, we managed to compile using both gcc and clang.

Issue #1351, opened 2019-11-25
get_rotation_matrix_from_xyz  isn't in Open3d.geometry?


prass-anyvision:
get_rotation_matrix_from_xyz(..)  isn't in Open3d.geometry?

Can't find that method. 

Examples refer to that : https://github.com/intel-isl/Open3D/blob/master/examples/Python/Basic/transformation.py

griegler:
Right, `get_rotation_matrix_from_xyz()` is currently not exposed in `open3d.geometry`, only in `open3d.geometry.Geometry3D` and its sub-classes. For now you can use `open3d.geometry.Geometry3D.get_rotation_matrix_from_xyz()`.

prass-anyvision:
Thanks - can I rotate with a rotation matrix or do I have to estimate euler angles or axis angles?

griegler:
Once you have a rotation matrix, you can use it in the `rotate` methods. The functions `get_rotation_matrix_from_*` are just convenience functions to get a rotation matrix from different rotation representations.

griegler:
#1355 exposes `get_rotation_matrix_from_*` in `open3d.geometry`.

prass-anyvision:
That was fast!

PR #1355, opened 2019-11-26
Expose get_rotation_matrix_from_* directly in open3d.geometry

griegler:
Currently, rotation matrices can be computed via instances of `open3d.geometry.Geometry3D`, or via `open3d.geometry.Geometry3D.get_rotation_matrix_from_*`. This PR exposes these functions directly via the module `open3d.geometry`.

Brought up in #1351.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1355)

### DIFF
--- a/src/Python/open3d_pybind/geometry/geometry.cpp
+++ b/src/Python/open3d_pybind/geometry/geometry.cpp
@@ -33,6 +33,24 @@
 using namespace open3d;
 
 void pybind_geometry_classes(py::module &m) {
+    // open3d.geometry functions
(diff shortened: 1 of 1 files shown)


rttariverdi67:
hi all,
I have tried to Expose get_rotation_matrix_from_* directly in open3d.geometry but still I have facing error!
what I do is :
assume "z_phi = np.random.uniform(-np.pi, np.pi)", 
insted of : *rotation_matrix = o3d.geometry.get_rotation_matrix_from_xyz([0.,0.,z_phi])*

PR #1362, opened 2019-11-27
Gui

Bai426:
Hi, I had a problem when I display a point cloud which is made of a color image and a depth image, but the result is not right. It looks seem being flipped. So I want to know how can I deal this problem in Open3D. 
image
image
image

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1362)


germanros1987:
Hi,

I am a bit confused. It seems that you cherry-picked some commits from the GUI branch to make a PR. Is this intentional?

Bai426:
No, I want to solve the problem when I used opened.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

theNded:
Issues should stay in the issue tab. 
Started to track this issue in the correct place.

PR #1374, opened 2019-12-03
minimal ml module

benjaminum:
This PR is mostly about changes to the build system.

- Adds a single Tensorflow op and code for building and loading
- New optional dependencies are TBB and Tensorflow

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1374)


griegler:
Is it possible/useful to add TBB as submodule, e.g., using this [fork](https://github.com/wjakob/tbb)?

benjaminum:
> Is it possible/useful to add TBB as submodule, e.g., using this [fork](https://github.com/wjakob/tbb)?

If we want to build it from source then yes. If we just want the headers we could also include the original repo. We have to be careful with the backward compatibility because we want to use the tbb lib installed on the system.

PR #1390, opened 2019-12-10
Added VisualizerWithVertexSelection

prewettg:
Added VisualizerWithVertexSelection (click or drag-rect to select vertices of a Geometry), added "editmesh" option to Visualizer example that uses it.  `bin/examples/Visualizer editmesh mesh.file` should let you select vertices. You can get the selection with GetPickedPoints().

VisualizerWithVertexSelection is based on VisualizerWithEditing, with cropping and voxel operations removed. Holding down shift will go into selection mode (like VisualizerWithEditing).  Clicking a point toggles whether it is selected. Dragging a rectangle (at least 3 pixels in both directions) while in selection mode adds all the points to the selection; pressing Delete or Backspace removes all the points in the rectangle from the selection. Ctrl-R clears the selection.

Many of the smaller changes are allowing the depth test to be set in RenderOptions. Visualizer has utility renderers, which I use for the selection visualization. I added the ability to set render options for them. The depth test is so that in selection mode I can render all the selectable points with GL_LESS and then the selected points on top of that with GL_LEQUAL, so that the selected points are also visible.

The point picking is done by drawing the geometry to the depth buffer only, then drawing the points to both the color and depth buffers. The points on the far side of the geometry will be depth-tested out. Unfortunately, this means that part of the point can be depth-tested out, too. In practice the points tend to look triangular instead of square because part of the point is behind the geometry. I tried using glPolygonOffset to move the geometry back, but that didn't help (and probably wouldn't have helped with steep or highly curved geometry anyway). I would like to depth-test the pixels of the square based on the center, but I'm not sure that is possible, as the pixels don't know the depth values of other pixels, and I don't think you can do the depth test in the geometry shader to remove points that fail. The only other thing I can think to do is to do a fuzzy depth test, where points only fail the depth test if they are more than a certain value behind the current value. That would require a shader just for this, and I'm not sure how feasible this method actually is. But for purposes of getting a demo movie for the release, I think this lets us move forward. If you have suggestions for improving the point drawing, please let me know.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1390)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,7 +6,7 @@ cmake_minimum_required(VERSION 3.1)
 set (CMAKE_CXX_STANDARD 11)
 
 add_custom_target(build_all_3rd_party_libs
-    COMMAND cmake -E echo "Custom target build_all_3rd_party_libs reached."
+    COMMAND ${CMAKE_COMMAND} -E echo "Custom target build_all_3rd_party_libs reached."
 )
 
 if (NOT CMAKE_BUILD_TYPE)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -98,6 +98,16 @@ set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/lib)
 set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/lib)
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/bin)
 
+# headless rendering
+if (ENABLE_HEADLESS_RENDERING)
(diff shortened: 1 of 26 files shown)


prewettg:
Ok, I updated for all the comments, except for the one with `view_control_ptr_->GetStatusString()`. It turns out that I can't do that, because view_control_ptr is defined by the base class (Visualizer), and is a ViewControl, but the VisualizerWithVertexSelection assigns a ViewControlWithEditor to it, and GetStatusString() is a function on the ViewControlWithEditor derived class.

Thanks for the reminder about static_pointer_cast, I end up casting a shared_ptr so rarely that I completely forgot about this.

yxlao:
@prewettg could you please check if updates are needed after #1392?

prewettg:
Yep, slight changes were needed for #1392

PR #1396, opened 2019-12-13
Multi-gpu and p2p transfer support 

yxlao:
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1396)


yxlao:
Merging now since all review comments have been addressed.

PR #1405, opened 2019-12-17
Pickpoint move

prewettg:
VisualizerWithVertexSelection will move selected points with right-drag. The depth will be queried on right mouse down, so that we can move so that the mouse stays with the point:  a pixel on the viewing plane (where the mouse moves) is larger than a pixel at the point's distance, due to perspective, and thus the point needs to move a larger distance than the mouse has traveled on the viewing plane.

It seems that points are not written to the depth buffer. This works fine when the points are in their initial position, but may give awkward results when re-moving. If no point was hit, it will use the last depth value. This will work great if the viewing position has not changed.  If it has changed, at least it is better than infinity, which is what we'd use if we used the actual depth value of 1.0. If the point is over geometry, the depth value of the geometry will be used, which will not be quite accurate, but should not be too bad.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1405)


prewettg:
1) You can update the geometry that the visualizer is using, as long as you call UpdateGeometry() afterwards.
2) You could do geometry->vertices_[selected_index] to get the coordinate. I changed GetPickedPoints() to return a struct with both the index and the coordinate, though, so you'll probably want to use that now.
3) I've added callbacks for selection changed and selection moved. Just assign a std::function to OnSelectedChanged or OnSelectionMoved.  I've tested changing geometry in OnSelectionMoved(), but if you want to change the geometry while the selection is moving I'll need to do more work.

prewettg:
Ok, I've added register_selection_changed_callback(), register_selection_moving_callback(), and register_selection_moved_callback() to Python. You can update the vertices (and normals) of the points and call update_geometry(None), but if you change the number of points it will clear the selection, because we can't rely on the indices of the points that were selected to remain the same.  I expect you have some Python code already, but I tested with this:
```
#!/usr/bin/env python
import open3d

def changed_cb():
    print("[python] changed")

def moved_cb():
    print("[python] moved")

mesh = open3d.io.read_triangle_mesh("path/to/madrid-animals-for-sarcophagus-cat-2.stl")
mesh.compute_vertex_normals()
vis = open3d.visualization.VisualizerWithVertexSelection()
def moving_cb():
    sel = vis.get_picked_points()
    for p in sel:
        mesh.vertices[p.index] = p.coord
    vis.update_geometry(None)
vis.create_window()
vis.add_geometry(mesh)  # must do AFTER create_window()
vis.register_selection_changed_callback(changed_cb)
vis.register_selection_moving_callback(moving_cb)
vis.register_selection_moved_callback(moved_cb)
vis.run()
vis.destroy_window()
```

PR #1407, opened 2019-12-17
Fixes for macOS

prewettg:
`cp` on macOS doesn't support the `-u` option and it doesn't look like there is anything equivalent.  Also added a couple of missing virtual destructors, maybe clang is more comprehensive in its warnings than GCC?

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1407)


prewettg:
Blast, I hadn't checked the CI for some reason. It looks like it's failing on all the gui branch not being stylistically correct. So the macOS changes are going to get swamped by style changes :(

PR #1429, opened 2020-01-08
Basic support for textures implemented

Xardvor:
Allocation and structure of vertex buffers for meshes got rewritten. Now it is array of structs instead of struct of arrays.
Current implementation could be optimized if we reuse indices when generating buffers.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1429)


Xardvor:
> Clang appears to make "has virtual functions but no virtual destructor" an error. GCC appears to have a warning for this (not sure which one, though, maybe -Wnon-virtual-dtor ?). If it is possible to force a specific warning as an error, it might be useful to add that to the build system to make GCC and Clang have similar behavior. (Not part of the request changes for this PR)

Yeah, we could have this. I usually don't put destructors in interface classes by default, only when needed. And like, c++ has this `override` keyword which would highlight any problems with virtual destructor in base class.

PR #1430, opened 2020-01-08
gui: Window now passes key events

prewettg:
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1430)

### DIFF
--- a/src/Open3D/GUI/Window.cpp
+++ b/src/Open3D/GUI/Window.cpp
@@ -467,9 +467,7 @@ void Window::OnMouseEvent(const MouseEvent &e) {
     for (auto it = impl_->children.rbegin();
          it != impl_->children.rend();  ++it) {
         if ((*it)->GetFrame().Contains(e.x, e.y)) {
-            if (e.type == MouseEvent::BUTTON_DOWN &&
-                e.button.button == MouseButton::LEFT)
-            {
+            if (e.type == MouseEvent::BUTTON_DOWN) {
                 impl_->focusWidget = it->get();
             }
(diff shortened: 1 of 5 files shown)


src/Open3D/GUI/Window.cpp
```diff
@@ -464,6 +467,11 @@ void Window::OnMouseEvent(const MouseEvent &e) {
     for (auto it = impl_->children.rbegin();
          it != impl_->children.rend();  ++it) {
         if ((*it)->GetFrame().Contains(e.x, e.y)) {
+            if (e.type == MouseEvent::BUTTON_DOWN &&
+                e.button.button == MouseButton::LEFT)
```
Xardvor: Maybe we need to move focus to widget also in RIGHT mouse button case?

Xardvor: Just to comment: if I, for example, made to modes for my SceneWidget, one for free fly camera and one for interacting with geometry. The later uses left mouse btn and keys and the first right mouse btn and keys. With current realization widget wouldn't receive keys for flying camera mode, because widget is unfocused.
So, either we need to make it to become focused on RMB too or give it some function to call to become focused.

prewettg: I assumed that native widgets don't get focus on right click, but not only do we need that for mouse movement, but native widgets do in fact get focus on right click. (Well, actually, it seems that some do and some don't, at least on macOS). I changed it to get focus on any BUTTON_DOWN (which will include middle down, because I'm pretty sure there is a middle-drag movement, too).

PR #1438, opened 2020-01-14
Added string ids for geometry

Xardvor:
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1438)

### DIFF
--- a/src/Open3D/Geometry/Geometry.h
+++ b/src/Open3D/Geometry/Geometry.h
@@ -26,6 +26,8 @@
 
 #pragma once
 
+#include <string>
+
 namespace open3d {
 namespace geometry {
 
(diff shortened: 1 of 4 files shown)


Xardvor:
> > _[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1438)_ status: 0 of 4 files reviewed, 5 unresolved discussions (waiting on @germanros1987, @prewettg, and @Xardvor)
> 
> _[src/Open3D/Geometry/Geometry.h, line 94 at r1](https://reviewable.io/reviews/intel-isl/open3d/1438#-Ly_1BczDGpD0IylmLrQ:-LyjnQsE7TpMBOdtNbhC:b-6quia9) ([raw file](https://github.com/intel-isl/open3d/blob/7e443938caec2948d9881d5541d0154cf4e10ae3/src/Open3D/Geometry/Geometry.h#L94)):_
> 
> Previously, germanros1987 wrote…
> > @yxlao do you agree with these new get/set methods?
> 
> lgtm, use `name_` for member variable, i.e.
> 
> ```c++
> void SetName(const std::string& name) { name_ = name; }
> ```

Yeah, sure. I forgot about this one -_-'

PR #1447, opened 2020-01-16
Some cameras are flipped (in the z-axis) in ViewControl::ConvertFromPinholeCameraParameters

pablospe:
Some cameras are flipped (in the z-axis) when the bounding_box center of the scene is not visible (i.e. center is behind the camera). This situation could happen when you are in an indoor scene, for instance, and looking at the walls instead of towards the center of the room.

The fix is simple, basically adding a absolute value to the result of the dot product (projection to the 'front_' vector).

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1447)


germanros1987:
@griegler could you please help review this PR?

germanros1987:
ping.

germanros1987:
@syncle do you have any advice?

PR #1451, opened 2020-01-18
Merge 'rendering' to 'gui'

prewettg:
Merge rendering -> gui.

Maksim, can you review this merge?  Everything looks okay to me, but the click-dragging the mouse in SceneWidget looks really weird, it's not rotating around an axis like I'm expecting, more like gimbal lock or something. I'm worried that I missed something, but if that's expected behavior, just merge it.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1451)


Xardvor:
On my machine camera works just as expected (by me). For 'gimbal lock' problem, I think you're talking about case when camera looks on object straight from top or bottom. And yes, I locked camera rotations in this case for purpose. I think it's, like, common sense for 3D apps. But, if you have some other experience, please tell, maybe we should change current approach

PR #1453, opened 2020-01-21
Depth visualize

Xardvor:
Modes for View objects added:
Color - default rendering mode
Depth - rendering depth only
Normals - rendering normals directions as colors

This PR adds support for Color and Depth modes

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1453)


src/Open3D/Visualization/Rendering/Filament/FilamentView.cpp
```diff
@@ -85,10 +112,64 @@ void FilamentView::SetViewport(std::int32_t x,
 }
 
 void FilamentView::SetClearColor(const Eigen::Vector3f& color) {
-    view_->setClearColor({color.x(), color.y(), color.z(), 1.f});
+    clearColor_ = color;
+
+    // We apply changes immediately only in color mode
+    // In other cases color will be set on mode switch
+    if (mode_ == Mode::Color) {
```
prewettg: This seems like a good thing to add to a comment in the header file, e.g. "Color is only applied when in Mode::Color", as I would expect it to work in all modes otherwise. (Of course the meaning of "color" in the other modes is a little different, but as a caller I would expect to know that.)

src/Open3D/Visualization/Rendering/Filament/FilamentView.cpp
```diff
@@ -58,8 +67,23 @@ FilamentView::~FilamentView() {
...
+    switch (mode) {
+        case Mode::Color:
+            view_->setClearColor(
+                    {clearColor_.x(), clearColor_.y(), clearColor_.z(), 1.f});
+            break;
+        case Mode::Depth:
+            view_->setClearColor(kDepthClearColor);
+            break;
+        case Mode::Normals:
+            break;
```
prewettg: Wouldn't Mode::Normals need some clear color, too? I assume that the shader won't be run for areas that don't have any objects, so they will get whatever clear color there is, which might be something odd (e.g. sky blue) from Mode::Color.

src/Open3D/GUI/SceneWidget.h
```diff
@@ -54,6 +54,7 @@ class SceneWidget : public Widget {
 
     void SetBackgroundColor(const Color& color);
     void SetDiscardBuffers(const visualization::View::TargetBuffers& buffers);
+    void SetViewMode(visualization::View::Mode mode);
```
prewettg: It seems like open3d::gui ought to abstract away open3d::visualization where possible. Unfortunately, this would mean duplicating the enum in open3d::gui::SceneWidget, which seems like a bit of a waste. I don't have a strong opinion here, but as a caller, I'm always a little frustrated when a function uses something I have to go look up in a completely different section of the code (even if it is just Cmd-RightClick in Xcode)

src/Open3D/Visualization/Rendering/Filament/FilamentView.cpp
Xardvor: I will add code for this mode in my other task for normals rendering

src/Open3D/GUI/SceneWidget.h
Xardvor: I see there two possible solutions:
1. I'll add another enum with modes to `SceneWidget`
2. I'll remove `SetDiscardBuffers` and `SetViewMode` methods and add `GetView` method.
I personally prefer second because both methods more about rendering than about widget. And we already has this `GetScene` method also.

prewettg: Either are fine with me, although I agree that #2 does sound like the better option.

src/Open3D/GUI/Materials/depth.mat
```diff
@@ -1,14 +1,25 @@
...
         prepareMaterial(material);
-        const float n = 0.1;
-        const float f = 100;
-        float linearZ = (2.0 * n) / (f + n - gl_FragCoord.z * (f - n));
+
+        const float near = materialParams.cameraNear;
+        const float far = materialParams.cameraFar;
+
+        //float z = depth * 2.0 - 1.0; // back to NDC
+        //float linearZ =  (2.0 * near) / (far + near - z * (far - near));
```
prewettg: I assume these comments are development code and should be removed?

PR #1499, opened 2020-02-07
Moved 3rdparty packages down a level to not pollute the include space

tykurtz:
# Description

This PR fixes some issues with which directories are included in CMake. Additionally, it adds the option to use system flann instead of the Open3D fork.

The issue was these commands,

```cmake
# In root CMakeLists.txt
include_directories(
    ${PROJECT_SOURCE_DIR}/3rdparty
)
```
and
```cmake
# In 3rdparty/CMakeLists.txt
set(poisson_INCLUDE_DIRS ${CMAKE_CURRENT_SOURCE_DIR}/)

list(APPEND 3RDPARTY_INCLUDE_DIRS
     ${poisson_INCLUDE_DIRS}
```

both put `Open3D/3rdparty` on the include path. This covered up some bugs with several include variables pointing to the wrong directory, and a typo in`${tinyfobjloader_INCLUDE_DIRS}`. 

This PR removes adding `Open3D/3rdparty` to the include path and isolates the include paths to the respective packages `_INCLUDE_DIRS`. This ensures that when building against system packages, the wrong headers don't get included on accident.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1499)


wolfgangl:
For more context of why using system FLANN is important, on Ubuntu 18.04, using open3d and PCL at the same time leads to segfaults in some cases.  

This is because PCL links to the system copy of FLANN, which is at version 1.9.1 while the 3rd party version currently included in open3d is 1.8.4.   Compiling with open3d with system FLANN fixes these crashes and allows PCL and open3d to co-exist.

tykurtz:
I'll setup a windows environment when I get home and dig into the build error.

germanros1987:
@tykurtz are you planning to add the fix or should we close this PR?

tykurtz:
> @tykurtz are you planning to add the fix or should we close this PR?

Yes, apologies for taking so long. I'm still planning on applying the fix. It's just taking a little bit longer than I expected because I'm not as familiar with development on windows. I'll try to get it done by EOW

germanros1987:
@tykurtz any news?

tykurtz:
@germanros1987 fixed the include issue on windows, should be good to go.

PR #1501, opened 2020-02-08
Fix bug.

Forest75:
Pass correct parameter to function update_geometry().

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1501)


scimad:
Did you notice, the odd looking argument name `arg0`?
```
TypeError: update_geometry(): incompatible function arguments. The following argument types are supported:
    1. (self: open3d.open3d.visualization.Visualizer, arg0: open3d.open3d.geometry.Geometry) -> bool
```

PR #1512, opened 2020-02-13
rendering scenes into a buffer

Xardvor:
New interface introduced: `RenderToBuffer`. It allows user to create requests to renderer for rendering next frame into a buffer. User can choose settings for view mode, camera, viewport, etc. In particular View and Camera settings can be manipulated for this rendering mode.
When working with Filament~ classes user also can use `FilamentRenderToBuffer` for headless rendering, without creating `FilamentRender` and its children. **This would not work on headless systems!**

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1512)


Xardvor:
appveyor CI fails, will wait for PR#1500, possibly it will fix all issues

PR #1520, opened 2020-02-17
Enable building docs for pure python submodules

benjaminum:
This PR adds code for retrieving the modules to be documented from index.rst.
It relies fully on inspect to avoid special treatment for the modules defined with pybind.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1520)


benjaminum:
With these changes I can build documentation for the ml module and open3d.core.
Since the changes are not needed at the moment in master this is not urgent.

PR #1531, opened 2020-02-21
Gui app feedback

prewettg:
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1531)


src/Open3D/GUI/Label.cpp
```diff
@@ -110,6 +118,10 @@ Widget::DrawResult Label::Draw(const DrawContext& context) {
     ImGui::SetCursorPos(
             ImVec2(frame.x - context.uiOffsetX, frame.y - context.uiOffsetY));
     ImGui::PushItemWidth(frame.width);
+    bool isDefaultColor = (impl_->color != DEFAULT_COLOR);
+    if (isDefaultColor) {
```
Xardvor: I think there is misspell here, shouldn't it be `if (!isDefaultColor)`?

src/Open3D/GUI/Materials/metal.mat
```diff
@@ -1,6 +1,7 @@
 material {
     name : Metal,
     shadingModel : lit,
+    culling : none,
```
Xardvor: Please, don't switch culling like this. If you need pretty rendering of back oriented faces, use `doubleSided`.

src/Open3D/GUI/Menu.cpp
```diff
@@ -69,59 +93,44 @@ void Menu::AddItem(const char *name,
...
 
 void Menu::AddMenu(const char *name, std::shared_ptr<Menu> submenu) {
-    impl_->items.push_back({NO_ITEM, name, "", submenu});
+    impl_->items.push_back({NO_ITEM, name, "", submenu, submenu->impl_.get()});
 }
 
 void Menu::AddSeparator() {
-    impl_->items.push_back({NO_ITEM, "", "", nullptr, false, false, true});
+    impl_->items.push_back(
+            {NO_ITEM, "", "", nullptr, nullptr, false, false, true});
```
Xardvor: Maybe introduce static function for creating separator? Something like `MenuItem::Separator()`, all this nullptrs and flags just blow my mind %)

src/Open3D/GUI/SceneWidget.cpp
```diff
@@ -36,9 +36,7 @@
 
 #include <Eigen/Geometry>
 
-#include <iostream>
-
-#define ENABLE_PAN 0
+#define ENABLE_PAN 1
```
Xardvor: Why do we need `#define` here? Do we need to turn off camera pan on some platforms?

src/Open3D/GUI/Menu.cpp
prewettg: This *is* the function for creating the separator, I'm not sure what passing it off to another function that does exactly the same would acheive...

src/Open3D/GUI/SceneWidget.cpp
prewettg: Originally it didn't entirely work. But I'm not convinced that we are going to like it, either. If you pan and then rotate, the rotate rotates about a different point (obviously) but I think it's going to be easy to get confused.

examples/Cpp/GuiDemo.cpp
```diff
@@ -56,27 +56,31 @@ class DemoWindow : public gui::Window {
...
-        debugSubmenu->AddItem("Voxinated", "", DEBUG_VOXINATED);
-        debugSubmenu->AddItem("Seluna", "", DEBUG_SELUNA);
-        viewMenu->AddMenu("Debug", debugSubmenu);
-        menubar_->AddMenu("View", viewMenu);
-        auto helpMenu = std::make_shared<gui::Menu>();
-        helpMenu->AddItem("About", "", HELP_ABOUT);
-        helpMenu->AddItem("Contact", "", HELP_CONTACT);
-        menubar_->AddMenu("Help", helpMenu);
-        SetMenubar(menubar_);
+        if (!gui::Application::GetInstance().GetMenubar()) {
```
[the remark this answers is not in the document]
prewettg: This could be the second or later instance of DemoWindow. Although I'm not sure there's any harm it creating a second menubar.

src/Apps/Open3DViewer/Open3DViewer_mac.mm
```diff
@@ -79,6 +79,19 @@ - (BOOL)application:(NSApplication *)sender openFile:(NSString *)filename {
     mOpenEmptyWindow = false;  // LoadAndCreateWindow() always opens a window
     return (LoadAndCreateWindow(filename.UTF8String));
 }
+
+- (void)applicationWillTerminate:(NSNotification *)notification {
+    // The app will terminate after this function exits. This will result
+    // in the Application object in main() getting destructed, but it still
+    // thinks it is running. So tell it to quit.
+    open3d::gui::Application::GetInstance().Quit();
```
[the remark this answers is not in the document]
prewettg: I don't think I understand what you're asking. `Quit()` posts some cleanup events to the SDL event loop, which then get processed in the next line.  Maybe the comment is misleading, though, so I've updated it.

src/Apps/Open3DViewer/Open3DViewer_mac.mm
```diff
@@ -79,6 +79,19 @@ - (BOOL)application:(NSApplication *)sender openFile:(NSString *)filename {
...
+    // Process events for a while. (Everything should happen in the first
+    // call, but call it multiple times just in case.)
+    int  i  = 0;
+    while (open3d::gui::Application::GetInstance().RunOneTick() && i < 20) {
```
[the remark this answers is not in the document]
prewettg: This is mostly paranoia. I assume that if you post to the SDL event loop and then read from the event loop that it would show up immediately. But in case it doesn't, loop through a few times. When the app has finished, RunOneTick() will return false and the loop will exit. On my machine it exits on i=0 (which is what I would expect).  There's not a threading issue and nothing with Filament, it is just that we need to make sure that the run loop actually processes the cleanup events. We can't stop and cleanup in the ordinary way because the cleanup involves closing windows, and all of that assumes that events are being processed. We could write special-purpose code to do it outside of the event loop, but I think that's even worse. The problem here is that SDL isn't really using Apple's event loop, so we have to exit our event loop before Apple exits theirs, which they seem to do immediately when the `-applicationWillTerminate:` function returns.
   I could do `while(RunOneTick());` if you'd prefer, although there's the risk that it would never return. (Unlikely, but I tend to assume the worst, hence the 20, just as a safeguard) I could also name the 20 if you'd prefer.

src/Open3D/GUI/Label.cpp
prewettg: Good point. I think I was trying to avoid inverted logic in the `if` but I can't think of a better name.

src/Open3D/GUI/Materials/metal.mat
prewettg: Fixed

germanros1987:


```diff

Issues
======
- Added 4
           

Clones added
============
- src/Open3D/GUI/FileDialog.cpp  1
- src/Open3D/GUI/FileDialogNative.cpp  1
         

Clones removed
==============
+ src/Open3D/GUI/Dialog.h  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/intel-isl/Open3D/pullRequest?prid=5044985&bid=16613832)

PR #1561, opened 2020-03-05
Tonemapping is switched off

Xardvor:
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1561)


Xardvor:
Appears that depth mode is too dark after switching off tonemapping, will add commit with fix

PR #1569, opened 2020-03-06
Moved materials switch to settings ui

Xardvor:
Also reduced materials options count:
* Lit
* Unlit
* Normal map
* Depth

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1569)

### DIFF
--- a/examples/Cpp/GuiDemo.cpp
+++ b/examples/Cpp/GuiDemo.cpp
@@ -131,8 +131,9 @@ public:
                 materials->AddItem("Blue ceramic"); */
         auto materials = std::shared_ptr<gui::Combobox>(
                 new gui::Combobox({"Gold", "Red plastic", "Blue ceramic"}));
-        materials->SetOnValueChanged([](const char *newValue) {
-            std::cout << "New material: " << newValue << std::endl;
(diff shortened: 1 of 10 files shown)


prewettg:
___
*[src/Open3D/GUI/Combobox.h, line 55 at r1](https://reviewable.io/reviews/intel-isl/open3d/1569#-M1lI7Hn3-pwuW-pJMpt:-M2-gu6NEloJXFvT90_0:bek4odk) ([raw file](https://github.com/intel-isl/open3d/blob/8a29eed04b35debfb3e90ed83db3cbf9fa61cf4a/src/Open3D/GUI/Combobox.h#L55)):*
<details><summary><i>Previously, Xardvor wrote…</i></summary><blockquote>

The reasoning is what I need index instead of entry name. I wasn't sure which is better - put second argument or just replace argument with widget which generated a callback. If you want, I could add second argument.
</blockquote></details>

My inclination would be to override with SetOnValueChanged(std::function<void(int)> onValueChanged), and you get the index with that one and the value with the other one, which is what Qt does. But, std::function<void(int, const char *)> is also a good option. I like either of those better than just returning the pointer.

Issue #1621, opened 2020-03-16
Type error in TriangleMesh::create_from_point_cloud_alpha_shape 


FedeMiorelli:
I am not able to use the function `create_from_point_cloud_alpha_shape` due to type errors.

Below is a simple code to reproduce, and the error displayed.

```

import open3d as o3d
import numpy as np

pc = o3d.geometry.PointCloud()
pc.points = o3d.utility.Vector3dVector(np.random.randn(100,3))
T_recon = o3d.geometry.TriangleMesh()

alpha = 1.0
T_recon.create_from_point_cloud_alpha_shape(pc, alpha)
```

```
Traceback (most recent call last):
  File "C:\Temp\test_open3d.py", line 16, in <module>
    T_recon.create_from_point_cloud_alpha_shape(pc, alpha)
TypeError: create_from_point_cloud_alpha_shape(): incompatible function arguments. The following argument types are supported:
    1. (pcd: open3d.open3d.geometry.PointCloud, alpha: float, tetra_mesh: open3d::geometry::TetraMesh, pt_map: List[int]) -> open3d.open3d.geometry.TriangleMesh

Invoked with: geometry::PointCloud with 100 points., 1.0
```

Open3d v0.9.0, Windows 10 64 bit, Python 3.7.6.0
Wheel installed from [https://www.lfd.uci.edu/~gohlke/pythonlibs/](https://www.lfd.uci.edu/~gohlke/pythonlibs/)

griegler:
Addressed in #1624.

FedeMiorelli:
@griegler Thank you!

Is there a workaround I could use from Python until this is released and packages are updated?
Like creating instances of the two additional parameters in the same way they would be specified inside C++ if they were not passed.

griegler:
You could use something like:

```python
tetra_mesh, pt_map = o3d.geometry.TetraMesh.create_from_point_cloud(pcd)
mesh = o3d.geometry.TriangleMesh.create_from_point_cloud_alpha_shape(pcd, alpha, tetra_mesh, pt_map)
```

FedeMiorelli:
That works, thanks

PR #1624, opened 2020-03-16
Fix of alpha shape pybind

griegler:
Addressing #1621.
In the Python example we call the surface reconstruction once with and once without pre-computed convex hull.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1624)


griegler:
@yxlao is there a problem with travis?

yxlao:
yes, the mac test was restarted

PR #1628, opened 2020-03-17
Suppress developer warning for OpenGL library with CMAKE_VERSION >= 3.11

akashsharma02:
A minor change in CMakeLists.txt
Looks like building glew/glfw from source always requires the legacy OpenGL library to be linked.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1628)


yxlao:
thanks, tested with CMake `13.16.4`

Issue #1636, opened 2020-03-19
segment_plane crashes when len(pcd.points) < ransac_n


christianbrugger:
Call segment_plane on a pointcloud with less points than `ransac_n` points leads to a segmentation fault.

**To Reproduce**
Here is an example with `ransac_n = 3` and 1 points:
```python
import open3d as o3d 
import numpy as np

a = np.zeros((1, 3), float)
pcd = o3d.geometry.PointCloud()
pcd.points = o3d.utility.Vector3dVector(a)

res = pcd.segment_plane(distance_threshold=0.1, ransac_n=3, num_iterations=100)
print(res)
```

This will print
```
Segmentation fault (core dumped)
```

**Expected behavior**
Raise a Python exception.

**Environment:**
 - OS: Ubuntu 16.04
 - Python version: 2.7.12
 - Open3D version: 0.9.0.0
 - Is this remote workstation?: no
 - How did you install Open3D?: pip

benjaminum:
Thanks for the code example. The bugfix #1637 is now in the master branch.

christianbrugger:
Thanks for the quick response Benjamin. I had a look at your commit. Can you make sure it also works with the following example?
```python
import open3d as o3d 
import numpy as np

a = np.zeros((100, 3), float)
pcd = o3d.geometry.PointCloud()
pcd.points = o3d.utility.Vector3dVector(a)

res = pcd.segment_plane(distance_threshold=0.1, ransac_n=200, num_iterations=100)
print(res)
```

benjaminum:
Yes, the new example generates a python exception.

christianbrugger:
Perfect, thanks for checking.

PR #1637, opened 2020-03-19
log error in PointCloud::SegmentPlane() if there are less than 3 points

benjaminum:
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1637)


benjaminum:
This addresses issue #1636

Issue #1689, opened 2020-04-07
RGBDOdometry with RGBD frames with convert_rgbd_to_intensity=False


akashsharma02:
**Describe the bug**
Why is it necessary for RGBD odometry to only support RGBD images which have convert_rgbd_to_intensity=True?

When I run RGBD odometry with two frames having color information, I get a Open3D warning for incorrect image sizes. I think, either the warning must be improved or support for this must be added. 

From code I observe that: 

```
inline bool CheckRGBDImagePair(const geometry::RGBDImage &source,
                                const geometry::RGBDImage &target) 
      return (CheckImagePair(source.color_, target.color_) &&
                  CheckImagePair(source.depth_, target.depth_) &&
                  CheckImagePair(source.color_, source.depth_) &&
                  CheckImagePair(target.color_, target.depth_) &&
                  CheckImagePair(source.color_, target.color_) &&
                  source.color_.num_of_channels_ == 1 &&
                  source.depth_.num_of_channels_ == 1 &&
                  target.color_.num_of_channels_ == 1 &&
                  target.depth_.num_of_channels_ == 1 &&
                  source.color_.bytes_per_channel_ == 4 &&
                  target.color_.bytes_per_channel_ == 4 &&
                  source.depth_.bytes_per_channel_ == 4 &&
                  target.depth_.bytes_per_channel_ == 4);
```

**Expected behavior**
Compute odometry regardless of num of channels in the source/target color images.

**Screenshots**

**Environment (please complete the following information):**
 - OS: Ubuntu 18.04 LTS
 - Python version: 3.7 (But using C++ API)
 - Open3D version: master
 - Is this remote workstation?: no
 - How did you install Open3D?: build_from_source (GCC 7 build, static library with build_dependency_from_source=off)

Issue #1356, opened 2019-11-26
Odometry (Python API) gives erroneous warning and doesn't compute odometry


scimad:
I am trying to compute odometry from two RGBD images. So, I tried to read two consecutive RGBD frames from Azure Kinect and use the `open3d.odometry.compute_rgbd_odometry()` function to get what I wanted. 
```
is_success, motion_matrix, info_matrix = o3d.odometry.compute_rgbd_odometry(rgbd, new_rgbd, 
pinhole_camera_intrinsic=my_camera_intrinsics)
```
I got this warning:

```[Open3D WARNING] [RGBDOdometry] Two RGBD pairs should be same in size.```

Of course the RGBD pairs are created in exact same fashion and also RGB and D are aligned. So, I tried to see if the odometry is working by using same source and target RGBD image.

```
is_success, motion_matrix, info_matrix = o3d.odometry.compute_rgbd_odometry(rgbd, rgbd, 
pinhole_camera_intrinsic=my_camera_intrinsics)
```

I still get the warning. The `is_success` also returns `False`. What's causing this? Is this a bug?

**Environment:**
 - OS: Ubuntu 18.04
 - Python version: 3.6.8
 - Open3D version: build from 802902c9

**Comment by theNded** [comment]
Can you please check if RGB and D are with the same size? Raw frames are not with the same size, you will need to use the [decoder](https://github.com/intel-isl/Open3D/blob/master/examples/Python/ReconstructionSystem/sensors/azure_kinect_mkv_reader.py) to align them.

**Comment by scimad** [comment]
Yeah, I am trying to modify on [this](https://github.com/intel-isl/Open3D/blob/master/examples/Python/ReconstructionSystem/sensors/azure_kinect_viewer.py) example and run with `--align_depth_to_color` argument. The frames are verified to have same size (height & width).

**Comment by scimad** [comment]
@theNded, Here's the complete code for reference.

```
import argparse
import open3d as o3d
import numpy as np


class PCDViewerWithCallback:

    def __init__(self, config, device, align_depth_to_color = True):
        self.flag_exit = False
        self.align_depth_to_color = align_depth_to_color

        self.sensor = o3d.io.AzureKinectSensor(config)
        if not self.sensor.connect(device):
            raise RuntimeError('Failed to connect to sensor')

    def escape_callback(self, vis):
        self.flag_exit = True
        return False

    def run_pointcloud(self):
        glfw_key_escape = 256
[...]

PR #1690, opened 2020-04-08
Fixes issue #1689

akashsharma02:
This PR fixes an issue I have raised: #1689.
 It should be okay IMO to internally convert RGB to intensity for RGBDOdometry.

Adding some more details to this problem at the request of @theNded 
There has been another issue raised for this earlier: #1356

The existing RGBD odometry requires that the input RGBD frames, each containing an RGB image is represented in intensity format, and disallows odometry for RGBD frames which maintain ```convert_rgb_to_intensity=False```. 

I think it should not be forced on the user to convert RGB frame to Intensity just to compute odometry because the user then would need to maintain a copy of the RGBD frame for visualization which is redundant. 

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1690)


germanros1987:


```diff

Clones added
============
- src/Open3D/Odometry/Odometry.cpp  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/intel-isl/Open3D/pullRequest?prid=5324245&bid=17423744)

theNded:
LGTM. @syncle do you like this minor modification?

PR #1732, opened 2020-04-21
Draw subsampled version of large point clouds when rotating

prewettg:
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1732)

### DIFF
--- a/src/Open3D/GUI/Application.cpp
+++ b/src/Open3D/GUI/Application.cpp
@@ -103,6 +103,7 @@ namespace gui {
 struct Application::Impl {
     std::string resourcePath;
     Theme theme;
+    double lastTime = 0.0;
     bool isGLFWinitalized = false;
     bool isRunning = false;
     bool shouldQuit = false;
@@ -210,6 +211,10 @@ void Application::Initialize() {
 void Application::Initialize(int argc, const char *argv[]) {
     impl_->resourcePath = findResourcePath(argc, argv);
     impl_->theme.fontPath = impl_->resourcePath + "/" + impl_->theme.fontPath;
+}
+
+double Application::Now() const {
+    return glfwGetTime();
 }
 
 std::shared_ptr<Menu> Application::GetMenubar() const { return impl_->menubar; }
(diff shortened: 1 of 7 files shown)


prewettg:
___
*[src/Open3D/GUI/Application.cpp, line 343 at r2](https://reviewable.io/reviews/intel-isl/open3d/1732#-M5c8RjS-fV2HY-f4TyW:-M5cBrNwDChMzQ7VWIxy:b-kq5ys9) ([raw file](https://github.com/intel-isl/open3d/blob/ae735f145a4dd6decae078a80b9d3c755dbd37df/src/Open3D/GUI/Application.cpp#L343)):*

It appears that TickEvents are not implemented (i.e., doing SetTickEventsEnabled(true)) and that TickEvents would never be run until this was added. Why limit running the events to when the frame is slow? Can we envision wanting TickEvents that aren't just for managing the FAST/BEST rendering path? If not then the Ticks should be internal implementation details and not publicly available.
</blockquote></details>

I was going to remove SetTickEventsEnabled(), as they are always enabled now. (At the time of the original implementation, back when we were using SDL, I was afraid it would use up an unnecessarily large amount of CPU, but I checked it this time, and it doesn't.) Tick events are, even now, used for more than managing FAST/BEST; they also handle movement/rotation via keyboard in Fly mode, although it is basically the same code path. I could imagine them being used for animations in the future, such as flying the camera over a predefined path, which seems like it was a feature in previous visualizers.

prewettg:
___
*[src/Open3D/GUI/SceneWidget.cpp, line 252 at r2](https://reviewable.io/reviews/intel-isl/open3d/1732#-M5c9xunEbhM4LEryHP4:-M5cMU6G6VgMspaRsJaR:b21p7n9) ([raw file](https://github.com/intel-isl/open3d/blob/ae735f145a4dd6decae078a80b9d3c755dbd37df/src/Open3D/GUI/SceneWidget.cpp#L252)):*

maybe put redraw = true in the hasKey lambda so it doesn't have to get be included in each if statement? If someone adds another if statement later there's no risk of forgetting to include redraw=true
</blockquote></details>

I don't think something named "hasKey" should have side-effects. I've reworked the code to generalize things a bit, which should address your issue.

prewettg:
___
*[src/Open3D/Visualization/Visualizer/GuiVisualizer.cpp, line 1326 at r2](https://reviewable.io/reviews/intel-isl/open3d/1732#-M5cAXPm7LnStOBPhBRg:-M5cMp9q9ERp079Xymx6:bbx7ve5) ([raw file](https://github.com/intel-isl/open3d/blob/ae735f145a4dd6decae078a80b9d3c755dbd37df/src/Open3D/Visualization/Visualizer/GuiVisualizer.cpp#L1326)):*

This number will differ based on the GPU. Is this set for lowest hardware we expect users to have?
</blockquote></details>

Yes, I tested this with the Iris+ 655 that I have and it works smoothly there. It would be great to have it be vary based on the ability of the GPU being used, but I don't know of any way to do that. Checking rendering time would be a decent way, but it doesn't look like Filament offers that functionality.

PR #1748, opened 2020-04-24
Changed Alpha and Stencil bit defaults to fix hang/crash with Nvidia …

errissa:
…driver on Linux. Tested on Linux with Nvidia binary driver and Intel integrated graphics. Not yet tested on Windows or Mac.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1748)

### DIFF
--- a/src/Open3D/GUI/Window.cpp
+++ b/src/Open3D/GUI/Window.cpp
@@ -198,9 +198,13 @@ Window::Window(const std::string& title,
     glfwWindowHint(GLFW_CONTEXT_VERSION_MINOR, 2);
     glfwWindowHint(GLFW_OPENGL_FORWARD_COMPAT, GL_TRUE);
     glfwWindowHint(GLFW_OPENGL_PROFILE, GLFW_OPENGL_CORE_PROFILE);
+    // NOTE: Setting alpha and stencil bits to match GLX standard default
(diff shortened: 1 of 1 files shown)


prewettg:
___
*[src/Open3D/GUI/Window.cpp, line 201 at r1](https://reviewable.io/reviews/intel-isl/open3d/1748#-M5i7MlF2D19dq4QJHae:-M5i7MlF2D19dq4QJHaf:baeq579) ([raw file](https://github.com/intel-isl/open3d/blob/f8ca092e879e929a31f27ce26fee0da180c149c3/src/Open3D/GUI/Window.cpp#L201)):*
> ```cpp
>     glfwWindowHint(GLFW_OPENGL_FORWARD_COMPAT, GL_TRUE);
>     glfwWindowHint(GLFW_OPENGL_PROFILE, GLFW_OPENGL_CORE_PROFILE);
>     glfwWindowHint(GLFW_ALPHA_BITS, 0);
> ```

This seems to work on macOS (AMD) and linux (Intel). It would be nice to have some comments explaining what these are for. On the surface, it would seem that setting no alpha bits would preclude alpha blending, but that works fine. Also, I can easily imagine myself coming across this code, saying "what is this for?!", testing to see if removing it breaks anything, and not noticing a failure on Linux with nVidia.

PR #1749, opened 2020-04-24
Read header of geometry files to determine what geometry it contains

prewettg:
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1749)

### DIFF
--- a/src/Open3D/IO/ClassIO/FileFormatIO.cpp
+++ b/src/Open3D/IO/ClassIO/FileFormatIO.cpp
@@ -1,0 +1,61 @@
+// ----------------------------------------------------------------------------
+// -                        Open3D: www.open3d.org                            -
+// ----------------------------------------------------------------------------
+// The MIT License (MIT)
+//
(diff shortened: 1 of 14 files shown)


prewettg:
___
*[src/Open3D/IO/FileFormat/FileGLTF.cpp, line 94 at r2](https://reviewable.io/reviews/intel-isl/open3d/1749#-M5mgVVL1g9wT-Uz3h5t:-M5xltw78watdNPfWKP9:b-mns6ay) ([raw file](https://github.com/intel-isl/open3d/blob/87f5d887524f865397cc6ae74274fa8a8f8d212e/src/Open3D/IO/FileFormat/FileGLTF.cpp#L94)):*

Does GLTF always have triangles? It's not possible to have a points only GLTF fileT?
</blockquote></details>

I don't know, but if it is possible, we don't support it, because there is no ReadPointCloudFromGLTF(), only ReadTriangleMeshFromGLTF(). Since we don't support it, I figured there was no point in detecting that situation.

PR #1778, opened 2020-05-01
Merge from 1577

errissa:
Not strictly speaking a merge from 1577 though it started that way. Brought in functionality from 1577 to support loading texture maps and other material properties from OBJ files. Added support in defaultLit and defaultUnlit materials for those texture maps and also added support in Open3DViewer to use the texture maps if present.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1778)


prewettg:
___
*[src/Open3D/Geometry/TriangleMesh.h, line 746 at r3](https://reviewable.io/reviews/intel-isl/open3d/1778#-M6bOWFYCeAZ42Y1oeUM:-M6bOWFYCeAZ42Y1oeUN:b-t5ih1r) ([raw file](https://github.com/intel-isl/open3d/blob/558506bf9fa65d6890b5f8c129cde1032f12a203/src/Open3D/Geometry/TriangleMesh.h#L746)):*
<details><summary><i alt="summary">Quoted 9 lines of code…</i></summary>

> ```
>             union {
>                 float f4[4] = {0};
>                 float f3[3];
>                 float f2[2];
>                 float f;
> 
>                 float x, y, z, w;
>                 float r, g, b, a;
>             };
> ```
</details>


I like this, but everywhere else the code tends to use Eigen for colors and what-have-you.

prewettg:
*[a discussion](https://reviewable.io/reviews/intel-isl/open3d/1778#-M6bRj1wCZ-hZAKFw2aC:-M6bRj1wCZ-hZAKFw2aD:bhrswa3) (no related file):*
I'm not sure how to do this with the Reviewable UI, but I'd like to mark my comments as "I'd like to see this if it isn't going to hold up the release" except for the macro comments which is "I'd really, really like to see anything else but a macro here, but I won't hold up a release, albeit with grumblings and mumblings if we end up with a macro."

---

trancelestial:
Hi, is there are way to supply reflectance map in .mtl file similar to the way base color, roughness and metallic are supplied using Kd/map_Kd, Pr/map_Pr and Pm/map_Pm texture references? I saw that in default lit material there is reflectance as a parameter but couldn't find if there is a way to use non default value when I want to use custom material (.obj+.mtl). Thanks in advance!

errissa:
Currently the OBJ loader does not support specifying a reflectance value or loading a reflectance map. The only way to change the default value would be to specify it programmatically.

Issue #1770, opened 2020-04-30
Debianization of Open3D


roehling:
I'm currently working on bringing Open3D into Debian, and I found a number of improvements to the build system which make it easier to build Open3D for a (Linux) distribution. Are you generally interested in receiving these patches as PRs?

Also, I noticed that you do not explicitly set an SONAME for the shared library. For the current release, I'm leaning towards 0.9 as SOVERSION, but it would be good to have an "official" ABI version attached to each release.

griegler:
In my opinion, our cmake scripts are a bit messy at the moment. So yes, it would be great if you could contribute a PR to improve them.

There exists `src/Open3D/version.txt`, but I think @yxlao could provide more information for the second question/request.

roehling:
I'll start with a few simple PRs and work my way up to the more complex changes I made to the CMake scripts (which you will want to review more thoroughly first)

roehling:
I was a bit work to turn my "make this Debian friendly" tweaks to the CMake scripts into something platform-independent, but I think I got it.

ssheorey:
@roehling now that we have a Debian package, is it OK to close this issue? Please let us know / submit PRs if any updates are needed for future Open3D versions.

roehling:
Yes, we can close this

PR #1782, opened 2020-05-01
Refactor CMake buildsystem

roehling:
This PR is part of issue #1770 to make it easier to package Open3D for
(Linux) distributions.

The old CMake scripts made heavy use of variables to propagate dependencies. Actually, CMake has pretty decent capabilities to do this on its own, which is why I reframed the whole build configuration in terms of CMake targets.

**EDIT**: I believe the PR is ready for review. Here is why think it should be merged:

* Exports a single `Open3D::Open3D` CMake target for convenient library usage. Takes care of include paths, preprocessor macros and links all dependencies.
* Source tree is embedding-friendly. Just use `add_subdirectory(path/to/open3d/sources)` in a project that depends on Open3D, and the `Open3D::Open3D` target becomes available.
* Distribution-friendly dependency management. (Almost) all dependencies can be switched between the vendored version in `3rdparty` and the system version. Thanks to CMake targets, this is transparent to Open3D developers and users.
* No meddling with FLAGS variables and almost no more hidden dependencies on CMake variables.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1782)


roehling:
I did remove [this line](https://github.com/intel-isl/Open3D/blob/2af072ac841f73433cdfceb478230870a46e8884/CMakeLists.txt#L263) that enables optimization for Debug Mode, because I figured if you want release optimizations with debug symbols, you could just as well use the build type RelWithDebInfo (which I made the new default).

If you prefer the old behavior, I can add an override for `CMAKE_CXX_FLAGS_DEBUG` instead.

germanros1987:
Hi @roehling,

This PR is great and we would like to thank you for the hard work done. 
The reason for which PR has not been merged yet is that we have a large upcoming PR to merge the gui branch into master. This gui branch brings an important update in terms of rendering and GUI. However, this is not compatible with your new CMake structure.

We were wondering if you would be willing to have a video call to discuss this. In order to organize a call, please join our Discord network: https://discord.gg/D35BGvn and reach out to any member (.e.g.  germanros).

Thanks!

roehling:
I think the PR is ready to merge again. There's something weird going on with the MacOS build; I tried to investigate as well as I could without actually owning a Mac, and I think there might be something fishy with some preinstalled libraries in the Github build image. It does work everywhere else and with a full build from source, so I leave it as is.

There are a few other issues, such as the weird way the Open3DViewer gets installed, but those are better saved for a different PR.

yxlao:
Will perform the following tests (used as my personal notes):
```shell
git fetch https://github.com/roehling/Open3D.git refactor-cmake-buildsystem:cmake-refactor
git co cmake-refactor
```
- [ ] Ubuntu: build filament from source
- [ ] Ubuntu: do not build filament from source
- [ ] Ubuntu: build .so, example project linking
- [ ] Ubuntu: build .a, example project linking
- [ ] Win: static link C++ runtime
- [ ] Win: dynamic link C++ runtime
- [ ] Win: build .dll, example project linking
- [ ] Win: build .lib, example project linking
- [ ] Mac: build deps (e.g. jpeg) from source
- [ ] Mac: do not build deps (e.g. jpeg) from source
- [ ] Mac: build filament from source
- [ ] Mac: build .so/.dylib, example project linking
- [ ] Mac: build .a, example project linking

yxlao:
Q1: Shall we set `BUILD_EIGEN3=ON` by default? What will be the disadvantage if we do so?

Current issue:
If I configure with
```shell
cmake -DBUILD_CUDA_MODULE=ON ..
```
and, by default `-DBUILD_EIGEN3=OFF`. This creates the following error:
```cpp
[ 46%] Building CUDA object src/Open3D/Core/CMakeFiles/Core.dir/Kernel/NonZeroCUDA.cu.o                                                                  
[ 46%] Building CUDA object src/Open3D/Core/CMakeFiles/Core.dir/Kernel/ReductionCUDA.cu.o                                                                
[ 53%] Built target Visualization                                                                                                                        
In file included from /home/yixing/repo/Open3D/src/Open3D/Utility/Console.h:29:0,                                                                        
                 from /home/yixing/repo/Open3D/src/Open3D/Core/SizeVector.h:34,                                                                          
                 from /home/yixing/repo/Open3D/src/Open3D/Core/Kernel/Reduction.h:29,                                                                    
                 from /home/yixing/repo/Open3D/src/Open3D/Core/Kernel/ReductionCUDA.cu:27:                                                               
/usr/include/eigen3/Eigen/Core:42:14: fatal error: math_functions.hpp: No such file or directory                                                         
     #include <math_functions.hpp>                                                                                                                       
              ^~~~~~~~~~~~~~~~~~~~                                                                                                                       
compilation terminated.          
```
which seems to be related to eigen version. 

My system Eigen is:
```
libeigen3-dev is already the newest version (3.3.4-4).
```

3rdparty/find_dependencies.cmake
```diff
@@ -0,0 +1,874 @@
...
+
+# Filament
+if(ENABLE_GUI)
+    if(BUILD_FILAMENT)
+        message(STATUS "Building third-party library Filament from source")
+        if(MSVC OR (CMAKE_C_COMPILER_ID STREQUAL "Clang" AND CMAKE_CXX_COMPILER_ID STREQUAL "Clang"))
+            set(FILAMENT_C_COMPILER "${CMAKE_C_COMPILER}")
+            set(FILAMENT_CXX_COMPILER "${CMAKE_CXX_COMPILER}")
+        else()
+            find_program(CLANG_CC NAMES clang-7 clang-8 clang-9 clang-10 clang)
```
yxlao, line 809: Q2: Shall we really allow falling back to `clang`? In my case, I have `clang-6` installed only, and the compilation fails in the middle without proper indication that the compiler is not supported. Maybe we can call `clang --version` and extract the version?

roehling:
There's a CUDA related bug in Eigen 3.3.4; I encountered this, too.

The main reason you'd want Open3D to use the system Eigen library is consistency, especially if you have multiple third-party libraries in your project which all use Eigen. But then again, you may run into bugs and outdated library versions like this.

I have no strong preference either way. Do you find it more likely that people mix and match different Eigen versions and have difficult-to-trace bugs, or people failing to understand that their system version is buggy and they need to use the shipped version? I'd choose the default option that generates fewer support questions down the line.

3rdparty/find_dependencies.cmake
roehling: Sure, we can do that. Come to think of it, I should probably switch the order so the newest available compiler version is used.

yxlao:
For Eigen, it is common for users to mix the system and included version (e.g. https://github.com/intel-isl/Open3D/issues/849#issuecomment-467423228). I would prefer to set it to set `BUILD_EIGEN3=ON` by default.

yxlao:
Q3: The `RelWithDebInfo` works well on ubuntu. What's the situation on Windows?

This part we had before wasn't elegant: we need to copy the file manually and we have to assume the `Release` folder. Do you see some possibility to improve this?
https://github.com/intel-isl/Open3D/blob/c25018853cb12dd939b51170d25f60fbc1c1406e/docs/_static/C++/CMakeLists.txt#L57-L72

And, it seems like we can only set Windows build type at build time, not configure time in CMake. Is this true? Do we also have `RelWithDebInfo`?
https://github.com/intel-isl/Open3D/blob/c25018853cb12dd939b51170d25f60fbc1c1406e/.appveyor.yml#L38

3rdparty/find_dependencies.cmake
yxlao: This might be useful: https://stackoverflow.com/a/16209709/1255535

roehling:
Visual Studio is a [multi-configuration generator](https://cmake.org/cmake/help/latest/prop_gbl/GENERATOR_IS_MULTI_CONFIG.html). If you need to know which configuration is currently built, the generator expression `$<CONFIG>` should work.

yxlao:
Q4: On macOS, `-DBUILD_SHARED_LIBS=OFF` crates some issue for the external library to link against open3d.

```shell
mkdir build && cd build
cmake -DCMAKE_INSTALL_PREFIX=~/open3d_install -DBUILD_SHARED_LIBS=OFF ..
cmake install -j
cd ../docs/_static/C++
cmake -DCMAKE_INSTALL_PREFIX=~/open3d_install ..
make -j
```
Got the following warning:
```shell
ld: warning: direct access in function 'open3d::visualization::VisualizerWithVertexSelect
ion::AddPickedPoints(std::__1::vector<int, std::__1::allocator<int> >)' from file '/Users
/ylao/open3d_install/lib/libOpen3D.a(VisualizerWithVertexSelection.cpp.o)' to global weak
 symbol 'guard variable for open3d::utility::Logger::i()::instance' from file 'CMakeFiles
/TestVisualizer.dir/TestVisualizer.cpp.o' means the weak symbol cannot be overridden at r
untime. This was likely caused by different translation units being compiled with differe
nt visibility settings.
```

Full terminal log: https://asciinema.org/a/wT7BrOQfF6u9sD3PMWd5xlVvT

roehling:
The linker warnings on Mac are a nuisance. I think they come from [this line](https://github.com/intel-isl/Open3D/blob/26ffff0db5590d1c7e6d9295921f58cb101cc564/CMakeLists.txt#L277), which I translated to [this line](https://github.com/roehling/Open3D/blob/e950e1e471a14774d6a5a6f497439309d08a3352/CMakeLists.txt#L300).

I'm not sure why those are needed. I left it out at first and then put it back in because I kept getting build failures in the CI. Maybe it's not needed now after all?

yxlao:
is there a way to hide those warnings?

roehling:
None that I know of. You'd need to force the same visibility on all object files, which would mean that you force hidden visibility for projects that use Open3D, and even if you can do it (with public target_compile_options), I don't think that's a good idea.

roehling:
I think I'll try and remove that visibility setting again, just to see if it causes build failures. Then, we would get rid of the linker warnings.

roehling:
I changed the visibility settings, which does create warnings when linking the static Open3D library against the Python module, but not when building applications with Open3D. Arguably, it's better to have the warnings once during the Open3D build than every time someone uses the library, so I'd suggest we leave it that way.

The long-term solution would be to compile everything with hidden visibility and explicitly export the Open3D symbols which are needed, windows DLL style. That is definitely out of the scope of this PR.

yxlao:
Q5: Some issues on Windows

1. **Static C++, dll**

    ```shell
    cmake -G "Visual Studio 16 2019" -A x64 -DSTATIC_WINDOWS_RUNTIME=ON -DBUILD_SHARED_LIBS=ON -DENABLE_JUPYTER=OFF -DCMAKE_INSTALL_PREFIX=C:\open3d_install -DPYTHON_EXECUTABLE=%USERPROFILE%\Miniconda3\envs\open3d3\python.exe ..

    cmake --build . --parallel %NUMBER_OF_PROCESSORS% --target install
    ```

    Error:

    ```shell
    LINK : fatal error LNK1189: library limit of 65535 objects exceeded [C:\Users\yixing\repo\Open3D\build\src\Open3D\Open3D.vcxproj]
    ```

2. **Static C++, lib**

    ```shell
    cmake -G "Visual Studio 16 2019" -A x64 -DSTATIC_WINDOWS_RUNTIME=ON -DBUILD_SHARED_LIBS=OFF -DENABLE_JUPYTER=OFF -DCMAKE_INSTALL_PREFIX=C:\open3d_install -DPYTHON_EXECUTABLE=%USERPROFILE%\Miniconda3\envs\open3d3\python.exe ..
    ```

    Error:

    ```shell
    -- Configuring done
    CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
    Please set them or make sure they are set and tested correctly in the CMake files:
    ZLIB_LIBRARY
        linked by target "png" in directory C:/Users/yixing/repo/Open3D/3rdparty/libpng

    -- Generating done
    CMake Generate step failed.  Build files cannot be regenerated correctly.
    ```

    However, the build compiles successfully in the end.

    Possible solution: reverse the order of png and zlib, and use `target_link_libraries(3rdparty_png INTERFACE 3rdparty_zlib)` at the end.

    ```cmake
    message(STATUS "Building third-party library zlib from source")
    add_subdirectory(${Open3D_3RDPARTY_DIR}/zlib)
    import_3rdparty_library(3rdparty_zlib INCLUDE_DIR ${Open3D_3RDPARTY_DIR}/zlib LIBRARIES ${ZLIB_LIBRARY})
    add_dependencies(3rdparty_zlib ${ZLIB_LIBRARY})
    set(ZLIB_TARGET "3rdparty_zlib")
    message(STATUS "Building third-party library libpng from source")
    add_subdirectory(${Open3D_3RDPARTY_DIR}/libpng)
    import_3rdparty_library(3rdparty_png INCLUDE_DIR ${Open3D_3RDPARTY_DIR}/libpng/ LIBRARIES ${PNG_LIBRARIES})
    add_dependencies(3rdparty_png ${PNG_LIBRARIES})
    target_link_libraries(3rdparty_png INTERFACE 3rdparty_zlib)
    set(PNG_TARGET "3rdparty_png")
    ```

3. **Dynamic C++, dll**

    ```shell
    cmake -G "Visual Studio 16 2019" -A x64 -DSTATIC_WINDOWS_RUNTIME=OFF -DBUILD_SHARED_LIBS=ON -DENABLE_JUPYTER=OFF -DCMAKE_INSTALL_PREFIX=C:\open3d_install -DPYTHON_EXECUTABLE=%USERPROFILE%\Miniconda3\envs\open3d3\python.exe ..

    cmake --build . --parallel %NUMBER_OF_PROCESSORS% --target install
    ```

    Error:

    ```shell
    LINK : fatal error LNK1189: library limit of 65535 objects exceeded [C:\Users\yixing\repo\Open3D\build\src\Open3D\Open3D.vcxproj]
    ```

4. **Static C++, lib**

    ```shell
    cmake -G "Visual Studio 16 2019" -A x64 -DSTATIC_WINDOWS_RUNTIME=OFF -DBUILD_SHARED_LIBS=OFF -DENABLE_JUPYTER=OFF -DCMAKE_INSTALL_PREFIX=C:\open3d_install -DPYTHON_EXECUTABLE=%USERPROFILE%\Miniconda3\envs\open3d3\python.exe ..
    ```

    Same error:

    ```shell
    -- Configuring done
    CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
    Please set them or make sure they are set and tested correctly in the CMake files:
    ZLIB_LIBRARY
        linked by target "png" in directory C:/Users/yixing/repo/Open3D/3rdparty/libpng

    -- Generating done
    CMake Generate step failed.  Build files cannot be regenerated correctly.
    ```
    
    However, the build compiles successfully in the end.

yxlao:
On `master` there was a commit that changed the threshold for `test_color_map.py`. We can change it back in this PR if needed.

roehling:
I reversed the order for zlib and libpng.

Increasing the tolerance was a stop-gap measure to let the test pass. Judging from [this paragraph](https://github.com/libjpeg-turbo/libjpeg-turbo/blob/master/README.md#what-about-libjpeg-v9), the libJPEG v9 API color transforms may be at fault. Maybe there's a way to fix this in the JPEG loader, so the results are identical regardless of the library version?

I've never seen the 65535 object limit before, but my internet search suggests that this may be caused by a problem with older CMake versions. Or maybe the CMake cache in the build folder is corrupted? I've had problems before with stale cache entries, especially when moving between different versions of `CMakeLists.txt` files.

roehling:
After some more digging, I found that a compliant JPEG decompression library may have a [1-bit difference per pixel component](https://en.wikipedia.org/wiki/JPEG#Required_precision), which might explain the difference between libjpeg and its fork libjpeg-turbo.

A 1-bit difference for an 8-bit color value is a tolerance of ~3.9e-3; in the worst case, if the decoder is consistently biased, the mean vector can have the same tolerance. Therefore, I'm changing the rtol value back and add a comment why.

yxlao:
The png/zlib and tolerance parts look good to me.

My CMake is 3.17.0 on windows. The interesting thing is that the `library limit of 65535 objects exceeded` error only happens if `BUILD_SHARED_LIBS=OFF`.

yxlao:
More data points:

```bash
# This works
cmake -G "Visual Studio 16 2019" -A x64 -DSTATIC_WINDOWS_RUNTIME=ON -DBUILD_SHARED_LIBS=ON -DENABLE_JUPYTER=OFF -DCMAKE_INSTALL_PREFIX=C:\open3d_install -DPYTHON_EXE CUTABLE=%USERPROFILE%\Miniconda3\envs\open3d3\python.exe ..
cmake --build . --verbose --parallel %NUMBER_OF_PROCESSORS% --config Release --target install

# But with the following warning, seem like the config not propagated correctly
LINK : warning LNK4098: defaultlib 'LIBCMT' conflicts with use of other libs; use /NODEFAULTLIB:library
```

roehling:
I think this applies to our 65535 objects problem:

https://developercommunity.visualstudio.com/content/problem/220174/fix-msvc-65535-symbol-limit-in-lib-files-lnk1189.html

yxlao:
We're getting closer to solving the issues. It is good news that Appveyor can recreate the 65535 objects problem on `Debug` build.

I did some additional experiments and added a patch in https://github.com/intel-isl/Open3D/pull/1908. If the tests pass, you can cherry-pick 7acbd4b as a starting point and run experiments and make improvements on top of that.

## Part I. Solving the `/bigobj` issue.
#### Exp 1. This works (putting `/bigobj` after `project(Open3D VERSION ${OPEN3D_VERSION} LANGUAGES C CXX)`):

```cmake
...
project(Open3D VERSION ${OPEN3D_VERSION} LANGUAGES C CXX)
...
set(CMAKE_C_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} /bigobj")
set(CMAKE_C_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} /bigobj")
set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} /bigobj")
set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} /bigobj")
...
```

#### Exp 2. This doesn't work (putting `/bigobj` before `project(Open3D VERSION ${OPEN3D_VERSION} LANGUAGES C CXX)`)

```cmake
...
set(CMAKE_C_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} /bigobj")
set(CMAKE_C_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} /bigobj")
set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} /bigobj")
set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} /bigobj")
...
project(Open3D VERSION ${OPEN3D_VERSION} LANGUAGES C CXX)
...
```

#### Exp 3. This doesn't work:

```cmake
string(APPEND CMAKE_C_FLAGS " /bigobj")
string(APPEND CMAKE_CXX_FLAGS " /bigobj")
```

#### Exp 4. This doesn't work:

```cmake
set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /bigobj")
set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /bigobj")
```

#### Exp 5. About `if(MSVC)`

`if(MSVC)` does not work before the `project()` statement. So it is removed.

## Part II. Solving the `defaultlib 'LIBCMT' conflicts`

Turns out that we need to [enable a policy](https://stackoverflow.com/a/56490614) for it to work.

```cmake
if(POLICY CMP0091)
    cmake_policy(SET CMP0091 NEW)
endif()
```

yxlao:
I also updated `docs/_static/C++/CMakeLists.txt` to link against the `Open3D::Open3D` target directly: https://github.com/intel-isl/Open3D/commit/35def567abdc66a2688f52b379c9d7fa9f20c2ac#diff-f850cb778290514b5aaed6801c2379ca, please feel free to make any changes to the file.

yxlao:
Looks like most of the tests are passing, except the project linking to open3d.
Can modify to something like: https://github.com/intel-isl/Open3D/blob/yixing/cmake-refactor/docs/_static/C++/CMakeLists.txt so that the dll can be copied and we can directly link to `Open3D::Open3D`?

roehling:
Yes, I meant to do that already, but I forgot and then figured I'll let the tests run anyway.

roehling:
I just ran into an interesting CMake quirk: functions and macros will record the current policies at definition time, so subsequent changes won't have any effect when the functions or macros are called.

yxlao:
Thanks @roehling ! The new fixes work great for me. There're (hopefully the last) two issues remaining.

#### Issue 1: CMake drops duplicated `-gencode`
In my machine where it has two very different CUDA cards, I ran into the issue described in https://stackoverflow.com/a/54505975/1255535, where CMake drops the second `-gencode` which causes `nvcc` to fail.

This [solution](https://stackoverflow.com/a/57193326/1255535) works for me (requires CMake 3.12+): 
```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index f8ba3911..b2089ffd 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -319,7 +319,7 @@ function(open3d_set_global_properties target)
         target_compile_definitions(${target} PRIVATE UNIX)
         target_compile_options(${target} PRIVATE "-Wno-deprecated-declarations" "$<$<COMPILE_LANGUAGE:CXX>:-Wno-unused-result>")
     endif()
-    target_compile_options(${target} PRIVATE "$<$<COMPILE_LANGUAGE:CUDA>:--expt-extended-lambda>" "$<$<COMPILE_LANGUAGE:CUDA>:${CUDA_GENCODES}>")
+    target_compile_options(${target} PRIVATE "$<$<COMPILE_LANGUAGE:CUDA>:--expt-extended-lambda>" "$<$<COMPILE_LANGUAGE:CUDA>:SHELL:${CUDA_GENCODES}>")
 endfunction()
 
 macro(add_source_group module_name)
```

#### Issue 2: Change CMake to 3.12+?
If I switch back to CMake 3.10, which is the latest cmake offered in ubuntu 18.04's default repo, I got the following error:

```shell
CMake Error at CMakeLists.txt:233 (target_link_libraries):
  Object library target "Camera" may not link to anything.
Call Stack (most recent call first):
  src/Open3D/Camera/CMakeLists.txt:8 (open3d_link_3rdparty_libraries)
```

Looks like this is related to https://gitlab.kitware.com/cmake/cmake/-/issues/18010. 

#### Summary

Since Issue 1 and Issue 2 both require 3.12+, to minimize other changes, I guess it is worthwhile to upgrade to 
```cmake
cmake_minimum_required(VERSION 3.12.0)
```
in both the root `CMakeLists.txt` and `docs/_static/C++/CMakeLists.txt`. What do you think?

roehling:
I don't see any simple way to work around these issues, so I concur that we should make CMake 3.12 the minimum required version. And Kitware provides [backported CMake versions](https://apt.kitware.com/) for the Ubuntu LTS releases.

yxlao:
Sure, sounds good, could you make the above changes to conclude the PR?

roehling:
Already working on it :)

yxlao:
Looks good. Thank you @roehling! This is an amazing contribution. Merging.

roehling:
Great! Thank you @yxlao for your help to get everything work smoothly!

PR #1896, opened 2020-05-25
GUI: added threading and loading dialog to app

prewettg:
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1896)

### DIFF
--- a/src/Open3D/Visualization/Visualizer/GuiVisualizer.cpp
+++ b/src/Open3D/Visualization/Visualizer/GuiVisualizer.cpp
@@ -1764,68 +1764,91 @@ bool GuiVisualizer::SetIBL(const char *path) {
     return result;
 }
 
-bool GuiVisualizer::LoadGeometry(const std::string &path) {
-    auto geometry = std::shared_ptr<geometry::Geometry3D>();
+void GuiVisualizer::LoadGeometry(const std::string &path) {
+    // Post the dialog creation back to the main thread so that the OS window
+    // gets created if it is not already.
+    gui::Application::GetInstance().PostToMainThread([this, path]() {
+        auto &theme = GetTheme();
+        auto loading_dlg = std::make_shared<gui::Dialog>("Loading");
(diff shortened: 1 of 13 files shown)


src/Open3D/Visualization/Visualizer/GuiVisualizer.cpp
```diff
@@ -1764,68 +1765,107 @@ bool GuiVisualizer::SetIBL(const char *path) {
...
+        loading_dlg->AddChild(vert);
+        ShowDialog(loading_dlg);
+    });
 
-    auto geometry_type = io::ReadFileGeometryType(path);
+    gui::Application::GetInstance().RunInThread([this, path, progressbar]() {
+        auto UpdateProgress = [this, progressbar](float value) {
+            gui::Application::GetInstance().PostToMainThread(
+                    this,
+                    [progressbar, value]() { progressbar->SetValue(value); });
```
pvarvak, line 1788: This will work, but some loaders will call update_progress on every point they load, potentially millions of times before GUI gets to process these events (and while GUI processes those millions of events, it will hold posted_lock_ which will block UpdateProgress which will block the loader, making it wait for GUI to finish before it can continue loading).  Ideally ProgressBar would only post only 1 PostToMainThread call only to flag PostRedraw on the window, and until it gets redrawn it would only update value (atomically).

src/Open3D/Visualization/Visualizer/GuiVisualizer.h
```diff
@@ -62,7 +62,7 @@ class GuiVisualizer : public gui::Window {
 
     bool SetIBL(const char* path);
 
-    bool LoadGeometry(const std::string& path);
+    void LoadGeometry(const std::string& path);
```
pvarvak, line 65: /// This will load Geometry asynchronously, it will return immediately
(maybe we should call it QueueLoadGeometry?)

PR #1901, opened 2020-05-27
Fix bug reading multi-threading config

scimad:
The condition `if config["python_multi_threading"]:` is true even when `config["python_multi_threading"]` is set `False` or `false` or anything.

This PR fixes that bug present in the reconstruction example.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1901)


scimad:
@griegler I think this can be reviewed and merged.

PR #1923, opened 2020-06-03
cmake: pickup python from PYTHON_EXECUTABLE or from PATH

pvarvak:
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1923)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,6 +51,20 @@ if (PREFER_OSX_HOMEBREW)
     set(CMAKE_FIND_APPBUNDLE LAST)
 endif()
 
+# setup PYTHON_EXECUTABLE if not set
+find_program(PYTHON_IN_PATH "python")
+if (NOT PYTHON_EXECUTABLE)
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -51,6 +51,20 @@ if (PREFER_OSX_HOMEBREW)
     set(CMAKE_FIND_APPBUNDLE LAST)
 endif()
 
+# setup PYTHON_EXECUTABLE if not set
+find_program(PYTHON_IN_PATH "python")
```
yxlao, line 55: 1. Can we integrate and unify the following python path related code?
    - https://github.com/intel-isl/Open3D/blob/master/3rdparty/CMake/FindPythonExecutable.cmake
    - https://github.com/intel-isl/Open3D/blob/master/src/Python/CMakeLists.txt#L6
2. In `.travis.yml` and `util/scripts/run-travis.sh`, could we remove the `-DPYTHON_EXECUTABLE` flags and test that the builds still work?

PR #1960, opened 2020-06-15
avoid deduplication of the -gencode option

benjaminum:
This PR fixes a problem if cuda code is compiled for multiple archs

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1960)


benjaminum:
master seems broken if the cuda module is enabled and there is more than one gencode option.
Please check if master fails as the CI does not capture this problem.

PR #1981, opened 2020-06-20
do not link main library when building the tf ops lib

benjaminum:
Linking the main library will cause side effects when using graph mode and cuda.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1981)

### DIFF
--- a/cpp/open3d/ml/Helper.h
+++ b/cpp/open3d/ml/Helper.h
@@ -1,0 +1,67 @@
+// ----------------------------------------------------------------------------
+// -                        Open3D: www.open3d.org                            -
+// ----------------------------------------------------------------------------
+// The MIT License (MIT)
+//
+// Copyright (c) 2020 www.open3d.org
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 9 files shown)


benjaminum:
> LGTM fro this PR. Please feel free to merge.
> 
> Question: is it true that in the future, TF ops lib will not link against the main Open3D lib? Linking with the main Open3D lib could bring useful features like neighbor search, hashing, etc.
> 
> > Reviewed 9 of 9 files at r1.
> > _[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1981#-:-MAO-cKxCcsZTBwgAnVT:b1q0vgb)_ status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @pvarvak)

I see this PR as a quick bugfix to make the ml-module in master work. We have to investigate what exactly is causing this problem. Unfortunately the CI cannot capture this at the moment.

If it is a problem with the compile flags we have to check if we can change them.
If it is a problem with a specific module we could link the objects that we need and exclude the one that causes the problem.

PR #2007, opened 2020-06-25
Add CPACK-based building of Debian package

errissa:
This PR creates a 'package' target to make that builds a Debian package. No special configuration is required, the target is created as part of the cmake configuration process. However, the package target has to be build explicitly with `make package` if desired. It's not built by default.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2007)

### DIFF
--- a/cpp/apps/CMakeLists.txt
+++ b/cpp/apps/CMakeLists.txt
@@ -113,7 +113,7 @@ macro(APP SRC_DIR APP_NAME TARGET_NAME)
             set(CPACK_PACKAGE_VERSION_MINOR "10")
             set(CPACK_PACKAGE_VERSION_PATCH "1")
             set(CPACK_DEBIAN_PACKAGE_ARCHITECTURE "amd64")
-            set(CPACK_DEBIAN_PACKAGE_DEPENDS "libc, libc++1, libgomp1, libpng16-16, libglfw3")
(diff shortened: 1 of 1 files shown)


errissa:
@yxlao Fixed. The libc dependency is not necessary. I had meant to remove it because it seems to cause problems on some systems.

yxlao:
The `libc` fix works for me.

1. Now the debian package is called "open3d-viewer-0.10.1-Linux-open3d_component.deb", can we remove the "-open3d_component"?
2. Instead of `make package`, let's use some other name, to indicate that we're making a "debian" package for the "viewer app"? Something like `make viewer-debian-package`. 
3. Currently, I am getting the following warnings at uninstallation:
    ```
    dpkg: warning: while removing open3d-viewer-open3d_component, directory '/home/yixing/open3d_install' not empty so not removed
    dpkg: warning: while removing open3d-viewer-open3d_component, directory '/home/yixing/.local/share/mime/packages' not empty so not removed
    dpkg: warning: while removing open3d-viewer-open3d_component, directory '/home/yixing/.local/share/icons/hicolor' not empty so not removed
    dpkg: warning: while removing open3d-viewer-open3d_component, directory '/home/yixing/.local/share/applications' not empty so not removed
    ```
    Is this expected? Will the warning goes away if `CMAKE_INSTALL_PREFIX` is set properly? (my `CMAKE_INSTALL_PREFIX` is `/home/yixing/open3d_install`) .

germanros1987:
@errissa we need you to take action on this.

PR #2013, opened 2020-06-26
Removed visualization::gui::util namespace

prewettg:
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2013)


prewettg:
The reason it's in a separate file is that it needs imgui.h for ImVec4. Since the function returns a value, we can't forward-declare the class, since we won't know how big the return value is. We could put it in Color.h, but then basically anything using `gui` is going to pull in imgui.h, which doesn't seem right, since imgui.h is an implementation detail.

PR #2020, opened 2020-06-29
Linear algebra module

theNded:
Placeholder for the algebra module, ideally should only be merged after the followings are done
## Functionalities
- [x] Matrix multiplication (mm)
- [ ] Batch mm
- [ ] Inverse
- [x] Solver with LU decomposition
- [ ] SVD
- [ ] ~~Conv2D (or should we put it in another module? it involves many other libraries like oneDNN and cuDNN)~~
## Interface
- [ ] OO functions in Tensor class
- [ ] Python binding
## Config
- [ ] Automatic download / detect of MKL, MAGMA (shipped with CUDA and cublas)
- [ ] Verified linking against `${MAGMA_LIBRARIES} ${MKL_LIBRARIES} ${CUDA_CUBLAS_LIBRARIES}`
## Tests
- [ ] UnitTest

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2020)


benjaminum:
@theNded Do we use 2d convs somewhere in some algorithms? If not I think we can leave it out at the moment.

theNded:
> @theNded Do we use 2d convs somewhere in some algorithms? If not I think we can leave it out at the moment.

We need several filters for image processing
https://github.com/intel-isl/Open3D/blob/master/cpp/open3d/geometry/Image.h#L72
we can either write a handcrafted backend or a general Conv2D for later usage... maybe we need to postpone this for a image-processing module (or something similar...)

yxlao:
Just a comment: the todo list is great -- it is also okay to leave some tasks to future PRs. We just need to make sure what we have for this PR is well tested and documented, and mark the `TODO`s in the code.

PR #2045, opened 2020-07-06
Use GitHub actions again for macOS CI

yxlao:
- Enable GitHub actions for macOS CI, disable Travis.
- Skip `brew update` for macOS on CI.
- Use `ccache` for Ubuntu and macOS builds
  - Prints `ccache` status before and each run. It shows that `cache` is indeed activated. e.g. https://github.com/intel-isl/Open3D/runs/843543441#step:6:24 and https://github.com/intel-isl/Open3D/runs/843543441#step:6:6213

This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2045)

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -28,6 +28,19 @@ jobs:
       uses: actions/checkout@v2
       with:
         submodules: true
+    - name: Setup cache
+      uses: actions/cache@v2
+      with:
+        # Ref: https://github.com/apache/incubator-mxnet/pull/18459/files
+        path: ~/.ccache
+        # We include the commit sha in the cache key, as new cache entries are
+        # only created if there is no existing entry for the key yet.
+        key: ${{ runner.os }}-ccache-${{ github.sha }}
+        # Restore any ccache cache entry, if none for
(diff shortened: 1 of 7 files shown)


benjaminum:
@yxlao macos on travis is very slow. Would it make sense to switch to the github macos runner?

yxlao:
@benjaminum I also prefer the GitHub macOS actions, however, we have had stability issues with randomly failing tests. That's why we only used GitHub macOS actions for a while and switched back to Travis.

Currently, Travis is breaking on master; additionally, it sometimes hits the 50 min total time limit and the test will fail. So sure, we can give Github actions another try.

benjaminum:
> @benjaminum I also prefer the GitHub macOS actions, however, we have had stability issues with randomly failing tests. That's why we only used GitHub macOS actions for a while and switched back to Travis.
> 
> Currently, Travis is breaking on master; additionally, it sometimes hits the 50 min total time limit and the test will fail. So sure, we can give Github actions another try.

I did not know about the stability issues but it is a good idea to try github again because we will likely run into the travis timelimit more often.

yxlao:
@barnjamin ready for review now

PR #2098, opened 2020-07-21
Bugfixes found while integrating with Python

prewettg:
These bugs are not necessarily Python-specific, just that I found them while creating Python examples.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2098)

### DIFF
--- a/cpp/open3d/visualization/gui/Window.cpp
+++ b/cpp/open3d/visualization/gui/Window.cpp
@@ -408,6 +408,14 @@ void Window::SetTitle(const char* title) {
 //       after MakeDrawContextCurrent() has been called), otherwise
 //       ImGUI won't be able to access the font.
 Size Window::CalcPreferredSize() {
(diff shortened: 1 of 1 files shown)


prewettg:
___
*[cpp/open3d/visualization/gui/Window.cpp, line 415 at r1](https://reviewable.io/reviews/intel-isl/open3d/2098#-MCqXoZa7bxp-dQMEPvR:-MD0o2gqA7scANfkLv5K:b-o6sy38) ([raw file](https://github.com/intel-isl/open3d/blob/7dcf37cb82d86ae55d95555600b6decb810ccee6/cpp/open3d/visualization/gui/Window.cpp#L415)):*

Since a size of (0,0) is likely to cause a crash down the line, should the method check the size at the end before returning? Is there any chance that even with children the bbox ends up (0,0,0,0)?
</blockquote></details>

Scaling should never be 0, so size should never be zero. In fact, if scaling is zero, that's a major problem.

PR #2102, opened 2020-07-23
New Scene Interface [not ready for master]

errissa:
The Open3DViewer does not fully function with the new interfaces. However, the Scene and FilamentScene class interfaces are now stable and can be reviewed.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2102)


cpp/open3d/visualization/rendering/Scene.h
```diff
@@ -41,83 +42,110 @@ class AxisAlignedBoundingBox;
...
-    // Passing empty id disables indirect lightning
-    virtual void SetIndirectLight(const IndirectLightHandle& id) = 0;
+    // Camera
+    virtual void AddCamera(const std::string& camera_name,
+                           std::shared_ptr<Camera> cam) = 0;
+    virtual void RemoveCamera(const std::string& camera_name) = 0;
+    virtual void SetActiveCamera(const std::string& camera_name) = 0;
+
+    // Scene geometry
+    virtual bool AddGeometry(const std::string& object_name,
```
benjaminum, line 77: I would prefer SetGeometry() which would replace an existing geometry with the same name. Of course remove+add is the same but set seems more user friendly.

cpp/open3d/visualization/rendering/Scene.h
```diff
@@ -41,83 +42,110 @@ class AxisAlignedBoundingBox;
...
+                             const geometry::Geometry3D& geometry,
+                             const Material& material) = 0;
+    virtual bool AddGeometry(const std::string& object_name,
```
benjaminum, line 80: Do we want to support geometry3d and Model? Having just one function should be enough and simplify the interface.

cpp/open3d/visualization/rendering/Scene.h
```diff
@@ -41,83 +42,110 @@ class AxisAlignedBoundingBox;
...
+    virtual void ShowGeometry(const std::string& object_name, bool show) = 0;
+    virtual void OverrideMaterial(const std::string& object_name,
+                                  const Material& material) = 0;
+    virtual void QueryGeometry(std::vector<std::string>& geometry) = 0;
+    virtual void SetGeometryTransform(const std::string& object_name,
+                                      const Transform& transform) = 0;
+    virtual Transform GetGeometryTransform(const std::string& object_name) = 0;
+    virtual geometry::AxisAlignedBoundingBox GetGeometryBoundingBox(
+            const std::string& object_name) = 0;
+    virtual void OverrideMaterialAll(const Material& material,
```
benjaminum, line 92: Will the material be shared or will there be copies for each geometry?
As I understand a material defines how we visualize the geometry, right?
Assuming we have a sequence of objects (obj1, ..., obj1000) can we change the material for all of these quickly?

cpp/open3d/visualization/rendering/Scene.h
```diff
@@ -41,83 +42,110 @@ class AxisAlignedBoundingBox;
...
+                              const Eigen::Vector3f& position,
+                              const Eigen::Vector3f& direction,
+                              float intensity,
+                              float falloff,
+                              float inner_cone_angle,
+                              float outer_cone_angle,
+                              bool cast_shadows) = 0;
+    virtual Light& GetLight(const std::string& light_name) = 0;
+    virtual void RemoveLight(const std::string& light_name) = 0;
+    virtual void UpdateLight(const std::string& light_name,
```
benjaminum, line 113: How is this used? Like this?
```
auto light = GetLight("bla");
light.some_property = 5;
UpdateLight("bla", light); // will this be necessary for filament to get the changes?
```

cpp/open3d/visualization/rendering/Scene.h
```diff
@@ -41,83 +42,110 @@ class AxisAlignedBoundingBox;
 namespace visualization {
 namespace rendering {
 
+class Renderer;
 class View;
+class Model;
+class Material;
```
benjaminum, line 48: Material seems to hold most of the properties for visualization. 
Would it make sense to wrap it in a more generic class ```Properties```, which can store more properties like ```visible, rendermode(wire,solid), use_lod, material, ...```? We could also have ```Properties``` as a base class and subclass for Geometry, Lights, and Cameras.
With this we can then write imgui code to let the user change all of these properties.

PR #2146, opened 2020-08-07
clean up util folder

yxlao:
- Remove unused scripts
- Remove docker (outdated, not maintained)
- Standardize naming

The only remaining files are as follows:
```
Open3D/util
├── check_cpp_style.cmake
├── check_style.py
├── install_deps_macos.sh
├── install_deps_ubuntu.sh
├── make_documentation.sh
└── run_ci.sh
```

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2146)


pwais:
@yxlao The online documentation still references headless docker support: http://www.open3d.org/docs/release/tutorial/docker/docker-gui.html    Did the feature never work?  Dockerized headless rendering is incredibly useful.

yxlao:
@pwais dockerized headless CPU rendering should be feasible. Our documentation build runs headless rendering https://github.com/intel-isl/Open3D/blob/master/.github/workflows/documentation.yml. It's possible to cook your own Dockerfile by following the scripts in documentation.yml.

pwais:
@yxlao yup I saw the CI stuff 👍  but I'd rather not have to try to cook up my own Dockerfile when there appears to be one that works (tho it got removed :( ) in this commit...   Is the functionality referenced in the current online docs going to be restored, or are the docs out of date and going to be deleted? 

Lastly, I'm not sure if you're familiar with this part of Open3D, but is there any way to configure Open3d do render-to-image _without_ using Xvfb/mesa?  I.e. is there any way to get Open3D to use "CPU-only functionality" to do rendering (sort of like putting pytorch or tensorflow into CPU mode)?   If not, no big deal, but I wanted to ask given that it would be another way to do "headless" or CPU-only rendering.  Also, this appears to be an Intel project, so I was hoping there might have been some feature here.  I've looked through many of the examples and haven't found anything though.

pwais:
oh huh, is the python package `open3d` on pypi right now actually built with headless rendering support? 
https://github.com/intel-isl/Open3D/blob/8a7a77c10d4f0339fb5c52b913235f6e5a23b937/cpp/pybind/_build_config.py.in#L10

PR #2185, opened 2020-08-17
Fix assimp build issue on some Ubuntu systems

errissa:
Ensures that assimp doesn't try to link against system minizip library

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2185)


yxlao:
closing since #2188 is merged

PR #2186, opened 2020-08-17
Various bug fixes resulting from vis refactor

errissa:
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2186)


prewettg:
The rotate-model functionality (the "model" button) still seems to not work, was that something this PR was intended to fix?

prewettg:
___
*[cpp/open3d/visualization/visualizer/GuiSettingsModel.cpp, line 193 at r2](https://reviewable.io/reviews/intel-isl/open3d/2186#-MF2pc-Z74w3PTwtQk8J:-MF3ywqK9lkM0kMJHIY_:blatoh5) ([raw file](https://github.com/intel-isl/open3d/blob/0fe941dcce07f87bd1a8777054efa3e679329aff/cpp/open3d/visualization/visualizer/GuiSettingsModel.cpp#L193)):*

Yes, there are 4 settings + all of the lighting parameters that don't use `true`. I changed the `NotifyChanged(material_type_changed)` to `NotifyChanged(material_change)` because with the vis-refactor/Scene changes, updating material properties is potentially an expensive operation. (For example, this can be observed when changing the sun direction with a complex model loaded.) So, the semantics are the same as before insofar as there is only a single special case about which the callback might want to know, it's just the special case covers a few more property changes.

I agree that if we need to track changes with more granularity that `NotifyChanged` needs to be modified. I was thinking a uint32_t with bit flags. But for now, I believe `NotifyChanged` sufficiently captures the expensive special case which is changing material properties without being confusing.
</blockquote></details>

I guess since I'm the one that was lazy in the first place, I can't complain...

PR #2212, opened 2020-08-21
Support cuda memory cacher

theNded:
Add cuda cacher for more efficient memory recycle.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2212)


benjaminum (changes requested):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2212)* status: 4 of 5 files reviewed, 2 unresolved discussions (waiting on @benjaminum, @theNded, and @yxlao)

___
*[cpp/open3d/core/MemoryManagerCUDA.cu, line 103 at r2](https://reviewable.io/reviews/intel-isl/open3d/2212#-MFMjj7z9_y-kf-9ee5Y:-MFWt6-zAaGaOklk6BZO:b98fkqf) ([raw file](https://github.com/intel-isl/open3d/blob/89d2f758f7f197772fa7c87325a882a76bed55b3/cpp/open3d/core/MemoryManagerCUDA.cu#L103)):*

We will only use cudaMalloc internally for now, that is also true for pytorch memory cacher. We can keep this in mind and support it on demand.
</blockquote></details>

cudaMalloc is fine. Just wanted to know if we can support different pointer alignments. From the code it seems we cannot use the cacher if we mix different alignments, right?

germanros1987:


```diff

Clones added
============
- cpp/open3d/core/MemoryManager.h  3
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/intel-isl/Open3D/pullRequest?prid=6049312&bid=19817444)

benjaminum (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2212)* status: 2 of 9 files reviewed, 5 unresolved discussions (waiting on @benjaminum, @theNded, and @yxlao)

___
*[cpp/open3d/core/MemoryManagerCUDA.cu, line 103 at r2](https://reviewable.io/reviews/intel-isl/open3d/2212#-MFMjj7z9_y-kf-9ee5Y:-MG2p0UL9avW5i67yVc-:b-5qytbu) ([raw file](https://github.com/intel-isl/open3d/blob/89d2f758f7f197772fa7c87325a882a76bed55b3/cpp/open3d/core/MemoryManagerCUDA.cu#L103)):*

I think yes for now -- here this alignment is used for splitting and merging memory blocks manually. For other alignments it will be error-prone to split and organize. One way is to cache these memory blocks but add a flag indicating that they are not splitable. Do you think it is urgent to support? Otherwise I can add this on demand afterwards.
</blockquote></details>

It is not urgent we can add this later when we need it.

yxlao:
merging, as all review comments are addressed, the alignment feature can be added later

PR #2262, opened 2020-08-31
Skip GCE CI if secrets are not available (forks)

ssheorey:
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2262)


yxlao:
thank you @ssheorey , that was a quick fix!

PR #2282, opened 2020-09-03
Expose method to update vertex attributes

errissa:
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2282)


prewettg (changes requested):
> Reviewed 3 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2282)* status: all files reviewed, 4 unresolved discussions (waiting on @errissa)

___
*[cpp/open3d/visualization/rendering/Scene.h, line 60 at r1](https://reviewable.io/reviews/intel-isl/open3d/2282#-MGKvmle7C1ZJeHc5N7T:-MGKvmle7C1ZJeHc5N7U:b-t6eaj7) ([raw file](https://github.com/intel-isl/open3d/blob/d3662ae99a0c2898017cef7304ec49e0bb9f982e/cpp/open3d/visualization/rendering/Scene.h#L60)):*
> ```objc
> class Scene {
> public:
>     const uint32_t kUpdatePointsFlag = 0x1;
> ```

Optional: I think `flag = (1 << n);` (where n is the flag number) makes it a little easier to correctly add flags in the future, and clearer that flags are being defined. Less mental hex math.
___
*[cpp/open3d/visualization/rendering/Scene.h, line 64 at r1](https://reviewable.io/reviews/intel-isl/open3d/2282#-MGKvbNGDRDI0whb5yKu:-MGKvbNGDRDI0whb5yKv:bhia9le) ([raw file](https://github.com/intel-isl/open3d/blob/d3662ae99a0c2898017cef7304ec49e0bb9f982e/cpp/open3d/visualization/rendering/Scene.h#L64)):*
> ```objc
>     const uint32_t kUpdateColorsFlag = 0x4;
>     const uint32_t kUpdateUv0Flag = 0x8;
>     const uint32_t kUpdateUv1Flag = 0x10;
> ```

kUpdateUv1Flag is defined, but not handled. Seems like we should either implement it or remove it.
___
*[cpp/open3d/visualization/rendering/filament/FilamentScene.cpp, line 396 at r1](https://reviewable.io/reviews/intel-isl/open3d/2282#-MGKuVfe5nnld3DzuAkn:-MGKuVfe5nnld3DzuAko:b-8tkt9j) ([raw file](https://github.com/intel-isl/open3d/blob/d3662ae99a0c2898017cef7304ec49e0bb9f982e/cpp/open3d/visualization/rendering/filament/FilamentScene.cpp#L396)):*
> ```cpp
> }
> 
> void FilamentScene::UpdateGeometry(const std::string& object_name,
> ```

Is it feasible to reuse code with regular PointCloud, and with PointCloudBuffers / TriangleMeshBuffers?
___
*[cpp/open3d/visualization/rendering/filament/FilamentScene.cpp, line 416 at r1](https://reviewable.io/reviews/intel-isl/open3d/2282#-MGKumDRAVBCj-d-1GX2:-MGKumDRAVBCj-d-1GX3:b-vslyb8) ([raw file](https://github.com/intel-isl/open3d/blob/d3662ae99a0c2898017cef7304ec49e0bb9f982e/cpp/open3d/visualization/rendering/filament/FilamentScene.cpp#L416)):*
> ```cpp
>         }
> 
>         if (update_flags & kUpdateColorsFlag && point_cloud.HasPointColors()) {
> ```

If kUpdatePointsFlag is specified, it seems like we should update all the other buffers, too, because if the number of points changes, then they'll certainly be wrong. Although, maybe we can get rid of kUpdatePointsFlag, since if we're going to read all the other fields, the caller might as well just RemoveGeometry() and AddGeometry(). Or maybe kUpdatePointsFlag can be limited to the case where the number of points don't change, enforced by whatever passes for an assert?

errissa (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2282)* status: 1 of 3 files reviewed, 4 unresolved discussions (waiting on @prewettg)

___
*[cpp/open3d/visualization/rendering/Scene.h, line 60 at r1](https://reviewable.io/reviews/intel-isl/open3d/2282#-MGKvmle7C1ZJeHc5N7T:-MGL1mga3s8ZXNxCzPyU:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/d3662ae99a0c2898017cef7304ec49e0bb9f982e/cpp/open3d/visualization/rendering/Scene.h#L60)):*

Optional: I think `flag = (1 << n);` (where n is the flag number) makes it a little easier to correctly add flags in the future, and clearer that flags are being defined. Less mental hex math.
</blockquote></details>

Done.
___
*[cpp/open3d/visualization/rendering/Scene.h, line 64 at r1](https://reviewable.io/reviews/intel-isl/open3d/2282#-MGKvbNGDRDI0whb5yKu:-MGL1djO7LdU9vwO-alg:bvpeo2n) ([raw file](https://github.com/intel-isl/open3d/blob/d3662ae99a0c2898017cef7304ec49e0bb9f982e/cpp/open3d/visualization/rendering/Scene.h#L64)):*

kUpdateUv1Flag is defined, but not handled. Seems like we should either implement it or remove it.
</blockquote></details>

Removed. Done.
___
*[cpp/open3d/visualization/rendering/filament/FilamentScene.cpp, line 396 at r1](https://reviewable.io/reviews/intel-isl/open3d/2282#-MGKuVfe5nnld3DzuAkn:-MGLXkaXBAiL8YQZPvfN:b-tvxtuk) ([raw file](https://github.com/intel-isl/open3d/blob/d3662ae99a0c2898017cef7304ec49e0bb9f982e/cpp/open3d/visualization/rendering/filament/FilamentScene.cpp#L396)):*

Is it feasible to reuse code with regular PointCloud, and with PointCloudBuffers / TriangleMeshBuffers?
</blockquote></details>

Unfortunately not. The vertex attribute arrays are separate buffers for the TPointClouds but the PointClouds and TriangleMesh vertex attribute arrays are interleaved so they will need separate implementations of UpdateGeometry.
___
*[cpp/open3d/visualization/rendering/filament/FilamentScene.cpp, line 416 at r1](https://reviewable.io/reviews/intel-isl/open3d/2282#-MGKumDRAVBCj-d-1GX2:-MGLY4pj2PHN_AHbdNGq:b-cea9hy) ([raw file](https://github.com/intel-isl/open3d/blob/d3662ae99a0c2898017cef7304ec49e0bb9f982e/cpp/open3d/visualization/rendering/filament/FilamentScene.cpp#L416)):*

If kUpdatePointsFlag is specified, it seems like we should update all the other buffers, too, because if the number of points changes, then they'll certainly be wrong. Although, maybe we can get rid of kUpdatePointsFlag, since if we're going to read all the other fields, the caller might as well just RemoveGeometry() and AddGeometry(). Or maybe kUpdatePointsFlag can be limited to the case where the number of points don't change, enforced by whatever passes for an assert?
</blockquote></details>

I had meant to have a check for the number of points changing. The Filament vertex buffers can't be resized so UpdateGeometry should only be called when the data but not the number of points changes. So, in that context it's conceivable someone could update the points without updating the other attributes.

I added a check (and associated note) for point size changing.

errissa (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2282)* status: 1 of 3 files reviewed, 4 unresolved discussions (waiting on @prewettg)

___
*[cpp/open3d/visualization/rendering/filament/FilamentScene.cpp, line 416 at r1](https://reviewable.io/reviews/intel-isl/open3d/2282#-MGKumDRAVBCj-d-1GX2:-MGL_fW8D0Kpxgtpaari:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/d3662ae99a0c2898017cef7304ec49e0bb9f982e/cpp/open3d/visualization/rendering/filament/FilamentScene.cpp#L416)):*

I had meant to have a check for the number of points changing. The Filament vertex buffers can't be resized so UpdateGeometry should only be called when the data but not the number of points changes. So, in that context it's conceivable someone could update the points without updating the other attributes.

I added a check (and associated note) for point size changing.
</blockquote></details>

Done.

PR #2324, opened 2020-09-12
Fix for "symbol already registered"

ssheorey:
Loading  cuda pybind, checking cuda availability and then loading cpu pybind causes "symbol already registered"  error. This re-appeared in specific combinations after the CUDA_VISIBLE_DEVICES fix. This uses ctypes to prevent symbol registration.
New `open3d_core_cuda_device_count() `C interface can be called through ctypes. Directly calling C++ interface is not portable due to name mangling.

Tested with these combinations:
GPU + CUDA : ISL workstation
GPU + CUDA + CUDA_VISIBLE_DEVICES=none : ISL workstation
No GPU + no CUDA : docker
GPU + no CUDA : docker
No GPU + CUDA : docker

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2324)

### DIFF
--- a/cpp/open3d/core/CUDAUtils.cpp
+++ b/cpp/open3d/core/CUDAUtils.cpp
@@ -42,7 +42,7 @@ int DeviceCount() {
     try {
         std::shared_ptr<CUDAState> cuda_state = CUDAState::GetInstance();
         return cuda_state->GetNumDevices();
-    } catch (const std::runtime_error & e) {    // GetInstance can throw
+    } catch (const std::runtime_error& e) {  // GetInstance can throw
         return 0;
     }
 #else
--- a/cpp/open3d/core/CUDAUtils.cpp
+++ b/cpp/open3d/core/CUDAUtils.cpp
@@ -39,8 +39,12 @@ namespace cuda {
 
 int DeviceCount() {
 #ifdef BUILD_CUDA_MODULE
-    std::shared_ptr<CUDAState> cuda_state = CUDAState::GetInstance();
-    return cuda_state->GetNumDevices();
+    try {
+        std::shared_ptr<CUDAState> cuda_state = CUDAState::GetInstance();
+        return cuda_state->GetNumDevices();
+    } catch (const std::runtime_error & e) {    // GetInstance can throw
+        return 0;
+    }
 #else
     return 0;
 #endif
@@ -66,3 +70,8 @@ void ReleaseCache() {
 }  // namespace cuda
 }  // namespace core
 }  // namespace open3d
+
+// C interface to provide un-mangled function to Python ctypes
+extern "C" int open3d_core_cuda_device_count() {
+    return open3d::core::cuda::DeviceCount();
+}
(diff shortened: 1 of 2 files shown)


yxlao (commented):
> Reviewed 1 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2324)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @ssheorey and @yxlao)

___
*[python/open3d/__init__.py, line 63 at r2](https://reviewable.io/reviews/intel-isl/open3d/2324#-MHA9-IrCYsO0J323xvu:-MHA9-IrCYsO0J323xvv:b-pfal1i) ([raw file](https://github.com/intel-isl/open3d/blob/14bf2af5e95bbad9e598418893ac6ce938267984/python/open3d/__init__.py#L63)):*
> ```Python
>         _pybind_cuda = _CDLL(
>             next((_Path(__file__).parent / 'cuda').glob('pybind*')))
>         if _pybind_cuda.open3d_core_cuda_device_count() > 0:
> ```

This is interesting, so no python-binding code is needed, and we can directly call extern C functions?

ssheorey (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2324)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @yxlao)

___
*[python/open3d/__init__.py, line 63 at r2](https://reviewable.io/reviews/intel-isl/open3d/2324#-MHA9-IrCYsO0J323xvu:-MHCMQtx5lAb-cn097nq:bdhh2t7) ([raw file](https://github.com/intel-isl/open3d/blob/14bf2af5e95bbad9e598418893ac6ce938267984/python/open3d/__init__.py#L63)):*

This is interesting, so no python-binding code is needed, and we can directly call extern C functions?
</blockquote></details>

Both C and C++ functions can be called directly from Python with `ctypes`. C++ function calling is not portable, since we have to use the mangled names. Also datatypes will be messy, so pybind11 is definitely much better in general. `ctypes` lets us avoid registering the C dll symbols as Python module symbols.

yxlao (approved):
> Reviewed 1 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2324)* status: :shipit: complete! all files reviewed, all discussions resolved

ray-lee-94:
File "C:\Users\dell\Open3D\build\lib\python_package\open3d\__init__.py", line 72, in <module>
    if _pybind_cuda.open3d_core_cuda_device_count() > 0:
  File "E:\.conda\envs\nav\lib\ctypes\__init__.py", line 387, in __getattr__
    func = self.__getitem__(name)
  File "E:\.conda\envs\nav\lib\ctypes\__init__.py", line 392, in __getitem__
    func = self._FuncPtr((name_or_ordinal, self))
AttributeError: function 'open3d_core_cuda_device_count' not found

PR #2340, opened 2020-09-15
Fix NanoFlannIndex base class memory leak

yxlao:
`NanoFlannIndexHolderBase` does not have a virtual destructor, causing the `unique_ptr` to its children classes not being able to destruct property.

Quick test case that causes memory leak:
```cpp
void RunTest() {
    core::Tensor ref = core::Tensor::Ones({100, 3}, core::Dtype::Float64);
    core::nns::NanoFlannIndex index(ref);
}

TEST(NearestNeighborSearch, KnnSearch) {
    while (true) {
        RunTest();
    }
}
```

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2340)

### DIFF
--- a/cpp/open3d/core/nns/NanoFlannIndex.h
+++ b/cpp/open3d/core/nns/NanoFlannIndex.h
@@ -53,7 +53,9 @@ namespace nns {
 enum Metric { L1, L2, Linf };
 
 /// Base struct for Index holder
-struct NanoFlannIndexHolderBase {};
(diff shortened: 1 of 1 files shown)


benjaminum (approved):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2340)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @germanros1987)

PR #2392, opened 2020-09-29
set cmake minimum to 3.15

benjaminum:
The generator expressions CXX_COMPILER_ID in cpp/CMakeLists.txt:13 do not support multiple ids before 3.15.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2392)


yxlao (approved):

yxlao:
LGTM as ubuntu 20.04 has 3.15+, and for ubuntu 18.04 we need to install a new cmake version manually anyway.

PR #2398, opened 2020-09-30
Remove unused variable from SolveJacobianSystemAndObtainExtrinsicMatrix

nachovizzo:
While looking into the Eigen utilities  I found this small unused variable. Is not a big thing, but better to remove it :)

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2398)


yxlao (approved):

yxlao:
thanks

Issue #2395, opened 2020-09-29
Is `#ifdef _OPENMP` really a must?


nachovizzo:
This is a question I've been asking myself for a while. But as far as I understood, one does not really need this:
```cpp
#ifdef _OPENMP
#pragma omp parallel
    {
#endif
```

AFIK, the idea is that if the compiler doesn't know how to eat the `#pragma` then it will ignore. Or is it because this generates a warning?

So if I have this naive example
```cpp
#include <iostream>
#include <vector>

int main() {
  const int N = 1 << 30;
  std::vector<int> x;
  x.reserve(N);
#pragma omp parallel for default(none) private(x)
  for (int i = 0; i < N; ++i) {
    x.emplace_back(i);
  }

  return 0;
}
```

With this CMakeLists.txt
```cmake
cmake_minimum_required(VERSION 3.10)
project(pragma_omp)

set(CMAKE_BUILD_TYPE Release)
set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
set(CMAKE_CXX_STANDARD 14)
set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra")

option(ENABLE_OPENMP "Enable OpenMP" OFF)
if(ENABLE_OPENMP)
  find_package(OpenMP REQUIRED)
  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${OpenMP_CXX_FLAGS}")
endif()

add_executable(main main.cpp)
```

The only thing that the guard is preventing is this warning
```sh
/main.cpp:8: warning: ignoring #pragma omp parallel [-Wunknown-pragmas]
    8 | #pragma omp parallel for default(none) private(x)
      | 
```

Is this the reason behind the guard?

I'm just curious. Thanks!

griegler:
Yes, the guard is only because of the warning. In the current build setup we regard warnings as errors.

nachovizzo:
What if we add `-Wno-unknown-pragmas` to the C++ flags This disables the pragma warnings and will make the code more readable. I'm totally willing to make a PR for this since takes no brainpower.

Let me know what you guys think.

griegler:
Looping in @yxlao

nachovizzo:
@griegler @yxlao I just created a draft PR in case you want to have a look at how these changes would look

yxlao:
@nachovizzo sounds good

PR #2408, opened 2020-10-01
Nacho/remove openmp guards

nachovizzo:
This PR is a draft to support the discussion open in #2395. This also solves 2 small builds "bugs" within the `C++` examples.

The idea overall is to improve code readability while also failing the build if `OpenMP` is supported and the propper `CXX_FLAGS` are not being set for the target, for example:
```cmake
EXAMPLE_CPP(TrimMeshBasedOnPointCloud ${CMAKE_PROJECT_NAME} ${OPENMP_TARGET})
EXAMPLE_CPP(EvaluateFeatureMatch      ${CMAKE_PROJECT_NAME} ${OPENMP_TARGET} ${FLANN_TARGET})
```

## Readability Improve

Taking as an example `open3d::utility::eigen::ComputeJTJandJTr` https://github.com/intel-isl/Open3D/blob/eaef3bd91489c6351a046571c00fbf0e0aa86d45/cpp/open3d/utility/Eigen.cpp#L165

The code changes from:
### Before
```cpp
template <typename MatType, typename VecType>
std::tuple<MatType, VecType, double> ComputeJTJandJTr(
        std::function<void(int, VecType &, double &)> f,
        int iteration_num,
        bool verbose /*=true*/) {
    MatType JTJ;
    VecType JTr;
    double r2_sum = 0.0;
    JTJ.setZero();
    JTr.setZero();
#ifdef _OPENMP
#pragma omp parallel
    {
#endif
        MatType JTJ_private;
        VecType JTr_private;
        double r2_sum_private = 0.0;
        JTJ_private.setZero();
        JTr_private.setZero();
        VecType J_r;
        double r;
#ifdef _OPENMP
#pragma omp for nowait
#endif
        for (int i = 0; i < iteration_num; i++) {
            f(i, J_r, r);
            JTJ_private.noalias() += J_r * J_r.transpose();
            JTr_private.noalias() += J_r * r;
            r2_sum_private += r * r;
        }
#ifdef _OPENMP
#pragma omp critical
        {
#endif
            JTJ += JTJ_private;
            JTr += JTr_private;
            r2_sum += r2_sum_private;
#ifdef _OPENMP
        }
    }
#endif
    if (verbose) {
        LogDebug("Residual : {:.2e} (# of elements : {:d})",
                 r2_sum / (double)iteration_num, iteration_num);
    }
    return std::make_tuple(std::move(JTJ), std::move(JTr), r2_sum);
}
```

to
### After
```cpp
template <typename MatType, typename VecType>
std::tuple<MatType, VecType, double> ComputeJTJandJTr(
        std::function<void(int, VecType &, double &)> f,
        int iteration_num,
        bool verbose /*=true*/) {
    MatType JTJ;
    VecType JTr;
    double r2_sum = 0.0;
    JTJ.setZero();
    JTr.setZero();
#pragma omp parallel
    {
        MatType JTJ_private;
        VecType JTr_private;
        double r2_sum_private = 0.0;
        JTJ_private.setZero();
        JTr_private.setZero();
        VecType J_r;
        double r;
#pragma omp for nowait
        for (int i = 0; i < iteration_num; i++) {
            f(i, J_r, r);
            JTJ_private.noalias() += J_r * J_r.transpose();
            JTr_private.noalias() += J_r * r;
            r2_sum_private += r * r;
        }
#pragma omp critical
        {
            JTJ += JTJ_private;
            JTr += JTr_private;
            r2_sum += r2_sum_private;
        }
    }
    if (verbose) {
        LogDebug("Residual : {:.2e} (# of elements : {:d})",
                 r2_sum / (double)iteration_num, iteration_num);
    }
    return std::make_tuple(std::move(JTJ), std::move(JTr), r2_sum);
}
```

## Potential Drawbacks

As far as know there are no drawbacks. The only problem would be if at some point there is an `unknown-pragma` which is not about `OpenMP` and `OpenMP` support is disabled, then the warning will be ignored. A fix for this would be not to threat that particular warning as an error, but this I don't know how to do honestly.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2408)


nachovizzo:
@griegler @yxlao 13fbdb2 also shows that the `ifdef` approach was having some issues. The build was failing because the compiler didn't recognize the `#pragma omp`, therefore, it wasn't really configured to use `OpenMP` at all. The same as the 2 examples mentioned above

yxlao (changes requested):
Looks good in general. It does provide much cleaner code with the cost of ignoring pragma warnings. The tradeoff seems to be worth it.

cpp/tests/utility/Eigen.cpp
```diff
@@ -184,9 +184,7 @@ TEST(Eigen, ComputeJTJandJTr) {
     ref_JTr << 0.477778, -0.262092, -0.162745, -0.545752, -0.643791, -0.883007;
 
     auto testFunction = [&](int i, Eigen::Vector6d &J_r, double &r) {
-#ifdef _OPENMP
 #pragma omp critical
```
yxlao, line 187: Seems like this `#pragma omp critical` can be removed. Other lambda functions used in `utility::ComputeJTJandJTr` do not have this pragma surrounded. It works for me on Ubuntu. Let's try to see if it can pass the CI.

cpp/tests/utility/Eigen.cpp
```diff
@@ -227,9 +225,7 @@ TEST(Eigen, ComputeJTJandJTr_vector) {
             [&](int i,
                 std::vector<Eigen::Vector6d, utility::Vector6d_allocator> &J_r,
                 std::vector<double> &r) {
-#ifdef _OPENMP
 #pragma omp critical
```
yxlao, line 228: Same, remove this `#pragma omp critical` .

yxlao (changes requested):

cpp/tests/CMakeLists.txt
```diff
@@ -44,6 +44,7 @@ find_package(Threads)
 target_link_libraries(tests PRIVATE Threads::Threads ${CMAKE_PROJECT_NAME} ${JSONCPP_TARGET} ${GOOGLETEST_TARGET})
 open3d_show_and_abort_on_warning(tests)
 open3d_set_global_properties(tests)
+open3d_link_3rdparty_libraries(tests)
```
yxlao, line 47: I would remove this and change the tests. Typically, we only link `tests` to Open3D but not 3rd-party libraries, such that the `tests` works like a user's program linking to Open3D.

cpp/tests/utility/Eigen.cpp
nachovizzo: Agree, will do

cpp/tests/utility/Eigen.cpp
nachovizzo: Done

cpp/tests/CMakeLists.txt
nachovizzo: Got it, without the `#pragmas` we don't even need to pass the `-fopenmp` flag to the compiler, so we don't need this anymore

yxlao (approved):

PR #2416, opened 2020-10-02
Force redraw when menu visibility changes

errissa:
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2416)


prewettg (changes requested):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2416)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on @errissa and @prewettg)

___
*[cpp/open3d/visualization/visualizer/GuiVisualizer.cpp, line 969 at r3](https://reviewable.io/reviews/intel-isl/open3d/2416#-MIeRfvhCpuUDDMEVTMc:-MIeRfvhCpuUDDMEVTMd:b-gj7xev) ([raw file](https://github.com/intel-isl/open3d/blob/448ca90dd79076096e67fec0e87c6da070c42134/cpp/open3d/visualization/visualizer/GuiVisualizer.cpp#L969)):*
> ```cpp
> }
> 
> void GuiVisualizer::OnSubMenuVisibilityChanged() {
> ```

This forces everyone who has a menu to know to inherit from Window and override this function with this code. For instance, vis-gui.py fails, because the pybinds don't know they need to override this. (But the pybind code also doesn't know the reference to the widget, which would mean we'd have to push it up to the python code.) I offer the following as possible solutions:
1) In DrawOnce, instead of calling this, iterate over all the children, attempt to dynamic cast to SceneWidget and call ForceRedraw() if it succeeds.
2) Like (1), but create some sort of interface or add ForceRedraw to Widget, whose default implementation does nothing, but SceneWidget's stays the same to avoid the dynamic cast.
3) In DrawOnce, instead of calling this, add a redraw_3d param/flag to DrawContext which DrawOnce will set. DrawOnce can then return REDRAW, and when SceneWidget::Draw() checks the flag it know to force a redraw.

(Also, you can set GUI_USE_NATIVE_MENUS to 0 in Menu.h to force ImGUI menus on macOS [and Windows, in the hopeful future], so technically the #ifdef should be `#if !defined(__APPLE__) || !GUI_USE_NATIVE_MENUS` to avoid future developers, i.e. probably me, from wondering what is going on when they try to use this for debugging :)

errissa (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2416)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @prewettg)

___
*[cpp/open3d/visualization/visualizer/GuiVisualizer.cpp, line 969 at r3](https://reviewable.io/reviews/intel-isl/open3d/2416#-MIeRfvhCpuUDDMEVTMc:-MIifVZX9KtwGgKsLnVc:b-p060c) ([raw file](https://github.com/intel-isl/open3d/blob/448ca90dd79076096e67fec0e87c6da070c42134/cpp/open3d/visualization/visualizer/GuiVisualizer.cpp#L969)):*

This forces everyone who has a menu to know to inherit from Window and override this function with this code. For instance, vis-gui.py fails, because the pybinds don't know they need to override this. (But the pybind code also doesn't know the reference to the widget, which would mean we'd have to push it up to the python code.) I offer the following as possible solutions:
1) In DrawOnce, instead of calling this, iterate over all the children, attempt to dynamic cast to SceneWidget and call ForceRedraw() if it succeeds.
2) Like (1), but create some sort of interface or add ForceRedraw to Widget, whose default implementation does nothing, but SceneWidget's stays the same to avoid the dynamic cast.
3) In DrawOnce, instead of calling this, add a redraw_3d param/flag to DrawContext which DrawOnce will set. DrawOnce can then return REDRAW, and when SceneWidget::Draw() checks the flag it know to force a redraw.

(Also, you can set GUI_USE_NATIVE_MENUS to 0 in Menu.h to force ImGUI menus on macOS [and Windows, in the hopeful future], so technically the #ifdef should be `#if !defined(__APPLE__) || !GUI_USE_NATIVE_MENUS` to avoid future developers, i.e. probably me, from wondering what is going on when they try to use this for debugging :)
</blockquote></details>

I implemented your suggestion 1. It seemed the most direct solution to this specific issue.

errissa (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2416)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @prewettg)

___
*[cpp/open3d/visualization/visualizer/GuiVisualizer.cpp, line 969 at r3](https://reviewable.io/reviews/intel-isl/open3d/2416#-MIeRfvhCpuUDDMEVTMc:-MIig-yVDlGmWjB3NM2T:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/448ca90dd79076096e67fec0e87c6da070c42134/cpp/open3d/visualization/visualizer/GuiVisualizer.cpp#L969)):*

I implemented your suggestion 1. It seemed the most direct solution to this specific issue.
</blockquote></details>

Done.

PR #2420, opened 2020-10-02
Package libc++.so and libc++abi.so in Ubuntu wheel

ssheorey:
Preloaded with ctypes.CDLL to isolate from user's system
Any other required DLLs can be packaged with the same mechanism
Ensure libc++ and libc++abi are from the same source.
Preload libc++ and libc++abi only if not found in the system
Always download filament binary for Linux wheel

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2420)


cpp/pybind/CMakeLists.txt
```diff
@@ -102,6 +102,13 @@ set_target_properties(${PACKAGE_NAME} PROPERTIES
                       FOLDER "Python"
                       LIBRARY_OUTPUT_DIRECTORY "${PYTHON_COMPILED_MODULE_DIR}"
                       ARCHIVE_OUTPUT_DIRECTORY "${PYTHON_COMPILED_MODULE_DIR}")
+# Include additional libraries that may be absent from the user system
+# eg: libc++.so and libc++abi.so (needed by filament) for Linux.
+# libc++.so is a linker script including libc++.so.1 and libc++abi.so, so append 1 to libc++.so
```
yxlao, line 107: In my case, they are symbolic links:

```
lrwxrwxrwx 1 root root   14 Nov 22  2018 libc++abi.so -> libc++abi.so.1
lrwxrwxrwx 1 root root   16 Nov 22  2018 libc++abi.so.1 -> libc++abi.so.1.0
-rw-r--r-- 1 root root 203K Nov 22  2018 libc++abi.so.1.0

-rw-r--r-- 1 root root   28 Nov 22  2018 libc++.so
lrwxrwxrwx 1 root root   13 Nov 22  2018 libc++.so.1 -> libc++.so.1.0
-rw-r--r-- 1 root root 757K Nov 22  2018 libc++.so.1.0
```

So in this system, we need ` libc++abi.so.1.0` and `libc++.so.1.0`. Would it be possible for CMake to resolve the symbolic links, instead of manually append the `.` at the end?

cpp/pybind/CMakeLists.txt
```diff
@@ -102,6 +102,13 @@ set_target_properties(${PACKAGE_NAME} PROPERTIES
...
+set(PYTHON_EXTRA_LIBRARIES "")
+if (BUILD_GUI AND $<PLATFORM_ID:Linux>)
```
yxlao, line 109: Seems like `$<PLATFORM_ID:Linux>` might not be working on some system. I tried:
```
if (BUILD_GUI AND $<PLATFORM_ID:Linux>)
    message(FATAL_ERROR "here")
    list(APPEND PYTHON_EXTRA_LIBRARIES ${CPP_LIBRARY}.1 ${CPPABI_LIBRARY})
endif()
```
and cmake didn't crash. Also I couldn't find the copied `libc+.so` files in the wheel package. Could we take a second look at this?

We can add a `if (LINUX)` in the top-level CMake. Similar to https://github.com/google/filament/blob/3d121eb873fd138de10a458a65973d1b4af8a84c/CMakeLists.txt#L82, we can have:

```
if (UNIX AND NOT APPLE)
    set(LINUX TRUE)
endif()
```

We have a bunch of `if (UNIX AND NOT APPLE)` in the code base, so this can simplify them as well.

cpp/pybind/make_python_package.cmake
```diff
@@ -29,6 +29,14 @@ foreach(COMPILED_MODULE_PATH ${COMPILED_MODULE_PATH_LIST})
         endif()
     endforeach()
 endforeach()
+# Include additional libraries that may be absent from the user system
+# eg: libc++.so and libc++abi.so (needed by filament)
+foreach(PYTHON_EXTRA_LIB ${PYTHON_EXTRA_LIBRARIES})
+  get_filename_component(PYTHON_EXTRA_LIB_REAL ${PYTHON_EXTRA_LIB} REALPATH)
```
yxlao, line 35: 4-space indent

yxlao (changes requested):

3rdparty/find_dependencies.cmake
```diff
@@ -901,11 +901,13 @@ if(BUILD_GUI)
     set(FILAMENT_MATC "${FILAMENT_ROOT}/bin/matc")
     target_link_libraries(3rdparty_filament INTERFACE Threads::Threads ${CMAKE_DL_LIBS})
     if(UNIX AND NOT APPLE)
-        find_library(CPP_LIBRARY c++)
-        if(CPP_LIBRARY)
-            # Ensure that libstdc++ gets linked first
-            target_link_libraries(3rdparty_filament INTERFACE -lstdc++ ${CPP_LIBRARY})
-        endif()
+        find_library(CPPABI_LIBRARY c++abi PATH_SUFFIXES llvm-7/lib llvm-8/lib
```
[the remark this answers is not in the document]
ssheorey: > In theory, we should be using the same version `libc++` and `libc++abi` as the clang compiler. E.g. `clang-7` should go with `llvm-7/lib` and etc.

That will need some careful design to account for all cases (user installed clang, multiple versions installed, downloaded filament, etc.). I think relying on libc++ ABI version compatibility should be OK.

> At least, we can search the llvm libs in the reverse order `llvm-11/lib` -> `llvm-10/lib` ->, etc., which is the same order as we search the clang compiler. (Also add `clang-11` if it is appropriate.)

Added, and reversed order.

> Another issue is that, in my machine (Ubuntu 18.4 + `clang-7` + `libc++-7-dev` + `libc++abi-7-dev`), CMake actually doesn't respect the `PATH_SUFFIXES`, it gets:
> 
> ```shell
> -- CPPABI_LIBRARY: /usr/lib/x86_64-linux-gnu/libc++abi.so                                                                                                
> -- CPP_LIBDIR: /usr/lib/x86_64-linux-gnu                                                                                                                 
> -- CPP_LIBRARY: /usr/lib/x86_64-linux-gnu/libc++.so
> ```
> 
> which, are not under `llvm-7/lib`, but they are symbolic links:
> 
> ```shell
> libc++abi.so -> ../llvm-7/lib/libc++abi.so
> libc++.so -> ../llvm-7/lib/libc++.so
> ```
> 
> so it worked fine but it would be good to force CMake to respect the `PATH_SUFFIXES`.

This ensures that the system default `libc++` is used for the build (so that its the same that will be linked at runtime) There may be multiple `libc++` versions installed on the system; the selected version (eg Ubuntu version specific default, or user chosen with `update-alternatives` is symlinked to /usr/lib/x86_64-linux-gnu/). `PATH_SUFFIXES` are searched if a system default `libc++` is not set.

cpp/pybind/CMakeLists.txt
ssheorey: `libc++.so` is not a symlink, unlike `libc++abi.so`, so it will not resolve to `libc++.so.1`. It's a script that needs to be parsed by the linker to find the actual library. `cmake` does not have functionality to do this.
The current method fixes this in a hacky way, but it will work as long as `libc++.so` has the same major ABI version. This is not needed for `libc++abi.so`, which is a symlink.

cpp/pybind/CMakeLists.txt
ssheorey: Thanks for catching that. I assumed the generator expression would be evaluated in the `if()`, but they are only evaluated during generation, not during configuration.
I found a cleaner method to check the platform, with the `CMAKE_SYSTEM_NAME` variables
https://gitlab.kitware.com/cmake/community/-/wikis/doc/tutorials/How-To-Write-Platform-Checks

yxlao (approved):

PR #2427, opened 2020-10-05
Feature/dec preserve edges

### LINKED ISSUE #2418: SimplifyQuadricDecimation boundary weight
When decimating preserving boundary vertices can help produce more visually pleasing results. Currently, no additional consideration is give to vertices that are on the boundary of a mesh.

I would like to add a `boundary_weight` argument to `simplify_quadric_decimation`. The value would default to 1. Values greater then 1 would weight the error for vertices on the boundary of a mesh.

OpenMesh has a more generalized method of mixing and matching error metrics. I think such an approach is interesting, but it seems more complex both to implement and to use. With that said, an alternate API could use something like a callback that passes vertex information and the current cost and expects a new cost to be returned. This could enable use cases beyond boundary preservation.

**Additional context**

I started looking into implementing this, but I'm a bit confused about the different between `TriangleMesh` class and the `HalfEdgeTriangleMesh` class. Is it possible to determine boundaries for `TriangleMesh`, if so, are there existing methods that can be used?

```cpp
    std::vector<Quadric> Qs(vertices_.size());
    for (size_t vidx = 0; vidx < vertices_.size(); ++vidx) {
        std::vector<int> edges = mesh->BoundaryHalfEdgesFromVertex(vidx);
        bool isBoundary = !edges.empty();
        for (int tidx : vert_to_triangles[vidx]) {
            Qs[vidx] += Quadric(triangle_planes[tidx], triangle_areas[tidx]);
            if (isBoundary) {
                Qs[vidx] *= 1e6;
            }
        }
    }
```

FreakTheMighty:
This pull request adds a `boundary_weight` parameter that can be used to preserve edge boundaries when performing `SimplifyQuadricDecimation`.

Closes #2418 

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2427)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
 * Corrected documentation of PointCloud.h
 * Added ISS Keypoint Detector
 * Added an RPC interface for external visualizers running in a separate process
-* Added `maximum_error` parameter to `simplify_quadric_decimation`
+* Added `maximum_error` and `boundary_weight` parameter to `simplify_quadric_decimation`
 
 ## 0.9.0
 
--- a/cpp/open3d/geometry/TriangleMesh.h
+++ b/cpp/open3d/geometry/TriangleMesh.h
@@ -417,8 +417,9 @@ public:
     /// will be reached.
     /// \param maximum_error defines the maximum error where a vertex is allowed
     /// to be merged
+    /// \param boundary_weight a weight applied to edge vertices used to preserve boundaries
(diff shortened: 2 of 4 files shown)


griegler (approved):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2427)* status: :shipit: complete! all files reviewed, all discussions resolved

FreakTheMighty:
Does the mac os build need to be retried?

PR #2445, opened 2020-10-09
add docs for arm build

yxlao:
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2445)


docs/arm.rst
```diff
@@ -0,0 +1,115 @@
...
+    ./bin/tests --gtest_filter="-*Reduce*Sum*"
+
+    # Install Open3D python package (optional)
+    make install-pip-package -j$(nproc)
+    python -c "import open3d; print(open3d)"
+
+    # Run Open3D GUI (available on when -DBUILD_GUI=ON)
+    ./bin/Open3D/Open3D
+
+
```
ssheorey, line 102: Is there an `install` step too?
```suggestion
    # Install Open3D binaries
    make install
```

docs/arm.rst
```diff
@@ -0,0 +1,115 @@
...
+Nvidia Jetson
+-------------
+
+Nvidia Jetson computers with 64-bit processor and OS are supported. You can
+compile Open3D with ``-DBUILD_CUDA_MODULE=ON`` and ``-DBUILD_GUI=ON`` and
+the Open3D GUI app should be functional.
+
```
ssheorey, line 109: Should we remind users here that we support CUDA 10.1 only?

ssheorey (approved):
Have we tried the build on actual hardware (Jetson, Rasperry Pi) or only on simulator?

docs/arm.rst
yxlao: In fact, if we don't consider the ML ops, the repo can work with a much wider range of CUDA versions. In my case, it is `10.2`.

docs/arm.rst
yxlao: yes, will add and mark as optional

yxlao:
> Have we tried the build on actual hardware (Jetson, Rasperry Pi) or only on simulator?

Yes, tested on Jetson AGX (GUI=ON and CUDA=ON) and Rasperry Pi 4 (GUI=OFF due to OpenGL ES).

docs/arm.rst
ssheorey: In that case, let's add:
```suggestion
We support CUDA v10.1, but other versions should work as well.
```

yxlao: done, added

PR #2474, opened 2020-10-14
consistent sidebar lower case

yxlao:
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2474)


ssheorey (approved):
Good idea to make case consistent.

PR #2485, opened 2020-10-15
Rework Python API for creating windows to avoid crash on exit in some simple situations.

prewettg:
In particular:
```
def main():
    gui.Application.instance.inititalize()
    w = Window("Crash")
    gui.Application.instance.add_window(w)
    gui.Application.instance.run()  # C++ cleans up filament, but Python still has a shared_ptr ref in w
    # crash here, Python, frees its pointer, Window's refcount goes to zero, but the filament object it references is destroyed
if __name__ == "__main__":
   main()
```
This also prevents Python users from forgetting to call add_window() which would lead to problems with layouts. And it should have the side-effect of preventing crash-on-exit if the Python script exits prematurely after the window is created due to an exception.

The API has been changed. The above code should be:
```
def main():
    gui.Application.instance.inititalize()
    w = gui.Application.instance.create_window("No crash")
    gui.Application.instance.run()
if __name__ == "__main__":
   main()
```

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2485)

### DIFF
--- a/cpp/pybind/visualization/gui/gui.cpp
+++ b/cpp/pybind/visualization/gui/gui.cpp
@@ -253,7 +253,11 @@ void pybind_gui_classes(py::module &m) {
                         return w.get();
                     },
                     "title"_a = std::string(), "width"_a = -1, "height"_a = -1,
-                    "x"_a = -1, "y"_a = -1, "flags"_a = 0)
+                    "x"_a = -1, "y"_a = -1, "flags"_a = 0,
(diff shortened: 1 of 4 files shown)


errissa (approved):
> Reviewed 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2485)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2521, opened 2020-10-20
Nacho/generalized icp

nachovizzo:
# Generalized ICP

This PR brings the power of gicp to Open3D.

Tasks to do before merge:

- [ ] Add a full tutorial on the pipelines/registration section
- [ ] Polish C++ implementation and make sure there is documentation

To discuss with Open3D team(@theNded):

- [ ] How to properly integrate covariance information on the `PointCloud` class. The current implementation adds some overhead (just because of the implementation per se) because it re-computes the covariance matrices for each iteration.
- [ ] How to expose the covariance matrix to provide different modeling (for example, instead of doing `plane-to-plane` ICP something else). I have some drafts on this branch, but I couldn't find an elegant solution for this.\

### KITTI benchmark, Sequence 00

- AVG runtime multicore: 14fps
- Other open-source implementations like [fast_gicp](https://github.com/SMRT-AIST/fast_gicp) achieve no more than 7/8 fps on the same computer.

## Sequence 00
experiment

|      Method      | Implementation  | avg. translational error | avg. rotational error |
| :--------------: | :-------------: | :----------------------: | :-------------------: |
|       gicp       |     Open3D      |         **1.18**         |       **0.568**       |
|       gicp       | KAIST/fast-gicp |           1.26           |         0.595         |
| PointToPoint ICP |     Open3D      |           3.18           |         1.24          |
| PointToPlane ICP |     Open3D      |           5.74           |         1.74          |

I also did some other experiments on different sequences that I will omit to
save space on this PR description.

More to come...

## Footnote

I consider this PR still a draft, although it's a good idea to start discussing the design of the implementation before I finish the implementation. I'm completely open to suggestions, so, I'd kindly ask to careful review this PR :) Thanks!

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2521)


germanros1987:


```diff

Issues
======
- Added 1
           

Clones added
============
- cpp/open3d/pipelines/registration/GeneralizedICP.h  2
- cpp/open3d/pipelines/registration/TransformationEstimation.h  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/intel-isl/Open3D/pullRequest?prid=6373770&bid=20680675)

germanros1987:


```diff

Issues
======
- Added 1
           

Clones added
============
- cpp/open3d/pipelines/registration/TransformationEstimation.h  2
- cpp/open3d/pipelines/registration/GeneralizedICP.h  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/intel-isl/Open3D/pullRequest?prid=6373770&bid=20680675)

germanros1987:


```diff

Issues
======
- Added 1
           

Clones added
============
- cpp/open3d/pipelines/registration/TransformationEstimation.h  2
- cpp/open3d/pipelines/registration/GeneralizedICP.h  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/intel-isl/Open3D/pullRequest?prid=6373770&bid=20680675)

nachovizzo:
@theNded any update on this PR?

theNded:
> @theNded any update on this PR?

I remember it was in draft mode. Is it ready for review now?

nachovizzo:
> > @theNded any update on this PR?
> 
> I remember it was in draft mode. Is it ready for review now?

Sorry, I was expecting to discuss this: 
image

But I wasn't really clear.
There are at least 4 things I don't like about the implementation:

1. I'm computing the covariance matrix for each point, at every single iteration of the ICP. This is done like this because there is no other way to do it with the current state of the API(which I'm not in favor to add tons of extra complexity) https://github.com/intel-isl/Open3D/blob/c8a636aceafab30fe8204cad7ab052cb625fad15/cpp/open3d/pipelines/registration/GeneralizedICP.cpp#L129
2. So far is the original implementation of the GICP, but it could be nice if we can expose the covariance matrix computation for `TransformationEstimationforGeneralizedICP`, this way user can potentially implement other registration pipelines on top of the GICP framework(I think this would be great for prototyping and for Open3D in general).
3. It would also be great to expose the `TransformationEstimationforGeneralizedICP` to the outside world. I'm using this on a research project(https://github.com/nachovizzo/Open3D/blob/d7a8c1cbeba715968ceefdabb7d94d6865219576/cpp/open3d/pipelines/registration/GeneralizedICP.cpp#L94) the drawback is that for that particular case, I had to re-compute the covariance matrix now for both sources and target PointClouds.
4. With the current `API` there is no way to access the current transformation of the iteration from the `ComputeTransformation` methods. Again, I couldn't foresee any clean way of doing this(and also without modifying tons of code). The problem with this is within this line https://github.com/intel-isl/Open3D/blob/c8a636aceafab30fe8204cad7ab052cb625fad15/cpp/open3d/pipelines/registration/GeneralizedICP.cpp#L141. The math is wrong and it should be `const Eigen::Matrix3d M = Ct + T.inverse() * Cs * T;` Where `T` is the current estimation of the ICP. I tried to change this, but we should modify the `TransformationEstimation` abstract class, and add defaults everywhere which I don't like https://github.com/intel-isl/Open3D/blob/c8a636aceafab30fe8204cad7ab052cb625fad15/cpp/open3d/pipelines/registration/TransformationEstimation.h#L85

One we found a fix for those 4 problems I guess I will find interior peace :lotus_position_man: And I also think it would be great if `Open3D` can have the GICP the way I dream, this could really a door for experimentation with different registration scenarios with not too much pain(If we expose the covariance estimation, then anyone who knows how to use python can try out different scenarios, everything will be running in C++ under the hood, so still fast and reliable)

Thanks and sorry for the verbosity!

theNded:
@nachovizzo Sorry for the delayed response and thanks again for the great contribution! Currently my working bandwidth is almost full, so I can only give you some minor feedbacks here. Please let me know if there are other concerns.

1. For the point cloud with a covariance matrix, current inheritance makes sense as a transition solution. We are shifting to a dictionary-based point cloud, meaning we can add properties with (str, tensor) pairs soon. With the new design it should be no problem adding a property.

2. I think it will be fine to add a method `ComputeCovariance` in the base class `TransformationEstimation`, and default its behavior to be void. In the comment you may explain that it's reserved for ICP variants that require point-wise covariance, such as G-ICP.

3. You can expose it in `TransformationEstimation.h`. ColoredICP did not expose its interface for some legacy reasons I think.

4. You may add a member variable `transformation_` in `TransformationEstimationforGeneralizedICP` to store transformation, like the color term introduced in Colored ICP: https://github.com/intel-isl/Open3D/blob/c8a636aceafab30fe8204cad7ab052cb625fad15/cpp/open3d/pipelines/registration/ColoredICP.h#L74. You can access this variable in `ComputeRMSE`, and update it properly at the end of `ComputeTransformation`. I think it should work, but you may take extra care and check if it goes wrong between iterations.

nachovizzo:
@theNded sorry for the delay. I did some experiments with the code and I've found that the solution for all the problems I had is to add
```cpp
    /// Covariance Matrix for each point
    std::vector<Eigen::Matrix3d> covariances_;
```
In the PointCloud class. this way:

- the covariance can be computed from outside the registration pipeline in any way a user wants:
```python
pcd.covariances = crazy_covariance_estimation_method(pcd.points, pcd.normals)
```
- If the user doesn't provide any covariance information(or normals) then, the GICP method will compute the same as the one described in the original paper(fallback behavior)

- the information doesn't get lost when we copy the source cloud:  https://github.com/intel-isl/Open3D/blob/75dd3674511afc05dd752d9b12cb474042781bde/cpp/open3d/pipelines/registration/Registration.cpp#L162

- there is no need to maintain the increments of the transformation during each iteration since we can update the covariance information inside the `PointCloud` class, this will be done when we hit this line: https://github.com/intel-isl/Open3D/blob/75dd3674511afc05dd752d9b12cb474042781bde/cpp/open3d/pipelines/registration/Registration.cpp#L175

The Generalized ICP would be also much cleaner and 100% extensible, with no extra effort.

That being said, this is out of the scope of this PR. I will open **another** PR, where we can discuss the covariance data member, and once that's ready. I will re open this PR (just adding the GICP implementation).

That should bring all the power and flexibility of having a Generalized registration framework to Open3D

PR #2524, opened 2020-10-21
Changed o3d.pipeline to o3d and fixed few conditions

claudiup423:
The PR fixes the reconstruction_system example by change to the lastest convention.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2524)


yxlao:
It's likely that you're in an older version. As of `v0.11.0` it is `open3d.pipelines.*`, see: http://www.open3d.org/docs/release/python_api/open3d.pipelines.integration.html .

PR #2530, opened 2020-10-22
Fix Open3DScene with no background color set not clearing on draw

prewettg:
The bug is that unless you explicitly set the background color, the background is black and the view does not clear on draw, leading to streaking from all the previous draws.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2530)


errissa (approved):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2530)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2563, opened 2020-10-30
fix reading float from PCD, instead of float_t

### LINKED ISSUE #310: `.travis.yml` is not well formatted
I have been struggling to make a python package install in the Travis CI for some hours now!

The command `pyenv global 3.6` utilized to setup the Travis CI Linux-image seems to encapsulate Python inside a virtual environment. The problem is that CMake script cannot figure out the correct site directory to install `py3d.cpython-36m-x86_64-linux-gnu.so` python library.

However, I believe there is something wrong with the `.travis.yml` setting, and we should not use `pyenv global 3.6`. Another interesting thing is that `Ruby` image is loaded by this setting file, instead of Python, or GCC or similar things...

### How to reproduce
Both Python scripts utilized inside the CMake:
```python
import site

print([path for path in site.getsitepackages() if path.endswith('packages')][-1])

```
and
```python
from distutils import sysconfig

print(sysconfig.get_python_lib())
```
returns the virtualenv site-packages that is not gloabl site-package, and any installation to is does not get detected.

#### Actual output
/opt/pyenv/versions/3.6/lib/python3.6/site-packages

#### Expected output
/opt/python/3.6.3/lib/python3.6/site-packages

I even tried `echo "/opt/pyenv/versions/3.6/lib/python3.6/site-packages/" > /opt/python/3.6.3/lib/python3.6/site-packages/open3d.pth;` without any success.

#### Travis CI output
```
...
-- Installing: /opt/pyenv/versions/3.6/lib/python3.6/site-packages/py3d.cpython-36m-x86_64-linux-gnu.so
...
$ sudo python file_io.py
  File "file_io.py", line 5, in <module>
    from py3d import *
ImportError: No module named py3d
The command "sudo python file_io.py" exited with 1.
Done. Your build exited with 1.
```

### LINKED ISSUE #2556: Reading PCD - incorrect use of float_t
When reading a binary "float" value of size 4 from a PCD file, the code should use the type "float" instead of "float_t", which could as well be double or long double, depending on the architecture and compiler settings. In that case, the code would read too much data and return incorrect values.

https://github.com/intel-isl/Open3D/blob/c3cb0c421d23a93af0f232d007f07171c894d294/cpp/open3d/io/file_format/FilePCD.cpp#L277
https://github.com/intel-isl/Open3D/blob/c3cb0c421d23a93af0f232d007f07171c894d294/cpp/open3d/io/file_format/FilePCD.cpp#L278

Note that I found that issue during an unrelated code inspection. I have not reproduced any run-time problems that could result.

mhillenibm:
The type float_t can be float, double, or long double, depending on the
architecture and compiler settings. Thus, when reading a "float" value
of size 4 from a PCD file, the code should not use float_t but float.

Fixes #2556 
(edit: fix referenced issue)

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2563)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 * Ubuntu 20.04 (Focal) support.
 * Added Line3D/Ray3D/Segment3D classes with plane, point, closest-distance, and AABB tests
 * Add Open3D-ML to Open3D wheel
+* Fix a bug in PointCloud file format, use `float` instead of `float_t`
 
 ## 0.9.0
 
--- a/cpp/open3d/io/file_format/FilePCD.cpp
+++ b/cpp/open3d/io/file_format/FilePCD.cpp
@@ -274,7 +274,7 @@ double UnpackBinaryPCDElement(const char *data_ptr,
         }
     } else if (type == 'F') {
         if (size == 4) {
-            std::float_t data;
+            float data;
             memcpy(&data, data_ptr, sizeof(data));
             return (double)data;
         } else {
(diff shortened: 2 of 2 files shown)


mhillenibm:
> Thanks for submitting this pull request! The maintainers of this repository would appreciate if you could update the **CHANGELOG.md** based on your changes.

Updated patch, add entry to CHANGELOG.md

yxlao (approved):

yxlao:
thanks @mhillenibm

PR #2585, opened 2020-11-06
Add draw() as replacement for draw_geometries()

### LINKED ISSUE #2549: Install Open3D v0.11.1 on Jetson Nano gives "GLFW Error: GLX: No GLXFBConfigs returned"
_____

**Describe the bug**
I built Open3D from source and installed on a jetson nano. The build and installation seem to work well, running regular Open3D commands is working, but when i try to use the visualizer i get the following error

[Open3D ERROR] GLFW Error: GLX: No GLXFBConfigs returned

I tried running both via ssh and with monitor directly connected to the jetson, both options give the same result.

I also tried the Open3DViewer app, invoked from Open3D/build with  
./bin/Open3D/Open3D
the GUI starts and I can load an STL file, the graphics look great and everything works.

So graphics are working nicely in GUI app, but o3d.visualization.draw_geometries([..]) fails when invoked from a python script.
I´m guessing this indicates i missed some flag or dependency during the build.

any pointers on how to adress this issue would be greatly appreciated.

**To Reproduce**
1. download nano 4.4.1 image at https://developer.nvidia.com/embedded/jetpack
2. flash image to sd card and boot up in jetson nano -> this gives the nvidia jetpack 4.4.1 version of ubuntu 18
3.
```
sudo apt remove --purge cmake
hash -r
sudo snap install cmake --classic
sudo update-alternatives --install /usr/bin/python python /usr/bin/python3.6 1
sudo apt-get -y install python3-pip
sudo update-alternatives --install /usr/bin/pip pip /usr/bin/pip3 1
mkdir -p ~/install && cd ~/install
git clone --recursive https://github.com/intel-isl/Open3D
cd Open3D
#checkout tags/v.0.11.1
git checkout 31fdcadbe4149e833df86612e718c23868c5b72f

sudo apt-get update -y
sudo apt-get install -y apt-utils build-essential 
sudo apt-get install -y python3-dev
sudo apt-get install -y xorg-dev libglu1-mesa-dev
sudo apt-get install -y libblas-dev liblapack-dev liblapacke-dev
sudo apt-get install -y libsdl2-dev libc++-7-dev libc++abi-7-dev libxi-dev
sudo apt-get install -y clang-7

mkdir build
cd build

export CUDA_PATH=/usr/local/cuda-10.2/
export CUDA_INC_PATH=/usr/local/cuda-10.2/include
cmake \
    -DCMAKE_CUDA_COMPILER=/usr/local/cuda-10.2/bin/nvcc \
    -DCMAKE_BUILD_TYPE=Release \
    -DBUILD_SHARED_LIBS=ON \
    -DBUILD_CUDA_MODULE=ON \
    -DBUILD_GUI=ON \
    -DBUILD_TENSORFLOW_OPS=OFF \
    -DBUILD_PYTORCH_OPS=OFF \
    -DBUILD_UNIT_TESTS=ON \
    -DCMAKE_INSTALL_PREFIX=~/open3d_install \
    -DPYTHON_EXECUTABLE=$(which python) \
    ..

make -j$(nproc)
make tests -j$(nproc)
./bin/tests --gtest_filter="-*Reduce*Sum*"

pip install Cython
sudo apt-get install -y libjpeg-dev
python -c "import open3d; print(open3d)"`


**CMAKE Output**
the cmake output look like this,
-- THIRD_PARTY_DOWNLOAD_DIR is not specified, will download directly.
-- Open3D 0.11.1+31fdcadb
CMake Warning at CMakeLists.txt:180 (message):
  ARM CPU detected, setting USE_BLAS=ON.


CMake Warning at CMakeLists.txt:184 (message):
  ARM CPU detected, setting BUILD_FILAMENT_FROM_SOURCE=ON.


-- Using python from PYTHON_EXECUTABLE variable: /usr/bin/python
-- (PYTHON_EXECUTABLE matches python from PATH)
-- CMAKE_CUDA_COMPILER: /usr/local/cuda-10.2/bin/nvcc
-- CMAKE_CUDA_COMPILER_VERSION: 
-- Building CUDA enabled
-- Autodetected CUDA architecture(s):  5.3
-- CUDA_GENCODES: -gencode arch=compute_53,code=sm_53
-- CMAKE_CUDA_TOOLKIT_INCLUDE_DIRECTORIES: /usr/local/cuda-10.2/targets/aarch64-linux/include
-- CMAKE_CUDA_IMPLICIT_INCLUDE_DIRECTORIES: /usr/include/c++/7;/usr/include/aarch64-linux-gnu/c++/7;/usr/include/c++/7/backward;/usr/lib/gcc/aarch64-linux-gnu/7/include;/usr/local/include;/usr/lib/gcc/aarch64-linux-gnu/7/include-fixed;/usr/include/aarch64-linux-gnu;/usr/include
-- Building library Assimp from source
-- Building with OpenMP
-- Building library 3rdparty_eigen3 from source
-- Building library 3rdparty_flann from source
-- Building library 3rdparty_nanoflann from source
-- Building library 3rdparty_glew from source
-- Building library 3rdparty_glfw3 from source
-- Found Vulkan: /usr/lib/aarch64-linux-gnu/libvulkan.so  
-- Using X11 for window creation
-- Building third-party library JPEG from source
-- NASM assembler not found - libjpeg-turbo performance may suffer
-- libturbojpeg: WITH_CRT_DLL=OFF
-- Building library 3rdparty_jsoncpp from source
-- Building library 3rdparty_lzf from source
-- Building library 3rdparty_tritriintersect from source
-- Building third-party library zlib from source
-- Building third-party library libpng from source
-- Building library 3rdparty_rply from source
-- Building library 3rdparty_tinyfiledialogs from source
-- Building library 3rdparty_tinygltf from source
-- Building library 3rdparty_tinyobjloader from source
-- Building library 3rdparty_qhull_r from source
-- Building library 3rdparty_qhullcpp from source
-- Building library 3rdparty_fmt from source
-- pybind11 v2.5.0
-- Building library 3rdparty_poisson from source
-- Building library 3rdparty_googletest from source
-- Building library 3rdparty_imgui from source
-- Building third-party library Filament from source
-- Filament can only be built with Clang >= 7
-- Using /usr/bin/clang++-7 to build Filament
-- CLANG_LIBDIR: /usr/lib/llvm-7/lib
-- CPP_LIBRARY: /usr/lib/llvm-7/lib/libc++.so
-- CPPABI_LIBRARY: /usr/lib/llvm-7/lib/libc++abi.so
-- A library with LAPACK API found.
-- A cache variable, namely LAPACKE_DIR, has been set to specify the install directory of LAPACKE
-- Looking for LAPACKE_dgeqrf
-- Looking for LAPACKE_dgeqrf - not found
-- Looking for lapacke : test with lapack fails
-- Looking for LAPACKE_dgeqrf
-- Looking for LAPACKE_dgeqrf - found
-- Using system BLAS/LAPACK
-- BUILD_JUPYTER_EXTENSION is set to OFF
-- Using Python version: 3.6
-- Using Python version major: 3
-- PYPI_PACKAGE_NAME: open3d
-- ========================================================================
-- Open3D 0.11.1 Configuration Summary
-- ========================================================================
-- Enabled Features:
--   OpenMP .................................. ON
--   Headless Rendering ...................... OFF
--   Azure Kinect Support .................... OFF
--   CUDA Support ............................ ON
--   Build GUI ............................... ON
--   Build Shared Library .................... ON
--   Build Unit Tests ........................ ON
--   Build Examples .......................... 
--   Build Python Module ..................... ON
--   - with Jupyter Notebook Support ......... OFF
--   Build TensorFlow Ops .................... OFF
--   Build PyTorch Ops ....................... OFF
--   Build Benchmarks ........................ OFF
--   Bundle Open3D-ML ........................ OFF
--   Build RPC interface ..................... OFF
--   Force GLIBCXX_USE_CXX11_ABI= ............ 0
-- ========================================================================
-- Third-Party Dependencies:
--   EIGEN3 .................................. yes (build from source)
--   FILAMENT ................................ yes (build from source)
--   FLANN ................................... yes (build from source)
--   FMT ..................................... yes (build from source)
--   GLEW .................................... yes (build from source)
--   GLFW .................................... yes (build from source)
--   GOOGLETEST .............................. yes (build from source)
--   IMGUI ................................... yes (build from source)
--   JPEG .................................... yes (build from source)
--   JSONCPP ................................. yes (build from source)
--   LIBLZF .................................. yes (build from source)
--   OPENGL .................................. yes (build from source)
--   PNG ..................................... yes (build from source)
--   PYBIND11 ................................ yes (build from source)
--   QHULL ................................... yes (build from source)
--   LIBREALSENSE ............................ no
--   TINYFILEDIALOGS ......................... yes (build from source)
--   TINYGLTF ................................ yes (build from source)
--   TINYOBJLOADER ........................... yes (build from source)
-- ========================================================================
-- Configuring done
-- Generating done
-- Build files have been written to: /home/ivan/install/Open3D/build

- Operating system: Ubuntu 18.04, nvidia jetpack (nano) 4.4.1
- Python version: Python 3.6
- Open3D version: v0.11.1
- Compiler version (if built from source): see cmake output
```

prewettg:
(also resolves #2549)

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2585)


benjaminum (changes requested):
Most comments are about the scene tree. For the functions called by the user we should rename arguments "geometry" to "obj" to make it shorter and more general.

> Reviewed 17 of 17 files at r1.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2585#-:-MMW-y_JCdHXgnMtxhYW:b-nb6tf3)* status: all files reviewed, 12 unresolved discussions (waiting on @prewettg)

___
*[cpp/open3d/visualization/visualizer/DrawVisualizer.h, line 62 at r1](https://reviewable.io/reviews/intel-isl/open3d/2585#-MMVmxfv7oJ71G2x3Cor:-MMVmxfv7oJ71G2x3Cos:b6mx38f) ([raw file](https://github.com/intel-isl/open3d/blob/a8085fe4b3ac5a1d2a23cbd67432fb13ac5363cf/cpp/open3d/visualization/visualizer/DrawVisualizer.h#L62)):*
> ```objc
>     enum class Shader { STANDARD, NORMALS, DEPTH };
> 
>     struct DrawObject {
> ```

No change request, just a comment:
This is good! Would it help you if we have an interface class for objects that can be passed to the visualizer? All objects that we want to draw have to subclass a Drawable class.
___
*[cpp/open3d/visualization/visualizer/DrawVisualizer.h, line 108 at r1](https://reviewable.io/reviews/intel-isl/open3d/2585#-MMVk2Oq58bPWJewzad_:-MMVk2Oq58bPWJewzada:b29zbgs) ([raw file](https://github.com/intel-isl/open3d/blob/a8085fe4b3ac5a1d2a23cbd67432fb13ac5363cf/cpp/open3d/visualization/visualizer/DrawVisualizer.h#L108)):*
> ```objc
>     void SetShader(Shader shader);
> 
>     void AddGeometry(const std::string& name,
> ```

I think there was a misunderstanding with the layers. Can we do groups with the object name? E.g., name="group/subgroup/myobject".
___
*[cpp/open3d/visualization/visualizer/DrawVisualizer.h, line 113 at r1](https://reviewable.io/reviews/intel-isl/open3d/2585#-MMVliSm9YAvEb3bq0ed:-MMVliSm9YAvEb3bq0ee:b4bmun7) ([raw file](https://github.com/intel-isl/open3d/blob/a8085fe4b3ac5a1d2a23cbd67432fb13ac5363cf/cpp/open3d/visualization/visualizer/DrawVisualizer.h#L113)):*
> ```objc
>                      const std::string& group = "",
>                      double time = 0.0,
>                      bool is_visible = true);
> ```

No change request, just a comment:
Passing properties like "is_visible" is good. In the future we may want something more generic because the number of properties will grow.
___
*[cpp/open3d/visualization/visualizer/DrawVisualizer.h, line 122 at r1](https://reviewable.io/reviews/intel-isl/open3d/2585#-MMVmInQ15ZAtOT_TBkf:-MMVmInQ15ZAtOT_TBkg:bluki3l) ([raw file](https://github.com/intel-isl/open3d/blob/a8085fe4b3ac5a1d2a23cbd67432fb13ac5363cf/cpp/open3d/visualization/visualizer/DrawVisualizer.h#L122)):*
> ```objc
>                      bool is_visible = true);
> 
>     void RemoveGeometry(const std::string& name);
> ```

We can call this function `RemoveObject()`  and use it also for objects that are not geometries like cameras or lights.
___
*[cpp/open3d/visualization/visualizer/DrawVisualizer.h, line 124 at r1](https://reviewable.io/reviews/intel-isl/open3d/2585#-MMVmbsp-Yah4FTtew-m:-MMVmbsqC_5bCreKnkdd:b-2nnlv3) ([raw file](https://github.com/intel-isl/open3d/blob/a8085fe4b3ac5a1d2a23cbd67432fb13ac5363cf/cpp/open3d/visualization/visualizer/DrawVisualizer.h#L124)):*
> ```objc
>     void RemoveGeometry(const std::string& name);
> 
>     void ShowGeometry(const std::string& name, bool show);
> ```

Same here `ShowObject()` or `SetObjectVisibility()`
___
*[cpp/open3d/visualization/visualizer/DrawVisualizer.h, line 126 at r1](https://reviewable.io/reviews/intel-isl/open3d/2585#-MMVmq3T4fytZ1ayj07O:-MMVmq3T4fytZ1ayj07P:bgah7al) ([raw file](https://github.com/intel-isl/open3d/blob/a8085fe4b3ac5a1d2a23cbd67432fb13ac5363cf/cpp/open3d/visualization/visualizer/DrawVisualizer.h#L126)):*
> ```objc
>     void ShowGeometry(const std::string& name, bool show);
> 
>     DrawObject GetGeometry(const std::string& name) const;
> ```

`GetObject`. How can I get the object that I have set at time=t?
___
*[cpp/open3d/visualization/visualizer/DrawVisualizer.h, line 133 at r1](https://reviewable.io/reviews/intel-isl/open3d/2585#-MMVnmxw53fWmC4FXC4Y:-MMVnmxw53fWmC4FXC4Z:b8naqak) ([raw file](https://github.com/intel-isl/open3d/blob/a8085fe4b3ac5a1d2a23cbd67432fb13ac5363cf/cpp/open3d/visualization/visualizer/DrawVisualizer.h#L133)):*
> ```objc
>     void ShowSkybox(bool show);
>     void ShowAxes(bool show);
>     void EnableGroup(const std::string& group, bool enable);
> ```

Probably not necessary, because it will be the same as `ShowObject()`.
___
*[cpp/open3d/visualization/visualizer/DrawVisualizer.cpp, line 595 at r1](https://reviewable.io/reviews/intel-isl/open3d/2585#-MMVqKLg-EF1nt_GPFfF:-MMVqKLg-EF1nt_GPFfG:brimz07) ([raw file](https://github.com/intel-isl/open3d/blob/a8085fe4b3ac5a1d2a23cbd67432fb13ac5363cf/cpp/open3d/visualization/visualizer/DrawVisualizer.cpp#L595)):*
> ```cpp
>         // Geometry list
>         settings.geometries_panel =
>                 new CollapsableVert("Geometries", v_spacing, margins);
> ```

Geometries -> Scene (or Scene tree because we have already used scene).
___
*[cpp/pybind/visualization/drawvisualizer.cpp, line 93 at r1](https://reviewable.io/reviews/intel-isl/open3d/2585#-MMVy8uz82warFgNmw64:-MMVy8uz82warFgNmw65:b-2w4aok) ([raw file](https://github.com/intel-isl/open3d/blob/a8085fe4b3ac5a1d2a23cbd67432fb13ac5363cf/cpp/pybind/visualization/drawvisualizer.cpp#L93)):*
> ```cpp
>                  "one parameter, the DrawVisualizer instance, and does not "
>                  "return any value.")
>             .def("add_geometry",
> ```

Can we name this just "add"? or "set" if we can use it to overwrite existing objects. As a python user we do not care much about the type and want to do `add([mesh, camera, light])`.
___
*[examples/python/gui/draw.py, line 92 at r1](https://reviewable.io/reviews/intel-isl/open3d/2585#-MMVzcR_4oQ77dDXXAJs:-MMVzcR_4oQ77dDXXAJt:b1udsoz) ([raw file](https://github.com/intel-isl/open3d/blob/a8085fe4b3ac5a1d2a23cbd67432fb13ac5363cf/examples/python/gui/draw.py#L92)):*
> ```Python
> 
> 
> def actions():
> ```

This feature is just awesome!
___
*[examples/python/gui/draw.py, line 146 at r1](https://reviewable.io/reviews/intel-isl/open3d/2585#-MMVzyQy5-tgjJq7vnW-:-MMVzyQy5-tgjJq7vnW0:b-hal5br) ([raw file](https://github.com/intel-isl/open3d/blob/a8085fe4b3ac5a1d2a23cbd67432fb13ac5363cf/examples/python/gui/draw.py#L146)):*
> ```Python
> 
> 
> def groups():
> ```

The check state does not propagate to the children. Is this intentional?
___
*[python/open3d/visualization/draw.py, line 5 at r1](https://reviewable.io/reviews/intel-isl/open3d/2585#-MMW-ZLY1kGbL_awPaE9:-MMW-ZLY1kGbL_awPaEA:b-g31wzf) ([raw file](https://github.com/intel-isl/open3d/blob/a8085fe4b3ac5a1d2a23cbd67432fb13ac5363cf/python/open3d/visualization/draw.py#L5)):*
> ```Python
> 
> 
> def draw(geometry=None,
> ```

We need to sync here to make the interface of the ExternalVisualizer compatible.

errissa (changes requested):
> Reviewed 16 of 39 files at r2, 1 of 13 files at r3, 1 of 13 files at r4.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2585)* status: 19 of 44 files reviewed, 16 unresolved discussions (waiting on @benjaminum, @errissa, and @prewettg)

___
*[cpp/open3d/visualization/gui/Materials/unlitPolygonOffset.mat, line 14 at r4](https://reviewable.io/reviews/intel-isl/open3d/2585#-MNexX5C2BuRlVr1XB3O:-MNexX5C2BuRlVr1XB3P:b-6o3wew) ([raw file](https://github.com/intel-isl/open3d/blob/2b2f9c3a5bd4d2b22c3d818b1328d39fa3e1a29e/cpp/open3d/visualization/gui/Materials/unlitPolygonOffset.mat#L14)):*
> ```
> vertex {
>     void materialVertex(inout MaterialVertexInputs material) {
>         float4 toCamera = 0.001f * (float4(getWorldCameraPosition(), 1.0f) - material.worldPosition);
> ```

This will move the vertex toward/away from camera an amount proportional to the distance between camera and vertex. the bigger the distance the bigger the offset. this is similar to the using the `factor` parameter of glPolygonOffset. The `units` parameter of glPolygonOffset is a constant offset.
___
*[cpp/open3d/visualization/gui/Materials/unlitPolygonOffset.mat, line 16 at r4](https://reviewable.io/reviews/intel-isl/open3d/2585#-MNex3YsDeBSXMwswKYb:-MNex3YuEJTdmEgIVAbk:bx9nc89) ([raw file](https://github.com/intel-isl/open3d/blob/2b2f9c3a5bd4d2b22c3d818b1328d39fa3e1a29e/cpp/open3d/visualization/gui/Materials/unlitPolygonOffset.mat#L16)):*
> ```
>         float4 toCamera = 0.001f * (float4(getWorldCameraPosition(), 1.0f) - material.worldPosition);
>         material.worldPosition += toCamera;
>         
> ```

are these red spaces? If so, the unnecessary spaces should be removed. check-style would catch that if the .mat files were C++ files.
___
*[cpp/open3d/visualization/rendering/filament/FilamentCamera.cpp, line 81 at r4](https://reviewable.io/reviews/intel-isl/open3d/2585#-MNezH_Z7a1diN5ryzIH:-MNezH_Z7a1diN5ryzII:b-wmrmnb) ([raw file](https://github.com/intel-isl/open3d/blob/2b2f9c3a5bd4d2b22c3d818b1328d39fa3e1a29e/cpp/open3d/visualization/rendering/filament/FilamentCamera.cpp#L81)):*
> ```cpp
>     auto e_matrix = t.matrix();
>     return filament::math::mat4(filament::math::mat4::row_major_init{
>             e_matrix(0, 0), e_matrix(0, 1), e_matrix(0, 2), e_matrix(0, 3),
> ```

why this commented out code?
___
*[cpp/open3d/visualization/rendering/filament/FilamentCamera.cpp, line 159 at r4](https://reviewable.io/reviews/intel-isl/open3d/2585#-MNeze9J79mSNUGQCwJb:-MNeze9K-aLwLOGY9XAd:bj2xqu9) ([raw file](https://github.com/intel-isl/open3d/blob/2b2f9c3a5bd4d2b22c3d818b1328d39fa3e1a29e/cpp/open3d/visualization/rendering/filament/FilamentCamera.cpp#L159)):*
> ```cpp
> 
> double FilamentCamera::GetFieldOfView() const {
>     if (projection_.is_defined_by_planes) {
> ```

This comment confuses me a little. is the purpose of the `is_defined_by_planes` flag to set orthographic projections? You can also set perspective projections using planes which would have meaningful FOVs.

prewettg (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2585)* status: 19 of 44 files reviewed, 16 unresolved discussions (waiting on @benjaminum, @errissa, and @prewettg)

___
*[cpp/open3d/visualization/gui/Materials/unlitPolygonOffset.mat, line 14 at r4](https://reviewable.io/reviews/intel-isl/open3d/2585#-MNexX5C2BuRlVr1XB3O:-MNfUXbT8wer2x7g2Bov:b-qdp9ka) ([raw file](https://github.com/intel-isl/open3d/blob/2b2f9c3a5bd4d2b22c3d818b1328d39fa3e1a29e/cpp/open3d/visualization/gui/Materials/unlitPolygonOffset.mat#L14)):*

This will move the vertex toward/away from camera an amount proportional to the distance between camera and vertex. the bigger the distance the bigger the offset. this is similar to the using the `factor` parameter of glPolygonOffset. The `units` parameter of glPolygonOffset is a constant offset.
</blockquote></details>

That was my intent. Not sure I understand the question/problem?
___
*[cpp/open3d/visualization/gui/Materials/unlitPolygonOffset.mat, line 16 at r4](https://reviewable.io/reviews/intel-isl/open3d/2585#-MNex3YsDeBSXMwswKYb:-MNfULgJ-lb28d6JiJMh:b194zjo) ([raw file](https://github.com/intel-isl/open3d/blob/2b2f9c3a5bd4d2b22c3d818b1328d39fa3e1a29e/cpp/open3d/visualization/gui/Materials/unlitPolygonOffset.mat#L16)):*

are these red spaces? If so, the unnecessary spaces should be removed. check-style would catch that if the .mat files were C++ files.
</blockquote></details>

Yep, seem to be spaces. I use Emacs to edit, and the autoindent puts in spaces, and I don't always remember to remove them, apparently. Removed
___
*[cpp/open3d/visualization/visualizer/DrawVisualizer.h, line 108 at r1](https://reviewable.io/reviews/intel-isl/open3d/2585#-MMVk2Oq58bPWJewzad_:-MNfXS6J2Mke6SKkH-jY:b3odha5) ([raw file](https://github.com/intel-isl/open3d/blob/a8085fe4b3ac5a1d2a23cbd67432fb13ac5363cf/cpp/open3d/visualization/visualizer/DrawVisualizer.h#L108)):*

I think there was a misunderstanding with the layers. Can we do groups with the object name? E.g., name="group/subgroup/myobject".
</blockquote></details>

Depends on time. It may have to wait until we do a full scene graph
___
*[cpp/open3d/visualization/visualizer/DrawVisualizer.h, line 113 at r1](https://reviewable.io/reviews/intel-isl/open3d/2585#-MMVliSm9YAvEb3bq0ed:-MNfXbncC2YKekXK0l-5:b-47oghg) ([raw file](https://github.com/intel-isl/open3d/blob/a8085fe4b3ac5a1d2a23cbd67432fb13ac5363cf/cpp/open3d/visualization/visualizer/DrawVisualizer.h#L113)):*

No change request, just a comment:
Passing properties like "is_visible" is good. In the future we may want something more generic because the number of properties will grow.
</blockquote></details>

I'd prefer not to prematurely optimize for genericity. If we can't articulate what kinds of generic things are needed now, I think it's better to wait until we know what we need and refactor. This may be a part of the scene graph, anyway.
___
*[cpp/open3d/visualization/visualizer/DrawVisualizer.h, line 122 at r1](https://reviewable.io/reviews/intel-isl/open3d/2585#-MMVmInQ15ZAtOT_TBkf:-MNfXzMHCCZDjVBNZA1x:b91tqtm) ([raw file](https://github.com/intel-isl/open3d/blob/a8085fe4b3ac5a1d2a23cbd67432fb13ac5363cf/cpp/open3d/visualization/visualizer/DrawVisualizer.h#L122)):*

We can call this function `RemoveObject()`  and use it also for objects that are not geometries like cameras or lights.
</blockquote></details>

This makes more sense for the scene graph. This current version is meant as more of a draw_geometries() replacement
___
*[cpp/open3d/visualization/visualizer/DrawVisualizer.h, line 124 at r1](https://reviewable.io/reviews/intel-isl/open3d/2585#-MMVmbsp-Yah4FTtew-m:-MNfY5wE4LyiIgDdJFni:b-3ju4d8) ([raw file](https://github.com/intel-isl/open3d/blob/a8085fe4b3ac5a1d2a23cbd67432fb13ac5363cf/cpp/open3d/visualization/visualizer/DrawVisualizer.h#L124)):*

Same here `ShowObject()` or `SetObjectVisibility()`
</blockquote></details>

This makes more sense for the scene graph. This current version is meant as more of a draw_geometries() replacement.
___
*[cpp/open3d/visualization/visualizer/DrawVisualizer.h, line 126 at r1](https://reviewable.io/reviews/intel-isl/open3d/2585#-MMVmq3T4fytZ1ayj07O:-MNfY7ZH9yq8Xfpto4xb:b-dnzcxl) ([raw file](https://github.com/intel-isl/open3d/blob/a8085fe4b3ac5a1d2a23cbd67432fb13ac5363cf/cpp/open3d/visualization/visualizer/DrawVisualizer.h#L126)):*

`GetObject`. How can I get the object that I have set at time=t?
</blockquote></details>

I think callers should keep track of that themselves. But maybe a scene graph query would be useful.
___
*[cpp/open3d/visualization/visualizer/DrawVisualizer.cpp, line 595 at r1](https://reviewable.io/reviews/intel-isl/open3d/2585#-MMVqKLg-EF1nt_GPFfF:-MNfYixZ--HnMW1thOhT:b-hkp859) ([raw file](https://github.com/intel-isl/open3d/blob/a8085fe4b3ac5a1d2a23cbd67432fb13ac5363cf/cpp/open3d/visualization/visualizer/DrawVisualizer.cpp#L595)):*

Geometries -> Scene (or Scene tree because we have already used scene).
</blockquote></details>

Deferring until we actually get a proper scene graph.

prewettg (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2585)* status: 16 of 44 files reviewed, 16 unresolved discussions (waiting on @benjaminum, @errissa, and @prewettg)

___
*[cpp/open3d/visualization/rendering/filament/FilamentCamera.cpp, line 81 at r4](https://reviewable.io/reviews/intel-isl/open3d/2585#-MNezH_Z7a1diN5ryzIH:-MNfbxQ8EHu8gXemdx-H:bbwg5s) ([raw file](https://github.com/intel-isl/open3d/blob/2b2f9c3a5bd4d2b22c3d818b1328d39fa3e1a29e/cpp/open3d/visualization/rendering/filament/FilamentCamera.cpp#L81)):*

why this commented out code?
</blockquote></details>

Historical reasons. Removed
___
*[cpp/open3d/visualization/rendering/filament/FilamentCamera.cpp, line 159 at r4](https://reviewable.io/reviews/intel-isl/open3d/2585#-MNeze9J79mSNUGQCwJb:-MNfc1NiB6lYvVDlwphX:bp09w1z) ([raw file](https://github.com/intel-isl/open3d/blob/2b2f9c3a5bd4d2b22c3d818b1328d39fa3e1a29e/cpp/open3d/visualization/rendering/filament/FilamentCamera.cpp#L159)):*

This comment confuses me a little. is the purpose of the `is_defined_by_planes` flag to set orthographic projections? You can also set perspective projections using planes which would have meaningful FOVs.
</blockquote></details>

True, hadn't thought about that, although given the interface to Camera, the only way you can specify planes is perpendicular ones... I changed the names in the structures to be more precise.

benjaminum (changes requested):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2585)* status: 16 of 44 files reviewed, 17 unresolved discussions (waiting on @benjaminum, @errissa, and @prewettg)

___
*[cpp/open3d/visualization/visualizer/DrawVisualizer.h, line 126 at r1](https://reviewable.io/reviews/intel-isl/open3d/2585#-MMVmq3T4fytZ1ayj07O:-MNg7eE-AFv2lW-ohlnO:b142uqg) ([raw file](https://github.com/intel-isl/open3d/blob/a8085fe4b3ac5a1d2a23cbd67432fb13ac5363cf/cpp/open3d/visualization/visualizer/DrawVisualizer.h#L126)):*

I think callers should keep track of that themselves. But maybe a scene graph query would be useful.
</blockquote></details>

Yes, but they need to be able to pass the time parameter.
___
*[cpp/open3d/visualization/visualizer/O3DVisualizer.h, line 120 at r5](https://reviewable.io/reviews/intel-isl/open3d/2585#-MNgdkpQ-MXhyClqOZ4W:-MNgdkpQ-MXhyClqOZ4X:b-2gnypu) ([raw file](https://github.com/intel-isl/open3d/blob/108815a3f3ab71cf2bd29a74b4b2107d8884e708/cpp/open3d/visualization/visualizer/O3DVisualizer.h#L120)):*
> ```objc
>                      rendering::Material* material = nullptr,
>                      const std::string& group = "",
>                      double order = 0.0,
> ```

time, please.

prewettg (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2585)* status: 16 of 44 files reviewed, 17 unresolved discussions (waiting on @benjaminum, @errissa, and @prewettg)

___
*[cpp/open3d/visualization/visualizer/O3DVisualizer.h, line 120 at r5](https://reviewable.io/reviews/intel-isl/open3d/2585#-MNgdkpQ-MXhyClqOZ4W:-MNij-Nj4X2eyw9kFGVO:b-eer4cu) ([raw file](https://github.com/intel-isl/open3d/blob/108815a3f3ab71cf2bd29a74b4b2107d8884e708/cpp/open3d/visualization/visualizer/O3DVisualizer.h#L120)):*

time, please.
</blockquote></details>

German nixed the idea of time. Instead he wants an ordering, so that if you give objects with t=[0, 1, 2, 7] you don't get five seconds of nothing. So it's no longer clock time, but if your timestep is constant (presumably most cases), "order" and "time" are essentially the same thing.

benjaminum (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2585)* status: 16 of 44 files reviewed, 17 unresolved discussions (waiting on @benjaminum, @errissa, @germanros1987, and @prewettg)

___
*[cpp/open3d/visualization/visualizer/O3DVisualizer.h, line 120 at r5](https://reviewable.io/reviews/intel-isl/open3d/2585#-MNgdkpQ-MXhyClqOZ4W:-MNilXka7LrnAwCxK48-:b-8ifzh8) ([raw file](https://github.com/intel-isl/open3d/blob/108815a3f3ab71cf2bd29a74b4b2107d8884e708/cpp/open3d/visualization/visualizer/O3DVisualizer.h#L120)):*

German nixed the idea of time. Instead he wants an ordering, so that if you give objects with t=[0, 1, 2, 7] you don't get five seconds of nothing. So it's no longer clock time, but if your timestep is constant (presumably most cases), "order" and "time" are essentially the same thing.
</blockquote></details>

Yes, it is the same and even if we call it time there wouldn't be five seconds of nothing. Calling it time makes the parameter easy to understand because in most of the cases we will use time to order the objects. @germanros1987 can we call it time? If I could wish for one of the comments to be addressed, it would be this one.

prewettg (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2585)* status: 16 of 46 files reviewed, 17 unresolved discussions (waiting on @benjaminum, @errissa, @germanros1987, and @prewettg)

___
*[cpp/open3d/visualization/visualizer/O3DVisualizer.h, line 120 at r5](https://reviewable.io/reviews/intel-isl/open3d/2585#-MNgdkpQ-MXhyClqOZ4W:-MNj2RIZ8cdNkfvaAMJM:bx7nik0) ([raw file](https://github.com/intel-isl/open3d/blob/108815a3f3ab71cf2bd29a74b4b2107d8884e708/cpp/open3d/visualization/visualizer/O3DVisualizer.h#L120)):*

Yes, it is the same and even if we call it time there wouldn't be five seconds of nothing. Calling it time makes the parameter easy to understand because in most of the cases we will use time to order the objects. @germanros1987 can we call it time? If I could wish for one of the comments to be addressed, it would be this one.
</blockquote></details>

But "time" is specifically what this is not. If I specify objects at t=[0, 1, 2, 3], I would expect to see object 1 show up at 1.0 seconds, object 2 at 2.0 seconds, etc. But what will happen is that object 1 shows up at 100 ms (that being the default frame time), object 2 at 200 ms, etc. Likewise, if I am specifying time, then I would expect with t=[0, 1, 5] that object 0 is displayed for 1 second, object 1 for 4 seconds (1.0 - 4.99), and object 2 for one frame. But in actuality what will happen is that object 0 will be displayed for 100 ms, then object 1 for 100 ms, and object 2 for 100 ms. So instead of a 5+ second animation as suggested by "time", we actually get a 300 ms animation (assuming the default frame time of 100 ms). Actually, the length of the animation is entirely dependent on len(unique_times) * frame_time, which is completely opposite of the expectation of a parameter named "time", at least to my way of thinking. To my mind, if a parameter is called "time", then implied units is "seconds" (possibly also "milliseconds");  in this case the unit is "frame number as given by the index into the list of sorted unique values, not the actual value specified by the parameter", which I submit are two completely different things. At least with a parameter named "order" you are under no illusions that it corresponds to anything related to the passing of events as commonly measured in seconds.

benjaminum (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2585)* status: 16 of 46 files reviewed, 10 unresolved discussions (waiting on @benjaminum, @errissa, @germanros1987, and @prewettg)

___
*[cpp/open3d/visualization/visualizer/O3DVisualizer.h, line 120 at r5](https://reviewable.io/reviews/intel-isl/open3d/2585#-MNgdkpQ-MXhyClqOZ4W:-MNj63zG4EjG9mCcYCUJ:b-9mqylm) ([raw file](https://github.com/intel-isl/open3d/blob/108815a3f3ab71cf2bd29a74b4b2107d8884e708/cpp/open3d/visualization/visualizer/O3DVisualizer.h#L120)):*

But "time" is specifically what this is not. If I specify objects at t=[0, 1, 2, 3], I would expect to see object 1 show up at 1.0 seconds, object 2 at 2.0 seconds, etc. But what will happen is that object 1 shows up at 100 ms (that being the default frame time), object 2 at 200 ms, etc. Likewise, if I am specifying time, then I would expect with t=[0, 1, 5] that object 0 is displayed for 1 second, object 1 for 4 seconds (1.0 - 4.99), and object 2 for one frame. But in actuality what will happen is that object 0 will be displayed for 100 ms, then object 1 for 100 ms, and object 2 for 100 ms. So instead of a 5+ second animation as suggested by "time", we actually get a 300 ms animation (assuming the default frame time of 100 ms). Actually, the length of the animation is entirely dependent on len(unique_times) * frame_time, which is completely opposite of the expectation of a parameter named "time", at least to my way of thinking. To my mind, if a parameter is called "time", then implied units is "seconds" (possibly also "milliseconds");  in this case the unit is "frame number as given by the index into the list of sorted unique values, not the actual value specified by the parameter", which I submit are two completely different things. At least with a parameter named "order" you are under no illusions that it corresponds to anything related to the passing of events as commonly measured in seconds.
</blockquote></details>

Please have a look at the visualization examples+viewer that I compiled.

errissa (changes requested):
> Reviewed 4 of 39 files at r2, 3 of 13 files at r3, 5 of 13 files at r4, 3 of 4 files at r5, 16 of 17 files at r6, 3 of 11 files at r7.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2585)* status: 39 of 47 files reviewed, 8 unresolved discussions (waiting on @errissa, @germanros1987, and @prewettg)

___
*[cpp/open3d/visualization/gui/Materials/unlitPolygonOffset.mat, line 14 at r4](https://reviewable.io/reviews/intel-isl/open3d/2585#-MNexX5C2BuRlVr1XB3O:-MO-Kc0L-Fh2TuM8ZbjG:bngmth6) ([raw file](https://github.com/intel-isl/open3d/blob/2b2f9c3a5bd4d2b22c3d818b1328d39fa3e1a29e/cpp/open3d/visualization/gui/Materials/unlitPolygonOffset.mat#L14)):*

That was my intent. Not sure I understand the question/problem?
</blockquote></details>

There's no problem, just wanted to make sure you understood that this isn't an exact replacement for glPolygonOffset and that the solution you implemented is what you wanted. It seems to work when I tested.
___
*[cpp/open3d/visualization/visualizer/O3DVisualizerSelections.h, line 107 at r7](https://reviewable.io/reviews/intel-isl/open3d/2585#-MO-LzL8CdSn4dX3-dos:-MO-LzL8CdSn4dX3-dot:b-pstgp) ([raw file](https://github.com/intel-isl/open3d/blob/4a7b8b159698123782a3576a0e70420341adfceb/cpp/open3d/visualization/visualizer/O3DVisualizerSelections.h#L107)):*
> ```objc
>     };
> 
>     int point_size_ = 3;
> ```

why point size of 3 as the default?

benjaminum (approved):
LGTM. Some changes will be addressed in another PR. @prewettg please check @errissa comments before merging.

> Reviewed 1 of 10 files at r8.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2585#-:-MO2Dgsr2pnwTTJ3XimI:b3dzif3)* status: 28 of 45 files reviewed, 8 unresolved discussions (waiting on @errissa, @germanros1987, and @prewettg)

errissa (commented):
> Reviewed 8 of 11 files at r7, 7 of 10 files at r8, 7 of 8 files at r10.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2585)* status: 45 of 46 files reviewed, 8 unresolved discussions (waiting on @errissa, @germanros1987, and @prewettg)

errissa (approved):

Issue #2123, opened 2020-07-30
Questions about RANSACConvergenceCriteria


XuyangBai:
Thanks for your great work. I am wondering the effect of `max_iteration` and `max_validation` in `RANSACConvergenceCriteria `. As shown [here](https://github.com/intel-isl/Open3D/blob/de9b015078aa3d86b3f54ba904aa309d2e054060/cpp/open3d/pipelines/registration/Registration.cpp#L217), when using the RANSAC based on correspondence, the aim of `max_iteration` and `max_validation` is the same, and the minimum value of them will be the iteration number. While for RANSAC based on feature matching ([code](https://github.com/intel-isl/Open3D/blob/de9b015078aa3d86b3f54ba904aa309d2e054060/cpp/open3d/pipelines/registration/Registration.cpp#L276)), since I am not an expert on OpenMP, I am not sure about the logic of the algorithm.  It seems every iteration in the for loop will increase the `itr ` by 1 and increase `total_validation` by one, so these two parameters are still having the same effect? If not, could you explain what is an iteration, and what is a validation?

Another question is that is there any plans to support the early stop scheme? For now it seems the algorithm runs as many iterations as we set (at least for ransac based on correspondence), which might be extremely slow.

Thanks a lot.

theNded:
This is a very good point... we also don't quite understand why the initial author implemented it in this way. The roles `max_iteration` and `max_validation` play are also the same in feature based method. Therefore you may set these variables the same, or set one of them to be really large (to be ignored) and tune the other.

Return criteria is also a good suggestion. It is currently used in [RegistrationICP](https://github.com/intel-isl/Open3D/blob/de9b015078aa3d86b3f54ba904aa309d2e054060/cpp/open3d/pipelines/registration/Registration.cpp#L188) but not in RANSAC. You may only change several lines to enable this feature -- would you be willing to contribute and make a PR?

XuyangBai:
[...]

PR #2636, opened 2020-11-17
Update RANSAC

theNded:
Addresses #2123 and supports early termination of RANSAC.
Replace the confusing `max_validation` with conventional `confidence`.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2636)


yxlao (changes requested):
> Reviewed 6 of 6 files at r2.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2636)* status: all files reviewed, 12 unresolved discussions (waiting on @syncle, @theNded, and @yxlao)

___
*[cpp/open3d/pipelines/registration/Registration.h, line 186 at r2](https://reviewable.io/reviews/intel-isl/open3d/2636#-MMYCqCQ70gaTx3RxMZu:-MMYCqCQ70gaTx3RxMZv:bej55zl) ([raw file](https://github.com/intel-isl/open3d/blob/f0e94605dac8e4e8b5b971888aa7999d58fd1ce3/cpp/open3d/pipelines/registration/Registration.h#L186)):*
> ```objc
>         const TransformationEstimation &estimation =
>                 TransformationEstimationPointToPoint(false),
>         int ransac_n = 6,
> ```

`ransac_n` should be the minimal # of points
___
*[cpp/open3d/pipelines/registration/Registration.h, line 190 at r2](https://reviewable.io/reviews/intel-isl/open3d/2636#-MMY9H-R-wu8GgcjlnFI:-MMY9H-R-wu8GgcjlnFJ:b-kzq3z9) ([raw file](https://github.com/intel-isl/open3d/blob/1730bcc83d84095d74ab7a8496284e914f1cd1b8/cpp/open3d/pipelines/registration/Registration.h#L190)):*
> ```objc
>                 TransformationEstimationPointToPoint(false),
>         int ransac_n = 3,
>         const std::vector<std::reference_wrapper<const CorrespondenceChecker>>
> ```

Would it be simpler to use it without `referecne_wrapper`, or are there additional concerns? For `CorrespondenceChecker`, the class objects are pretty small (typically a few `double` or `bool`), we can afford the copy. 
___
*[cpp/open3d/pipelines/registration/Registration.cpp, line 30 at r2](https://reviewable.io/reviews/intel-isl/open3d/2636#-MMYFhUT7quLutQx2NCJ:-MMYFhUT7quLutQx2NCK:b-r3em9t) ([raw file](https://github.com/intel-isl/open3d/blob/1730bcc83d84095d74ab7a8496284e914f1cd1b8/cpp/open3d/pipelines/registration/Registration.cpp#L30)):*
> ```cpp
> 
> #include <cstdlib>
> #include <limits>
> ```

doesn't look like these two headers are used
___
*[cpp/open3d/pipelines/registration/Registration.cpp, line 196 at r2](https://reviewable.io/reviews/intel-isl/open3d/2636#-MMYCJgK0NnZerrpP8rz:-MMYCJgK0NnZerrpP8s-:brvpqw5) ([raw file](https://github.com/intel-isl/open3d/blob/1730bcc83d84095d74ab7a8496284e914f1cd1b8/cpp/open3d/pipelines/registration/Registration.cpp#L196)):*
> ```cpp
>         const CorrespondenceSet &corres,
>         double max_correspondence_distance,
>         const TransformationEstimation &estimation
> ```

This may be renamed to `TransformationEstimator`, in the future PR. What do you think?
___
*[cpp/open3d/pipelines/registration/Registration.cpp, line 234 at r2](https://reviewable.io/reviews/intel-isl/open3d/2636#-MMYFypD4RSFUPZqaVxC:-MMYFypD4RSFUPZqaVxD:b22e1c5) ([raw file](https://github.com/intel-isl/open3d/blob/1730bcc83d84095d74ab7a8496284e914f1cd1b8/cpp/open3d/pipelines/registration/Registration.cpp#L234)):*
> ```cpp
>                     if (!checker.get().Check(source, target, ransac_corres,
>                                              transformation)) {
>                         check = false;
> ```

just do `continue;` here should also work, so that we don't need the `check`
___
*[cpp/open3d/pipelines/registration/Registration.cpp, line 247 at r2](https://reviewable.io/reviews/intel-isl/open3d/2636#-MMYWln-AoIAKc0uC0Hz:-MMYKWQE7Fn0NSTSWta_:b-6s0tol) ([raw file](https://github.com/intel-isl/open3d/blob/1730bcc83d84095d74ab7a8496284e914f1cd1b8/cpp/open3d/pipelines/registration/Registration.cpp#L247)):*
> ```cpp
> 
>                 if (result.fitness_ > best_result_local.fitness_ ||
>                     (result.fitness_ == best_result_local.fitness_ &&
> ```

- Is it relatively uncommon for `result.fitness_ == best_result_local.fitness_` to be exactly the same? In this case, the `rmse` is used rarely. Is this expected?
- Does it make sense to add a comparator operator for `RegistrationResult`? e.g.
    ```cpp
    bool RegistrationResult::IsBetterThan(const RegistrationResult &other);
    ```
    It will simplify code in two locations.
___
*[cpp/open3d/pipelines/registration/Registration.cpp, line 306 at r2](https://reviewable.io/reviews/intel-isl/open3d/2636#-MMY_QZk3ow9P3zSQnYS:-MMY_QZk3ow9P3zSQnYT:bsta1kn) ([raw file](https://github.com/intel-isl/open3d/blob/1730bcc83d84095d74ab7a8496284e914f1cd1b8/cpp/open3d/pipelines/registration/Registration.cpp#L306)):*
> ```cpp
>     pipelines::registration::CorrespondenceSet corres_ij(num_src_pts);
> 
> #pragma omp for nowait
> ```

- shall we keep the barrier `#pragma omp parallel`?
- for the two `omp for`, why do we specify `nowait`?
___
*[cpp/open3d/pipelines/registration/Registration.cpp, line 340 at r2](https://reviewable.io/reviews/intel-isl/open3d/2636#-MMYZtcy2ageim01m8k-:-MMYZtcy2ageim01m8k0:b-1mns44) ([raw file](https://github.com/intel-isl/open3d/blob/1730bcc83d84095d74ab7a8496284e914f1cd1b8/cpp/open3d/pipelines/registration/Registration.cpp#L340)):*
> ```cpp
>         }
> 
>         if (corres_mutual.size() >= 10) {
> ```

could you add code comment on where does the `10` come from?
___
*[cpp/open3d/pipelines/registration/Registration.cpp, line 343 at r2](https://reviewable.io/reviews/intel-isl/open3d/2636#-MMYZyD9A4ZNz9CVsvy7:-MMYZyD9A4ZNz9CVsvy8:brw7fuk) ([raw file](https://github.com/intel-isl/open3d/blob/1730bcc83d84095d74ab7a8496284e914f1cd1b8/cpp/open3d/pipelines/registration/Registration.cpp#L343)):*
> ```cpp
>             utility::LogDebug("{:d} correspondences remain after mutual filter",
>                               corres_mutual.size());
>             return RegistrationRANSACBasedOnCorrespondence(
> ```

would it be simpler to just assign it to `corres_ji`?

e.g.
```
if (corres_mutual.size() >= 10) {
    corres_ij = corres_mutual;
} else {
    print("warn: to few, fall back");
}

```
___
*[cpp/open3d/pipelines/registration/Registration.cpp, line 352 at r2](https://reviewable.io/reviews/intel-isl/open3d/2636#-MMYaR7f7p0drgLtUHnI:-MMYaR7f7p0drgLtUHnJ:bsikcb5) ([raw file](https://github.com/intel-isl/open3d/blob/1730bcc83d84095d74ab7a8496284e914f1cd1b8/cpp/open3d/pipelines/registration/Registration.cpp#L352)):*
> ```cpp
>     }
> 
>     return RegistrationRANSACBasedOnCorrespondence(
> ```

very nice to reuse the code 
___
*[cpp/pybind/pipelines/registration/registration.cpp, line 532 at r2](https://reviewable.io/reviews/intel-isl/open3d/2636#-MMYaJjTArUnxE6z77QB:-MMYaJjTArUnxE6z77QC:bgiktri) ([raw file](https://github.com/intel-isl/open3d/blob/1730bcc83d84095d74ab7a8496284e914f1cd1b8/cpp/pybind/pipelines/registration/registration.cpp#L532)):*
> ```cpp
>                  "Desired probability of success for RANSAC. Used for "
>                  "estimating early termination by k = log(1 - "
>                  "confidence)/log(1 - inlier_ratio^{ransac_n}"},
> ```

add `.`
___
*[examples/cpp/RegistrationRANSAC.cpp, line 92 at r2](https://reviewable.io/reviews/intel-isl/open3d/2636#-MMYanGk--4obOXJICnn:-MMYanGl4wL3yZ3cBLpX:b9ycxxv) ([raw file](https://github.com/intel-isl/open3d/blob/1730bcc83d84095d74ab7a8496284e914f1cd1b8/examples/cpp/RegistrationRANSAC.cpp#L92)):*
> ```cpp
>     }
> 
>     bool mutual_filter;
> ```

`bool xx = false;` (https://stackoverflow.com/a/4622240/1255535)

theNded (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2636)* status: 1 of 6 files reviewed, 12 unresolved discussions (waiting on @syncle and @yxlao)

___
*[cpp/open3d/pipelines/registration/Registration.h, line 186 at r2](https://reviewable.io/reviews/intel-isl/open3d/2636#-MMYCqCQ70gaTx3RxMZu:-MMYgQ0E8v8ARnWgZFUa:b15c5br) ([raw file](https://github.com/intel-isl/open3d/blob/f0e94605dac8e4e8b5b971888aa7999d58fd1ce3/cpp/open3d/pipelines/registration/Registration.h#L186)):*

`ransac_n` should be the minimal # of points
</blockquote></details>

This is the geometric model size, which is fixed 3. # points check is in the function body.
___
*[cpp/open3d/pipelines/registration/Registration.h, line 190 at r2](https://reviewable.io/reviews/intel-isl/open3d/2636#-MMY9H-R-wu8GgcjlnFI:-MMYgrp92izZm_QFZZkf:b-7l0mvx) ([raw file](https://github.com/intel-isl/open3d/blob/1730bcc83d84095d74ab7a8496284e914f1cd1b8/cpp/open3d/pipelines/registration/Registration.h#L190)):*

Would it be simpler to use it without `referecne_wrapper`, or are there additional concerns? For `CorrespondenceChecker`, the class objects are pretty small (typically a few `double` or `bool`), we can afford the copy. 
</blockquote></details>

This was the legacy implementation.  If we want to remove `reference_wrapper`, we have to change the const `Check` method in all the inheritances of `CorrespondenceChecker` to non-const. We can do it in another PR after a redesign.
___
*[cpp/open3d/pipelines/registration/Registration.cpp, line 30 at r2](https://reviewable.io/reviews/intel-isl/open3d/2636#-MMYFhUT7quLutQx2NCJ:-MMYicwh-JJrr2LRRmZa:bss4a2b) ([raw file](https://github.com/intel-isl/open3d/blob/1730bcc83d84095d74ab7a8496284e914f1cd1b8/cpp/open3d/pipelines/registration/Registration.cpp#L30)):*

doesn't look like these two headers are used
</blockquote></details>

Removed.
___
*[cpp/open3d/pipelines/registration/Registration.cpp, line 196 at r2](https://reviewable.io/reviews/intel-isl/open3d/2636#-MMYCJgK0NnZerrpP8rz:-MMYingZEctUcX3J09e6:brxp9sq) ([raw file](https://github.com/intel-isl/open3d/blob/1730bcc83d84095d74ab7a8496284e914f1cd1b8/cpp/open3d/pipelines/registration/Registration.cpp#L196)):*

This may be renamed to `TransformationEstimator`, in the future PR. What do you think?
</blockquote></details>

Sounds good.
___
*[cpp/open3d/pipelines/registration/Registration.cpp, line 234 at r2](https://reviewable.io/reviews/intel-isl/open3d/2636#-MMYFypD4RSFUPZqaVxC:-MMYj4rA-luI-TBjUsha:b-op8ckw) ([raw file](https://github.com/intel-isl/open3d/blob/1730bcc83d84095d74ab7a8496284e914f1cd1b8/cpp/open3d/pipelines/registration/Registration.cpp#L234)):*

just do `continue;` here should also work, so that we don't need the `check`
</blockquote></details>

`continue` will be applied to `for checker` loop right?
___
*[cpp/open3d/pipelines/registration/Registration.cpp, line 247 at r2](https://reviewable.io/reviews/intel-isl/open3d/2636#-MMYWln-AoIAKc0uC0Hz:-MMYlggT0Rvy_XiBpHpb:b-a0r7zl) ([raw file](https://github.com/intel-isl/open3d/blob/1730bcc83d84095d74ab7a8496284e914f1cd1b8/cpp/open3d/pipelines/registration/Registration.cpp#L247)):*

- Is it relatively uncommon for `result.fitness_ == best_result_local.fitness_` to be exactly the same? In this case, the `rmse` is used rarely. Is this expected?
- Does it make sense to add a comparator operator for `RegistrationResult`? e.g.
    ```cpp
    bool RegistrationResult::IsBetterThan(const RegistrationResult &other);
    ```
    It will simplify code in two locations.
</blockquote></details>

For small point clouds (<1000) I think this case should be not uncommon. 
Done.
___
*[cpp/open3d/pipelines/registration/Registration.cpp, line 306 at r2](https://reviewable.io/reviews/intel-isl/open3d/2636#-MMY_QZk3ow9P3zSQnYS:-MMYmlyRBrk7EZHyIq1i:bac6ffr) ([raw file](https://github.com/intel-isl/open3d/blob/1730bcc83d84095d74ab7a8496284e914f1cd1b8/cpp/open3d/pipelines/registration/Registration.cpp#L306)):*

- shall we keep the barrier `#pragma omp parallel`?
- for the two `omp for`, why do we specify `nowait`?
</blockquote></details>

Changed to `omp parallel for`, as we don't have `critical` statements here.
___
*[cpp/open3d/pipelines/registration/Registration.cpp, line 340 at r2](https://reviewable.io/reviews/intel-isl/open3d/2636#-MMYZtcy2ageim01m8k-:-MMYnszO59G7fHhHrQLZ:b933npz) ([raw file](https://github.com/intel-isl/open3d/blob/1730bcc83d84095d74ab7a8496284e914f1cd1b8/cpp/open3d/pipelines/registration/Registration.cpp#L340)):*

could you add code comment on where does the `10` come from?
</blockquote></details>

Changed to some 'empirical' descriptions
___
*[cpp/open3d/pipelines/registration/Registration.cpp, line 343 at r2](https://reviewable.io/reviews/intel-isl/open3d/2636#-MMYZyD9A4ZNz9CVsvy7:-MMYnvPHDr6kykbcvXP6:b-oqdwm1) ([raw file](https://github.com/intel-isl/open3d/blob/1730bcc83d84095d74ab7a8496284e914f1cd1b8/cpp/open3d/pipelines/registration/Registration.cpp#L343)):*

would it be simpler to just assign it to `corres_ji`?

e.g.
```
if (corres_mutual.size() >= 10) {
    corres_ij = corres_mutual;
} else {
    print("warn: to few, fall back");
}

```
</blockquote></details>

I prefer to keep it as it is, `corres_ij == corres_mutual` may change the semantic.
___
*[cpp/open3d/pipelines/registration/Registration.cpp, line 352 at r2](https://reviewable.io/reviews/intel-isl/open3d/2636#-MMYaR7f7p0drgLtUHnI:-MMYjTmZ3t_hkg0OtAV4:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/1730bcc83d84095d74ab7a8496284e914f1cd1b8/cpp/open3d/pipelines/registration/Registration.cpp#L352)):*

very nice to reuse the code 
</blockquote></details>

Done.
___
*[cpp/pybind/pipelines/registration/registration.cpp, line 532 at r2](https://reviewable.io/reviews/intel-isl/open3d/2636#-MMYaJjTArUnxE6z77QB:-MMYjgZf6ZZ-LbHc7K7I:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/1730bcc83d84095d74ab7a8496284e914f1cd1b8/cpp/pybind/pipelines/registration/registration.cpp#L532)):*

add `.`
</blockquote></details>

Done.
___
*[examples/cpp/RegistrationRANSAC.cpp, line 92 at r2](https://reviewable.io/reviews/intel-isl/open3d/2636#-MMYanGk--4obOXJICnn:-MMYjm_22A3b3zQjIc8h:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/1730bcc83d84095d74ab7a8496284e914f1cd1b8/examples/cpp/RegistrationRANSAC.cpp#L92)):*

`bool xx = false;` (https://stackoverflow.com/a/4622240/1255535) 
</blockquote></details>

Done.

yxlao (approved):
> Reviewed 3 of 5 files at r3.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2636)* status: 4 of 6 files reviewed, 2 unresolved discussions (waiting on @syncle, @theNded, and @yxlao)

___
*[cpp/open3d/pipelines/registration/Registration.h, line 190 at r2](https://reviewable.io/reviews/intel-isl/open3d/2636#-MMY9H-R-wu8GgcjlnFI:-MMYv5sZ-CxmlA0AhEn6:bq0ncok) ([raw file](https://github.com/intel-isl/open3d/blob/1730bcc83d84095d74ab7a8496284e914f1cd1b8/cpp/open3d/pipelines/registration/Registration.h#L190)):*

This was the legacy implementation.  If we want to remove `reference_wrapper`, we have to change the const `Check` method in all the inheritances of `CorrespondenceChecker` to non-const. We can do it in another PR after a redesign.
</blockquote></details>

sounds good
___
*[cpp/open3d/pipelines/registration/Registration.cpp, line 234 at r2](https://reviewable.io/reviews/intel-isl/open3d/2636#-MMYFypD4RSFUPZqaVxC:-MMYvA_97OQ-7mE9SQ4G:b-hiukit) ([raw file](https://github.com/intel-isl/open3d/blob/1730bcc83d84095d74ab7a8496284e914f1cd1b8/cpp/open3d/pipelines/registration/Registration.cpp#L234)):*

`continue` will be applied to `for checker` loop right?
</blockquote></details>

you're right, this looks good

cpp/open3d/pipelines/registration/Registration.cpp
```diff
@@ -194,152 +192,160 @@ RegistrationResult RegistrationRANSACBasedOnCorrespondence(
...
-        }      // end of for-loop
+            }  // if < exit_itr_local
+        }      // for loop
 #pragma omp critical
         {
-            if (result_private.fitness_ > result.fitness_ ||
-                (result_private.fitness_ == result.fitness_ &&
-                 result_private.inlier_rmse_ < result.inlier_rmse_)) {
-                result = result_private;
+            if (best_result_local.IsBetterRANSACThan(best_result)) {
```
sambarluc, line 259: In relation to the comment above, I'm not sure I get the value of identifying the best result thread-locally first, and then reducing it globally, in particular because the reduction here a rather check and operation, so you might just end up having idle threads, that could instead work on other samples in the loop above.

cpp/open3d/pipelines/registration/Registration.cpp
```diff
@@ -194,152 +192,160 @@ RegistrationResult RegistrationRANSACBasedOnCorrespondence(
...
+                best_result = best_result_local;
+            }
+            if (exit_itr_local > exit_itr) {
```
sambarluc, line 262: I also really don't get this. Why is the shared variable only updated if `exit_itr_local` is larger? Furthermore, is the logging below correct, in view of my comment above regarding the use of different stopping criteria among different threads?

theNded: In general, in this implementation, each thread runs its own allocated iterations (a portion of preset 10K iterations). The results are only synchronized at the end of their loop. So it is correct that each thread is using different criteria, and the final exit condition is a rough estimation. The best result, however, is the best we have.

We can use more complex blocking strategies to make results perfectly precise, but it will
1) introduce more complex design
2) make it much harder to debug
3) slightly harm the performance (due to repetitive synchronization)

That's the reason why I chose a simplified implementation: it gives us correct transformation results fast and easily, while providing a rough estimation of exit condition that is non-essential.

cpp/open3d/pipelines/registration/Registration.cpp
theNded: See below

cpp/open3d/pipelines/registration/Registration.cpp
```diff
@@ -194,152 +192,160 @@ RegistrationResult RegistrationRANSACBasedOnCorrespondence(
...
-                        finished_validation = true;
+
+                if (result.IsBetterRANSACThan(best_result_local)) {
+                    best_result_local = result;
+
+                    // Update exit condition if necessary
+                    double exit_itr_d =
+                            std::log(1.0 - criteria.confidence_) /
+                            std::log(1.0 - std::pow(result.fitness_, ransac_n));
+                    exit_itr_local =
```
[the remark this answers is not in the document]
theNded: See below

cpp/open3d/pipelines/registration/Registration.cpp
sambarluc: Thanks for your replies, I see your point.

cpp/open3d/pipelines/registration/Registration.cpp
```diff
@@ -194,152 +192,160 @@ RegistrationResult RegistrationRANSACBasedOnCorrespondence(
...
-                    if (total_validation >= criteria.max_validation_)
-                        finished_validation = true;
+
+                if (result.IsBetterRANSACThan(best_result_local)) {
+                    best_result_local = result;
+
+                    // Update exit condition if necessary
+                    double exit_itr_d =
+                            std::log(1.0 - criteria.confidence_) /
+                            std::log(1.0 - std::pow(result.fitness_, ransac_n));
```
sambarluc, line 249: One more thing, sorry. This can relatively easy overflow, and fitness can be zero, sometimes. Shouldn't there be some guard around it?

PR #2638, opened 2020-11-17
fix for implicit capture of this in cpp20

al-tu:
fix for compilation with -std=c++2a
`error: implicit capture of 'this' via '[=]' is deprecated in C++20 [-Werror=deprecated]`

related documents:
http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/p0409r2.html
http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0806r2.html

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2638)


al-tu:
seems like it's an backwards-incompatible change and a proper fix would be to just list everything that has to be captured by value explicitly

yxlao (approved):

yxlao:
thanks @al-tu

PR #2665, opened 2020-11-26
cmake-3.17 required

reyanshsolis:
* [Updated] CMake min requirement from 3.15 to 3.17 as required by faiss.
[3.18 is not supported by current CI Tests configuration]
* [Updated] CMake-3.18-aarch64 for arm64-filament-only CI, install wget

* [Modified] "Ubuntu 18.04 and below" to "Ubuntu 18.04" as the package is incompatible with previous versions of Ubuntu, as libc++7-dev required by the package is not available in previous versions of Ubuntu.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2665)

### DIFF
--- a/.github/workflows/openblas.yml
+++ b/.github/workflows/openblas.yml
@@ -127,6 +127,13 @@ jobs:
             python --version
             pip install pytest=="6.0.1" -U
             pip install wheel=="0.35.1" -U
+            # Get pre-compiled CMake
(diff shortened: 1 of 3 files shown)


yxlao (approved):
LGTM after updating cmake version in `docs/compilation.rst`

PR #2666, opened 2020-11-26
Simplify macos dependencies

yxlao:
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2666)


ssheorey (approved):
I understand we are using self-built versions of `tbb` and `libusb`, so its not necessary for the use to do a system install. So no pkg-config needed either. Looks good.

PR #2698, opened 2020-12-03
CI upgrades (caching to GCS, auto-cancel redundant)

ssheorey:
Documentation build scripts moved to ci_utils
Auto-cancel redundant workflows
**Removed**: [Add Python 3.9 wheel building and tests] Not supported by conda-build yet.
PRs only test Python 3.6
master tests Python 3.6, 3.7, 3.8
Use GCS bucket for ccache for openblas (arm64 container). Build is about 1:15hrs now instead of >5hrs.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2698)


yxlao (approved):
> Reviewed 2 of 6 files at r1, 4 of 5 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2698)* status: all files reviewed, 2 unresolved discussions (waiting on @germanros1987 and @ssheorey)

___
*[.github/workflows/openblas.yml, line 82 at r3](https://reviewable.io/reviews/intel-isl/open3d/2698#-MNez-1SE68_bAVMFKJR:-MNez-1SE68_bAVMFKJS:b-95nh03) ([raw file](https://github.com/intel-isl/open3d/blob/8c8ba0513b959c85194ea86f6bc81076b127917d/.github/workflows/openblas.yml#L82)):*
> ```YAML
>             ccache -s
> 
>   arm64-filament-only:
> ```

I added cache upload/download for the `arm64-filament-only` job, but do we still need this job? If the `arm64:` job can reliably run under 6 hours with the new caching mechanism, we can probably remove the `arm64-filament-only`.  

The only worst-case scenario is for some reason we lost all cache and the `arm64:` job had to start from scratch.
___
*[util/ci_utils.sh, line 408 at r3](https://reviewable.io/reviews/intel-isl/open3d/2698#-MNexuEjC8lV3u1svP62:-MNexuEjC8lV3u1svP63:b-gzqbba) ([raw file](https://github.com/intel-isl/open3d/blob/8c8ba0513b959c85194ea86f6bc81076b127917d/util/ci_utils.sh#L408)):*
> ```Shell
>         python -m pip install -r "${OPEN3D_ML_ROOT}/requirements-tensorflow.txt"
>     else
>         echo OPEN3D_ML_ROOT="$OPEN3D_ML_ROOT" not specified or invalid. Skipping ML dependencies.
> ```

nit: `echo "OPEN3D_ML_ROOT=$OPEN3D_ML_ROOT not specified or invalid. Skipping ML dependencies."` shall also work, shall we keep the double quotes for echo statements? same for the other qoutes

PR #2710, opened 2020-12-06
CUDA/CPU compatible TSDF integration

theNded:
This PR introduces volumetric TSDF integration and surface extraction supported by both CPU/CUDA. It also provides assistance functions including
- tgemeotry/legacy image/triangle mesh conversion
- point cloud from depth image
- ~~point cloud voxel downsample via hashmap~~

On CUDA, pure geometric integration time is ~2ms/frame. On CPU it is varying from 40ms to 120ms per frame.

TBD:
- [x] color integration
- [x] fix potential block allocation issues
- [x] fix normals in marching cubes
- [x] pybind and unit tests
- [x] merge hashmap
- [x] merge changes in tgeometry from tensorlist to tensor

Potential improvements:
- [x] Simplify nested `DISPATCH` and conditional compilation in `GeneralEWSharedImpl.h`
- [x] Add detailed instructions to `tintegration` examples, including hints for selecting `voxel_size` and `block_count` for extremely large scenes. Create a jupyter tutorial
- [x] Complete comments everywhere
- [x] Add simple and reproducible test cases

Pending potential bug:
- [ ] Integration is 1.5x slow when the voxel grid is in a different context.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2710)


theNded:
@yxlao I think somehow dispatch + conditional compilation does not work for Windows. Can please help with this?

yxlao:
@theNded yes, i am debugging on windows now

theNded:
> @theNded yes, i am debugging on windows now

Thanks! As a workaround we can roll back and separate CPU/CUDA implementation in different files, that will introduce a lot of duplicates though.

yxlao (changes requested):
Submitting part I of the review. 

Part II, to be reviewed:
```
GeneralEWCPU.cpp
GeneralEWCUDA.cu
GeneralEWMacros.h
GeneralEWSharedImpl.h
GeneralIndexer.h
TSDFVoxelGrid.cpp
```

> Reviewed 17 of 33 files at r1, 8 of 8 files at r2, 4 of 4 files at r3, 5 of 10 files at r4.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2710#-:-MOfeZcs4KBzxou981_V:b-27umbk)* status: 33 of 39 files reviewed, 25 unresolved discussions (waiting on @theNded and @yxlao)

___
*[cpp/open3d/core/hashmap/Hashmap.h, line 119 at r2](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOc2SS73f_9cZwfRR7j:-MOc2SS73f_9cZwfRR7k:bclhush) ([raw file](https://github.com/intel-isl/open3d/blob/6f1cd4ec2687edd615dde5f973a5fe4730519e04/cpp/open3d/core/hashmap/Hashmap.h#L119)):*
> ```objc
>     Hashmap Copy(const Device& device);
>     Hashmap CPU();
>     Hashmap CUDA();
> ```

do we need a device id?
___
*[cpp/open3d/core/kernel/CUDALauncher.cuh, line 149 at r4](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOf6Ia64L_K_ZbwtwJL:-MOf6Ia64L_K_ZbwtwJM:by2wb98) ([raw file](https://github.com/intel-isl/open3d/blob/8c617caa42a82e492375d2a51d9eac5891fddda0/cpp/open3d/core/kernel/CUDALauncher.cuh#L149)):*
> ```Cuda
>                 <<<grid_size, default_block_size, 0>>>(n, element_kernel);
>         OPEN3D_GET_LAST_CUDA_ERROR("LaunchIntegrateKernel failed.");
>     }
> ```

typo: `LaunchGeneralKernel failed` 
___
*[cpp/open3d/core/kernel/GeneralEW.h, line 32 at r4](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOfHnZaEQE0NsByzHUr:-MOfHnZaEQE0NsByzHUs:b-n12nl) ([raw file](https://github.com/intel-isl/open3d/blob/8c617caa42a82e492375d2a51d9eac5891fddda0/cpp/open3d/core/kernel/GeneralEW.h#L32)):*
> ```objc
> 
> #include "open3d/core/Tensor.h"
> #include "open3d/utility/Console.h"
> ```

console.h and hepler.h not required
___
*[cpp/open3d/core/kernel/GeneralEWSharedImpl.h, line 1 at r4](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOfUM190hCJe7jHek1l:-MOfUM190hCJe7jHek1m:bpcymo3) ([raw file](https://github.com/intel-isl/open3d/blob/8c617caa42a82e492375d2a51d9eac5891fddda0/cpp/open3d/core/kernel/GeneralEWSharedImpl.h#L1)):*
> ```objc
> // ----------------------------------------------------------------------------
> ```

future PR: consider renaming and moving this file
e.g. we can have the general launcher in `/kernel`, but the lambda functions to be launched can be in other locations
___
*[cpp/open3d/core/kernel/GeneralIndexer.h, line 47 at r4](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOfZHPR2sDrj94j8-E0:-MOfZHPR2sDrj94j8-E1:b-27b9vs) ([raw file](https://github.com/intel-isl/open3d/blob/8c617caa42a82e492375d2a51d9eac5891fddda0/cpp/open3d/core/kernel/GeneralIndexer.h#L47)):*
> ```objc
>                      const Tensor& extrinsic = Tensor::Eye(
>                              4, core::Dtype::Float32, core::Device("CPU:0")),
>                      float scale = 1.0f) {
> ```

Run `intrinsic.AssertShape` and `extrinsic.AssertShape`.
Run `intrinsic.AssertDtype` and `extrinsic.AssertDtype` since we are assuming them to be `float`
___
*[cpp/open3d/core/kernel/GeneralIndexer.h, line 115 at r4](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOfbDjP3O96J5a7SvMq:-MOfbDjP3O96J5a7SvMr:b4sdsm4) ([raw file](https://github.com/intel-isl/open3d/blob/8c617caa42a82e492375d2a51d9eac5891fddda0/cpp/open3d/core/kernel/GeneralIndexer.h#L115)):*
> ```objc
> };
> 
> /// Convert between ND coordinates and their corresponding linear offsets
> ```

We should probably mention this needs to be contiguous.
___
*[cpp/open3d/t/geometry/Image.cpp, line 81 at r4](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOfNzEM9Wu_rvWLf3gH:-MOfNzEM9Wu_rvWLf3gI:b-dmmmgh) ([raw file](https://github.com/intel-isl/open3d/blob/8c617caa42a82e492375d2a51d9eac5891fddda0/cpp/open3d/t/geometry/Image.cpp#L81)):*
> ```cpp
> 
>     if (image_legacy.IsEmpty()) {
>         return Image();
> ```

i would also give it the device for consistency `return Image(0, 0, 1, Float32, device);`
___
*[cpp/open3d/t/geometry/PointCloud.h, line 276 at r4](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOfV85gAga0MkfUv2UR:-MOfV85gAga0MkfUv2US:buxak92) ([raw file](https://github.com/intel-isl/open3d/blob/8c617caa42a82e492375d2a51d9eac5891fddda0/cpp/open3d/t/geometry/PointCloud.h#L276)):*
> ```objc
> 
>     /// Create a PointCloud from a depth image.
>     static PointCloud CreateFromDepthImage(
> ```

let's copy over the full docstrings from the legacy `pointcloud.h`
___
*[cpp/open3d/t/geometry/PointCloud.cpp, line 138 at r4](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOfcA8i8Qnn2teAZMJF:-MOfcA8i8Qnn2teAZMJG:b64uxkv) ([raw file](https://github.com/intel-isl/open3d/blob/8c617caa42a82e492375d2a51d9eac5891fddda0/cpp/open3d/t/geometry/PointCloud.cpp#L138)):*
> ```cpp
> }
> 
> /// Create a PointCloud from a depth image
> ```

no need for docstrings here
___
*[cpp/open3d/t/geometry/TSDFVoxelGrid.h, line 57 at r4](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOfJZQZ2z4lzjrOX8QM:-MOfJZQZ2z4lzjrOX8QN:b-gk8phr) ([raw file](https://github.com/intel-isl/open3d/blob/8c617caa42a82e492375d2a51d9eac5891fddda0/cpp/open3d/t/geometry/TSDFVoxelGrid.h#L57)):*
> ```objc
> /// Users may specialize their own channels that can be reinterpreted from the
> /// internal Tensor.
> class TSDFVoxelGrid {
> ```

Is it intentional that `TSDFVoxelGrid` does not inherit `Geometry`? 
___
*[cpp/open3d/t/geometry/TSDFVoxelGrid.h, line 93 at r4](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOfPfRz0TfeTZeh78Wo:-MOfPfRz0TfeTZeh78Wp:b6zq08g) ([raw file](https://github.com/intel-isl/open3d/blob/8c617caa42a82e492375d2a51d9eac5891fddda0/cpp/open3d/t/geometry/TSDFVoxelGrid.h#L93)):*
> ```objc
>     TriangleMesh ExtractSurfaceMesh();
> 
>     TSDFVoxelGrid Copy(const core::Device &device);
> ```

Add missing docstrings. Add `.` at the end of the docstrings in his header.
___
*[cpp/open3d/t/geometry/TSDFVoxelGrid.h, line 96 at r4](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOfPlGTDEBr8TAMeYmP:-MOfPlGTDEBr8TAMeYmQ:bwn4ljc) ([raw file](https://github.com/intel-isl/open3d/blob/8c617caa42a82e492375d2a51d9eac5891fddda0/cpp/open3d/t/geometry/TSDFVoxelGrid.h#L96)):*
> ```objc
> 
>     TSDFVoxelGrid CPU();
>     TSDFVoxelGrid CUDA();
> ```

shall we give it the device ID?
___
*[cpp/open3d/t/geometry/TSDFVoxelGrid.h, line 101 at r4](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOfQlJP4GicZAPuSboK:-MOfQlJP4GicZAPuSboL:b-ewvfo0) ([raw file](https://github.com/intel-isl/open3d/blob/8c617caa42a82e492375d2a51d9eac5891fddda0/cpp/open3d/t/geometry/TSDFVoxelGrid.h#L101)):*
> ```objc
> 
> protected:
>     std::unordered_map<std::string, core::Dtype> attr_dtype_map_;
> ```

move this after all member functions
___
*[cpp/open3d/t/geometry/TSDFVoxelGrid.h, line 103 at r4](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOfQ-sz93cboJ8G_Q5B:-MOfQ-sz93cboJ8G_Q5C:bm8e15t) ([raw file](https://github.com/intel-isl/open3d/blob/8c617caa42a82e492375d2a51d9eac5891fddda0/cpp/open3d/t/geometry/TSDFVoxelGrid.h#L103)):*
> ```objc
>     std::unordered_map<std::string, core::Dtype> attr_dtype_map_;
> 
>     std::shared_ptr<core::Hashmap> GetVoxelBlockHashmap() {
> ```

Remove? We can directly access block_hashmap_.
___
*[cpp/open3d/t/geometry/TSDFVoxelGrid.h, line 107 at r4](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOfQdU7CyCBYbUsoWEh:-MOfQdU7CyCBYbUsoWEi:b7gmi63) ([raw file](https://github.com/intel-isl/open3d/blob/8c617caa42a82e492375d2a51d9eac5891fddda0/cpp/open3d/t/geometry/TSDFVoxelGrid.h#L107)):*
> ```objc
>     }
> 
>     /// Return (active_entries, 27) with \addrs and \masks for radius (3)
> ```

what are the \addrs and \masks ?
___
*[cpp/open3d/t/geometry/TSDFVoxelGrid.cpp, line 59 at r4](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOfdFaJBfCnvJOPTUy9:-MOfdFaJBfCnvJOPTUyA:bmgymvr) ([raw file](https://github.com/intel-isl/open3d/blob/8c617caa42a82e492375d2a51d9eac5891fddda0/cpp/open3d/t/geometry/TSDFVoxelGrid.cpp#L59)):*
> ```cpp
> 
>     int64_t total_bytes = 0;
>     for (auto &kv : attr_dtype_map_) {
> ```

Can we just directly check the 3 keys without using the for loop? This can also be merged with the check if `attr_dtype_map_.count("tsdf") == 0 `.
___
*[cpp/open3d/t/geometry/TSDFVoxelGrid.cpp, line 116 at r4](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOfUWJx644Iqdj0YZQ2:-MOfUWJx644Iqdj0YZQ3:b-1bocc5) ([raw file](https://github.com/intel-isl/open3d/blob/8c617caa42a82e492375d2a51d9eac5891fddda0/cpp/open3d/t/geometry/TSDFVoxelGrid.cpp#L116)):*
> ```cpp
> void TSDFVoxelGrid::Integrate(const Image &depth,
>                               const Image &color,
>                               const core::Tensor &intrinsics,
> ```

One recurring theme is how we deal with device consistency of depth/color/intrinsics/extrinsics, this problem also applies to other algorithms that we implement. We can work on a more systematic approach in the future. 
___
*[cpp/pybind/t/geometry/tsdf_voxelgrid.cpp, line 54 at r4](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOfIsv56GbKnpfUZn0h:-MOfIsv56GbKnpfUZn0i:b-88d6s0) ([raw file](https://github.com/intel-isl/open3d/blob/8c617caa42a82e492375d2a51d9eac5891fddda0/cpp/pybind/t/geometry/tsdf_voxelgrid.cpp#L54)):*
> ```cpp
>             "voxel_size"_a = 3.0 / 512, "sdf_trunc"_a = 0.04,
>             "block_resolution"_a = 16, "block_count"_a = 100,
>             "device"_a = core::Device("CUDA:0"));
> ```

let's be consistent with c++ initial values, use `CPU:0`?
___
*[cpp/tests/t/geometry/TSDFVoxelGrid.cpp, line 41 at r4](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOfH0EY6jFDznfqfVZJ:-MOfH0EY6jFDznfqfVZK:b-vb2poc) ([raw file](https://github.com/intel-isl/open3d/blob/8c617caa42a82e492375d2a51d9eac5891fddda0/cpp/tests/t/geometry/TSDFVoxelGrid.cpp#L41)):*
> ```cpp
> 
> template <class T, int M, int N, int A>
> core::Tensor FromEigen(const Eigen::Matrix<T, M, N, A>& matrix) {
> ```

move to `cpp/open3d/core/EigenConverter.h` ?
___
*[examples/cpp/TIntegrateRGBD.cpp, line 1 at r4](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOfG0-P90AbJrQVfyv6:-MOfG0-P90AbJrQVfyv7:b-m27nif) ([raw file](https://github.com/intel-isl/open3d/blob/8c617caa42a82e492375d2a51d9eac5891fddda0/examples/cpp/TIntegrateRGBD.cpp#L1)):*
> ```cpp
> #include <fmt/format.h>
> ```

is this needed?
___
*[examples/cpp/TIntegrateRGBD.cpp, line 3 at r4](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOfFpfh-xZBOqNk7f8v:-MOfFpfh-xZBOqNk7f8w:bjn747g) ([raw file](https://github.com/intel-isl/open3d/blob/8c617caa42a82e492375d2a51d9eac5891fddda0/examples/cpp/TIntegrateRGBD.cpp#L3)):*
> ```cpp
> #include <fmt/format.h>
> 
> #include "open3d/Open3D.h"
> ```

Similar to other examples, let's only include the `Open3d.h` header. 
If headers are missing, include them in `Open3d.h.in`.
___
*[examples/cpp/TIntegrateRGBD.cpp, line 32 at r4](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOfFjVG-M92QfO96mjg:-MOfFjVG-M92QfO96mjh:blzjeo4) ([raw file](https://github.com/intel-isl/open3d/blob/8c617caa42a82e492375d2a51d9eac5891fddda0/examples/cpp/TIntegrateRGBD.cpp#L32)):*
> ```cpp
>     utility::LogInfo("     --camera_intrinsic [intrinsic_path]");
>     utility::LogInfo("     --device [CPU:0]");
>     utility::LogInfo("     --output_filename [mesh]");
> ```

this is not used, remove?
___
*[examples/python/reconstruction_system/tintegrate_scene.py, line 14 at r2](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOc0RyT9p6gwsWtiSsV:-MOc0RyT9p6gwsWtiSsW:b-2rtu0r) ([raw file](https://github.com/intel-isl/open3d/blob/6f1cd4ec2687edd615dde5f973a5fe4730519e04/examples/python/reconstruction_system/tintegrate_scene.py#L14)):*
> ```Python
> sys.path.append("../utility")
> from file import *
> sys.path.append(".")
> ```

- Get the current `pwd` and append relative path, this enables calling this python file from a directory.
    ```bash
    # previously, we have to
    cd examples/python/reconstruction_system
    python tintegrate_scene.py
    
    # now, this is also possible
    python examples/python/reconstruction_system/tintegrate_scene.py
    ```
- Prefer import explicit name instead of `import *`: `from file import read_poses_from_log, get_rgbd_file_lists`


The outcome looks something like this:
```python
import numpy as np
import time
import open3d as o3d
import sys
import os
import argparse
pwd = os.path.dirname(os.path.realpath(__file__))
sys.path.append(pwd)
sys.path.append(os.path.join(pwd, "..", "utility"))
from file import read_poses_from_log, get_rgbd_file_lists
```
___
*[examples/python/reconstruction_system/tintegrate_scene.py, line 24 at r2](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOc1kjC-1CstrzcS5JE:-MOc1kjC-1CstrzcS5JF:b2ew6sp) ([raw file](https://github.com/intel-isl/open3d/blob/6f1cd4ec2687edd615dde5f973a5fe4730519e04/examples/python/reconstruction_system/tintegrate_scene.py#L24)):*
> ```Python
>         type=str,
>         help=
>         'path to the dataset. It should contain 16bit depth images in a folder named depth/, and rgb images in a folder named color/ or rgb/'
> ```

use multi-line string to keep it under 80 columns, e.g.
```
str = 'This is a very '
      'long string that '
      'spans 3 lines.'
```

same for the other comments
___
*[python/test/t/geometry/test_voxelgrid.py, line 108 at r4](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOfG_xoDn5hWNO8rGeg:-MOfG_xoDn5hWNO8rGeh:bwe7zf6) ([raw file](https://github.com/intel-isl/open3d/blob/8c617caa42a82e492375d2a51d9eac5891fddda0/python/test/t/geometry/test_voxelgrid.py#L108)):*
> ```Python
> 
>     pcd = volume.extract_surface_points().to_legacy_pointcloud()
>     pcd_gt = o3d.io.read_point_cloud(test_data_path + "tsdf_pcd.ply")
> ```

shall we move this to the same folder of the colors and depths, e.g. something like `RGBD/color/example_tsdf_pcd.ply`?

theNded (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2710)* status: 33 of 39 files reviewed, 25 unresolved discussions (waiting on @yxlao)

___
*[cpp/open3d/core/hashmap/Hashmap.h, line 119 at r2](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOc2SS73f_9cZwfRR7j:-MOgvrwlCdu-P2uxPWV1:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/6f1cd4ec2687edd615dde5f973a5fe4730519e04/cpp/open3d/core/hashmap/Hashmap.h#L119)):*

do we need a device id?
</blockquote></details>

Done.
___
*[cpp/open3d/core/kernel/CUDALauncher.cuh, line 149 at r4](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOf6Ia64L_K_ZbwtwJL:-MOgwVvv3dTKt3qaySCU:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/8c617caa42a82e492375d2a51d9eac5891fddda0/cpp/open3d/core/kernel/CUDALauncher.cuh#L149)):*

typo: `LaunchGeneralKernel failed` 
</blockquote></details>

Done.
___
*[cpp/open3d/core/kernel/GeneralEW.h, line 32 at r4](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOfHnZaEQE0NsByzHUr:-MOgwb2sDdtgJ1XMHGSa:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/8c617caa42a82e492375d2a51d9eac5891fddda0/cpp/open3d/core/kernel/GeneralEW.h#L32)):*

console.h and hepler.h not required
</blockquote></details>

Done.
___
*[cpp/open3d/core/kernel/GeneralEWSharedImpl.h, line 1 at r4](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOfUM190hCJe7jHek1l:-MOgx5fk5Aa0uypxKpXe:b-se0k6z) ([raw file](https://github.com/intel-isl/open3d/blob/8c617caa42a82e492375d2a51d9eac5891fddda0/cpp/open3d/core/kernel/GeneralEWSharedImpl.h#L1)):*

future PR: consider renaming and moving this file
e.g. we can have the general launcher in `/kernel`, but the lambda functions to be launched can be in other locations
</blockquote></details>

Agreed, we should aggregate kernels by functionalities and separate them into different files.
___
*[cpp/open3d/core/kernel/GeneralIndexer.h, line 47 at r4](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOfZHPR2sDrj94j8-E0:-MOh3RGZCmjuuAqld7yf:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/8c617caa42a82e492375d2a51d9eac5891fddda0/cpp/open3d/core/kernel/GeneralIndexer.h#L47)):*

Run `intrinsic.AssertShape` and `extrinsic.AssertShape`.
Run `intrinsic.AssertDtype` and `extrinsic.AssertDtype` since we are assuming them to be `float`
</blockquote></details>

Done.
___
*[cpp/open3d/core/kernel/GeneralIndexer.h, line 115 at r4](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOfbDjP3O96J5a7SvMq:-MOh3x7p8UMpcxmp_Xp3:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/8c617caa42a82e492375d2a51d9eac5891fddda0/cpp/open3d/core/kernel/GeneralIndexer.h#L115)):*

We should probably mention this needs to be contiguous.
</blockquote></details>

Done.
___
*[cpp/open3d/t/geometry/Image.cpp, line 81 at r4](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOfNzEM9Wu_rvWLf3gH:-MOh4EDXCaSRxX4E4ZGe:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/8c617caa42a82e492375d2a51d9eac5891fddda0/cpp/open3d/t/geometry/Image.cpp#L81)):*

i would also give it the device for consistency `return Image(0, 0, 1, Float32, device);`
</blockquote></details>

Done.
___
*[cpp/open3d/t/geometry/PointCloud.h, line 276 at r4](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOfV85gAga0MkfUv2UR:-MOh4hI9Ce5S6EGtDAVP:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/8c617caa42a82e492375d2a51d9eac5891fddda0/cpp/open3d/t/geometry/PointCloud.h#L276)):*

let's copy over the full docstrings from the legacy `pointcloud.h`
</blockquote></details>

Done.
___
*[cpp/open3d/t/geometry/PointCloud.cpp, line 138 at r4](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOfcA8i8Qnn2teAZMJF:-MOh4r8L6vwaxch_bjs_:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/8c617caa42a82e492375d2a51d9eac5891fddda0/cpp/open3d/t/geometry/PointCloud.cpp#L138)):*

no need for docstrings here
</blockquote></details>

Done.
___
*[cpp/open3d/t/geometry/TSDFVoxelGrid.h, line 57 at r4](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOfJZQZ2z4lzjrOX8QM:-MOh4s6B6zu_qflKm92q:bl5njy3) ([raw file](https://github.com/intel-isl/open3d/blob/8c617caa42a82e492375d2a51d9eac5891fddda0/cpp/open3d/t/geometry/TSDFVoxelGrid.h#L57)):*

Is it intentional that `TSDFVoxelGrid` does not inherit `Geometry`? 
</blockquote></details>

Yes, as transformation and visualization are non-trivial. In the future, we may add it to the geometry class once we have all the peripheral helpers.
___
*[cpp/open3d/t/geometry/TSDFVoxelGrid.h, line 93 at r4](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOfPfRz0TfeTZeh78Wo:-MOh5c4t4md4sijTswL1:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/8c617caa42a82e492375d2a51d9eac5891fddda0/cpp/open3d/t/geometry/TSDFVoxelGrid.h#L93)):*

Add missing docstrings. Add `.` at the end of the docstrings in his header.
</blockquote></details>

Done.
___
*[cpp/open3d/t/geometry/TSDFVoxelGrid.h, line 96 at r4](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOfPlGTDEBr8TAMeYmP:-MOh5jik3raX5372lwAW:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/8c617caa42a82e492375d2a51d9eac5891fddda0/cpp/open3d/t/geometry/TSDFVoxelGrid.h#L96)):*

shall we give it the device ID?
</blockquote></details>

Done.
___
*[cpp/open3d/t/geometry/TSDFVoxelGrid.h, line 101 at r4](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOfQlJP4GicZAPuSboK:-MOh5wb31Oo5xUOpnKzX:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/8c617caa42a82e492375d2a51d9eac5891fddda0/cpp/open3d/t/geometry/TSDFVoxelGrid.h#L101)):*

move this after all member functions
</blockquote></details>

Done.
___
*[cpp/open3d/t/geometry/TSDFVoxelGrid.h, line 103 at r4](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOfQ-sz93cboJ8G_Q5B:-MOh5ysO9z9GQH9W7X11:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/8c617caa42a82e492375d2a51d9eac5891fddda0/cpp/open3d/t/geometry/TSDFVoxelGrid.h#L103)):*

Remove? We can directly access block_hashmap_.
</blockquote></details>

Done.
___
*[cpp/open3d/t/geometry/TSDFVoxelGrid.h, line 107 at r4](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOfQdU7CyCBYbUsoWEh:-MOh7XyQ1IWkS7rfjQbN:b-g14xu6) ([raw file](https://github.com/intel-isl/open3d/blob/8c617caa42a82e492375d2a51d9eac5891fddda0/cpp/open3d/t/geometry/TSDFVoxelGrid.h#L107)):*

what are the \addrs and \masks ?
</blockquote></details>

Added descriptions.
___
*[cpp/open3d/t/geometry/TSDFVoxelGrid.cpp, line 59 at r4](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOfdFaJBfCnvJOPTUy9:-MOh7tloE6IonT_mtFl-:b90oytr) ([raw file](https://github.com/intel-isl/open3d/blob/8c617caa42a82e492375d2a51d9eac5891fddda0/cpp/open3d/t/geometry/TSDFVoxelGrid.cpp#L59)):*

Can we just directly check the 3 keys without using the for loop? This can also be merged with the check if `attr_dtype_map_.count("tsdf") == 0 `.
</blockquote></details>

Potentially there can be additional properties (e.g. mask probability). This structure may be more open to extension.
___
*[cpp/open3d/t/geometry/TSDFVoxelGrid.cpp, line 116 at r4](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOfUWJx644Iqdj0YZQ2:-MOh869y4f-lUHLEiALH:b-pbdqz3) ([raw file](https://github.com/intel-isl/open3d/blob/8c617caa42a82e492375d2a51d9eac5891fddda0/cpp/open3d/t/geometry/TSDFVoxelGrid.cpp#L116)):*

One recurring theme is how we deal with device consistency of depth/color/intrinsics/extrinsics, this problem also applies to other algorithms that we implement. We can work on a more systematic approach in the future. 
</blockquote></details>

Agreed.
___
*[cpp/pybind/t/geometry/tsdf_voxelgrid.cpp, line 54 at r4](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOfIsv56GbKnpfUZn0h:-MOh8ET36OvZ43HpqPc6:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/8c617caa42a82e492375d2a51d9eac5891fddda0/cpp/pybind/t/geometry/tsdf_voxelgrid.cpp#L54)):*

let's be consistent with c++ initial values, use `CPU:0`?
</blockquote></details>

Done.
___
*[cpp/tests/t/geometry/TSDFVoxelGrid.cpp, line 41 at r4](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOfH0EY6jFDznfqfVZJ:-MOh8ziY0WWb74hJoNQU:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/8c617caa42a82e492375d2a51d9eac5891fddda0/cpp/tests/t/geometry/TSDFVoxelGrid.cpp#L41)):*

move to `cpp/open3d/core/EigenConverter.h` ?
</blockquote></details>

Done.
___
*[examples/cpp/TIntegrateRGBD.cpp, line 1 at r4](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOfG0-P90AbJrQVfyv6:-MOh9KY09wAiX-qOAN15:ba3dfob) ([raw file](https://github.com/intel-isl/open3d/blob/8c617caa42a82e492375d2a51d9eac5891fddda0/examples/cpp/TIntegrateRGBD.cpp#L1)):*

is this needed?
</blockquote></details>

Removed
___
*[examples/cpp/TIntegrateRGBD.cpp, line 3 at r4](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOfFpfh-xZBOqNk7f8v:-MOh9M6L6MuRx2hgAHGR:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/8c617caa42a82e492375d2a51d9eac5891fddda0/examples/cpp/TIntegrateRGBD.cpp#L3)):*

Similar to other examples, let's only include the `Open3d.h` header. 
If headers are missing, include them in `Open3d.h.in`.
</blockquote></details>

Done.
___
*[examples/cpp/TIntegrateRGBD.cpp, line 32 at r4](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOfFjVG-M92QfO96mjg:-MOh9S7c08JejH2JXBL2:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/8c617caa42a82e492375d2a51d9eac5891fddda0/examples/cpp/TIntegrateRGBD.cpp#L32)):*

this is not used, remove?
</blockquote></details>

Done.
___
*[examples/python/reconstruction_system/tintegrate_scene.py, line 14 at r2](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOc0RyT9p6gwsWtiSsV:-MOh9aFgAIJmXaF9lf0N:b-wy4xbp) ([raw file](https://github.com/intel-isl/open3d/blob/6f1cd4ec2687edd615dde5f973a5fe4730519e04/examples/python/reconstruction_system/tintegrate_scene.py#L14)):*

- Get the current `pwd` and append relative path, this enables calling this python file from a directory.
    ```bash
    # previously, we have to
    cd examples/python/reconstruction_system
    python tintegrate_scene.py
    
    # now, this is also possible
    python examples/python/reconstruction_system/tintegrate_scene.py
    ```
- Prefer import explicit name instead of `import *`: `from file import read_poses_from_log, get_rgbd_file_lists`


The outcome looks something like this:
```python
import numpy as np
import time
import open3d as o3d
import sys
import os
import argparse
pwd = os.path.dirname(os.path.realpath(__file__))
sys.path.append(pwd)
sys.path.append(os.path.join(pwd, "..", "utility"))
from file import read_poses_from_log, get_rgbd_file_lists
```
</blockquote></details>

I planned to rewrite this part when refactoring the python reconstruction system (including utilities). What do you think?
___
*[examples/python/reconstruction_system/tintegrate_scene.py, line 24 at r2](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOc1kjC-1CstrzcS5JE:-MOhAHKyBaoRyiJmdqOy:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/6f1cd4ec2687edd615dde5f973a5fe4730519e04/examples/python/reconstruction_system/tintegrate_scene.py#L24)):*

use multi-line string to keep it under 80 columns, e.g.
```
str = 'This is a very '
      'long string that '
      'spans 3 lines.'
```

same for the other comments
</blockquote></details>

Done.
___
*[python/test/t/geometry/test_voxelgrid.py, line 108 at r4](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOfG_xoDn5hWNO8rGeg:-MOhAtk_6VMtzks1N3NK:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/8c617caa42a82e492375d2a51d9eac5891fddda0/python/test/t/geometry/test_voxelgrid.py#L108)):*

shall we move this to the same folder of the colors and depths, e.g. something like `RGBD/color/example_tsdf_pcd.ply`?
</blockquote></details>

Done.

yxlao (changes requested):
> Reviewed 4 of 10 files at r4, 19 of 19 files at r5.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2710)* status: all files reviewed, 17 unresolved discussions (waiting on @theNded and @yxlao)

___
*[cpp/open3d/core/kernel/GeneralEWCPU.cpp, line 93 at r5](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOiNgxA3ZxyuDV50Vmk:-MOiNgxA3ZxyuDV50Vml:bkjhypq) ([raw file](https://github.com/intel-isl/open3d/blob/97be2937db6af2a56b2812a24b128f05c707ff1c/cpp/open3d/core/kernel/GeneralEWCPU.cpp#L93)):*
> ```cpp
> 
>     tbb::concurrent_unordered_set<Coord3i, Coord3iHash> set;
>     CPULauncher::LaunchGeneralKernel(n, [&](int64_t workload_idx) {
> ```

This looks similar to `tbb::parallel_for(0, n, [&](int64_t workload_idx){}` since the input does not require an indexer.

Maybe we can just call it `ParallelFor`? e.g. `CPULauncher::ParallelFor(0, n, lamba_func)`, same for the CUDA version.
___
*[cpp/open3d/core/kernel/GeneralEWCPU.cpp, line 98 at r5](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOiOmmL764XBFVFL9NH:-MOiOmmL764XBFVFL9NI:bazob9o) ([raw file](https://github.com/intel-isl/open3d/blob/97be2937db6af2a56b2812a24b128f05c707ff1c/cpp/open3d/core/kernel/GeneralEWCPU.cpp#L98)):*
> ```cpp
>         float z = pcd_ptr[3 * workload_idx + 2];
> 
>         int64_t xb_lo =
> ```

Looks like we'll be better off using `int` for `xb_lo`, `xb_hi`, etc. They will need to be eventually converted to `int`. 
___
*[cpp/open3d/core/kernel/GeneralEWCPU.cpp, line 127 at r5](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOiPcVM0L1AdxkTuNb7:-MOiPcVM0L1AdxkTuNb8:b-9vapi6) ([raw file](https://github.com/intel-isl/open3d/blob/97be2937db6af2a56b2812a24b128f05c707ff1c/cpp/open3d/core/kernel/GeneralEWCPU.cpp#L127)):*
> ```cpp
>     for (auto it = set.begin(); it != set.end(); ++it, ++count) {
>         int64_t offset = count * 3;
>         block_coords_ptr[offset + 0] = static_cast<int>((*it).x_);
> ```

`it->x_` would be simpler.
___
*[cpp/open3d/core/kernel/GeneralEWSharedImpl.h, line 43 at r5](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOifzFhACTmvxdQIe89:-MOifzFhACTmvxdQIe8A:b-ubuf14) ([raw file](https://github.com/intel-isl/open3d/blob/97be2937db6af2a56b2812a24b128f05c707ff1c/cpp/open3d/core/kernel/GeneralEWSharedImpl.h#L43)):*
> ```objc
> namespace kernel {
> 
> #if defined(BUILD_CUDA_MODULE) && defined(__CUDACC__)
> ```

my bad, we can move this out of the namespaces since macros don't apply namespaces; and maybe call it `OPEN3D_ATOMIC_ADD` to avoid potential collisions with other libraries.
___
*[cpp/open3d/core/kernel/GeneralEWSharedImpl.h, line 77 at r5](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOihDYREII4ISzGm0B7:-MOihDYREII4ISzGm0B8:b-cx4juu) ([raw file](https://github.com/intel-isl/open3d/blob/97be2937db6af2a56b2812a24b128f05c707ff1c/cpp/open3d/core/kernel/GeneralEWSharedImpl.h#L77)):*
> ```objc
> /// 12-byte voxel structure.
> /// uint16_t for colors and weights, sacrifices minor accuracy but saves memory.
> OPEN3D_DEVICE const float kColorFactor = 255.0f;
> ```

we can move both inside to the class
```
static constexpr uint16_t kMaxUint16 = 65535;
static constexpr float kColorFactor = 255.0f;
```
___
*[cpp/open3d/core/kernel/GeneralEWSharedImpl.h, line 92 at r5](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOiiPO29BnFjmtIUM6R:-MOiiPO29BnFjmtIUM6S:b-4ktbrp) ([raw file](https://github.com/intel-isl/open3d/blob/97be2937db6af2a56b2812a24b128f05c707ff1c/cpp/open3d/core/kernel/GeneralEWSharedImpl.h#L92)):*
> ```objc
>     OPEN3D_HOST_DEVICE float GetWeight() { return static_cast<float>(weight); }
>     OPEN3D_HOST_DEVICE float GetR() {
>         return static_cast<float>(r / kColorFactor);
> ```

If we use `uint16_t` for colors, could you comment on why the `kColorFactor` is 255? (i.e. 255 is more related to`uint8_t` in traditional cases)
___
*[cpp/open3d/core/kernel/GeneralEWSharedImpl.h, line 162 at r5](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOijMPgBjZlZ3nE38Ms:-MOijMPgBjZlZ3nE38Mt:br4ijwr) ([raw file](https://github.com/intel-isl/open3d/blob/97be2937db6af2a56b2812a24b128f05c707ff1c/cpp/open3d/core/kernel/GeneralEWSharedImpl.h#L162)):*
> ```objc
> };
> 
> #define DISPATCH_BYTESIZE_TO_VOXEL(BYTESIZE, ...)            \
> ```

i would also move this to the top, since macros are not bound to a namespace
___
*[cpp/open3d/core/kernel/GeneralEWSharedImpl.h, line 199 at r5](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOiqcJoDMeVE_7FRoDE:-MOiqcJoDMeVE_7FRoDF:b-7vvms8) ([raw file](https://github.com/intel-isl/open3d/blob/97be2937db6af2a56b2812a24b128f05c707ff1c/cpp/open3d/core/kernel/GeneralEWSharedImpl.h#L199)):*
> ```objc
>     int64_t nb_idx = (dxb + 1) + (dyb + 1) * 3 + (dzb + 1) * 9;
> 
>     bool block_mask_i = *static_cast<bool*>(
> ```

some ideas for refactoring, the `static_cast` happens quite often in this code

if we add:

```cpp
    template <typename T>
    inline OPEN3D_HOST_DEVICE T GetValueFromCoord(int64_t x, int64_t y) const {
        int64_t workload;
        CoordToWorkload(x, y, &workload);
        return *(reinterpret_cast<T*>(static_cast<uint8_t*>(ptr_) +
                                      workload * element_byte_size_));
    }
```

then we can directly do 
```cpp
bool block_mask_i = nb_block_masks_indexer.GetValueFromCoord<bool>(curr_block_idx, nb_idx);
```

(this is optional, can be added in future PR if appropriate)
___
*[cpp/open3d/core/kernel/GeneralIndexer.h, line 149 at r5](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOiblWp0yNmlNnK1uyN:-MOiblWp0yNmlNnK1uyO:b-1ycup1) ([raw file](https://github.com/intel-isl/open3d/blob/97be2937db6af2a56b2812a24b128f05c707ff1c/cpp/open3d/core/kernel/GeneralIndexer.h#L149)):*
> ```objc
> 
>         SizeVector shape = ndarray.GetShape();
>         int64_t n = static_cast<int64_t>(shape.size());
> ```

`int n = ndarray.NumDims();` may be simpler.
___
*[cpp/open3d/core/kernel/GeneralIndexer.h, line 171 at r5](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOibvCL5iqaDeZWMfeo:-MOibvCL5iqaDeZWMfep:b-1ycup1) ([raw file](https://github.com/intel-isl/open3d/blob/97be2937db6af2a56b2812a24b128f05c707ff1c/cpp/open3d/core/kernel/GeneralIndexer.h#L171)):*
> ```objc
> 
>     /// Only used for simple shapes
>     NDArrayIndexer(const SizeVector& shape) {
> ```

`int n = ndarray.NumDims();` may be simpler.
___
*[cpp/open3d/t/geometry/TSDFVoxelGrid.cpp, line 59 at r4](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOfdFaJBfCnvJOPTUy9:-MOi0lbs4qRtp6ApXLa9:bbzaf3r) ([raw file](https://github.com/intel-isl/open3d/blob/8c617caa42a82e492375d2a51d9eac5891fddda0/cpp/open3d/t/geometry/TSDFVoxelGrid.cpp#L59)):*

Potentially there can be additional properties (e.g. mask probability). This structure may be more open to extension.
</blockquote></details>

yeah, but is the for loop providing some additional benefits?

e.g. to detect `mask_probability`, these seems to be equivalent:

```cpp
for (k, v in map) {
    if (k == tsdf) {do_something();}
    if (k == weight) {do_something();}
    if (k == color) {do_something();}
    if (k == mask_probability) {do_something();}
}
```

```cpp
if (tsdf in map) {do_something();}
if (weight in map) {do_something();}
if (color in map) {do_something();}
if (mask_probability in map) {do_something();}
```
___
*[cpp/open3d/t/geometry/TSDFVoxelGrid.cpp, line 284 at r5](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOif86X0kwrKdvLH7_v:-MOif86X0kwrKdvLH7_w:b-lu24dh) ([raw file](https://github.com/intel-isl/open3d/blob/97be2937db6af2a56b2812a24b128f05c707ff1c/cpp/open3d/t/geometry/TSDFVoxelGrid.cpp#L284)):*
> ```cpp
> 
> TSDFVoxelGrid TSDFVoxelGrid::Copy(const core::Device &device) {
>     TSDFVoxelGrid cpu_tsdf_voxelgrid(attr_dtype_map_, voxel_size_, sdf_trunc_,
> ```

`device_tsdf_voxelgrid`?
___
*[cpp/open3d/t/geometry/TSDFVoxelGrid.cpp, line 299 at r5](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOifDbE2lNtzkWDYY5j:-MOifDbE2lNtzkWDYY5k:bnja8p8) ([raw file](https://github.com/intel-isl/open3d/blob/97be2937db6af2a56b2812a24b128f05c707ff1c/cpp/open3d/t/geometry/TSDFVoxelGrid.cpp#L299)):*
> ```cpp
> 
> TSDFVoxelGrid TSDFVoxelGrid::CUDA(int device_id) {
>     if (GetDevice().GetType() == core::Device::DeviceType::CUDA) {
> ```

guess in this case it should be
```
if (GetDevice() == core::Device("CUDA", device_id))
```

theNded (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2710)* status: 37 of 42 files reviewed, 17 unresolved discussions (waiting on @yxlao)

___
*[cpp/open3d/core/kernel/GeneralEWCPU.cpp, line 93 at r5](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOiNgxA3ZxyuDV50Vmk:-MOixsND7DSDvpfwoUTj:bu1p19q) ([raw file](https://github.com/intel-isl/open3d/blob/97be2937db6af2a56b2812a24b128f05c707ff1c/cpp/open3d/core/kernel/GeneralEWCPU.cpp#L93)):*

This looks similar to `tbb::parallel_for(0, n, [&](int64_t workload_idx){}` since the input does not require an indexer.

Maybe we can just call it `ParallelFor`? e.g. `CPULauncher::ParallelFor(0, n, lamba_func)`, same for the CUDA version.


</blockquote></details>

True, let's refactor kernel launcher in a separate PR.
___
*[cpp/open3d/core/kernel/GeneralEWCPU.cpp, line 98 at r5](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOiOmmL764XBFVFL9NH:-MOiyBeMCQkrUnDAheY4:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/97be2937db6af2a56b2812a24b128f05c707ff1c/cpp/open3d/core/kernel/GeneralEWCPU.cpp#L98)):*

Looks like we'll be better off using `int` for `xb_lo`, `xb_hi`, etc. They will need to be eventually converted to `int`. 
</blockquote></details>

Done.
___
*[cpp/open3d/core/kernel/GeneralEWCPU.cpp, line 127 at r5](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOiPcVM0L1AdxkTuNb7:-MOiyHV-CL7jL8JkaWIG:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/97be2937db6af2a56b2812a24b128f05c707ff1c/cpp/open3d/core/kernel/GeneralEWCPU.cpp#L127)):*

`it->x_` would be simpler.
</blockquote></details>

Done.
___
*[cpp/open3d/core/kernel/GeneralEWSharedImpl.h, line 43 at r5](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOifzFhACTmvxdQIe89:-MOiynZqCMdORwrTwQmR:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/97be2937db6af2a56b2812a24b128f05c707ff1c/cpp/open3d/core/kernel/GeneralEWSharedImpl.h#L43)):*

my bad, we can move this out of the namespaces since macros don't apply namespaces; and maybe call it `OPEN3D_ATOMIC_ADD` to avoid potential collisions with other libraries.
</blockquote></details>

Done.
___
*[cpp/open3d/core/kernel/GeneralEWSharedImpl.h, line 77 at r5](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOihDYREII4ISzGm0B7:-MOiyvAA60jrbAONfMSc:bal3hwi) ([raw file](https://github.com/intel-isl/open3d/blob/97be2937db6af2a56b2812a24b128f05c707ff1c/cpp/open3d/core/kernel/GeneralEWSharedImpl.h#L77)):*

we can move both inside to the class
```
static constexpr uint16_t kMaxUint16 = 65535;
static constexpr float kColorFactor = 255.0f;
```
</blockquote></details>

Static float variable cannot be initialized in class; initializing a static function outside a class will cause compilation issues from nvcc. So this is the simplest solution here afaik.
___
*[cpp/open3d/core/kernel/GeneralEWSharedImpl.h, line 92 at r5](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOiiPO29BnFjmtIUM6R:-MOiziJq1nJVwdiwbGtQ:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/97be2937db6af2a56b2812a24b128f05c707ff1c/cpp/open3d/core/kernel/GeneralEWSharedImpl.h#L92)):*

If we use `uint16_t` for colors, could you comment on why the `kColorFactor` is 255? (i.e. 255 is more related to`uint8_t` in traditional cases)
</blockquote></details>

Done.
___
*[cpp/open3d/core/kernel/GeneralEWSharedImpl.h, line 162 at r5](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOijMPgBjZlZ3nE38Ms:-MOj-ADw87wG5nS-8CSZ:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/97be2937db6af2a56b2812a24b128f05c707ff1c/cpp/open3d/core/kernel/GeneralEWSharedImpl.h#L162)):*

i would also move this to the top, since macros are not bound to a namespace
</blockquote></details>

Done.
___
*[cpp/open3d/core/kernel/GeneralEWSharedImpl.h, line 199 at r5](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOiqcJoDMeVE_7FRoDE:-MOj0FD089LI2L7fgxJg:b-xx28b0) ([raw file](https://github.com/intel-isl/open3d/blob/97be2937db6af2a56b2812a24b128f05c707ff1c/cpp/open3d/core/kernel/GeneralEWSharedImpl.h#L199)):*

some ideas for refactoring, the `static_cast` happens quite often in this code

if we add:

```cpp
    template <typename T>
    inline OPEN3D_HOST_DEVICE T GetValueFromCoord(int64_t x, int64_t y) const {
        int64_t workload;
        CoordToWorkload(x, y, &workload);
        return *(reinterpret_cast<T*>(static_cast<uint8_t*>(ptr_) +
                                      workload * element_byte_size_));
    }
```

then we can directly do 
```cpp
bool block_mask_i = nb_block_masks_indexer.GetValueFromCoord<bool>(curr_block_idx, nb_idx);
```

(this is optional, can be added in future PR if appropriate)
</blockquote></details>

Yes I considered this too. We can refactor indexer along with general kernel launchers in the next PR.
___
*[cpp/open3d/core/kernel/GeneralIndexer.h, line 149 at r5](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOiblWp0yNmlNnK1uyN:-MOj-lD3BlogcehLdwiH:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/97be2937db6af2a56b2812a24b128f05c707ff1c/cpp/open3d/core/kernel/GeneralIndexer.h#L149)):*

`int n = ndarray.NumDims();` may be simpler.
</blockquote></details>

Done.
___
*[cpp/open3d/core/kernel/GeneralIndexer.h, line 171 at r5](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOibvCL5iqaDeZWMfeo:-MOj-u1IAxG4M4qTU4lk:b-ixsu3p) ([raw file](https://github.com/intel-isl/open3d/blob/97be2937db6af2a56b2812a24b128f05c707ff1c/cpp/open3d/core/kernel/GeneralIndexer.h#L171)):*

`int n = ndarray.NumDims();` may be simpler.
</blockquote></details>

This constructor does not take ndarray as an input, so we have to use the dirty version.
___
*[cpp/open3d/t/geometry/TSDFVoxelGrid.cpp, line 59 at r4](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOfdFaJBfCnvJOPTUy9:-MOj139l-oY8qxBJz3zY:b-1ie0hk) ([raw file](https://github.com/intel-isl/open3d/blob/8c617caa42a82e492375d2a51d9eac5891fddda0/cpp/open3d/t/geometry/TSDFVoxelGrid.cpp#L59)):*

yeah, but is the for loop providing some additional benefits?

e.g. to detect `mask_probability`, these seems to be equivalent:

```cpp
for (k, v in map) {
    if (k == tsdf) {do_something();}
    if (k == weight) {do_something();}
    if (k == color) {do_something();}
    if (k == mask_probability) {do_something();}
}
```

```cpp
if (tsdf in map) {do_something();}
if (weight in map) {do_something();}
if (color in map) {do_something();}
if (mask_probability in map) {do_something();}
```
</blockquote></details>

My bad, there should be `total_bytes += dtype.ByteSize()` for other cases, but this seems less reasonable as the user may also define it by themselves. Changed.
___
*[cpp/open3d/t/geometry/TSDFVoxelGrid.cpp, line 284 at r5](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOif86X0kwrKdvLH7_v:-MOj0ZbS9aXJ9aHS-1d_:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/97be2937db6af2a56b2812a24b128f05c707ff1c/cpp/open3d/t/geometry/TSDFVoxelGrid.cpp#L284)):*

`device_tsdf_voxelgrid`?
</blockquote></details>

Done.
___
*[cpp/open3d/t/geometry/TSDFVoxelGrid.cpp, line 299 at r5](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOifDbE2lNtzkWDYY5j:-MOj0mvy-mGLOp5l6KeX:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/97be2937db6af2a56b2812a24b128f05c707ff1c/cpp/open3d/t/geometry/TSDFVoxelGrid.cpp#L299)):*

guess in this case it should be
```
if (GetDevice() == core::Device("CUDA", device_id))
```
</blockquote></details>

Done.

yxlao (approved):
> Reviewed 5 of 5 files at r6.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2710)* status: all files reviewed, 1 unresolved discussion (waiting on @theNded)

___
*[cpp/open3d/core/kernel/GeneralEWSharedImpl.h, line 77 at r5](https://reviewable.io/reviews/intel-isl/open3d/2710#-MOihDYREII4ISzGm0B7:-MOj6wHTCu2_nFXWm6Ua:b6ti55h) ([raw file](https://github.com/intel-isl/open3d/blob/97be2937db6af2a56b2812a24b128f05c707ff1c/cpp/open3d/core/kernel/GeneralEWSharedImpl.h#L77)):*

Static float variable cannot be initialized in class; initializing a static function outside a class will cause compilation issues from nvcc. So this is the simplest solution here afaik.
</blockquote></details>

right, tested `constexpr float` instead of `const float`, that former one works inside the class

PR #2711, opened 2020-12-06
GCE GPU CI fixes

ssheorey:
Fixes for:

1. Github glitch can prevent GCE VM deletion: Auto delete GCE VM after timeout (set to 1.5 hours).
2. CUDA driver and library version mismatch due to updated drivers from NVIDIA. `apt` would install multiple CUDA library versions. Updated NVIDIA driver version from 440 to 455. New base VM image created on GCE: `open3d-gpu-ci-base-20201228`. C++ tests still ran fine, though.
3. Python tests on GPU:  Separate Python build and run test step in CI. Added Open3D-ML to docker image. Python wheel is transferred from GHA to GCE for Python tests on bionic and focal.
4. Remove redundant `make install-pip-package` from `build_all`. This is tested in `build_wheel`.
5. CI configurations appear with names in GHA for easier debugging: `['2-bionic', '3-ML-bionic', '4-ML-RPC-bionic', '5-ML-RPC-focal']`
6. Dockerfile maintainer changed to `open3d-info@osvf.org`. Is there a better email for Open3D?

This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2711)

### DIFF
--- a/util/docker/open3d-gpu/scripts/gce-ubuntu-docker-run.sh
+++ b/util/docker/open3d-gpu/scripts/gce-ubuntu-docker-run.sh
@@ -13,6 +13,7 @@ set -e
 # GITHUB_SHA
 ## CI test matrix:
 CI_CONFIG_ID=${CI_CONFIG_ID:=0}
+CI_CONFIG_ID=${CI_CONFIG_ID%%-*} # '3-ML-bionic' -> '3'
 
 # CI configuration specification
 SHARED=(OFF ON OFF ON OFF OFF)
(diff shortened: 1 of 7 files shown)


util/docker/open3d-gpu/scripts/gce-ubuntu-docker-run.sh
```diff
@@ -49,6 +49,7 @@ GCE_VM_BASE_OS=ubuntu20.04
 GCE_VM_IMAGE_SPEC=(--image-project=ubuntu-os-cloud --image-family=ubuntu-2004-lts)
 GCE_VM_CUSTOM_IMAGE_FAMILY=ubuntu-os-docker-gpu-2004-lts
 VM_IMAGE=open3d-gpu-ci-base-$(date +%Y%m%d)
+GCE_CI_TIMEOUT=720 # Self delete VM after timeout
```
yxlao, line 52: Could we add some comments on what does this timeout value control? Is the 720 (second?) sufficient for our purpose?

yxlao (approved):

util/docker/open3d-gpu/scripts/gce-ubuntu-docker-run.sh
ssheorey: Still testing this, so set the timeout at 720s. Didn't work, so it's actually not ready to be merged.
Finally, I'll set the timeout at 7200s (2 hours).

PR #2744, opened 2020-12-11
fix undefined reference in Debug build with gcc

benjaminum:
This PR fixes a link error in debug builds with gcc

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2744)


yxlao (approved):
Verified with `gcc (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0` debug build.

PR #2746, opened 2020-12-12
Better picking, fixes crash removing last selection set

prewettg:
Picking examines nearby points and takes the point with the highest score. Point clouds need to have a smaller range than meshes.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2746)

### DIFF
--- a/cpp/open3d/visualization/gui/PickPointsInteractor.cpp
+++ b/cpp/open3d/visualization/gui/PickPointsInteractor.cpp
@@ -26,6 +26,8 @@
 
 #include "open3d/visualization/gui/PickPointsInteractor.h"
 
+#include <unordered_map>
+
 #include "open3d/geometry/Image.h"
 #include "open3d/geometry/PointCloud.h"
 #include "open3d/geometry/TriangleMesh.h"
@@ -37,8 +39,6 @@
 #include "open3d/visualization/rendering/Open3DScene.h"
 #include "open3d/visualization/rendering/Scene.h"
 #include "open3d/visualization/rendering/View.h"
-
-#include <unordered_map>
 
 #define WANT_DEBUG_IMAGE 0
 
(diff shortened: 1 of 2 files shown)


errissa (changes requested):
> Reviewed 2 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2746)* status: all files reviewed, 1 unresolved discussion (waiting on @prewettg)

___
*[cpp/open3d/visualization/gui/PickPointsInteractor.cpp, line 318 at r1](https://reviewable.io/reviews/intel-isl/open3d/2746#-MOXTnl1ACE927pFHYFb:-MOXTnl2BTQ_AVFTJvIu:btm3h7h) ([raw file](https://github.com/intel-isl/open3d/blob/f72b011962fdf481d185179b851a0e35a0e55b54/cpp/open3d/visualization/gui/PickPointsInteractor.cpp#L318)):*
> ```cpp
>         indices.clear();
>         if (x1 - x0 == 1 && y1 - y0 == 1) {
>             struct Score {
> ```

What is the reason for using a structure here? Is there a possibility that the score becomes more complex than a single float value?

prewettg (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2746)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @errissa)

___
*[cpp/open3d/visualization/gui/PickPointsInteractor.cpp, line 318 at r1](https://reviewable.io/reviews/intel-isl/open3d/2746#-MOXTnl1ACE927pFHYFb:-MOXclV8AJKkEmi2CAyG:b-qolqfm) ([raw file](https://github.com/intel-isl/open3d/blob/f72b011962fdf481d185179b851a0e35a0e55b54/cpp/open3d/visualization/gui/PickPointsInteractor.cpp#L318)):*

What is the reason for using a structure here? Is there a possibility that the score becomes more complex than a single float value?
</blockquote></details>

No, it's there to enforce that the value is set to a known default (0), but I forgot the part that actually sets the value! Fixed in latest commit.

errissa (approved):
> Reviewed 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2746)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2767, opened 2020-12-15
[Do not merge] [Benchmarking NNS]

reyanshsolis:
## Benchmark HybridSearch NNS
To run test time for HybridSearch NNS, check out this branch `rishabh/BenchmarksNNS`
from the `Open3D directory` run the example on CUDA and CPU using the following respective command,
``` 
CUDA:
./build/bin/examples/BenchmarkNNS CUDA:0 examples/test_data/ICP/cloud_bin_0.pcd examples/test_data/ICP/cloud_bin_1.pcd
CPU:
./build/bin/examples/BenchmarkNNS CPU:0 examples/test_data/ICP/cloud_bin_0.pcd examples/test_data/ICP/cloud_bin_1.pcd
```
______________________________________________________________
Correspondences: (coming out to be different on CPU and GPU)
|                 | Tensor CPU | Tensor GPU |
| --------------- | ---------- | ---------- |
| Identity        | 7552       | 7550       |
| Some good guess | 34741      | 34738      |

______________________________________________________________
Screenshot result

Screenshot from 2020-12-16 03-02-46

______________________________________________________________

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2767)


junha-l:
@yxlao 
It turns out that the **matching time** of gpu knn search could be slower than cpu knn search if the dataset size and query size is larger than some threshold.
If we don't care about the **prepare time** (e.g. kdtree construction time) or we query multiple times with the same dataset and query data, cpu knn search is more efficient.

The execution time of gpu & cpu hybrid search time are like below.
I used Intel(R) Core(TM) i7-9700KF CPU and GeForce RTX 2080 Ti GPU.
image

Another idea to solve this issue is modifying the algorithm of the hybrid search.
We have implemented the hybrid search as knn search followed by radius thresholding.
Instead, we can try fixed radius search followed by knn thresholding. 
Since the benjamin's implementation efficiently handle large data, we might reduce the execution time

PR #2781, opened 2020-12-17
Add 3D Labels to SceneWidget

errissa:
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2781)


prewettg (changes requested):
> Reviewed 6 of 8 files at r1, 1 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2781)* status: all files reviewed, 5 unresolved discussions (waiting on @errissa)

___
*[cpp/open3d/visualization/gui/Label3D.h, line 36 at r2](https://reviewable.io/reviews/intel-isl/open3d/2781#-MOmG-Q12Ks1za5XireH:-MOmG-Q12Ks1za5XireI:b56wtdn) ([raw file](https://github.com/intel-isl/open3d/blob/43d5385fd71847660e4a161b7a79e7e8b69b5057/cpp/open3d/visualization/gui/Label3D.h#L36)):*
> ```objc
> namespace gui {
> 
> class Label3D {
> ```

I think we need some documentation (for the class? for the constructor?) saying that this is intended to be used with SceneWidget. But, also, see below.
___
*[cpp/open3d/visualization/gui/SceneWidget.cpp, line 894 at r2](https://reviewable.io/reviews/intel-isl/open3d/2781#-MOmGIyf7S9d5ndtPvca:-MOmGIyf7S9d5ndtPvcb:b-knm0xs) ([raw file](https://github.com/intel-isl/open3d/blob/43d5385fd71847660e4a161b7a79e7e8b69b5057/cpp/open3d/visualization/gui/SceneWidget.cpp#L894)):*
> ```cpp
> }
> 
> std::shared_ptr<Label3D> SceneWidget::AddLabel(const Eigen::Vector3f& pos,
> ```

This makes sense, but I think it would be better for our long-term vision of a scene graph to add the labels to Open3DScene and have SceneWidget::Draw() read the labels from Open3DScene. This way Open3DScene still stays the "model" and SceneWidget the "view"/"controller".
___
*[cpp/open3d/visualization/gui/SceneWidget.cpp, line 959 at r2](https://reviewable.io/reviews/intel-isl/open3d/2781#-MOmITYg9PzXx1Ue4nA_:-MOmITYg9PzXx1Ue4nAa:b-x718xd) ([raw file](https://github.com/intel-isl/open3d/blob/43d5385fd71847660e4a161b7a79e7e8b69b5057/cpp/open3d/visualization/gui/SceneWidget.cpp#L959)):*
> ```
>             ndc.x() *= f.width;
>             ndc.y() *= f.height;
> ```
Did you try this on Linux? I could be wrong, but I think you'll need to offset by f.x and f.y, in case the SceneWidget is not at (0, 0). On Windows/Linux the menu offset the SceneWidget so that it is definitely not at (0, 0).
___
*[cpp/open3d/visualization/gui/SceneWidget.cpp, line 968 at r2](https://reviewable.io/reviews/intel-isl/open3d/2781#-MOmGvlP-JS3cKDyPK3D:-MOmGvlP-JS3cKDyPK3E:btb6zg3) ([raw file](https://github.com/intel-isl/open3d/blob/43d5385fd71847660e4a161b7a79e7e8b69b5057/cpp/open3d/visualization/gui/SceneWidget.cpp#L968)):*
> ```cpp
>         }
> 
>         ImGui::End();
> ```

This will render the labels with all the UI elements. Conceptually it seems like these ought to be with the 3D elements (although pragmatically that seems like maybe more effort than it's worth just for that). But we could potentially have z-order issues, if you have multiple SceneWidgets in layouts. Although since SceneWidgets have problems with z-order currently anyway, I guess we can leave it for now unless it's easy to do or something.
___
*[cpp/open3d/visualization/rendering/Camera.h, line 109 at r2](https://reviewable.io/reviews/intel-isl/open3d/2781#-MOmHjFz8JYLHBwp51aK:-MOmHjFz8JYLHBwp51aL:b-4w8zrc) ([raw file](https://github.com/intel-isl/open3d/blob/43d5385fd71847660e4a161b7a79e7e8b69b5057/cpp/open3d/visualization/rendering/Camera.h#L109)):*
> ```objc
>     virtual Transform GetProjectionMatrix() const = 0;
> 
>     virtual Eigen::Vector2f GetNDC(const Eigen::Vector3f& pt) const = 0;
> ```

Can we change this to a more descriptive name? It took me a while to work out what it was (especially since "DC" frequently means "Device Context" in drawing code).  GetNormalizedDeviceCoordinates() would be an obvious choice, but that assumes the reader understands what filament means by "normalized device coordinates". How would `ConvertToNormalizedViewCoords()` be? Or `ConvertTo2DCoord()`?

errissa (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2781)* status: all files reviewed, 5 unresolved discussions (waiting on @prewettg)

___
*[cpp/open3d/visualization/gui/Label3D.h, line 36 at r2](https://reviewable.io/reviews/intel-isl/open3d/2781#-MOmG-Q12Ks1za5XireH:-MOnzU8W8hjgOl37r1hL:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/43d5385fd71847660e4a161b7a79e7e8b69b5057/cpp/open3d/visualization/gui/Label3D.h#L36)):*

I think we need some documentation (for the class? for the constructor?) saying that this is intended to be used with SceneWidget. But, also, see below.
</blockquote></details>

Done.
___
*[cpp/open3d/visualization/gui/SceneWidget.cpp, line 894 at r2](https://reviewable.io/reviews/intel-isl/open3d/2781#-MOmGIyf7S9d5ndtPvca:-MOnzgGxBwCLgFFpP48i:b-x2spsw) ([raw file](https://github.com/intel-isl/open3d/blob/43d5385fd71847660e4a161b7a79e7e8b69b5057/cpp/open3d/visualization/gui/SceneWidget.cpp#L894)):*

This makes sense, but I think it would be better for our long-term vision of a scene graph to add the labels to Open3DScene and have SceneWidget::Draw() read the labels from Open3DScene. This way Open3DScene still stays the "model" and SceneWidget the "view"/"controller".
</blockquote></details>

Agreed. Long term, they belong in Open3DScene or a scene graph and not directly in SceneWidget. Given that that future is not yet clearly defined, I put the labels in SceneWidget for this release.
___
*[cpp/open3d/visualization/gui/SceneWidget.cpp, line 959 at r2](https://reviewable.io/reviews/intel-isl/open3d/2781#-MOmITYg9PzXx1Ue4nA_:-MOnzV5--JXcUss6qM-n:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/43d5385fd71847660e4a161b7a79e7e8b69b5057/cpp/open3d/visualization/gui/SceneWidget.cpp#L959)):*

> ```
>             ndc.x() *= f.width;
>             ndc.y() *= f.height;
> ```
Did you try this on Linux? I could be wrong, but I think you'll need to offset by f.x and f.y, in case the SceneWidget is not at (0, 0). On Windows/Linux the menu offset the SceneWidget so that it is definitely not at (0, 0).
</blockquote></details>

Done.
___
*[cpp/open3d/visualization/gui/SceneWidget.cpp, line 968 at r2](https://reviewable.io/reviews/intel-isl/open3d/2781#-MOmGvlP-JS3cKDyPK3D:-MOnk4qX51-UiKM-vPfo:b-1fw4vi) ([raw file](https://github.com/intel-isl/open3d/blob/43d5385fd71847660e4a161b7a79e7e8b69b5057/cpp/open3d/visualization/gui/SceneWidget.cpp#L968)):*

This will render the labels with all the UI elements. Conceptually it seems like these ought to be with the 3D elements (although pragmatically that seems like maybe more effort than it's worth just for that). But we could potentially have z-order issues, if you have multiple SceneWidgets in layouts. Although since SceneWidgets have problems with z-order currently anyway, I guess we can leave it for now unless it's easy to do or something.
</blockquote></details>

I agree this is not ideal, but it is was implemented this way in the interest of time. We currently have no examples that use multiple SceneWidgets, but you right it could be a problem which should be addressed for the next release.
___
*[cpp/open3d/visualization/rendering/Camera.h, line 109 at r2](https://reviewable.io/reviews/intel-isl/open3d/2781#-MOmHjFz8JYLHBwp51aK:-MOnkNi24YO3sFsg_UoY:b-l2it78) ([raw file](https://github.com/intel-isl/open3d/blob/43d5385fd71847660e4a161b7a79e7e8b69b5057/cpp/open3d/visualization/rendering/Camera.h#L109)):*

Can we change this to a more descriptive name? It took me a while to work out what it was (especially since "DC" frequently means "Device Context" in drawing code).  GetNormalizedDeviceCoordinates() would be an obvious choice, but that assumes the reader understands what filament means by "normalized device coordinates". How would `ConvertToNormalizedViewCoords()` be? Or `ConvertTo2DCoord()`?
</blockquote></details>

Two things. 1) Anyone with a graphics background looking at a camera class will immediately understand the meaning of NDC although I grant you that otherwise it may require some research or head scratching. I will add a comment to explain.

2) I initially called it GetScreenCoords but it gives you the NDC which is [-1, 1] and therefore not really screen coordinates. So, I don't think ConvertTo2DCoord captures the meaning correctly.

prewettg (approved):
> Reviewed 4 of 4 files at r3.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2781)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2815, opened 2020-12-23
Fixes to O3DVisualizer shading modes, adds unlit mode

prewettg:
Fixed changing point size while not in standard shader mode resetting to standard shader mode;  non-standard shading mode not keeping point sizes;  added unlit shader mode.

Known bug: lines are not shown in depth mode (and even if they were, they'd only be one pixel GL lines)

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2815)


errissa (approved):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2815)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2819, opened 2020-12-24
Lazy load renderer when creating window in case Python crashes before window becomes visible

prewettg:
This fixes the following Python script hanging on Windows:
```
#!/usr/bin/env python
import open3d as o3d

def main():
    o3d.visualization.gui.Application.instance.initialize()
    w = o3d.visualization.gui.Application.instance.create_window("Crash", 640, 480)

    crash

if __name__ == "__main__":
    main()
```

On Windows it seems that if the program exits before the first draw message is received, the Filament render (or command?) thread has not started yet. The Window destructor will destroy the Filament renderer, which will clean up some stuff and then hang at filament/src/Renderer.cpp:130 (FRenderer::terminate() at Fence::waitAndDestroy()). It is not clear to me if this is a timing problem or something else. If the script exits shortly after the renderer is created in the draw, the script may still hang. (To test, try setting an on_layout callback and throwing an exception there.)

This PR does not change the behavior on macOS / Linux (which crash before and after this PR)

This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2819)

### DIFF
--- a/cpp/open3d/visualization/gui/Window.cpp
+++ b/cpp/open3d/visualization/gui/Window.cpp
@@ -361,7 +361,7 @@ Window::Window(const std::string& title,
     //   w = o3d.visualization.Application.instance
     //          .create_window("Crash", 640, 480)
     //   <anything that throws an exception>
-#endif // !WIN32
+#endif  // !WIN32
 }
 
 void Window::CreateRenderer() {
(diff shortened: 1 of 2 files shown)


errissa (approved):
> Reviewed 1 of 2 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2819)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2026, opened 2020-06-29
refactor color map optimization

yxlao:
- color map optimization is now a class
- the numerical results are exactly the same as before, see `python/test/test_color_map.py`
- separated rigid and non-rigid optimization options
- docstrings are updated, now docstring parsing non-overloaded `__init__ ` function are supported
- deleted cpp unit tests, most of the original unit test were commented out and unused
- add a ported [`std::optional<T>`](https://en.cppreference.com/w/cpp/utility/optional), to simplify code

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2026)

Issue #2712, opened 2020-12-07
Set image files region of interest in Colormap Optimization


drperpen:
Hi,
I am reconstructing from 2d images in SideFX Houdini and was thinking in using your Open3D Colormap with my mesh to compare it with my Houdini's color assignment results. However, my 2d images files from the cameras are large (8K). In Houdini I deal with this by creating region of interest/bounding boxes on the images, as the area of interest rarely is higher than 15% of the whole frame (cameras are far)

Is there a way or would you consider to implement a way to set only the area you need from the image file?

Thanks!

yxlao:
Yes, it is possible to automatically detect the regions of interest using an "adaptive anchor point" approach.  It is not included in the current `master` branch, but we do have some demo code. You may check this for your reference (though you might need to fix some build issues to get it to compile): https://github.com/intel-isl/Open3D/commit/2dd6a714efe6e255bac328c2a81259c032b6484a .

Here are the ideas:
1. Unified anchor points are placed in a square grid pattern over the image for image warping (`src/Core/ColorMap/ImageWarpingField.h`, the density is controlled by `number_of_vertical_anchors`.)
2. Only some of the grid cells contain the subjects, the anchor points associated with these grid cells receives non-zero warping field flow values
3. The rest of the anchor points does not contain the subject, thus their flow values are always zero
4. Now,
    - Original implementation: we consider all anchor points and build solve a linear system with a large matrix
    - "Adaptive anchor point" approach: only the anchor points associated with the activated grid cells are used to build a sparse matrix, and we solve the linear system with the sparse matrix

Given the same parameters, the results of both implementations are identical. However, the "adaptive anchor point" approach is significantly faster, which allows us to use finer grid cells (larger `number_of_vertical_anchors`) to achieve better results.

drperpen:
Thank you for your answer. I will look into that!

PR #2832, opened 2020-12-29
color map optimization refactoring

yxlao:
Both comments from #2026 have been addressed.

- [x] simplified interface
- [x] separated rigid and non-rigid optimization options
- [x] added option to log intermediate results for debugging
- [x] use utility::optional, more code reuse
- [x] remove unused unit tests
- [x] docstrings are updated, now docstring parsing non-overloaded __init__ function are supported

```python
# new API directly returns the optimized mesh
option = o3d.pipelines.color_map.NonRigidOptimizerOption(maximum_iteration=5)
mesh_optimized = o3d.pipelines.color_map.run_non_rigid_optimizer(mesh, 
                                                                 rgbd_images, 
                                                                 camera_trajectory, 
                                                                 option)
```

This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2832)


theNded (changes requested):
> Reviewed 25 of 25 files at r1.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2832)* status: all files reviewed, 3 unresolved discussions (waiting on @yxlao)

___
*[cpp/open3d/pipelines/color_map/ColorMapUtils.cpp, line 93 at r1](https://reviewable.io/reviews/intel-isl/open3d/2832#-MPgOAs480JAkZEsgbTz:-MPgOAs52epovrNLlgei:b9z0780) ([raw file](https://github.com/intel-isl/open3d/blob/1c387ae6a725d5d9799253fbc429ba1d225ca959/cpp/open3d/pipelines/color_map/ColorMapUtils.cpp#L93)):*
> ```cpp
>            std::vector<geometry::Image>,
>            std::vector<geometry::Image>>
> CreateGradientImages(const std::vector<geometry::RGBDImage>& images_rgbd) {
> ```

CreateUtilImagesFromRGBD, or something similar? Since it couples a lot of images
___
*[cpp/open3d/pipelines/color_map/NonRigidOptimizer.cpp, line 70 at r1](https://reviewable.io/reviews/intel-isl/open3d/2832#-MPgPRPs-rcuCNWhjum4:-MPgPRPs-rcuCNWhjum5:bqr1n8a) ([raw file](https://github.com/intel-isl/open3d/blob/1c387ae6a725d5d9799253fbc429ba1d225ca959/cpp/open3d/pipelines/color_map/NonRigidOptimizer.cpp#L70)):*
> ```cpp
> /// Utility/Eigen.h/cpp, but this function takes additional multiplication
> /// pattern that can produce JTJ having hundreds of rows and columns.
> template <typename VecInTypeDouble,
> ```

Note for future implementation: in SLAC and color map optimization I may switch to solving Jx = r instead of JtJ x = Jtr. That formulation is more thread-friendly, each thread will be responsible for one row instead of touching various matrix blocks that require read-write lock.
___
*[cpp/open3d/pipelines/color_map/NonRigidOptimizer.cpp, line 251 at r1](https://reviewable.io/reviews/intel-isl/open3d/2832#-MPgQ1Fz3gNC1UmAO-I8:-MPgQ1Fz3gNC1UmAO-I9:b-pzuykn) ([raw file](https://github.com/intel-isl/open3d/blob/1c387ae6a725d5d9799253fbc429ba1d225ca959/cpp/open3d/pipelines/color_map/NonRigidOptimizer.cpp#L251)):*
> ```cpp
> 
>     // Create all debugging directories. We don't delete any existing files but
>     // will overwrite them if the names are the same.
> ```

Nice choice. Maybe also add it to the comment of the variable in option.

yxlao (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2832)* status: 18 of 25 files reviewed, 3 unresolved discussions (waiting on @theNded)

___
*[cpp/open3d/pipelines/color_map/ColorMapUtils.cpp, line 93 at r1](https://reviewable.io/reviews/intel-isl/open3d/2832#-MPgOAs480JAkZEsgbTz:-MPgYTiR9cvWc0lj6kHk:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/1c387ae6a725d5d9799253fbc429ba1d225ca959/cpp/open3d/pipelines/color_map/ColorMapUtils.cpp#L93)):*

CreateUtilImagesFromRGBD, or something similar? Since it couples a lot of images
</blockquote></details>

Done.
___
*[cpp/open3d/pipelines/color_map/NonRigidOptimizer.cpp, line 70 at r1](https://reviewable.io/reviews/intel-isl/open3d/2832#-MPgPRPs-rcuCNWhjum4:-MPgYC7P01R0sFxlVZMx:b-im7hkw) ([raw file](https://github.com/intel-isl/open3d/blob/1c387ae6a725d5d9799253fbc429ba1d225ca959/cpp/open3d/pipelines/color_map/NonRigidOptimizer.cpp#L70)):*

Note for future implementation: in SLAC and color map optimization I may switch to solving Jx = r instead of JtJ x = Jtr. That formulation is more thread-friendly, each thread will be responsible for one row instead of touching various matrix blocks that require read-write lock.
</blockquote></details>

Sounds good. 

We had some previous results for the `Jx=r` approach. The issue is that when the point cloud is large, the `J` matrix can be very big so that it takes up more memory and time for the solver. On the other hand, the `JtJ` matrix is much smaller.



To make the `Jx=r` approach work, the are two possible approaches:

### Approach 1: sparse matrix for `J`
Use a sparse matrix to store `J`. This also requires a sparse solver. https://github.com/intel-isl/Open3D/blob/2dd6a714efe6e255bac328c2a81259c032b6484a/src/Core/ColorMap/ColorMapOptimizationJacobian.cpp#L208

### Approach 2: compress `J` and remove all zero columns. 

This is related to the "adaptive anchor point" approach mentioned in #2712 . This help to reduce the size of `J` while doesn't affect the results.


___
*[cpp/open3d/pipelines/color_map/NonRigidOptimizer.cpp, line 251 at r1](https://reviewable.io/reviews/intel-isl/open3d/2832#-MPgQ1Fz3gNC1UmAO-I8:-MPgYSrh3qcdoY3L_rUS:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/1c387ae6a725d5d9799253fbc429ba1d225ca959/cpp/open3d/pipelines/color_map/NonRigidOptimizer.cpp#L251)):*

Nice choice. Maybe also add it to the comment of the variable in option.
</blockquote></details>

Done.

theNded (changes requested):
> Reviewed 7 of 7 files at r2.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2832)* status: all files reviewed, 1 unresolved discussion (waiting on @yxlao)

___
*[cpp/open3d/pipelines/color_map/NonRigidOptimizer.cpp, line 70 at r1](https://reviewable.io/reviews/intel-isl/open3d/2832#-MPgPRPs-rcuCNWhjum4:-MPgr6AyEBbICLYCxQlG:byp66ag) ([raw file](https://github.com/intel-isl/open3d/blob/1c387ae6a725d5d9799253fbc429ba1d225ca959/cpp/open3d/pipelines/color_map/NonRigidOptimizer.cpp#L70)):*

Sounds good. 

We had some previous results for the `Jx=r` approach. The issue is that when the point cloud is large, the `J` matrix can be very big so that it takes up more memory and time for the solver. On the other hand, the `JtJ` matrix is much smaller.



To make the `Jx=r` approach work, the are two possible approaches:

### Approach 1: sparse matrix for `J`
Use a sparse matrix to store `J`. This also requires a sparse solver. https://github.com/intel-isl/Open3D/blob/2dd6a714efe6e255bac328c2a81259c032b6484a/src/Core/ColorMap/ColorMapOptimizationJacobian.cpp#L208

### Approach 2: compress `J` and remove all zero columns. 

This is related to the "adaptive anchor point" approach mentioned in #2712 . This help to reduce the size of `J` while doesn't affect the results.



</blockquote></details>

Thanks for the detailed explanation. I think if the linear system is sparse, we can exploit sparsity by either Jx = r or JtJ = Jtr, see section 3.3, http://www.cs.cmu.edu/~kaess/pub/Dellaert17fnt.pdf. But the illustration here seems dense, so maybe we will have to stick to the current solution.

theNded (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2832#-:-MPguXT07L-6DlInETMl:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @yxlao)

yxlao (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2832)* status: all files reviewed, 1 unresolved discussion (waiting on @theNded)

___
*[cpp/open3d/pipelines/color_map/NonRigidOptimizer.cpp, line 70 at r1](https://reviewable.io/reviews/intel-isl/open3d/2832#-MPgPRPs-rcuCNWhjum4:-MPgySI793rL7g1iD39r:b-y7zc9l) ([raw file](https://github.com/intel-isl/open3d/blob/1c387ae6a725d5d9799253fbc429ba1d225ca959/cpp/open3d/pipelines/color_map/NonRigidOptimizer.cpp#L70)):*

Thanks for the detailed explanation. I think if the linear system is sparse, we can exploit sparsity by either Jx = r or JtJ = Jtr, see section 3.3, http://www.cs.cmu.edu/~kaess/pub/Dellaert17fnt.pdf. But the illustration here seems dense, so maybe we will have to stick to the current solution.
</blockquote></details>

Yes, in our current example, `J` is not very sparse. When the subject is small and the background is large, `J` will be more sparse. In this case, most anchor points will receive warping flows of 0.

PR #2839, opened 2020-12-30
Added a VisualizerWithEditAndKeyCallback class

ajbrookhouse:
I was working on a project where I wanted to use the VisualizerWithEditing class to be able to select regions of a 3d point cloud and also register custom key callbacks.

I have added a VisualizerWithEditAndKeyCallback class, which contains functions from both VisualizerWithEditing and VisualizerWithKeyCallback functionalities, allowing the visualizer to be used with editing features and also use the register_key_callback function to register custom key callbacks. 

This class is implemented in c++ and also has updated the python bindings.

### DIFF
--- a/cpp/open3d/Open3D.h.in
+++ b/cpp/open3d/Open3D.h.in
@@ -126,6 +126,7 @@
 #include "open3d/visualization/visualizer/VisualizerWithEditing.h"
 #include "open3d/visualization/visualizer/VisualizerWithKeyCallback.h"
 #include "open3d/visualization/visualizer/VisualizerWithVertexSelection.h"
+#include "open3d/visualization/visualizer/VisualizerWithEditAndKeyCallback.h"
 
 // clang-format off
 @BUILD_AZURE_KINECT_COMMENT@#include "open3d/io/sensor/azure_kinect/AzureKinectRecorder.h"
--- a/cpp/open3d/visualization/visualizer/VisualizerWithEditAndKeyCallback.cpp
+++ b/cpp/open3d/visualization/visualizer/VisualizerWithEditAndKeyCallback.cpp
@@ -1,0 +1,772 @@
+// ----------------------------------------------------------------------------
+// -                        Open3D: www.open3d.org                            -
+// ----------------------------------------------------------------------------
+// The MIT License (MIT)
+//
+// Copyright (c) 2018 www.open3d.org
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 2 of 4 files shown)


yxlao:
@ajbrookhouse regarding the style checker, please follow http://www.open3d.org/docs/release/contribute/styleguide.html#style-checker.

ajbrookhouse:
Thank you, I fixed the style issue. 

Currently, the VisualizerWithEditAndKeyCallback  class I made contains copies of code from VisualizerWithEdit and VisualizerWithKeyCallback from the cpp and h files. I merged them manually with slight adjustment as needed.

However it would probably be better if the VisualizerWithEditAndKeyCallback class inherited from the two other classes. That way code isn't manually duplicated, as that can be messy to maintain. Do you think this is a change I should make? It would involve making the different types of Visualizers virtual classes to make multiple inheritance between them work better.

I just wanted to ask and see what you thought before trying that. I think it would be cleaner.

prewettg:
@ajbrookhouse Sorry for the late response, I just got back from a two-week vacation. I agree with you that it would be better to have both visualizers inherit from a common base class, although multiple inheritance can get messy, so maybe it would be simpler to just merge them (as you have done) and then typedef the old names to the new one, so that you always get the functionality of both. If you don't want the other features, you just won't use them... However, we're planning on deprecating these visualizers in favor of the new O3DVisualizer (new with 0.12), which is uses the new GUI and aims to merge the functionality of all these old visualizers into one hopefully full-featured and easy to use class. We aren't doing any active maintenance on the old viewers, so it's kind of up to you how much work you want to put into this. I'm certainly happy to approve your PR as it is, pending resolution of the minor bug below.

prewettg (changes requested):
I tried this out with `examples/python/visualization/customized_visualization_key_action.py` and just changed the name of the class, and it seems that something with the key events seems to have changed. The original class rotates as long as the space bar is held down, but the new class continues rotating even after the space bar is release.

> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2839#-:-MQnQafA0itC4vi1-whP:bdshuoq)* status: 0 of 4 files reviewed, all discussions resolved (waiting on @prewettg)

prewettg (changes requested):
> Reviewed 3 of 4 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2839)* status: :shipit: complete! all files reviewed, all discussions resolved

ajbrookhouse:
Hi @prewettg, thank you so much for your feedback. 

So, long story short, I was interested in using this library for the 3d classes like point cloud, mesh, etc, and being able to have a visualizer that can be run and displayed easily in a python program.

As I'm working more on my project there is more I'm realizing that my group wants to have in their visualizer. Essentially I'm planning on adding several features, like ways to have a 3d selection of which part of a model to show, if the whole model is bigger, also a 3d selection to have a get_points like feature that returns all points within the 3d selection. We also probably want to add stuff like having multiple point clouds being shown in this edit window, and having controls to hide/view certain clouds once they've been loaded. Ideally, as time goes on we're going to keep adding more. 

Originally, when I was just making a VisualizerWithEditAndKeyCallback, it seemed pretty clear how that would fit into your github project through a pull request, and I will still try to fix the bug you mentioned and get this pulled in. However, I'm going to be working soon on adding a lot of things to the visualizer class, for our personal use in our project. Is this something that you'd be interested having a pull request for once I finish and further develop?

prewettg:
@ajbrookhouse Sure, we'd love some PRs. Since the old visualizer is being deprecated, new development is on O3DVisualizer, so feel free to create PRs against that. It would be helpful to us if you can limit PRs to one feature per PR, that makes them easier to integrate. I think O3DVisualizer already includes one feature you listed, namely the ability to have multiple objects and toggle them on and off. It also has the ability to select points directly (but no polygon selection yet, and definitely no 3D volume selection). See [`examples/python/gui/draw.py`](https://github.com/intel-isl/Open3D/blob/master/examples/python/gui/draw.py). The example uses `open3d.visualization.draw()`, but that is just a wrapper around O3DVisualizer, as you can see in [`python/open3d/visualization/draw.py`](https://github.com/intel-isl/Open3D/blob/master/python/open3d/visualization/draw.py).
   As far as collaboration, I think it might be helpful to open a draft PR when you start a feature and ping me, so that we have an idea of what you're planning on doing, just so you can avoid spending time on something if it's already mostly done or something. That way you don't go through all the work to finish a large feature and discover we can't use for <reason>.

PR #2846, opened 2021-01-01
Add sudo to make install

pauljurczak:
This is necessary when installing to system files in case the previous sudo timeout already expired.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2846)

### DIFF
--- a/docs/compilation.rst
+++ b/docs/compilation.rst
@@ -107,7 +107,7 @@ To install Open3D C++ library:
 
 .. code-block:: bash
 
-    make install
+    sudo make install
 
 To link a C++ project against the Open3D C++ library, please refer to
 :ref:`create_cplusplus_project`.


docs/compilation.rst
```diff
@@ -107,7 +107,7 @@ To install Open3D C++ library:
 
 .. code-block:: bash
 
-    make install
+    sudo make install
```
yxlao, line 110: If the `CMAKE_INSTALL_PREFIX` dir does not need root access to write, `sudo` is not required. I would be good to describe the scenario when to use the `sudo` here. Something like:

```
# If ...
make install

# If ...
sudo make install
```

pauljurczak: That's true, but in my opinion it's easier for an inexperienced user just to type `sudo make install`, which is necessary in default case, rather than think, which option to use. It will work for all cases. The experienced user will know, that it may be redundant for their configuration.

Just following "offer only one way of doing things, if possible" paradigm.

yxlao: That's true. But from another perspective, we want to prevent inexperienced users from changing their root-protected directories until they are sure what they are doing. It's [non-trivial to uninstall](https://stackoverflow.com/a/44649542/1255535) the library if it is installed in the system directory. The problem is more prominent when there's an upgrade, where `make install` could result in unwanted files without proper uninstallation first. 

What about setting the default to the non-root version? That is, we modify the docs to ask the user to enter `CMAKE_INSTALL_PREFIX` by default. (In case the user forgets to do so, the system will automatically warn the user that they may want to use `sudo`.)

pauljurczak: Agree. This is a better option.

PR #2849, opened 2021-01-02
Implement np style arange for Tensor

theNded:
- [x] Omit `start` in python API

This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2849)

### DIFF
--- a/cpp/open3d/core/CMakeLists.txt
+++ b/cpp/open3d/core/CMakeLists.txt
@@ -1,4 +1,6 @@
 set(KERNEL_SRC
+    kernel/Arange.cpp
+    kernel/ArangeCPU.cpp
     kernel/IndexGetSet.cpp
     kernel/IndexGetSetCPU.cpp
     kernel/NonZero.cpp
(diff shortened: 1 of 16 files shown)


yxlao:
image
(Left: good; Right: bad)

CI is failing since Pillow 8.1.0 (released Jan 02) does not come with pre-compiled aarch64 package. The solution is to remove matplotlib as open3d's dependency and only enable it when building docs.

yxlao (approved):
> Reviewed 1 of 8 files at r1, 1 of 2 files at r2, 2 of 4 files at r3, 9 of 11 files at r4, 2 of 3 files at r5.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2849)* status: 14 of 16 files reviewed, 1 unresolved discussion (waiting on @theNded and @yxlao)

___
*[cpp/open3d/core/Tensor.h, line 335 at r2](https://reviewable.io/reviews/intel-isl/open3d/2849#-MQ5AkYJ2DqFYkeFfKwo:-MQ5AkYJ2DqFYkeFfKwp:b-r1zue0) ([raw file](https://github.com/intel-isl/open3d/blob/11835ebac7d2108b5bc88608092ea7befbc8acc2/cpp/open3d/core/Tensor.h#L335)):*
> ```objc
>     static Tensor Diag(const Tensor& input);
> 
>     /// Create a 1D tensor with evenly spaced values in the given interval.
> ```

shall we make this private or protected?

PR #2852, opened 2021-01-04
Unprojection function in the ViewControl

PieroV:
Hi!

I added a function to unproject 2D points to obtain a ray from the camera eye containing the 2D point.
It is useful to perform raycasting queries with clicks on the visualizer.
Also, since this operation is const, I added a const overload to obtain the ViewControl from the visualizer.

The function is aware that the camera may be orthogonal rather than perspective.
For these cases, the origin is in a plane that has look_at - eye_ as normal, and in which eye_ lies.

However, because of the overload, I had to modify the python bindings: I added a lambda instead of a cast, as from what I understand that is what you usually do in these cases. It compiled, but I could not create the pip package, so I could not test that everything still worked in Python.

### DIFF
--- a/cpp/pybind/visualization/visualizer.cpp
+++ b/cpp/pybind/visualization/visualizer.cpp
@@ -112,7 +112,8 @@ void pybind_visualizer(py::module &m) {
                  "reset_bounding_box"_a = true)
             .def("clear_geometries", &Visualizer::ClearGeometries,
                  "Function to clear geometries from the visualizer")
-            .def("get_view_control", &Visualizer::GetViewControl,
+            .def("get_view_control",
+                 [](Visualizer &vis) { return vis.GetViewControl(); },
                  "Function to retrieve the associated ``ViewControl``",
                  py::return_value_policy::reference_internal)
             .def("get_render_option", &Visualizer::GetRenderOption,
--- a/cpp/open3d/visualization/visualizer/ViewControl.cpp
+++ b/cpp/open3d/visualization/visualizer/ViewControl.cpp
@@ -290,6 +290,34 @@ void ViewControl::ChangeWindowSize(int width, int height) {
     SetProjectionParameters();
 }
 
+geometry::Ray3D ViewControl::UnprojectPoint(double x, double y) const {
+    if (!window_width_ || !window_height_) {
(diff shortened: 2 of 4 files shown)


PieroV:
Hi,
I am having some troubles in understanding why the CI fails on Ubuntu CI arm64.
I see that pillow failed to install, because it could not find jpeg (see lines 10777-10778):
```  The headers or library files could not be found for jpeg,
  a required dependency when compiling Pillow from source.```
but then I see (line 11028)
```[100%] Built target install-pip-package```
So I think that the creation/installation of the Python package fails, but the tests run anyway.
The same happened in PR #2833.

**Review by errissa — APPROVED** [review]
> Reviewed 2 of 4 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/2852)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @PieroV)

**Comment by PieroV** [comment]
> > Reviewed 2 of 4 files at r1, 2 of 2 files at r2, all commit messages.
> > _[Reviewable](https://reviewable.io/reviews/isl-org/open3d/2852)_ status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @PieroV)

Hello!

What do you need me to do? Squash and rebase?

File: cpp/pybind/visualization/visualizer.cpp [hunk 1]
```diff
@@ -112,9 +112,11 @@ void pybind_visualizer(py::module &m) {
                  "reset_bounding_box"_a = true)
             .def("clear_geometries", &Visualizer::ClearGeometries,
                  "Function to clear geometries from the visualizer")
-            .def("get_view_control", &Visualizer::GetViewControl,
-                 "Function to retrieve the associated ``ViewControl``",
-                 py::return_value_policy::reference_internal)
+            .def(
+                    "get_view_control",
+                    [](Visualizer &vis) { return vis.GetViewControl(); },
```
pablospe, line 117: What was the purpose of this change?

pablospe: These changes were reverted in this PR:
https://github.com/isl-org/Open3D/pull/6116/files

PR #2875, opened 2021-01-07
refactoring: add pipelines/kernel folder

yxlao:
Similar to `t/geometry/kernel`, we need `t/pipelines/kernel` folder and namespace as well.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2875)


theNded (approved):
> Reviewed 15 of 15 files at r1.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2875)* status: all files reviewed, 1 unresolved discussion (waiting on @yxlao)

___
*[cpp/open3d/t/pipelines/kernel/TransformationConverter.h, line 49 at r1](https://reviewable.io/reviews/intel-isl/open3d/2875#-MQiUEjZ54ZfjPCZIL3x:-MQiUEjZ54ZfjPCZIL3y:bm6us6i) ([raw file](https://github.com/intel-isl/open3d/blob/ccd8b497b92d70ad1d3ec3f1aea21a2ea8aaef28/cpp/open3d/t/pipelines/kernel/TransformationConverter.h#L49)):*
> ```objc
> /// \return Transformation, a tensor of shape {4, 4}, dtype Float32.
> core::Tensor PoseToTransformation(const core::Tensor &pose);
> ```

Just a note: we can make a separate PR to convert (N, 6) to (N, 4, 4), which may be useful for pose graph optimization and non-rigid transformation.

PR #2877, opened 2021-01-07
Standarize To() and Copy()

yxlao:
### Motivation
There are many instances where we first check the device and then copy the tensor. E.g.
```cpp
    if (device.GetType() == core::Device::DeviceType::CPU) {
        return tensor_cpu;
    } else {
        return tensor_cpu.Copy(device);
    }
```

This can now be done in one step via `To()`.
```cpp
    return tensor_cpu.To(device);
```

### Design principles

- `Tensor::Copy()` becomes `Tensor::Clone()` (similar to [`torch.clone`](https://pytorch.org/docs/stable/generated/torch.clone.html#torch.clone))
- `Tensor::To()` can take different parameters (similar to [`torch.Tensor.to`](https://pytorch.org/docs/stable/tensors.html#torch.Tensor.to)):
  - `Tensor::To(device)`
  - `Tensor::To(dtype)`
  - `Tensor::To(device, dtype)`
- The same design principles applied to 
  - [x] `HashMap`
  - [x] `TensorList`
  - [x] `t::Geometry::PointCloud`
  - [x] `t::Geometry::TSDFVoxelGrid`
- `CPU()` and `CUDA()` helper function
  - [x] `HashMap`
    - [x] pybind
  - [x] `t::Geometry::PointCloud`
    - [x] pybind
  - [x] `t::Geometry::TSDFVoxelGrid`
    - [x] pybind

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2877)


theNded (commented):
LGTM. One question: will we also provide `CPU()` and `CUDA(int)` as a sugar, or it is usually only standard in python?

> Reviewed 11 of 21 files at r1, 22 of 22 files at r2.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2877#-:-MQXVYkA4Jzw4piUyzkt:b-q8ofdr)* status: :shipit: complete! all files reviewed, all discussions resolved

yxlao:
> LGTM. One question: will we also provide CPU() and CUDA(int) as a sugar, or it is usually only standard in python?

@theNded sounds good, added cpu() and cuda() for the geometry classes.

theNded (approved):
> Reviewed 11 of 11 files at r3.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2877)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2882, opened 2021-01-08
update mkl download links

yxlao:
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2882)

### DIFF
--- a/3rdparty/mkl/mkl.cmake
+++ b/3rdparty/mkl/mkl.cmake
@@ -11,18 +11,45 @@
 include(ExternalProject)
 
 if(WIN32)
-    set(MKL_INCLUDE_URL      "https://anaconda.org/intel/mkl-include/2020.1/download/win-64/mkl-include-2020.1-intel_216.tar.bz2")
-    set(MKL_URL              "https://anaconda.org/intel/mkl-static/2020.1/download/win-64/mkl-static-2020.1-intel_216.tar.bz2")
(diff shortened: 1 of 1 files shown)


ssheorey (approved):
Btw, can we use `set_local_or_remote_url` with `ExternalProject`? By default, `ExternalProject` only supports either local or remote, not both.

yxlao:
> Btw, can we use `set_local_or_remote_url` with `ExternalProject`? By default, `ExternalProject` only supports either local or remote, not both.

Yes. `set_local_or_remote_url` selects either
- one local URL
- one or more remote URLs

The output of `set_local_or_remote_url` is then used by `ExternalProject`. So effectively, `ExternalProject` can now support both local and remote.

ssheorey:
> > Btw, can we use `set_local_or_remote_url` with `ExternalProject`? By default, `ExternalProject` only supports either local or remote, not both.
> 
> Yes. `set_local_or_remote_url` selects either:
> 
>     * one local URL, or
> 
>     * one or more remote URLs
> 
> 
> The output of `set_local_or_remote_url` is then used by `ExternalProject`. So effectively, `ExternalProject` can now support both local and remote.

Thanks, saw the example in `mkl.cmake`. Moving forward, we can switch more `ExternalProject` dependencies to use local download cache by switching from git download -> zip download and using `set_local_or_remote_url`.

Issue #2487, opened 2020-10-16
compile error on Debian GNU/Linux 10 (buster) Open3D v.0.11.0


nanotuxi:
**Describe the bug**
Trying to compile Open3d from git (master) v0.11.0 in a python venv throws an error
related to 
`Open3D/cpp/open3d/geometry/Line3D.h:117:9: error: ‘transform’ was not declared in this scope`

**To Reproduce**
1. Clone a fresh version of Open3d 0.11.0
`git clone --recursive https://github.com/intel-isl/Open3D`
2. cd Open3D && mkdir build && cd build
3. configure with `ccmake ..`
4. run `cmake ..`
5. run `make -j6`
This throws 
```
Scanning dependencies of target visualization
[ 55%] Building CXX object cpp/open3d/geometry/CMakeFiles/geometry.dir/Line3D.cpp.o
[ 55%] Building CXX object cpp/open3d/geometry/CMakeFiles/geometry.dir/MeshBase.cpp.o
[ 55%] Building CXX object cpp/open3d/geometry/CMakeFiles/geometry.dir/IntersectionTest.cpp.o
[ 55%] Building CXX object cpp/open3d/geometry/CMakeFiles/geometry.dir/TriangleMesh.cpp.o
[ 56%] Building CXX object cpp/open3d/geometry/CMakeFiles/geometry.dir/SurfaceReconstructionBallPivoting.cpp.o
[ 57%] Building CXX object cpp/open3d/geometry/CMakeFiles/geometry.dir/VoxelGridFactory.cpp.o
[ 57%] Building CXX object cpp/open3d/visualization/CMakeFiles/visualization.dir/shader/RGBDImageShader.cpp.o
In file included from ~/Open3D/cpp/open3d/geometry/IntersectionTest.h:32,
                 from ~/Open3D/cpp/open3d/geometry/SurfaceReconstructionBallPivoting.cpp:31:
~/Open3D/cpp/open3d/geometry/Line3D.h: In member function ‘virtual void open3d::geometry::Line3D::Transform(const Eigen::Transform<double, 3, 2>&)’:
~/Open3D/cpp/open3d/geometry/Line3D.h:117:9: error: ‘transform’ was not declared in this scope
         transform(t);
         ^~~~~~~~~
//reduce output error size to shorten reading manually
[... cut original error output manually...]

In file included from ~/Open3D/cpp/open3d/Open3D.h:40,
[...]

PR #2885, opened 2021-01-09
Fix external Eigen build.

saedrna:
Eigen v3.3.* does not have `transformation` function for `ParametrizedLine`. In addition, initializer list is also not fully supported. Therefore, it does not compile with external Eigen, which generally use the release version instead of the master one.

This issue is also discussed here #2487.

### DIFF
--- a/cpp/open3d/geometry/Line3D.cpp
+++ b/cpp/open3d/geometry/Line3D.cpp
@@ -1,5 +1,6 @@
 #include "Line3D.h"
 
+#include <array>
 #include <cmath>
 
 namespace open3d {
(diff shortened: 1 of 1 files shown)


ssheorey:
Hi @saedrna, thanks for submitting this PR. Being able to use the system / stable `Eigen` install is very useful. Can you look at the unit test failures? You can also use `make apply-style` to fix the style CI failure.

saedrna:
@ssheorey I would like to. But, I am afraid that I am not quite familiar with the CI stuffs. Maybe I will look into it when free. As you probably see, the changes are quite simple -- I just copy the implementation of `Transform` from the Eigen code here and add explicit constructor instead of initlizer list. I do not quite understand how these issues come from.

PR #2888, opened 2021-01-10
make install-pip-package force update

yxlao:


theNded:
CIs seem broken on Windows, can you please check the issue?

PR #2891, opened 2021-01-11
compile 3rd-party library in a separate step for arm64 CI

yxlao:
This essentially reverts https://github.com/intel-isl/Open3D/pull/2708 and adds minor tweaks, where we first compile the 3rd-party libraries and cache the compilation.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2891)

### DIFF
--- a/.github/workflows/openblas.yml
+++ b/.github/workflows/openblas.yml
@@ -153,9 +153,8 @@ jobs:
             ccache -s
 
       - name: Upload cache to GCS
-        if: ${{ github.ref == 'refs/heads/master' }}
         run: |
-          gsutil -m rsync -r -d ~/.ccache/ gs://open3d-ci-cache/ccache-arm64/
+          gsutil -m rsync -r -d ~/.ccache/ gs://open3d-ci-cache/ccache-arm64/ || true
 
   arm64:
     runs-on: ubuntu-20.04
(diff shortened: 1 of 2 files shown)


.github/workflows/openblas.yml
```diff
@@ -153,9 +153,8 @@ jobs:
             ccache -s
 
       - name: Upload cache to GCS
-        if: ${{ github.ref == 'refs/heads/master' }}
         run: |
-          gsutil -m rsync -r -d ~/.ccache/ gs://open3d-ci-cache/ccache-arm64/
+          gsutil -m rsync -r -d ~/.ccache/ gs://open3d-ci-cache/ccache-arm64/ || true
```
ssheorey, line 154: This should work. We were worried about simultaneous writes to the GCS bucket with the same cache file name causing corruption, but looks like writes to GCS are atomic:
https://stackoverflow.com/a/29321942

util/ci_utils.sh
```diff
@@ -499,3 +499,33 @@ build_docs() {
...
+    update-alternatives --install /usr/bin/cc cc /usr/bin/clang 100
+    update-alternatives --install /usr/bin/c++ c++ /usr/bin/clang++ 100
+    /usr/sbin/update-ccache-symlinks
+    echo 'export PATH="/usr/lib/ccache:$PATH"' | tee -a ~/.bashrc
+    virtualenv --python=$(which python) ${HOME}/venv
+    source ${HOME}/venv/bin/activate
+    which python
+    python --version
+    pip install pytest=="6.0.1" -U
+    pip install wheel=="0.35.1" -U
```
ssheorey, line 523: "$PYTEST_VER"
"$WHEEL_VER"

ssheorey (approved):

Issue #26, opened 2017-05-25
Python tutorial that uses ORB feature from OpenCV to initialize RGBDOdometry


syncle:
This issues is derived from #24. Lower priority. 
Need not to be done before the first launch.

Issue #25, opened 2017-05-25
Python binding for image processing module


syncle:
This issue is derived from #24

Issue #24, opened 2017-05-25
Basic image processing module


syncle:
During the process of implementing OdometryRGBD, we found a need of basic image processing modules (e.g., pixel pointer, downsampling, filtering, creating pyramid image) that are generally useful for OdometryRGBD, texturing meshes, and so on. 

Plans:
- Put image processing functions into core/geometry/image.cpp/h
- Some useful image functions are distributed in the individual projects, we are going to move them too after discussion.
- I am not going to introduce heavy library for image processing. Instead, I will implement reasonably efficient functions: such as separable 1D filtering instead of 2D filtering.
- I will add TestImageProcess.cpp/h in Test module for debugging and demonstrating implemented functions.
- If OpenCV is available, CMake tool-chain will link it for utilizing higher level functions such as ORB.

Jaesik

Issue #20, opened 2017-05-10
Beautify function names


qianyizh:
Things like:

* ``SelectionPolygonVolume::CropGeometry``
* ``SelectDownSample``, ``VoxelDownSample``, ``UniformDownSample``, ...
* ``ComputePointCloudToPointCloudDistance``, ...

Move the output to the return part of functions. For all ``Geometry`` type, return ``std::share_ptr`` because ``Visualizer`` needs to share the ownership. But for ``std::vector``, use the move and trivial copy mechanics to directly return.

PR #2898, opened 2021-01-14
Image processing Ops sample with IPP and NPP

ssheorey:
- Install IPP-IW from OpenCV Github release
- Use NPP from CUDA installation
- Image::Dilate() with NPP and IPP. No ARM support, limited datatype support on CPU/GPU.
- Image::To() and Image::LinearTransform() (in terms of IPP only)
- CUDA_CALL and IPP_CALL macros to call CUDA/IPP functions only if CUDA/IPP support is built
- Separate checking for (dtype, channels) support for each image op for NPP and IPP
- C++ Unit tests
- Python bindings

Note:
- [IPP is free for all use.](https://software.intel.com/en-us/articles/free-ipsxe-tools-and-libraries)
- IPP IW from OpenCV github is used for quick integration. IPP is also available directly from `apt` / `conda` / `pip`, but setting this up looks more complex.
- No support yet for USE_SYSTEM_IPP

Documentation:
- [Intel PP available functionality in C backend API](https://software.intel.com/content/www/us/en/develop/documentation/ipp-dev-reference/top/volume-2-image-processing.html#volume-2-image-processing)
- [Intel PP integration wrappers (actually used C++ API)](https://software.intel.com/content/www/us/en/develop/documentation/ippiw-dev-guide-and-reference/top/c-reference-1/image-processing-1.html)
- [NVIDIA PP](https://docs.nvidia.com/cuda/npp/index.html)

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2898)


yxlao (changes requested):
Seems like there are some file permission change `normal -> executable`:



> Reviewed 13 of 19 files at r1, 7 of 8 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2898#-:-MRNVjLr98CDZyia6uE7:b-85t7hn)* status: 20 of 21 files reviewed, 32 unresolved discussions (waiting on @reyanshsolis, @ssheorey, @theNded, and @yxlao)

___
*[3rdparty/find_dependencies.cmake, line 1140 at r2](https://reviewable.io/reviews/intel-isl/open3d/2898#-MRNBbzq4a_UVv9P_hpo:-MRNBbzq4a_UVv9P_hpp:ba70th5) ([raw file](https://github.com/intel-isl/open3d/blob/151079dafdb35b50f87b0cb2b1e9a7283f99e9a3/3rdparty/find_dependencies.cmake#L1140)):*
> ```CMake
> if (WITH_IPPICV)
>     include(${Open3D_3RDPARTY_DIR}/ippicv/ippicv.cmake)
>     if (WITH_IPPICV)    # If platform is unsupported, this will fail
> ```

shall we add arm64 detection to switch off `WITH_IPPICV` automatically?
___
*[3rdparty/find_dependencies.cmake, line 1143 at r2](https://reviewable.io/reviews/intel-isl/open3d/2898#-MRNFbej0xIn53owTZSb:-MRNFbej0xIn53owTZSc:bjit6rz) ([raw file](https://github.com/intel-isl/open3d/blob/151079dafdb35b50f87b0cb2b1e9a7283f99e9a3/3rdparty/find_dependencies.cmake#L1143)):*
> ```CMake
>         message(STATUS "IPP-ICV ${IPPICV_VERSION_STRING} available. Building interface wrappers IPP-IW.")
>         import_3rdparty_library(3rdparty_ippicv
>             INCLUDE_DIRS "${IPPIW_INCLUDE_DIR}" "${IPPICV_INCLUDE_DIR}"
> ```

We can concatenate `IPPICV_INCLUDE_DIR` and `IPPIW_INCLUDE_DIR` in `ippicv.cmake`, such that we only need to use `IPPICV_INCLUDE_DIR` here. This would be more consistent with the usage of other libraries.

Same for the `LIBRARIES` property.
___
*[3rdparty/ippicv/ippicv.cmake, line 26 at r2](https://reviewable.io/reviews/intel-isl/open3d/2898#-MRNGNZi5A-tO4grIT28:-MRNGNZi5A-tO4grIT29:b-ov5olo) ([raw file](https://github.com/intel-isl/open3d/blob/151079dafdb35b50f87b0cb2b1e9a7283f99e9a3/3rdparty/ippicv/ippicv.cmake#L26)):*
> ```CMake
> else()
>     set(WITH_IPPICV OFF)
>     message(WARNING "IPP-ICV disabled: Unsupported Platform.")
> ```

shall we use `FATAL_ERROR` here as well?
___
*[3rdparty/ippicv/ippicv.cmake, line 40 at r2](https://reviewable.io/reviews/intel-isl/open3d/2898#-MRNH2-R7dtMe7e7Wy9Z:-MRNH2-R7dtMe7e7Wy9_:br991z1) ([raw file](https://github.com/intel-isl/open3d/blob/151079dafdb35b50f87b0cb2b1e9a7283f99e9a3/3rdparty/ippicv/ippicv.cmake#L40)):*
Perhaps `PATCH_COMMAND` would be more clear here, since it is our own patch.

From https://cmake.org/cmake/help/v3.19/module/ExternalProject.html, it seems like `UPDATE_COMMAND` refers to source code version updates.

> by default the external project’s sources will be updated if the download method supports updates **(e.g. a git repository would be checked if the GIT_TAG does not refer to a specific commit).**
___
*[cpp/open3d/core/Tensor.h, line 761 at r2](https://reviewable.io/reviews/intel-isl/open3d/2898#-MRNIa_v7Fu-sYT2uhPh:-MRNIa_v7Fu-sYT2uhPi:b-l9lpxa) ([raw file](https://github.com/intel-isl/open3d/blob/151079dafdb35b50f87b0cb2b1e9a7283f99e9a3/cpp/open3d/core/Tensor.h#L761)):*
> ```objc
>     Tensor Abs_();
> 
>     /// Element wise clipping of tensor values so that resulting values lie in
> ```

- Nit: `Element-wise`
- Nit: `returning a new tensor.` (with `.`)
- Add `\param min_val` and `\param max_val`, such that `\p` works
___
*[cpp/open3d/core/Tensor.cpp, line 908 at r2](https://reviewable.io/reviews/intel-isl/open3d/2898#-MRNLb1D2oXPZO6PD8sY:-MRNLb1D2oXPZO6PD8sZ:bxpf1o4) ([raw file](https://github.com/intel-isl/open3d/blob/151079dafdb35b50f87b0cb2b1e9a7283f99e9a3/cpp/open3d/core/Tensor.cpp#L908)):*
> ```cpp
> Tensor Tensor::Clip(double min_val, double max_val) const {
>     Tensor dst_tensor(shape_, dtype_, GetDevice());
>     /* kernel::UnaryEW(*this, dst_tensor, kernel::UnaryEWOpCode::Abs) ; */
> ```

remove comments?
___
*[cpp/open3d/core/Tensor.cpp, line 914 at r2](https://reviewable.io/reviews/intel-isl/open3d/2898#-MRNLc8P23IKynpVK9UQ:-MRNLc8P23IKynpVK9UR:bxpf1o4) ([raw file](https://github.com/intel-isl/open3d/blob/151079dafdb35b50f87b0cb2b1e9a7283f99e9a3/cpp/open3d/core/Tensor.cpp#L914)):*
> ```cpp
> 
> Tensor Tensor::Clip_(double min_val, double max_val) {
>     /* kernel::UnaryEW(*this, *this, kernel::UnaryEWOpCode::Abs); */
> ```

remove comments?
___
*[cpp/open3d/t/geometry/CMakeLists.txt, line 12 at r2](https://reviewable.io/reviews/intel-isl/open3d/2898#-MRNLkds6jPTGh4_HvFn:-MRNLkds6jPTGh4_HvFo:b4xk36r) ([raw file](https://github.com/intel-isl/open3d/blob/151079dafdb35b50f87b0cb2b1e9a7283f99e9a3/cpp/open3d/t/geometry/CMakeLists.txt#L12)):*
> ```CMake
>     kernel/PointCloudCUDA.cu
>     kernel/TSDFVoxelGridCUDA.cu
>     kernel/nppImage.cpp
> ```

By our naming convention, we can call it `NPPImage` or `NppImage`. Same for `ipp`.
___
*[cpp/open3d/t/geometry/CMakeLists.txt, line 37 at r2](https://reviewable.io/reviews/intel-isl/open3d/2898#-MRNM7Rs27bWIJIVVG8c:-MRNM7Rs27bWIJIVVG8d:b-m1imf1) ([raw file](https://github.com/intel-isl/open3d/blob/151079dafdb35b50f87b0cb2b1e9a7283f99e9a3/cpp/open3d/t/geometry/CMakeLists.txt#L37)):*
> ```CMake
> if (WITH_IPPICV)
>     list(APPEND ALL_T_GEOMETRY_SRC ${T_GEOMETRY_KERNEL_IPP_SRC})
> endif()
> ```

This part looks quite convoluted. We actually only need one variable `T_GEOMETRY_SRC` and we can just append to it with `set` instead of `list(APPEND`. We also don't need to distinguish the `KERNEL` or `IPP` sources.

Something like:
```
set(T_GEOMETRY_SRC 
    cpu_source_0.cpp
    cpu_source_1.cpp
)

if(BUILD_CUDA_MODULE)
    set(T_GEOMETRY_SRC
        ${T_GEOMETRY_SRC}
        cuda_source_0.cpp
        cuda_source_1.cpp
    )
endif()

if(WITH_IPPICV)
    set(T_GEOMETRY_SRC
        ${T_GEOMETRY_SRC}
        ipp_source_0.cpp
        ipp_source_1.cpp
    )
endif()
```
___
*[cpp/open3d/t/geometry/CMakeLists.txt, line 50 at r2](https://reviewable.io/reviews/intel-isl/open3d/2898#-MRNP0vFCOyYNhQNG7br:-MRNP0vFCOyYNhQNG7bs:b-c62vfg) ([raw file](https://github.com/intel-isl/open3d/blob/151079dafdb35b50f87b0cb2b1e9a7283f99e9a3/cpp/open3d/t/geometry/CMakeLists.txt#L50)):*
> ```CMake
> 
> if (WITH_IPPICV)
>     target_compile_definitions(tgeometry PRIVATE WITH_IPPICV HAVE_IPPICV=true)
> ```

is there a way we can reduce this to one, e.g. only using `WITH_IPPICV`?

Future work: we may need to unify all macros in code to have `OPEN3D_` prefix to avoid collision with other libraries.
___
*[cpp/open3d/t/geometry/Image.h, line 170 at r2](https://reviewable.io/reviews/intel-isl/open3d/2898#-MRNTRoqEL4kuGyUsTMt:-MRNTRoqEL4kuGyUsTMu:bm566er) ([raw file](https://github.com/intel-isl/open3d/blob/151079dafdb35b50f87b0cb2b1e9a7283f99e9a3/cpp/open3d/t/geometry/Image.h#L170)):*
> ```objc
>             -std::numeric_limits<double>::max();
> 
>     /// Returns an Image with the specified \p Dtype.
> ```

`\p dtype`
___
*[cpp/open3d/t/geometry/Image.h, line 177 at r2](https://reviewable.io/reviews/intel-isl/open3d/2898#-MRNTTfq0pkab33iptEr:-MRNTTfq0pkab33iptEs:b-h0f0pj) ([raw file](https://github.com/intel-isl/open3d/blob/151079dafdb35b50f87b0cb2b1e9a7283f99e9a3/cpp/open3d/t/geometry/Image.h#L177)):*
> ```objc
>     /// \param copy If true, a new tensor is always created; if false, the copy
>     /// is avoided when the original tensor already has the targeted dtype.
>     Image ConvertTo(core::Dtype dtype,
> ```

Let's rename it to `Image::To` following https://github.com/intel-isl/Open3D/pull/2877 . 

It may be tricky to define `SCALE_DEFAULT`, would `optional` help here? E.g .
```
Image To(
    core::Dtype dtype,
    bool copy = false,
    utility::optional<double> scale = utility::nullopt,
    utility::optional<double> offset = utility::nullopt
) const;
```
(Moving `copy` to the front to be consistent with #2877. `offset` doesn't need to be `optional` but it might be good to keep the symmetry.)

Then the default scale can be handled in the function definition.
___
*[cpp/open3d/t/geometry/Image.cpp, line 102 at r3](https://reviewable.io/reviews/intel-isl/open3d/2898#-MRO82hZ2KAIiSsQLtFE:-MRO82hZ2KAIiSsQLtFF:br3zbcb) ([raw file](https://github.com/intel-isl/open3d/blob/655e9b7afaf14a377b219670c03df9d275a897c1/cpp/open3d/t/geometry/Image.cpp#L102)):*
> ```cpp
> 
>     auto new_data = data_.To(dtype, copy);
>     if (scale != 1.0) {
> ```

call `LinearTransform` here?
___
*[cpp/open3d/t/geometry/Image.cpp, line 103 at r3](https://reviewable.io/reviews/intel-isl/open3d/2898#-MRO7jRf6vOla50OR10R:-MRO7jRf6vOla50OR10S:b-37jlbx) ([raw file](https://github.com/intel-isl/open3d/blob/655e9b7afaf14a377b219670c03df9d275a897c1/cpp/open3d/t/geometry/Image.cpp#L103)):*
> ```cpp
>     auto new_data = data_.To(dtype, copy);
>     if (scale != 1.0) {
>         new_data *= scale;
> ```

Question: what are the considerations to cast dtype first before scaling, but not the reverse?
___
*[cpp/open3d/t/geometry/Image.cpp, line 154 at r3](https://reviewable.io/reviews/intel-isl/open3d/2898#-MRO8keI52cS-Ll8Yxvx:-MRO8keI52cS-Ll8Yxvy:b93dbfu) ([raw file](https://github.com/intel-isl/open3d/blob/655e9b7afaf14a377b219670c03df9d275a897c1/cpp/open3d/t/geometry/Image.cpp#L154)):*
> ```cpp
>         std::count(npp_supported.begin(), npp_supported.end(),
>                    std::make_pair(GetDtype(), GetChannels())) > 0) {
>         CUDA_CALL(npp::dilate, data_, dstim.data_, half_kernel_size);
> ```

- Seems like `CUDA_CALL` is used once, we may consider using `#ifdef` directly.
- Currently `CUDA_CALL` only takes care of a single function call, where the benefit is limited. It would be quite helpful if we have something that takes care of arbitrary multi-line code. Otherwise, it may be easier to just use the naive approach.
___
*[cpp/open3d/t/geometry/Image.cpp, line 159 at r3](https://reviewable.io/reviews/intel-isl/open3d/2898#-MROA9ua5lyED7Z46AWi:-MROA9ua5lyED7Z46AWj:ba5atub) ([raw file](https://github.com/intel-isl/open3d/blob/655e9b7afaf14a377b219670c03df9d275a897c1/cpp/open3d/t/geometry/Image.cpp#L159)):*
> ```cpp
>                std::count(ipp_supported.begin(), ipp_supported.end(),
>                           std::make_pair(GetDtype(), GetChannels())) > 0) {
>         IPP_CALL(ipp::dilate, data_, dstim.data_, half_kernel_size);
> ```

Same comments as `CUDA_CALL`.
___
*[cpp/open3d/t/geometry/Image.cpp, line 162 at r3](https://reviewable.io/reviews/intel-isl/open3d/2898#-MROAE4g1_ezqN6PRno5:-MROAE4g1_ezqN6PRno6:b-iaup5z) ([raw file](https://github.com/intel-isl/open3d/blob/655e9b7afaf14a377b219670c03df9d275a897c1/cpp/open3d/t/geometry/Image.cpp#L162)):*
> ```cpp
>     } else {
>         /*     call to core::kernel::UnaryWindowOp(); */
>         utility::LogError("Not implemented");
> ```

Is the ARM CI failing on this? If yes, we can actually mark the test as expected to fail on ARM. 
___
*[cpp/open3d/t/geometry/Image.cpp, line 164 at r3](https://reviewable.io/reviews/intel-isl/open3d/2898#-MROALZf4r_Cxos65sS3:-MROALZf4r_Cxos65sS4:b-t2dlbg) ([raw file](https://github.com/intel-isl/open3d/blob/655e9b7afaf14a377b219670c03df9d275a897c1/cpp/open3d/t/geometry/Image.cpp#L164)):*
> ```cpp
>         utility::LogError("Not implemented");
>     }
>     if (!dstim.data_.IsContiguous())
> ```

Add `{}`.  Maybe add comments on when `dstim.data_` can be non-contiguous (e.g. is it an internal error)? And, in this case, shall we make it contiguous? 
___
*[cpp/open3d/t/geometry/kernel/ippImage.h, line 63 at r3](https://reviewable.io/reviews/intel-isl/open3d/2898#-MROAwwYE5nnNFU-7hAY:-MROAwwYE5nnNFU-7hAZ:bh0ommy) ([raw file](https://github.com/intel-isl/open3d/blob/655e9b7afaf14a377b219670c03df9d275a897c1/cpp/open3d/t/geometry/kernel/ippImage.h#L63)):*
> ```objc
> }
> 
> void dilate(const open3d::core::Tensor &srcim,
> ```

`Dilate`
___
*[cpp/open3d/t/geometry/kernel/ippImage.cpp, line 48 at r3](https://reviewable.io/reviews/intel-isl/open3d/2898#-MROBMA87_Qz-3F0YCck:-MROBMA87_Qz-3F0YCcl:b-kqf952) ([raw file](https://github.com/intel-isl/open3d/blob/655e9b7afaf14a377b219670c03df9d275a897c1/cpp/open3d/t/geometry/kernel/ippImage.cpp#L48)):*
> ```cpp
>     // result in an exception if there are errors.
> 
>     // create nask
> ```

nit: 
```
// Create mask.
```
___
*[cpp/open3d/t/geometry/kernel/ippImage.cpp, line 49 at r3](https://reviewable.io/reviews/intel-isl/open3d/2898#-MROBW_QCab49j2YbkAR:-MROBW_QCab49j2YbkAS:b-27rn2i) ([raw file](https://github.com/intel-isl/open3d/blob/655e9b7afaf14a377b219670c03df9d275a897c1/cpp/open3d/t/geometry/kernel/ippImage.cpp#L49)):*
> ```cpp
> 
>     // create nask
>     core::Tensor mask(core::SizeVector{2 * half_kernel_size + 1,
> ```

`core::Tensor::One` can do it without `Fill`.
___
*[cpp/open3d/t/geometry/kernel/ippImage.cpp, line 54 at r3](https://reviewable.io/reviews/intel-isl/open3d/2898#-MROCKbECEoujq0xEgUz:-MROCKbECEoujq0xEgV-:bxdtwt8) ([raw file](https://github.com/intel-isl/open3d/blob/655e9b7afaf14a377b219670c03df9d275a897c1/cpp/open3d/t/geometry/kernel/ippImage.cpp#L54)):*
> ```cpp
>     mask.Fill(1);
> 
>     auto dt = srcim.GetDtype();
> ```

nit: `dtype` to be consistent with other locations.
___
*[cpp/open3d/t/geometry/kernel/ippImage.cpp, line 56 at r3](https://reviewable.io/reviews/intel-isl/open3d/2898#-MROBnUwCafCtgN4iGc-:-MROBnUwCafCtgN4iGc0:bury4q6) ([raw file](https://github.com/intel-isl/open3d/blob/655e9b7afaf14a377b219670c03df9d275a897c1/cpp/open3d/t/geometry/kernel/ippImage.cpp#L56)):*
> ```cpp
>     auto dt = srcim.GetDtype();
>     // Create IPP wrappers for all Open3D tensors
>     const ::ipp::IwiImage srcImage(
> ```

nit: the following name would be more consistent:
```
srcim    -> src_im
srcImage -> ipp_src_im
```

Same for `dst` and `mask`.
___
*[cpp/open3d/t/geometry/kernel/nppImage.h, line 36 at r3](https://reviewable.io/reviews/intel-isl/open3d/2898#-MRODONcEOcQ5dQkGy04:-MRODONcEOcQ5dQkGy05:b-nizcyy) ([raw file](https://github.com/intel-isl/open3d/blob/655e9b7afaf14a377b219670c03df9d275a897c1/cpp/open3d/t/geometry/kernel/nppImage.h#L36)):*
> ```objc
> namespace npp {
> 
> inline bool supported(core::Dtype dtype, size_t channels) {
> ```

Use `CamelCase` for function name. Also, shall we think of a more informative name? E.g. 
___
*[cpp/open3d/t/geometry/kernel/nppImage.h, line 48 at r3](https://reviewable.io/reviews/intel-isl/open3d/2898#-MROr7_f1v-PMXOIJXZ3:-MROr7_f1v-PMXOIJXZ4:b-qhz3cc) ([raw file](https://github.com/intel-isl/open3d/blob/655e9b7afaf14a377b219670c03df9d275a897c1/cpp/open3d/t/geometry/kernel/nppImage.h#L48)):*
> ```objc
> 
> #ifdef BUILD_CUDA_MODULE
> void dilate(const open3d::core::Tensor &srcim,
> ```

Same, `Dilate`
___
*[cpp/open3d/t/geometry/kernel/nppImage.cpp, line 49 at r3](https://reviewable.io/reviews/intel-isl/open3d/2898#-MROrK_a6l-BN01s9EgB:-MROrK_a6l-BN01s9EgC:b-8ggvdv) ([raw file](https://github.com/intel-isl/open3d/blob/655e9b7afaf14a377b219670c03df9d275a897c1/cpp/open3d/t/geometry/kernel/nppImage.cpp#L49)):*
> ```cpp
>     // result in an exception if there are errors.
> 
>     // create nask
> ```

`// Create mask.`
___
*[cpp/open3d/t/geometry/kernel/nppImage.cpp, line 50 at r3](https://reviewable.io/reviews/intel-isl/open3d/2898#-MROrN6M2DyC1kaIeRRi:-MROrN6N1gfdzBfmQend:bsm9cf3) ([raw file](https://github.com/intel-isl/open3d/blob/655e9b7afaf14a377b219670c03df9d275a897c1/cpp/open3d/t/geometry/kernel/nppImage.cpp#L50)):*
> ```cpp
> 
>     // create nask
>     core::Tensor mask(core::SizeVector{2 * half_kernel_size + 1,
> ```

`Tensor::Ones`
___
*[cpp/open3d/t/geometry/kernel/nppImage.cpp, line 54 at r3](https://reviewable.io/reviews/intel-isl/open3d/2898#-MROrOsI9LEKNK97XTUn:-MROrOsI9LEKNK97XTUo:b-xoin6w) ([raw file](https://github.com/intel-isl/open3d/blob/655e9b7afaf14a377b219670c03df9d275a897c1/cpp/open3d/t/geometry/kernel/nppImage.cpp#L54)):*
> ```cpp
>                       core::Dtype::UInt8, srcim.GetDevice());
>     mask.Fill(1);
>     NppiSize oMaskSize = {2 * half_kernel_size + 1, 2 * half_kernel_size + 1};
> ```

`snake_case` for variables.
___
*[cpp/open3d/t/geometry/kernel/nppImage.cpp, line 100 at r3](https://reviewable.io/reviews/intel-isl/open3d/2898#-MROrYn177uqT6PkWfwt:-MROrYn177uqT6PkWfwu:bavylcb) ([raw file](https://github.com/intel-isl/open3d/blob/655e9b7afaf14a377b219670c03df9d275a897c1/cpp/open3d/t/geometry/kernel/nppImage.cpp#L100)):*
> ```cpp
>             nppiDilateBorder_32f_C4R(NPP_ARGS);
>         }
>     } else
> ```

nit: else `{}`
___
*[cpp/pybind/t/geometry/image.cpp, line 133 at r3](https://reviewable.io/reviews/intel-isl/open3d/2898#-MROrsoI674-u5-K9hOb:-MROrsoI674-u5-K9hOc:b-xvo8mm) ([raw file](https://github.com/intel-isl/open3d/blob/655e9b7afaf14a377b219670c03df9d275a897c1/cpp/pybind/t/geometry/image.cpp#L133)):*
> ```cpp
> 
>     // Conversion.
>     image.def("convert_to", &Image::ConvertTo,
> ```

Rename to "to", `utility::nullopt` becomes `py::none()` in pybind
___
*[cpp/tests/t/geometry/Image.cpp, line 144 at r3](https://reviewable.io/reviews/intel-isl/open3d/2898#-MROtM8Z4bViTZsGnMXW:-MROtM8Z4bViTZsGnMXX:bxfg3rx) ([raw file](https://github.com/intel-isl/open3d/blob/655e9b7afaf14a377b219670c03df9d275a897c1/cpp/tests/t/geometry/Image.cpp#L144)):*
> ```cpp
>     t::geometry::Image output = input.ConvertTo(core::Dtype::Float32);
>     EXPECT_EQ(output.GetDtype(), core::Dtype::Float32);
>     EXPECT_THAT(output.AsTensor().ToFlatVector<float>(),
> ```

One approach is to use `Tensor::AlloClose()` so that we can compare tensors, instead of having to convert to a flat vector and compare the vectors. The other nice thing is that `AllClose` also checks for dtype and shape equality.

To compare float vectors, `ExpectEQ` shall be sufficient. It is defined in `cpp/tests/test_utility/Compare.h`. 
___
*[cpp/tests/t/geometry/Image.cpp, line 198 at r3](https://reviewable.io/reviews/intel-isl/open3d/2898#-MROteMg0fFWC7nmjEQ1:-MROteMg0fFWC7nmjEQ2:bb42lz8) ([raw file](https://github.com/intel-isl/open3d/blob/655e9b7afaf14a377b219670c03df9d275a897c1/cpp/tests/t/geometry/Image.cpp#L198)):*
> ```cpp
>         EXPECT_EQ(output.GetCols(), input.GetCols());
>         EXPECT_EQ(output.GetChannels(), input.GetChannels());
>         EXPECT_THAT(output.AsTensor().ToFlatVector<uint8_t>(),
> ```

We can extend `ExpectEQ` to support `uint8_t` and `uint16_t` if necessary.

ssheorey (commented):
Good catch. I think that happened when I tested on Windows and copied the changed files back to my Mac. Fixed.

> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2898#-:-MRRWd4U3D_LbOSzPsb7:b-krxdt4)* status: 7 of 22 files reviewed, 28 unresolved discussions (waiting on @reyanshsolis, @theNded, and @yxlao)

___
*[3rdparty/find_dependencies.cmake, line 1140 at r2](https://reviewable.io/reviews/intel-isl/open3d/2898#-MRNBbzq4a_UVv9P_hpo:-MRR_n70DOBpRbOXAhYo:b7fwuen) ([raw file](https://github.com/intel-isl/open3d/blob/151079dafdb35b50f87b0cb2b1e9a7283f99e9a3/3rdparty/find_dependencies.cmake#L1140)):*

shall we add arm64 detection to switch off `WITH_IPPICV` automatically?
</blockquote></details>

That currently happens inside `ippicv.cmake`. Added it here for clarity.
___
*[3rdparty/find_dependencies.cmake, line 1143 at r2](https://reviewable.io/reviews/intel-isl/open3d/2898#-MRNFbej0xIn53owTZSb:-MRRhcNY-ZlHEs-4b_Hx:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/151079dafdb35b50f87b0cb2b1e9a7283f99e9a3/3rdparty/find_dependencies.cmake#L1143)):*

We can concatenate `IPPICV_INCLUDE_DIR` and `IPPIW_INCLUDE_DIR` in `ippicv.cmake`, such that we only need to use `IPPICV_INCLUDE_DIR` here. This would be more consistent with the usage of other libraries.

Same for the `LIBRARIES` property.
</blockquote></details>

Done.
___
*[3rdparty/ippicv/ippicv.cmake, line 26 at r2](https://reviewable.io/reviews/intel-isl/open3d/2898#-MRNGNZi5A-tO4grIT28:-MRRYF0078KXYFop0kJl:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/151079dafdb35b50f87b0cb2b1e9a7283f99e9a3/3rdparty/ippicv/ippicv.cmake#L26)):*

shall we use `FATAL_ERROR` here as well?
</blockquote></details>

Done.
___
*[3rdparty/ippicv/ippicv.cmake, line 40 at r2](https://reviewable.io/reviews/intel-isl/open3d/2898#-MRNH2-R7dtMe7e7Wy9Z:-MRRYLEv0qtI_AaEk5JG:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/151079dafdb35b50f87b0cb2b1e9a7283f99e9a3/3rdparty/ippicv/ippicv.cmake#L40)):*

Perhaps `PATCH_COMMAND` would be more clear here, since it is our own patch.

From https://cmake.org/cmake/help/v3.19/module/ExternalProject.html, it seems like `UPDATE_COMMAND` refers to source code version updates.

> by default the external project’s sources will be updated if the download method supports updates **(e.g. a git repository would be checked if the GIT_TAG does not refer to a specific commit).**
</blockquote></details>

Done.
___
*[cpp/open3d/core/Tensor.h, line 761 at r2](https://reviewable.io/reviews/intel-isl/open3d/2898#-MRNIa_v7Fu-sYT2uhPh:-MRRiJuU5S52A9QI0ERp:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/151079dafdb35b50f87b0cb2b1e9a7283f99e9a3/cpp/open3d/core/Tensor.h#L761)):*

- Nit: `Element-wise`
- Nit: `returning a new tensor.` (with `.`)
- Add `\param min_val` and `\param max_val`, such that `\p` works
</blockquote></details>

Done.
___
*[cpp/open3d/core/Tensor.cpp, line 908 at r2](https://reviewable.io/reviews/intel-isl/open3d/2898#-MRNLb1D2oXPZO6PD8sY:-MRRivD_98CkpY7H34cw:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/151079dafdb35b50f87b0cb2b1e9a7283f99e9a3/cpp/open3d/core/Tensor.cpp#L908)):*

remove comments?
</blockquote></details>

Done.
___
*[cpp/open3d/core/Tensor.cpp, line 914 at r2](https://reviewable.io/reviews/intel-isl/open3d/2898#-MRNLc8P23IKynpVK9UQ:-MRRivmj5hg4y-49RIpJ:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/151079dafdb35b50f87b0cb2b1e9a7283f99e9a3/cpp/open3d/core/Tensor.cpp#L914)):*

remove comments?
</blockquote></details>

Done.
___
*[cpp/open3d/t/geometry/CMakeLists.txt, line 12 at r2](https://reviewable.io/reviews/intel-isl/open3d/2898#-MRNLkds6jPTGh4_HvFn:-MRRYmNl11HRRDwf_1T0:b-x3d1h4) ([raw file](https://github.com/intel-isl/open3d/blob/151079dafdb35b50f87b0cb2b1e9a7283f99e9a3/cpp/open3d/t/geometry/CMakeLists.txt#L12)):*

By our naming convention, we can call it `NPPImage` or `NppImage`. Same for `ipp`.
</blockquote></details>

Renamed to NPPImage and IPPImage
___
*[cpp/open3d/t/geometry/CMakeLists.txt, line 37 at r2](https://reviewable.io/reviews/intel-isl/open3d/2898#-MRNM7Rs27bWIJIVVG8c:-MRRmDjU9yhNwYt417p0:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/151079dafdb35b50f87b0cb2b1e9a7283f99e9a3/cpp/open3d/t/geometry/CMakeLists.txt#L37)):*

This part looks quite convoluted. We actually only need one variable `T_GEOMETRY_SRC` and we can just append to it with `set` instead of `list(APPEND`. We also don't need to distinguish the `KERNEL` or `IPP` sources.

Something like:
```
set(T_GEOMETRY_SRC 
    cpu_source_0.cpp
    cpu_source_1.cpp
)

if(BUILD_CUDA_MODULE)
    set(T_GEOMETRY_SRC
        ${T_GEOMETRY_SRC}
        cuda_source_0.cpp
        cuda_source_1.cpp
    )
endif()

if(WITH_IPPICV)
    set(T_GEOMETRY_SRC
        ${T_GEOMETRY_SRC}
        ipp_source_0.cpp
        ipp_source_1.cpp
    )
endif()
```
</blockquote></details>

Done.
___
*[cpp/open3d/t/geometry/CMakeLists.txt, line 50 at r2](https://reviewable.io/reviews/intel-isl/open3d/2898#-MRNP0vFCOyYNhQNG7br:-MRRta2Y3MFyO1pH-sQ2:b-tx2mh0) ([raw file](https://github.com/intel-isl/open3d/blob/151079dafdb35b50f87b0cb2b1e9a7283f99e9a3/cpp/open3d/t/geometry/CMakeLists.txt#L50)):*

is there a way we can reduce this to one, e.g. only using `WITH_IPPICV`?

Future work: we may need to unify all macros in code to have `OPEN3D_` prefix to avoid collision with other libraries.
</blockquote></details>

Done. Moved HAVE_IPPICV to a `static const bool` member of `t::Geoemetry::Image`
___
*[cpp/open3d/t/geometry/Image.h, line 170 at r2](https://reviewable.io/reviews/intel-isl/open3d/2898#-MRNTRoqEL4kuGyUsTMt:-MRRuiax9o4Oj0L3l_oC:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/151079dafdb35b50f87b0cb2b1e9a7283f99e9a3/cpp/open3d/t/geometry/Image.h#L170)):*

`\p dtype`
</blockquote></details>

Done.
___
*[cpp/open3d/t/geometry/Image.h, line 177 at r2](https://reviewable.io/reviews/intel-isl/open3d/2898#-MRNTTfq0pkab33iptEr:-MRRxpBlCRA43B7MW6CR:bicm4nc) ([raw file](https://github.com/intel-isl/open3d/blob/151079dafdb35b50f87b0cb2b1e9a7283f99e9a3/cpp/open3d/t/geometry/Image.h#L177)):*

Let's rename it to `Image::To` following https://github.com/intel-isl/Open3D/pull/2877 . 

It may be tricky to define `SCALE_DEFAULT`, would `optional` help here? E.g .
```
Image To(
    core::Dtype dtype,
    bool copy = false,
    utility::optional<double> scale = utility::nullopt,
    utility::optional<double> offset = utility::nullopt
) const;
```
(Moving `copy` to the front to be consistent with #2877. `offset` doesn't need to be `optional` but it might be good to keep the symmetry.)

Then the default scale can be handled in the function definition.
</blockquote></details>

Done (for scale. Kept offset default at 0).
___
*[cpp/open3d/t/geometry/Image.cpp, line 102 at r3](https://reviewable.io/reviews/intel-isl/open3d/2898#-MRO82hZ2KAIiSsQLtFE:-MRS0G_797lJ6cvtQ4pH:b-xeoz13) ([raw file](https://github.com/intel-isl/open3d/blob/655e9b7afaf14a377b219670c03df9d275a897c1/cpp/open3d/t/geometry/Image.cpp#L102)):*

call `LinearTransform` here?
</blockquote></details>

Rewrote `LinearTransform()` in terms of `To()`, since `To()` is more general.
___
*[cpp/open3d/t/geometry/Image.cpp, line 103 at r3](https://reviewable.io/reviews/intel-isl/open3d/2898#-MRO7jRf6vOla50OR10R:-MRup7jI2eX-wDckgjo-:bps8mo1) ([raw file](https://github.com/intel-isl/open3d/blob/655e9b7afaf14a377b219670c03df9d275a897c1/cpp/open3d/t/geometry/Image.cpp#L103)):*

Question: what are the considerations to cast dtype first before scaling, but not the reverse?
</blockquote></details>

Yes, that's a good point. The correct way to do data type conversion for Images also involves a `saturate_cast` (e.g. in OpenCV). This ensures that overflow and underflow are clamped to the min and max values of the data type ranges properly. I've rewritten `To()` to use the IPP `iwiScale()` that correctly implements this. For NPP, the API is a bit unfriendly (Scale does not admit user supplied `scale` parameter, need separate calls for Mul and Add, cannot scale with a float value for `int` data types), so for now I'm copying data to the CPU, using IPP `iwiScale` and copying it back to CUDA. I'll replace this with a Tensor based implementation in a follow up PR.
___
*[cpp/open3d/t/geometry/Image.cpp, line 154 at r3](https://reviewable.io/reviews/intel-isl/open3d/2898#-MRO8keI52cS-Ll8Yxvx:-MRSSsc822LNVpPep0xJ:bqd155u) ([raw file](https://github.com/intel-isl/open3d/blob/655e9b7afaf14a377b219670c03df9d275a897c1/cpp/open3d/t/geometry/Image.cpp#L154)):*

- Seems like `CUDA_CALL` is used once, we may consider using `#ifdef` directly.
- Currently `CUDA_CALL` only takes care of a single function call, where the benefit is limited. It would be quite helpful if we have something that takes care of arbitrary multi-line code. Otherwise, it may be easier to just use the naive approach.
</blockquote></details>

`CUDA_CALL` and `IPP_CALL` will be used by each of the image processing ops later (this PR only implements a sample op). These macros replace the `#ifdef  BUILD_CUDA_MODULE ... cuda_function(); ...  #endif` (3 lines) with a single line and result in cleaner code in the case of a single function call. For multiple lines with CUDA specific code, the current `#ifdef ... #endif` should be fine, since we won't save relatively many lines of code. 
___
*[cpp/open3d/t/geometry/Image.cpp, line 159 at r3](https://reviewable.io/reviews/intel-isl/open3d/2898#-MROA9ua5lyED7Z46AWi:-MRSUgFV4iq8YERLi6-v:b-v22umm) ([raw file](https://github.com/intel-isl/open3d/blob/655e9b7afaf14a377b219670c03df9d275a897c1/cpp/open3d/t/geometry/Image.cpp#L159)):*

Same comments as `CUDA_CALL`.
</blockquote></details>

We should have more uses for IPP_CALL as well as we implement more image processing ops.
___
*[cpp/open3d/t/geometry/Image.cpp, line 162 at r3](https://reviewable.io/reviews/intel-isl/open3d/2898#-MROAE4g1_ezqN6PRno5:-MRRZFtB67b1h44lUwKb:b-s6gdhm) ([raw file](https://github.com/intel-isl/open3d/blob/655e9b7afaf14a377b219670c03df9d275a897c1/cpp/open3d/t/geometry/Image.cpp#L162)):*

Is the ARM CI failing on this? If yes, we can actually mark the test as expected to fail on ARM. 
</blockquote></details>

On ARM, an exception is expected by GoogleTest, so the CI should still pass. It timed out at 6 hours the last 2 times on this branch, though. I'll look into it.
___
*[cpp/open3d/t/geometry/Image.cpp, line 164 at r3](https://reviewable.io/reviews/intel-isl/open3d/2898#-MROALZf4r_Cxos65sS3:-MRSVMBFEUTJQkXGvIgN:beew9fe) ([raw file](https://github.com/intel-isl/open3d/blob/655e9b7afaf14a377b219670c03df9d275a897c1/cpp/open3d/t/geometry/Image.cpp#L164)):*

Add `{}`.  Maybe add comments on when `dstim.data_` can be non-contiguous (e.g. is it an internal error)? And, in this case, shall we make it contiguous? 
</blockquote></details>

That's not necessary. Removed.
___
*[cpp/open3d/t/geometry/kernel/ippImage.h, line 63 at r3](https://reviewable.io/reviews/intel-isl/open3d/2898#-MROAwwYE5nnNFU-7hAY:-MRSVpnz8SCgYpxqMAw1:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/655e9b7afaf14a377b219670c03df9d275a897c1/cpp/open3d/t/geometry/kernel/ippImage.h#L63)):*

`Dilate`
</blockquote></details>

Done.
___
*[cpp/open3d/t/geometry/kernel/ippImage.cpp, line 49 at r3](https://reviewable.io/reviews/intel-isl/open3d/2898#-MROBW_QCab49j2YbkAR:-MRSWUHi9oT3ihxJCSVx:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/655e9b7afaf14a377b219670c03df9d275a897c1/cpp/open3d/t/geometry/kernel/ippImage.cpp#L49)):*

`core::Tensor::One` can do it without `Fill`.
</blockquote></details>

Done.
___
*[cpp/open3d/t/geometry/kernel/nppImage.h, line 36 at r3](https://reviewable.io/reviews/intel-isl/open3d/2898#-MRODONcEOcQ5dQkGy04:-MRSXmYHEbyzjtizNFnK:bll9vi5) ([raw file](https://github.com/intel-isl/open3d/blob/655e9b7afaf14a377b219670c03df9d275a897c1/cpp/open3d/t/geometry/kernel/nppImage.h#L36)):*

Use `CamelCase` for function name. Also, shall we think of a more informative name? E.g. 
</blockquote></details>

We don't need that function any more, since checking for support should be done separately for each Image op. Removed.
___
*[cpp/open3d/t/geometry/kernel/nppImage.h, line 48 at r3](https://reviewable.io/reviews/intel-isl/open3d/2898#-MROr7_f1v-PMXOIJXZ3:-MRSYQVAADnvMg0Xdaai:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/655e9b7afaf14a377b219670c03df9d275a897c1/cpp/open3d/t/geometry/kernel/nppImage.h#L48)):*

Same, `Dilate`
</blockquote></details>

Done.
___
*[cpp/open3d/t/geometry/kernel/nppImage.cpp, line 49 at r3](https://reviewable.io/reviews/intel-isl/open3d/2898#-MROrK_a6l-BN01s9EgB:-MRSYRw42ypL1yJBbndx:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/655e9b7afaf14a377b219670c03df9d275a897c1/cpp/open3d/t/geometry/kernel/nppImage.cpp#L49)):*

`// Create mask.`
</blockquote></details>

Done.
___
*[cpp/open3d/t/geometry/kernel/nppImage.cpp, line 50 at r3](https://reviewable.io/reviews/intel-isl/open3d/2898#-MROrN6M2DyC1kaIeRRi:-MRSYSbq8hAiXQpCLDfV:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/655e9b7afaf14a377b219670c03df9d275a897c1/cpp/open3d/t/geometry/kernel/nppImage.cpp#L50)):*

`Tensor::Ones`
</blockquote></details>

Done.
___
*[cpp/open3d/t/geometry/kernel/nppImage.cpp, line 54 at r3](https://reviewable.io/reviews/intel-isl/open3d/2898#-MROrOsI9LEKNK97XTUn:-MRSY_0s8TLwEPzG6OkJ:b3qtlkx) ([raw file](https://github.com/intel-isl/open3d/blob/655e9b7afaf14a377b219670c03df9d275a897c1/cpp/open3d/t/geometry/kernel/nppImage.cpp#L54)):*

`snake_case` for variables.
</blockquote></details>

I was using names from the NPP and IPP example code. Updated all to `snake_case`.
___
*[cpp/pybind/t/geometry/image.cpp, line 133 at r3](https://reviewable.io/reviews/intel-isl/open3d/2898#-MROrsoI674-u5-K9hOb:-MRS_aYn8x2v9JXTfOEt:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/655e9b7afaf14a377b219670c03df9d275a897c1/cpp/pybind/t/geometry/image.cpp#L133)):*

Rename to "to", `utility::nullopt` becomes `py::none()` in pybind
</blockquote></details>

Done.
___
*[cpp/tests/t/geometry/Image.cpp, line 144 at r3](https://reviewable.io/reviews/intel-isl/open3d/2898#-MROtM8Z4bViTZsGnMXW:-MRfVfXH2DeApQe9PHql:b-jpepwh) ([raw file](https://github.com/intel-isl/open3d/blob/655e9b7afaf14a377b219670c03df9d275a897c1/cpp/tests/t/geometry/Image.cpp#L144)):*

One approach is to use `Tensor::AlloClose()` so that we can compare tensors, instead of having to convert to a flat vector and compare the vectors. The other nice thing is that `AllClose` also checks for dtype and shape equality.

To compare float vectors, `ExpectEQ` shall be sufficient. It is defined in `cpp/tests/test_utility/Compare.h`. 
</blockquote></details>

The `GoogleMock` `EXPECT_THAT` and `ElementsAreArray` has a useful error message format for easy debugging. With `AllClose()`this debugging info may not be available in the error message. Here's an example:

/Users/ssheorey/Documents/Open3D/Code/Open3D/cpp/tests/t/geometry/Image.cpp:161: Failure
Value of: output.AsTensor().ToFlatVector<uint8_t>()
Expected: has 32 elements where
element #0 is equal to '\x1' (1),
element #1 is equal to '\x1' (1),
element #2 is equal to '\x1' (1),
element #3 is equal to '\0',
element #4 is equal to '\0',
element #5 is equal to '\x1' (1),
element #6 is equal to '\x1' (1),
element #7 is equal to '\x1' (1),
element #8 is equal to '\x1' (1),
element #9 is equal to '\x1' (1),
element #10 is equal to '\x1' (1),
element #11 is equal to '\x1' (1),
element #12 is equal to '\0',
element #13 is equal to '\x1' (1),
element #14 is equal to '\x1' (1),
element #15 is equal to '\x1' (1),
element #16 is equal to '\x1' (1),
element #17 is equal to '\x1' (1),
element #18 is equal to '\x1' (1),
element #19 is equal to '\x1' (1),
element #20 is equal to '\0',
element #21 is equal to '\x1' (1),
element #22 is equal to '\x1' (1),
element #23 is equal to '\x1' (1),
element #24 is equal to '\0',
element #25 is equal to '\x1' (1),
element #26 is equal to '\x1' (1),
element #27 is equal to '\x1' (1),
element #28 is equal to '\0',
element #29 is equal to '\0',
element #30 is equal to '\0',
element #31 is equal to '\0'
  Actual: { '\0', '\0', '\0', '\0', '\0', '\0', '\0', '\0', '\x1' (1), '\x1' (1), '\0', '\0', '\0', '\0', '\x1' (1), '\0', '\0', '\0', '\x1' (1), '\0', '\0', '\0', '\0', '\0', '\0', '\0', '\0', '\0', '\0', '\0', '\0', '\0' }, whose element #0 doesn't match
___
*[cpp/tests/t/geometry/Image.cpp, line 198 at r3](https://reviewable.io/reviews/intel-isl/open3d/2898#-MROteMg0fFWC7nmjEQ1:-MRuobL_2KFQym13qJzw:bssvtcz) ([raw file](https://github.com/intel-isl/open3d/blob/655e9b7afaf14a377b219670c03df9d275a897c1/cpp/tests/t/geometry/Image.cpp#L198)):*

We can extend `ExpectEQ` to support `uint8_t` and `uint16_t` if necessary.
</blockquote></details>

I can switch to ExpectEQ if its a priority, else will take a look at this later.

reyanshsolis (changes requested):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2898)* status: 5 of 25 files reviewed, 29 unresolved discussions (waiting on @ssheorey, @theNded, and @yxlao)

___
*[cpp/open3d/t/geometry/Image.h, line 177 at r2](https://reviewable.io/reviews/intel-isl/open3d/2898#-MRNTTfq0pkab33iptEr:-MS3BtvCDrXAQS1N1PhM:b-ayu27) ([raw file](https://github.com/intel-isl/open3d/blob/151079dafdb35b50f87b0cb2b1e9a7283f99e9a3/cpp/open3d/t/geometry/Image.h#L177)):*

Done (for scale. Kept offset default at 0).
</blockquote></details>

Can we support overloading "To" functions for device transfers? 
To(device), To(device, dtype, copy), To(dtype, copy, scale, offset). 
___
*[cpp/open3d/t/geometry/Image.h, line 164 at r5](https://reviewable.io/reviews/intel-isl/open3d/2898#-MS3CW3A1UIbDr3p9izm:-MS3CW3A1UIbDr3p9izn:b-9zmyu1) ([raw file](https://github.com/intel-isl/open3d/blob/6fe2f0294054e15dc3c50e5b2add14af1099107a/cpp/open3d/t/geometry/Image.h#L164)):*
> ```objc
>     /// \param half_kernel_size A dilation mask of size 2*half_kernel_size+1 is
>     /// used.
>     Image Dilate(int half_kernel_size = 1) const;
> ```

Can we also add in-place functions for such operations, such as Dilate_ that will modify *this itself ?

ssheorey (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2898)* status: 5 of 27 files reviewed, 29 unresolved discussions (waiting on @reyanshsolis, @theNded, and @yxlao)

___
*[cpp/open3d/t/geometry/Image.h, line 177 at r2](https://reviewable.io/reviews/intel-isl/open3d/2898#-MRNTTfq0pkab33iptEr:-MS4P8PP6LGO4-mwU4cn:bv04u8e) ([raw file](https://github.com/intel-isl/open3d/blob/151079dafdb35b50f87b0cb2b1e9a7283f99e9a3/cpp/open3d/t/geometry/Image.h#L177)):*

Can we support overloading "To" functions for device transfers? 
To(device), To(device, dtype, copy), To(dtype, copy, scale, offset). 
</blockquote></details>

Added `To(device)`, `CPU()`, `CUDA()`, `Clone()` (+ tests) so we now have equivalent functionality in `PointCloud` and `Image`.  Can add `To(device, dtype)` in another PR if there is need.
___
*[cpp/open3d/t/geometry/Image.h, line 164 at r5](https://reviewable.io/reviews/intel-isl/open3d/2898#-MS3CW3A1UIbDr3p9izm:-MS4Q49C6DOfUl6oE-oM:bgg24n4) ([raw file](https://github.com/intel-isl/open3d/blob/6fe2f0294054e15dc3c50e5b2add14af1099107a/cpp/open3d/t/geometry/Image.h#L164)):*

Can we also add in-place functions for such operations, such as Dilate_ that will modify *this itself ?
</blockquote></details>

Dilate (like Filter) is a neighbourhood operation and can't be done in place.

reyanshsolis (changes requested):
> Reviewed 13 of 20 files at r5, 9 of 9 files at r6.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2898)* status: all files reviewed, 29 unresolved discussions (waiting on @ssheorey, @theNded, and @yxlao)

___
*[cpp/open3d/t/geometry/Image.h, line 177 at r2](https://reviewable.io/reviews/intel-isl/open3d/2898#-MRNTTfq0pkab33iptEr:-MS6135v8JyDaZEmUl87:b-ewk8my) ([raw file](https://github.com/intel-isl/open3d/blob/151079dafdb35b50f87b0cb2b1e9a7283f99e9a3/cpp/open3d/t/geometry/Image.h#L177)):*

Added `To(device)`, `CPU()`, `CUDA()`, `Clone()` (+ tests) so we now have equivalent functionality in `PointCloud` and `Image`.  Can add `To(device, dtype)` in another PR if there is need.
</blockquote></details>

Thanks. This is helpful.
___
*[cpp/open3d/t/geometry/Image.h, line 164 at r5](https://reviewable.io/reviews/intel-isl/open3d/2898#-MS3CW3A1UIbDr3p9izm:-MS61Ovk31TgTDOjkgzh:b43ejbi) ([raw file](https://github.com/intel-isl/open3d/blob/6fe2f0294054e15dc3c50e5b2add14af1099107a/cpp/open3d/t/geometry/Image.h#L164)):*

Dilate (like Filter) is a neighbourhood operation and can't be done in place.
</blockquote></details>

right! understood, thanks for the clarification. 
___
*[cpp/open3d/t/geometry/kernel/IPPImage.h, line 47 at r6](https://reviewable.io/reviews/intel-isl/open3d/2898#-MS62jzO1usuLWCe6xAc:-MS62jzP9lKo5EZsZyOf:b-jwjxry) ([raw file](https://github.com/intel-isl/open3d/blob/b3e2c2a39461178645ed1def8f7c3b3c66c0a75e/cpp/open3d/t/geometry/kernel/IPPImage.h#L47)):*
> ```objc
>     if (dtype == core::Dtype::UInt8 || dtype == core::Dtype::Bool) {
>         return ipp8u;
>     } else if (dtype == core::Dtype::UInt16) {
> ```

`else if (dtype == core::Dtype::UInt16) { return ipp16u; }` repeated twice.

ssheorey (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2898)* status: all files reviewed, 29 unresolved discussions (waiting on @reyanshsolis, @theNded, and @yxlao)

___
*[cpp/open3d/t/geometry/kernel/IPPImage.h, line 47 at r6](https://reviewable.io/reviews/intel-isl/open3d/2898#-MS62jzO1usuLWCe6xAc:-MS68tX1C4sTcfqGBFMc:b4x3gpo) ([raw file](https://github.com/intel-isl/open3d/blob/b3e2c2a39461178645ed1def8f7c3b3c66c0a75e/cpp/open3d/t/geometry/kernel/IPPImage.h#L47)):*

`else if (dtype == core::Dtype::UInt16) { return ipp16u; }` repeated twice. 
</blockquote></details>

Thanks. Will remove.

yxlao (approved):
> Reviewed 12 of 20 files at r5, 8 of 9 files at r6, 2 of 2 files at r7.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2898)* status: all files reviewed, 29 unresolved discussions (waiting on @reyanshsolis, @theNded, and @yxlao)

PR #2902, opened 2021-01-15
Update realsense.rst

Rubix982:
The file `download_dataset.py` is formatted according to python3's FString syntax. The command used here should be `python3`, instead of `python`.

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/2902)

### DIFF
--- a/docs/tutorial/sensor/realsense.rst
+++ b/docs/tutorial/sensor/realsense.rst
@@ -39,7 +39,7 @@ Sample RealSense bag files
 
 You con download sample RealSense bag datasets with this script::
 
-    python examples/python/reconstruction_system/scripts/download_dataset.py L515_test
+    python3 examples/python/reconstruction_system/scripts/download_dataset.py L515_test
 
 Check the script for more RS bag datasets.
 


Rubix982:
Sure, I'll update the CHANGELOG.md if this PR is approved. Thanks!

yxlao:
@Rubix982 thanks for the PR. Typically we expect the user to use a virtual environment with python3. We are considering adding a global warning message to warn users if their python version is not compatible.

Rubix982:
@yxlao I actually had to follow this pattern. I ran the `pip` command for installing the dependencies ( `pip install open3d` ), then, when I ran the python command for the file, I noticed that it demanded I use a Python version `>=3.6.5`. So I went back, deleted my `virtualenv`, then I ran the `pip` command again, this time replacing time with `pip3`, and also installing `cython` as well, then I ran the `python3`, command, which worked perfectly now.

germanros1987:
@yxlao I guess we could make explicit that our examples are python3 only...

Rubix982:
@germanros1987, It depends on whether the project decides to be backwards compatible or just shift to `python3`

PR #2905, opened 2021-01-15
Recenter rotation to point on double-click

prewettg:
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2905)


cpp/open3d/visualization/rendering/filament/FilamentCamera.cpp
```diff
@@ -253,9 +242,14 @@ Camera::Transform FilamentCamera::GetViewMatrix() const {
     return FilamentToCameraTransform(ftransform);
 }
 
-Camera::Transform FilamentCamera::GetProjectionMatrix() const {
-    auto ftransform = camera_->getProjectionMatrix();  // mat4 (not mat4f)
-    return FilamentToCameraTransform(ftransform);
+Camera::ProjectionMatrix FilamentCamera::GetProjectionMatrix() const {
+    auto ft = camera_->getProjectionMatrix();  // mat4 (not mat4f)
+    Camera::ProjectionMatrix::MatrixType proj;
+    proj << float(ft(0, 0)), float(ft(0, 1)), float(ft(0, 2)), float(ft(0, 3)),
```
pablospe, line 248: maybe with double(...) or static_cast<double>(...) ?

cpp/open3d/visualization/rendering/filament/FilamentRenderToBuffer.cpp
```diff
@@ -133,9 +137,11 @@ void FilamentRenderToBuffer::SetDimensions(const std::uint32_t width,
     width_ = width;
     height_ = height;
 
-    buffer_size_ = width * height * 3 * sizeof(std::uint8_t);
+    buffer_size_ = width * height * n_channels_ * sizeof(std::uint8_t);
     if (buffer_) {
         buffer_ = static_cast<std::uint8_t*>(realloc(buffer_, buffer_size_));
+    } else {
+        buffer_ = static_cast<std::uint8_t*>(malloc(buffer_size_));
```
pablospe, line 144: shouldn't be uint16_t? If the depth buffer are encoding float numbers...

prewettg: No, we're reading from the color buffer, which is a uint8_t*. The shader encodes the (most significant 24-bits of the) float into the R, G, and B channels. As far as OpenGL and Filament are concerned, this is an ordinary RGBA32 color buffer.

cpp/open3d/visualization/rendering/filament/FilamentCamera.cpp
prewettg: If you look into the guts, Camera::ProjectionMatrix::MatrixType ends up being a float matrix, not a double. Also, I personally hate static_cast (and think that the usually stated reason for using it, "easier to search for" is lame, because I have never once done or heard of anyone searching for casts, while on the negative side it uses underscores which are hard to type, masquerade as spaces visually which has a one pixel line as differentiator between a real space which ends a token, and takes up a lot of space in lines which are usually already pretty cramped for space), and in this case it makes things much less readable, since you can fit the 4 columns of the matrix in one line.

pablospe:
Great this has been merge to master! When is the next version 0.12.1 is coming out?

cpp/pybind/visualization/rendering/rendering.cpp
```diff
@@ -119,7 +123,12 @@ void pybind_rendering_classes(py::module &m) {
                     "be accessed after that point.")
             .def("render_to_image", &PyOffscreenRenderer::RenderToImage,
                  "Renders scene to an image, blocking until the image is "
-                 "returned");
+                 "returned")
+            .def("render_to_depth_image",
+                 &PyOffscreenRenderer::RenderToDepthImage,
+                 "Renders scene depth buffer to a float image, blocking until "
+                 "the image is returned. Pixels range from 0 (near plane) to "
+                 "1 (far plane)");
```
pablospe, line 131: @prewettg 
Pixels range from 0 (near plane) to 1 (far plane). Then, how do you obtain the actual depth value? Using the near and far plane? It seems there is a missing interface in python for this, or how can you get the near/far plane of the camera (in python)?

PR #2907, opened 2021-01-17
Jsoncpp with ExternalProject

yxlao:
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2907)


ssheorey (approved):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2907)* status: 0 of 25 files reviewed, 1 unresolved discussion (waiting on @yxlao)

___
*[3rdparty/jsoncpp/jsoncpp.cmake, line 7 at r2](https://reviewable.io/reviews/intel-isl/open3d/2907#-MRYFUvjDWkcPmhCKqOD:-MRYFUvjDWkcPmhCKqOE:b-vpajlo) ([raw file](https://github.com/intel-isl/open3d/blob/13e4b93b69d9b1483992d0940cc1eeb61181c1fb/3rdparty/jsoncpp/jsoncpp.cmake#L7)):*
> ```CMake
>     PREFIX jsoncpp
>     GIT_REPOSITORY https://github.com/open-source-parsers/jsoncpp.git
>     GIT_TAG 1.9.4
> ```

Switching from git to URL download (`https://github.com/open-source-parsers/jsoncpp/archive/1.9.4.zip`) allows using local URL for caching 3rd party downloads. Also download should be faster since we don't download git history.

yxlao (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2907)* status: 0 of 25 files reviewed, 1 unresolved discussion (waiting on @ssheorey)

___
*[3rdparty/jsoncpp/jsoncpp.cmake, line 7 at r2](https://reviewable.io/reviews/intel-isl/open3d/2907#-MRYFUvjDWkcPmhCKqOD:-MRYiKLY18PzjvdpR-cK:bdsg2qd) ([raw file](https://github.com/intel-isl/open3d/blob/13e4b93b69d9b1483992d0940cc1eeb61181c1fb/3rdparty/jsoncpp/jsoncpp.cmake#L7)):*

Switching from git to URL download (`https://github.com/open-source-parsers/jsoncpp/archive/1.9.4.zip`) allows using local URL for caching 3rd party downloads. Also download should be faster since we don't download git history.
</blockquote></details>

That's a good point. In this case, the main limitation is the `git apply` command which only works in a `.git` directory. If we download the `.zip` file, we will need to find another patching mechanism that works across the different OS.

I've added a `GIT_SHALLOW ON` to avoid downloading the git history. Also in this case, `jsoncpp` is relatively small. 

In the future, agreed that we could by default download the zip file.

PR #2920, opened 2021-01-20
[Draft] Changes rotation to be like PPTK

prewettg:
Rotation was rotation(dx * right-axis + dy * up-aixs), is not rotation(right-axis) * rotation(up-axis).

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2920)


prewettg:
Obsoleted by #2920

PR #2934, opened 2021-01-22
function for casting the tensor data pointer to a datatype

ShubhamAgarwal12:
Cast output of the tensor.GetDataPointer() implicitly in a function.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2934)


cpp/open3d/core/nns/FaissIndex.cpp
```diff
@@ -109,16 +109,15 @@ std::pair<Tensor, Tensor> FaissIndex::SearchKnn(const Tensor &query_points,
     int64_t num_query_points = query_points.GetShape()[0];
     knn = std::min(knn, (int)GetDatasetSize());
 
-    auto *data_ptr = static_cast<const float *>(query_points.GetDataPtr());
+    auto *data_ptr = query_points.GetDataPtr<float>();
```
yxlao, line 112: Could you double-check if we actually need/not-need the `const`? 
- Here the `const` is missing but in other locations, I see that we still keep the `const`.
- We also have two getters already: `const T* GetDataPtr()` and `T* GetDataPtr()`, can the `const` be somehow determined automatically by the compiler?

yxlao (changes requested):

cpp/open3d/core/nns/FaissIndex.cpp
ShubhamAgarwal12: This is True. The const can be automatically determined by the compiler. I've removed the const when using GetDataPtr

cpp/open3d/core/Tensor.h
```diff
@@ -1093,6 +1093,16 @@ class Tensor {
         return strides_[shape_util::WrapDim(dim, NumDims())];
     }
 
+    template <typename T>
+    inline T* GetDataPtr() {
+        return static_cast<T*>(data_ptr_);
```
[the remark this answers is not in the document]
ShubhamAgarwal12: Done.

yxlao (approved):

PR #2935, opened 2021-01-22
Python example of using a texture on a Material

errissa:
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2935)

### DIFF
--- a/examples/python/gui/textured-model.py
+++ b/examples/python/gui/textured-model.py
@@ -5,9 +5,12 @@ import open3d.visualization.gui as gui
 import open3d.visualization.rendering as rendering
 import sys, os
 
+
 def main():
     if len(sys.argv) < 2:
-        print ("Usage: texture-model.py [model directory]\n\t This example will load [model direcotry].obj plus any of albedo, normal, ao, metallic and roughness textures present.")
+        print(
+            "Usage: texture-model.py [model directory]\n\t This example will load [model direcotry].obj plus any of albedo, normal, ao, metallic and roughness textures present."
+        )
         exit()
 
     model_dir = sys.argv[1]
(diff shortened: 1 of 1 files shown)


prewettg (changes requested):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2935)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @errissa)

___
*[examples/python/gui/textured-model.py, line 17 at r2](https://reviewable.io/reviews/intel-isl/open3d/2935#-MRfUy7h8sk9rvQjZtL3:-MRfUy7h8sk9rvQjZtL4:b-b326r0) ([raw file](https://github.com/intel-isl/open3d/blob/6e03974c4e7ad9e6485f02c4e8972ab07d199a3f/examples/python/gui/textured-model.py#L17)):*
> ```Python
> 
>     model_dir = sys.argv[1]
>     model_name = os.path.join(model_dir, os.path.basename(model_dir) + ".obj")
> ```

Since we have examples/test_data/monkey, I think should default to using that, so that the example runs out of the box.
___
*[examples/python/gui/textured-model.py, line 40 at r2](https://reviewable.io/reviews/intel-isl/open3d/2935#-MRfV9mg1B1q3HKGPruu:-MRfV9mg1B1q3HKGPruv:btn5xi5) ([raw file](https://github.com/intel-isl/open3d/blob/6e03974c4e7ad9e6485f02c4e8972ab07d199a3f/examples/python/gui/textured-model.py#L40)):*
> ```Python
> 
>     o3d.visualization.draw([{
>         "name": "cube",
> ```

Maybe not "cube", how about "model" or the filename?

PR #2970, opened 2021-01-30
Refactored RPC interface to be more amenable to variation in visualizers (fixes add not updating)

prewettg:
The initial problem was to get RPC add-geometry to redraw the scene in response to an update. However, this is a little troublesome because a) the scene doesn't know about the SceneWidget (which needs to note that it needs to actually redraw the scene for the next redraw and not just copy a bitmap) and the widgets don't know about their owning window (which is what requests a redraw from the OS), and b) different visualizers need to do different things even on adding geometry: O3DVisualizer needs to add to the scene using a lit material if it has normals or unlit otherwise, and to the list of geometries, but GuiVisualizer needs to add to the scene using the current material (or maybe replace the current object). So it seemed like it would be more flexible for the RPC Receiver to have a callback function so that the owner can process the add request appropriately, rather than the Receiver adding to the scene and expecting the higher level objects to figure out what happened.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2970)

### DIFF
--- a/cpp/open3d/visualization/visualizer/GuiVisualizer.cpp
+++ b/cpp/open3d/visualization/visualizer/GuiVisualizer.cpp
@@ -863,8 +863,18 @@ void GuiVisualizer::Layout(const gui::Theme &theme) {
 
 void GuiVisualizer::StartRPCInterface(const std::string &address, int timeout) {
 #ifdef BUILD_RPC_INTERFACE
-    impl_->receiver_ = std::make_shared<Receiver>(
-            this, impl_->scene_wgt_->GetScene(), address, timeout);
+    auto on_geometry = [this](std::shared_ptr<geometry::Geometry3D> geom,
+                              const std::string &path, int time,
+                              const std::string &layer) {
+        // Rather than duplicating the logic to figure out the correct material,
+        // just add with the default material and pretend the user changed the
+        // current material and update everyone's material.
+        impl_->scene_wgt_->GetScene()->AddGeometry(path, geom.get(),
+                                                   rendering::Material());
+        impl_->UpdateFromModel(GetRenderer(), true);
+    };
+    impl_->receiver_ =
+            std::make_shared<Receiver>(address, timeout, this, on_geometry);
     try {
         utility::LogInfo("Starting to listen on {}", address);
         impl_->receiver_->Start();
--- a/cpp/open3d/visualization/visualizer/O3DVisualizer.cpp
+++ b/cpp/open3d/visualization/visualizer/O3DVisualizer.cpp
@@ -1687,8 +1687,17 @@ Open3DScene *O3DVisualizer::GetScene() const {
 
 void O3DVisualizer::StartRPCInterface(const std::string &address, int timeout) {
 #ifdef BUILD_RPC_INTERFACE
-    impl_->receiver_ = std::make_shared<Receiver>(
-            this, impl_->scene_->GetScene(), address, timeout);
+    auto on_geometry = [this](std::shared_ptr<geometry::Geometry3D> geom,
+                              const std::string &path, int time,
+                              const std::string &layer) {
+        impl_->AddGeometry(path, geom, nullptr, nullptr, layer, time, true);
+        if (impl_->objects_.size() == 1) {
+            impl_->ResetCameraToDefault();
+        }
+    };
+
+    impl_->receiver_ =
+            std::make_shared<Receiver>(address, timeout, this, on_geometry);
     try {
         utility::LogInfo("Starting to listen on {}", address);
         impl_->receiver_->Start();
--- a/cpp/open3d/visualization/visualizer/Receiver.cpp
+++ b/cpp/open3d/visualization/visualizer/Receiver.cpp
@@ -195,7 +195,7 @@ std::shared_ptr<zmq::message_t> Receiver::ProcessMessage(
             }
         }
 
-        HandleGeometry(mesh, msg.path, msg.time, msg.layer);
+        SetGeometry(mesh, msg.path, msg.time, msg.layer);
 
     } else {
         // create a PointCloud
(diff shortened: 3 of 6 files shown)


benjaminum (changes requested):
I am not sure about the communication between the objects. For the RPC Receiver it should be only necessary to talk to the Scene. Wouldn't it be better to let the Scene know about all the viewers (SceneWidget)?

> Reviewed 6 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2970#-:-MSHEETV2IySY9kOWpz7:b-ayz5io)* status: all files reviewed, 4 unresolved discussions (waiting on @prewettg)

___
*[cpp/open3d/visualization/visualizer/GuiVisualizer.cpp, line 874 at r1](https://reviewable.io/reviews/intel-isl/open3d/2970#-MSHAE6ID7aJnpxnZIyx:-MSHAE6ID7aJnpxnZIyy:btsxc0s) ([raw file](https://github.com/intel-isl/open3d/blob/397ba3ef381b29cb872e0e85dd6ae2e59dd310b0/cpp/open3d/visualization/visualizer/GuiVisualizer.cpp#L874)):*
> ```cpp
>         impl_->scene_wgt_->GetScene()->AddGeometry(path, geom.get(),
>                                                    rendering::Material());
>         impl_->UpdateFromModel(GetRenderer(), true);
> ```

Is this triggering the update of the widget showing the objects?
___
*[cpp/open3d/visualization/visualizer/O3DVisualizer.cpp, line 1693 at r1](https://reviewable.io/reviews/intel-isl/open3d/2970#-MSHCtbv0UGtJUBvL4Ba:-MSHCtbv0UGtJUBvL4Bb:bed7tqs) ([raw file](https://github.com/intel-isl/open3d/blob/397ba3ef381b29cb872e0e85dd6ae2e59dd310b0/cpp/open3d/visualization/visualizer/O3DVisualizer.cpp#L1693)):*
> ```cpp
>                               const std::string &path, int time,
>                               const std::string &layer) {
>         impl_->AddGeometry(path, geom, nullptr, nullptr, layer, time, true);
> ```

Why not use the add function of the Scene?
___
*[cpp/open3d/visualization/visualizer/O3DVisualizer.cpp, line 1695 at r1](https://reviewable.io/reviews/intel-isl/open3d/2970#-MSHC4-94s2i8xL1e77a:-MSHC4-94s2i8xL1e77b:bj2z4ze) ([raw file](https://github.com/intel-isl/open3d/blob/397ba3ef381b29cb872e0e85dd6ae2e59dd310b0/cpp/open3d/visualization/visualizer/O3DVisualizer.cpp#L1695)):*
> ```cpp
>         impl_->AddGeometry(path, geom, nullptr, nullptr, layer, time, true);
>         if (impl_->objects_.size() == 1) {
>             impl_->ResetCameraToDefault();
> ```

This is just to center the camera on the object so the user does not have to do this manually, right?
___
*[cpp/open3d/visualization/visualizer/Receiver.h, line 69 at r1](https://reviewable.io/reviews/intel-isl/open3d/2970#-MSHAZnp61dyYlUKjXUa:-MSHAZnp61dyYlUKjXUb:b-g4x2fv) ([raw file](https://github.com/intel-isl/open3d/blob/397ba3ef381b29cb872e0e85dd6ae2e59dd310b0/cpp/open3d/visualization/visualizer/Receiver.h#L69)):*
> ```objc
>     OnGeometryFunc on_geometry_;
> 
>     void HandleGeometry(std::shared_ptr<geometry::Geometry3D> geom,
> ```

I would prefer Set because there will be also an RPC for getting geometry in the future.

prewettg (commented):
I initially tried that, but realized that doesn't work well. For instance, the O3DVisualizer would need to scan through the scene and figure out what changed. But the Scene also contains things the O3DVisualizer doesn't know about, like the axes and the background color/image. Also, O3DVisualizer needs to do things like store information about the geometries outside the scene. In the future when we have an actual scene graph with flexible properties, perhaps it can act as the model, but for now the scene is only part of the O3DVisualizer model.

> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2970#-:-MSU8A1E57_mofFLDmnV:bn8kqwo)* status: 4 of 6 files reviewed, 4 unresolved discussions (waiting on @benjaminum)

___
*[cpp/open3d/visualization/visualizer/GuiVisualizer.cpp, line 874 at r1](https://reviewable.io/reviews/intel-isl/open3d/2970#-MSHAE6ID7aJnpxnZIyx:-MSUxih03qf2VxDRerrg:b-9omy7o) ([raw file](https://github.com/intel-isl/open3d/blob/397ba3ef381b29cb872e0e85dd6ae2e59dd310b0/cpp/open3d/visualization/visualizer/GuiVisualizer.cpp#L874)):*

Is this triggering the update of the widget showing the objects?
</blockquote></details>

It adds a geometry to the scene, updates necessary internals and triggers the update of the widget.
___
*[cpp/open3d/visualization/visualizer/O3DVisualizer.cpp, line 1693 at r1](https://reviewable.io/reviews/intel-isl/open3d/2970#-MSHCtbv0UGtJUBvL4Ba:-MSU6h8h2J_r4krL2iVb:b-fdoc2e) ([raw file](https://github.com/intel-isl/open3d/blob/397ba3ef381b29cb872e0e85dd6ae2e59dd310b0/cpp/open3d/visualization/visualizer/O3DVisualizer.cpp#L1693)):*

Why not use the add function of the Scene?
</blockquote></details>

Because it's too low-level. This AddGeometry() updates the list of geometries, updates the tree controls so that the checkbox callback works so that clicking the checkbox shows/hides the geometry, keeps track of the original material so that it can be restored after switching to depth or normal coloring mode, etc.
___
*[cpp/open3d/visualization/visualizer/O3DVisualizer.cpp, line 1695 at r1](https://reviewable.io/reviews/intel-isl/open3d/2970#-MSHC4-94s2i8xL1e77a:-MSU78os-KD_xhF5l40_:b4je7ku) ([raw file](https://github.com/intel-isl/open3d/blob/397ba3ef381b29cb872e0e85dd6ae2e59dd310b0/cpp/open3d/visualization/visualizer/O3DVisualizer.cpp#L1695)):*

This is just to center the camera on the object so the user does not have to do this manually, right?
</blockquote></details>

Right. My simple test just did an empty o3d.visualization.draw([], rpc_interface=True), so the camera is at (0, 0, 0). Then I had a separate script that added a torus. Since the camera is still at (0, 0, 0), nothing visibly changed, which seemed wrong to me. But if you've already got something there, then it's quite likely that the user has rotated or changed the camera, so it shouldn't be reset.
___
*[cpp/open3d/visualization/visualizer/Receiver.h, line 69 at r1](https://reviewable.io/reviews/intel-isl/open3d/2970#-MSHAZnp61dyYlUKjXUa:-MSUxrntEy9WK4N7sB7v:bb6un1j) ([raw file](https://github.com/intel-isl/open3d/blob/397ba3ef381b29cb872e0e85dd6ae2e59dd310b0/cpp/open3d/visualization/visualizer/Receiver.h#L69)):*

I would prefer Set because there will be also an RPC for getting geometry in the future. 
</blockquote></details>

Done

benjaminum (approved):
LGTM. Thanks for the explanations.

> Reviewed 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2970#-:-MSWROZ8Ao3nvo_Fcro2:btybteq)* status: all files reviewed, 4 unresolved discussions (waiting on @benjaminum)

PR #3002, opened 2021-02-05
Adds ImageWidget and ability to efficiently change Scene background image

prewettg:
See `examples/python/gui/video.py` for an example that demonstrates both video in both 2D and 3D widgets

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3002)


ssheorey (commented):
Thanks for the video widget, it will be very useful. Some questions:
- How do we use it for live video capture (indefinite duration, no seeking)?
- How is the behavior if `get_frame()` is delayed, i.e. for example takes 50ms to return in a 30fps video? It would be good to document any expectations on the user provided functions.
- Does this work with `t::geometry::Image` as well? If not, the extra data copy for converting `tImage` to `Image` could be a performance bottleneck.

germanros1987:
@prewettg could you resolve all those conflicts before Rene start reviewing it?

glassTransition:
I see that this isn't closed yet, but I'd like to start planning a path forward and have some questions.

I'm using Python and want to display 2D images (depth and RGB images from a Realsense camera as well as a point cloud) in near-realtime. Would you recommend adding single images to the list and displaying them from there? You certainly can't be recommending saving them to files and reading them back in with the ImageLabel?

I think many people would want to do what I'm trying to do and am not understanding why it isn't supported as a first class function? I would like to display both 2D and 3D frames as they're being captured. I certainly may be missing something fundamental and apologize if so.

Thank you for any help you can provide.

ssheorey:
Hi @glassTransition real time capture from RealSense cameras and display of point clouds and sensor images is under progress. Please see https://github.com/intel-isl/Open3D-ML/pull/239.

> I see that this isn't closed yet, but I'd like to start planning a path forward and have some questions.
> 
> I'm using Python and want to display 2D images (depth and RGB images from a Realsense camera as well as a point cloud) in near-realtime. Would you recommend adding single images to the list and displaying them from there? You certainly can't be recommending saving them to files and reading them back in with the ImageLabel?
> 
> I think many people would want to do what I'm trying to do and am not understanding why it isn't supported as a first class function? I would like to display both 2D and 3D frames as they're being captured. I certainly may be missing something fundamental and apologize if so.
> 
> Thank you for any help you can provide.

glassTransition:
> Hi @glassTransition real time capture from RealSense cameras and display of point clouds and sensor images is under progress. Please see [intel-isl/Open3D-ML#239](https://github.com/intel-isl/Open3D-ML/pull/239).

Thank you! 

I saw the request about simultaneous displays rather than toggling on the other thread. That's what I'm after too so I'll follow that thread for updates.

I also need access to the tracking camera data so I can update the pointcloud as I move the depth camera and capture more points. I have this working with the GUI but my depth displays are in another window right now. I am monitoring the keyboard for user input to trigger capturing more points.

My ultimate goal is to automatically update the scan as I move the camera based on history of the FOV. I may use multiple depth cameras as well to increase coverage.

ssheorey:
> I also need access to the tracking camera data so I can update the pointcloud as I move the depth camera and capture more points. I have this working with the GUI but my depth displays are in another window right now. I am monitoring the keyboard for user input to trigger capturing more points.
> 
> My ultimate goal is to automatically update the scan as I move the camera based on history of the FOV. I may use multiple depth cameras as well to increase coverage.

We don't have access to the RealSense IMU as yet. Please open a new issue for it. Also, PRs are welcome!
Multiple depth cameras should work.

errissa (changes requested):
> Reviewed 2 of 12 files at r3, 29 of 63 files at r5, 2 of 3 files at r6, 50 of 50 files at r7.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3002)* status: all files reviewed, 2 unresolved discussions (waiting on @prewettg)

___
*[cpp/open3d/visualization/rendering/filament/FilamentResourceManager.cpp, line 410 at r7](https://reviewable.io/reviews/intel-isl/open3d/3002#-MXY7WZt63d79ayh7OKJ:-MXY7WZu6ws6Opo0FQKX:b-k6whfi) ([raw file](https://github.com/intel-isl/open3d/blob/6e80970bdcf0a686367ea6e3b5068af13219c247/cpp/open3d/visualization/rendering/filament/FilamentResourceManager.cpp#L410)):*
> ```cpp
> }
> 
> bool FilamentResourceManager::UpdateTexture(
> ```

I realize this was added to supported the video display in which case the image dimensions don't change. But any reason not to reallocate the texture if the image dimensions do change?

errissa (approved):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3002)* status: all files reviewed, 1 unresolved discussion (waiting on @prewettg)

PR #3021, opened 2021-02-10
PointRCNN ops

Prantl:
Added torch version of ops (+ tests) required for PointRCNN (GPU only)

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3021)


germanros1987:
@yxlao i think you should be able to review this pretty quickly. Could you help? It's ready to go

cpp/open3d/ml/contrib/RoIPoolKernel.cu
```diff
@@ -0,0 +1,300 @@
...
+#include <math.h>
+#include <stdio.h>
+
+#include "open3d/ml/contrib/RoIPoolKernel.h"
+
+#define THREADS_PER_BLOCK 256
+#define DIVUP(m, n) ((m) / (n) + ((m) % (n) > 0))
+// #define DEBUG
+
+__device__ inline int pt_in_box3d(float x,
```
yxlao, line 16: Similar to other `contrib` ops, could you put the code inside

```
namespace open3d {
namespace ml {
namespace contrib {
```

Same for other files.

cpp/open3d/ml/contrib/RoIPoolKernel.cu
```diff
@@ -0,0 +1,300 @@
...
+            batch_size, pts_num, boxes_num, feature_in_len, sampled_pts_num,
+            xyz, pts_idx, pts_feature, pooled_features, pooled_empty_flag);
+
+    cudaFree(pts_assign);
+    cudaFree(pts_idx);
+
+#ifdef DEBUG
+    cudaDeviceSynchronize();  // for using printf in kernel function
+#endif
+}
```
yxlao, line 300: Add a newline at the end. Without the newline, github shows as a "stop" sign.

image

Same for other files that don't have the newline at the end.

python/test/ml_ops/test_gathering.py
```diff
@@ -0,0 +1,51 @@
...
+ml_torch_gpu_only = pytest.mark.parametrize('ml', [
+    v for k, v in mltest._ml_modules.items()
+    if mltest.is_gpu_device_name(v.device) and v.module.__name__ == 'torch'
+])
+
+
+@ml_torch_gpu_only
+def test_gathering(ml):
+
+    values0 = np.load('python/test/ml_ops/data/gathering/values0.npy')
```
yxlao, line 44: At least we shall use relative paths here, this way the tests can be executed from any directory. e.g. https://github.com/intel-isl/Open3D/blob/2ded6fac2895716a7c68e73eac5c51b4d2e34c7f/python/test/t/io/test_realsense.py#L37

However, since we have many `.npy` files here, I recommend downloading them from google cloud and remove them from the directory. I'll be sending you the links to the files separately.

germanros1987:
@yxlao I think we are ready for a second round if you are

yxlao (approved):

PR #3026, opened 2021-02-11
Read depth directly instead of using depth shader and reading back color

errissa:
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3026)


errissa:
Closed to reopen against master branch.

PR #3043, opened 2021-02-16
Allow ccache to cache CUDA compilation

yxlao:
Following https://github.com/ccache/ccache/blob/master/doc/INSTALL.md#installation, after `cache` is installed, we need to either:
- Method 1: Prefix `ccache` when calling the compiler. e.g. `ccache gcc` instead of `gcc`. This is the more portable way and it is exactly what `CMAKE_X_COMPILER_LAUNCHER` is used for in this PR.
- Method 2: Or, we can also configure the symbolic links, following the instruction above.
  - If you install `ccache` from `apt-get`, the configuration is done automatically. But `nvcc` is not configured and `ccache` is usually too old to support newer `nvcc` flags.
  - If you compile `ccache` from source, you'll need to do the configuration by yourself. However, in my test, if you configure `nvcc` symbolic links manually, cmake will not be able to detect CUDA compiler. 
- Therefore, Method 2 currently does not work well with CUDA. Method 1 is more portable and doesn't require users to make additional configurations (as long as they have a sufficiently new cache installed).
- Method 1&2 can co-exist. 

This PR enables "Method 1" automatically via the cmake `CMAKE_CUDA_COMPILER_LAUNCHER` flag. 

- With latest `ccache` 4.2 (compile-from-source): works with CUDA/C/C++. 
- With `ccache` 3.4.1 from Ubuntu 18.04: works with C/C++, so no gain here, same as before.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3043)


ssheorey (approved):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3043)* status: 0 of 12 files reviewed, 1 unresolved discussion (waiting on @yxlao)

___
*[3rdparty/faiss/faiss_build.cmake, line 23 at r2](https://reviewable.io/reviews/intel-isl/open3d/3043#-MThA9raC-D50dDcgjCO:-MThA9rbE5D_F7-2ruxd:b-jma94e) ([raw file](https://github.com/intel-isl/open3d/blob/6e0600c340da7d995ed6bc6e5647db5dadf9e837/3rdparty/faiss/faiss_build.cmake#L23)):*
> ```CMake
>         -DCMAKE_C_COMPILER_LAUNCHER=${CMAKE_C_COMPILER_LAUNCHER}
>         -DCMAKE_CXX_COMPILER_LAUNCHER=${CMAKE_CXX_COMPILER_LAUNCHER}
>         -DCMAKE_CUDA_COMPILER_LAUNCHER=${CMAKE_CUDA_COMPILER_LAUNCHER}
> ```

Do we also need a`-DCMAKE_CUDA_COMPILER=...` arg?

ssheorey:
I was able to install and run `ccache 4.1` from the Hirsute (21.04) repo on Focal (20.04). They've added `ccache 4.2` to the 21.04 `proposed` section today, so chances look good that we can just use `ccache` binaries from a future version of Ubuntu.

yxlao:
updated `CMAKE_CUDA_COMPILER`

PR #3048, opened 2021-02-16
Change UpdateGeometry to work on updates where vertex count changes

errissa:
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3048)


ssheorey (approved):
For the case that future point clouds may have more vertices than the first one (but `max_vertices` is known), can we allocate memory for `max_vertices` in `add_geometry()`?
Otherwise, I can try adding dummy vertices in the first point cloud. How do I do that in a way that's not visible in the display?

PR #3050, opened 2021-02-17
Add function to randomly downsample from pointcloud

scimad:
This PR has a function that downsamples the pointcloud by selecting random points from a point cloud without any rules (such as *every k point* or *voxel downsampling*). I think for some use cases, sampling completely random points could be useful.

Such as in my case.

I was trying to visualize an RBG image's pose on same 3D space with the model / point cloud. To visualize RGB image in 3D, I used pixel to image coordinate transformation and converted each point on image coordinate to a point on pointcloud.To give you the idea:

image

Uniform sampling however wouldn't work well for such visualization as it appeared like the following.

image

This PR allous us to sample random points from point cloud at a given sampling ratio and the result looks like this:

image

Let me know if I need to update anything.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3050)


yxlao (approved):
thanks @scimad

PR #3069, opened 2021-02-21
LGTM / CodeQL security checks

ssheorey:
Uses CodeQL github action.
Adds about 40 minutes to Ubuntu CI. But total time doesn't increase since building Python wheel takes longer.

Code scanning results are available here: 
https://github.com/intel-isl/Open3D/security/code-scanning?query=tool%3ACodeQL
[Select your branch]

**Contributors:**
Alerts in 3rd party code: **Dismiss as Won't Fix.** They should not re-appear on the next scan.
Deleted alerts return on the next scan unless fixed.

**Github docs:**
https://docs.github.com/en/github/finding-security-vulnerabilities-and-errors-in-your-code/managing-code-scanning-alerts-for-your-repository

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/3069)

### DIFF
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -73,9 +73,10 @@ jobs:
           ccache -M 2G  # See .github/workflows/readme.md for ccache strategy.
 
       - name: Initialize CodeQL
+        if: matrix.CONFIG=='ON'
         uses: github/codeql-action/init@v1
         with:
-          languages: cpp, python
+          languages: cpp, python, javascript
           queries: security-and-quality
 
       - name: Config and build
@@ -86,6 +87,7 @@ jobs:
           ccache -s
 
       - name: Perform CodeQL Analysis
+        if: matrix.CONFIG=='ON'
         uses: github/codeql-action/analyze@v1
 
 
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -85,6 +85,8 @@ jobs:
           ccache -s
           ./util/run_ci.sh
           ccache -s
+          # Clean folders to prevent duplicate CodeQL alerts
+          rm -rf build/lib/python_package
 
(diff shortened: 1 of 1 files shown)


yxlao (commented):
We shall probably ignore the `3rdparty` directory in our scans. E.g with https://docs.github.com/en/github/finding-security-vulnerabilities-and-errors-in-your-code/configuring-code-scanning#specifying-directories-to-scan

ssheorey:
> We shall probably ignore the `3rdparty` directory in our scans. E.g with https://docs.github.com/en/github/finding-security-vulnerabilities-and-errors-in-your-code/configuring-code-scanning#specifying-directories-to-scan

That only works for interpreted languages (Python, Javascript). For compiled languages (C++), it produces alerts on all the code that's built. Deleting the built 3rdparty code before analysis may avoid 3rdparty alerts. Testing this.

ssheorey:
Only way to ignore alerts in 3rd party code seems to be to mark them as **Dismiss: Won't fix**. They should not re-appear on repeat CI runs. I've marked all 3rd party code alerts like this (>1000). We'll need to update these alerts when upgrading / adding 3rdparty code.

yxlao:
LGTM so far. A question, from https://github.com/intel-isl/Open3D/security/code-scanning?query=tool%3ACodeQL, looks like we're only getting warnings for python code. Do we have warnings for non-3rd-party C++ code?

ssheorey:
> LGTM so far. A question, from https://github.com/intel-isl/Open3D/security/code-scanning?query=tool%3ACodeQL, looks like we're only getting warnings for python code. Do we have warnings for non-3rd-party C++ code?

Yes, that's strange. I can see issues on the [lgtm.com website](https://lgtm.com/projects/g/intel-isl/Open3D/alerts/?mode=list&lang=cpp), but not here. Will check.

PR #3072, opened 2021-02-21
Implemented ability to render to a bitmap window and interact with UI

prewettg:
This abstracts the GLFW calls into a WindowSystem class so that the window system can be swapped out with a bitmap window system.

Example code for macOS that use BitmapWindowSystem with Application to launch a bitmap window with UI; each draw is read back and displayed as a bitmap on a native macOS window:

```#include "open3d/Open3D.h"
#include "open3d/visualization/gui/BitmapWindowSystem.h"
#include "open3d/visualization/gui/Native.h"

#import <Cocoa/Cocoa.h>

using namespace open3d;
using namespace open3d::visualization;

static const int INITIAL_WIDTH = 1024;
static const int INITIAL_HEIGHT = 768;

auto gWindowSystem = std::make_shared<gui::BitmapWindowSystem>();

gui::MouseEvent fromMouseEvent(NSEvent *e, gui::MouseEvent::Type type,
                               CGFloat height) {
    float scaling = e.window.backingScaleFactor;
    int mods = 0;
    if (e.modifierFlags & NSEventModifierFlagShift) {
        mods |= int(gui::KeyModifier::SHIFT);
    }
    if (e.modifierFlags & NSEventModifierFlagCommand) {
        mods |= int(gui::KeyModifier::CTRL);  // Cmd key is mapped to CTRL on mac
    }
    if (e.modifierFlags & NSEventModifierFlagControl) {
        mods |= int(gui::KeyModifier::ALT);  // Ctrl key is mapped to ALT on mac
    }
    if (e.modifierFlags & NSEventModifierFlagOption) {
        mods |= int(gui::KeyModifier::META);  // Opt key is mapped to META on mac
    }
    return {type,
            int(e.locationInWindow.x * scaling),
            int((height - e.locationInWindow.y) * scaling),
            mods};
}

gui::KeyEvent fromKeyEvent(NSEvent *e, gui::KeyEvent::Type type) {
    gui::KeyEvent ke;
    ke.type = type;
    ke.key = gui::KeyName([e.characters characterAtIndex:0]);
    ke.isRepeat = (e.ARepeat == YES);
    return ke;
}

@interface MirrorView : NSView
{
    gui::Window *window;
    gui::WindowSystem::OSWindow osWindow;
}
@property (retain) NSBitmapImageRep* current_frame;

- (void)setCurrentFrame:(std::shared_ptr<geometry::Image>) frame;
- (void)onClose:(NSNotification *)notif;
@end

@implementation MirrorView
- (instancetype)initWithFrame:(NSRect)frameRect window:(gui::Window*) win {
    self = [super initWithFrame:frameRect];
    if (self) {
        self->window = win;
        self->osWindow = window->GetOSWindow();
        self.current_frame = nil;
    }
    return self;
}

- (void)setCurrentFrame:(std::shared_ptr<geometry::Image>) frame {
    unsigned char *bytes = frame->data_.data();
    self.current_frame =
        [[NSBitmapImageRep alloc]
         initWithBitmapDataPlanes: &bytes
                       pixelsWide: frame->width_
                       pixelsHigh: frame->height_
                    bitsPerSample: 8 * frame->bytes_per_channel_
                  samplesPerPixel: frame->num_of_channels_
                         hasAlpha: NO
                         isPlanar: NO
                   colorSpaceName: NSDeviceRGBColorSpace
                      bytesPerRow: frame->width_ * frame->num_of_channels_ * frame->bytes_per_channel_
                     bitsPerPixel: 8 * frame->bytes_per_channel_ * frame->num_of_channels_];
    self.needsDisplay = YES;
}

- (void)viewDidEndLiveResize {
    float scaleFactor = self.window.backingScaleFactor;
    gWindowSystem->SetWindowSize(self->osWindow,
                                 int(self.window.frame.size.width * scaleFactor),
                                 int(self.window.frame.size.height * scaleFactor));
    auto windowRect = NSMakeRect(0, 0, self.window.frame.size.width,
                                 self.window.frame.size.height);
    self.frame = [NSWindow contentRectForFrameRect:windowRect
                                         styleMask:NSTitledWindowMask];
}

- (void)onClose:(NSNotification *)notif {
    self->window->Close();
    [NSNotificationCenter.defaultCenter removeObserver:self];
}

- (void)mouseDown:(NSEvent *)e {
    auto me = fromMouseEvent(e, gui::MouseEvent::BUTTON_DOWN,
                             self.frame.size.height);
    me.button.button = gui::MouseButton::LEFT;
    gWindowSystem->PostMouseEvent(self->osWindow, me);
}

- (void)mouseDragged:(NSEvent *)e {
    auto me = fromMouseEvent(e, gui::MouseEvent::DRAG,
                             self.frame.size.height);
    me.move.buttons = int(gui::MouseButton::LEFT);
    gWindowSystem->PostMouseEvent(self->osWindow, me);
}

- (void)mouseUp:(NSEvent *)e {
    auto me = fromMouseEvent(e, gui::MouseEvent::BUTTON_UP,
                             self.frame.size.height);
    me.button.button = gui::MouseButton::LEFT;
    gWindowSystem->PostMouseEvent(self->osWindow, me);
}

- (void)rightMouseDown:(NSEvent *)e {
    [super rightMouseDown:e];  // docs say must super for -rightMouseDown:
    auto me = fromMouseEvent(e, gui::MouseEvent::BUTTON_DOWN,
                             self.frame.size.height);
    me.button.button = gui::MouseButton::RIGHT;
    gWindowSystem->PostMouseEvent(self->osWindow, me);
}

- (void)rightMouseDragged:(NSEvent *)e {
    auto me = fromMouseEvent(e, gui::MouseEvent::DRAG,
                             self.frame.size.height);
    me.move.buttons = int(gui::MouseButton::RIGHT);
    gWindowSystem->PostMouseEvent(self->osWindow, me);
}

- (void)rightMouseUp:(NSEvent *)e {
    auto me = fromMouseEvent(e, gui::MouseEvent::BUTTON_UP,
                             self.frame.size.height);
    me.button.button = gui::MouseButton::RIGHT;
    gWindowSystem->PostMouseEvent(self->osWindow, me);
}

- (void)scrollWheel:(NSEvent *)e {
    auto me = fromMouseEvent(e, gui::MouseEvent::WHEEL,
                             self.frame.size.height);
    me.wheel.dx = e.deltaX;
    me.wheel.dy = e.deltaY;
    me.wheel.isTrackpad = true;  // assume laptop if MacOS
    gWindowSystem->PostMouseEvent(self->osWindow, me);
}

- (BOOL)acceptsFirstResponder {return YES;}
- (BOOL)canBecomeKeyView {return YES;}

- (void)keyDown:(NSEvent *)e {
    auto ke = fromKeyEvent(e, gui::KeyEvent::Type::DOWN);
    gWindowSystem->PostKeyEvent(self->osWindow, ke);
}

- (void)keyUp:(NSEvent *)e {
    auto ke = fromKeyEvent(e, gui::KeyEvent::Type::UP);
    gWindowSystem->PostKeyEvent(self->osWindow, ke);
}

- (void)drawRect:(NSRect)dirtyRect {
    [super drawRect:dirtyRect];
    if (self.current_frame != nil) {
        if ([self.current_frame drawInRect: self.bounds] == NO) {
            std::cerr << "[error] Could not draw frame" << std::endl;
        }
    }
}
@end

@interface AppDelegate : NSObject <NSApplicationDelegate>
@end

@interface AppDelegate ()
{
}
@property (retain) NSWindow *window;
@end

@implementation AppDelegate
- (void)applicationDidFinishLaunching:(NSNotification *)notification {
}

@end

int main(int argc, const char *argv[]) {
    // Initialize Cocoa
    AppDelegate *delegate = [[AppDelegate alloc] init];
    NSApplication *application = [NSApplication sharedApplication];
    [application setDelegate:delegate];
    [NSApp finishLaunching];

    gui::MacTransformIntoApp();

    // Create the Open3D offscreen window first, so that the real window knows
    // about it.
    auto &o3d_app = gui::Application::GetInstance();
    o3d_app.SetWindowSystem(gWindowSystem);
    o3d_app.Initialize();

    auto draw = std::make_shared<visualizer::O3DVisualizer>("Test", INITIAL_WIDTH, INITIAL_HEIGHT);
    auto torus = geometry::TriangleMesh::CreateTorus();
    torus->ComputeVertexNormals();
    draw->AddGeometry("torus", torus);
    draw->ResetCameraToDefault();
    draw->ShowSettings(true);
    o3d_app.AddWindow(draw);

    // Normally the delegate would create the initial window, but that doesn't
    // work so well because we need to tell the real window about the offscreen
    // window, so that it can pass events.
    NSRect contentRect = NSMakeRect(0, 0, INITIAL_WIDTH, INITIAL_HEIGHT);
    NSWindow *window;
    window = [[NSWindow alloc]
              initWithContentRect:contentRect
                        styleMask:NSTitledWindowMask|NSClosableWindowMask |
                                  NSFullSizeContentViewWindowMask |
                                  NSResizableWindowMask
                          backing:NSBackingStoreBuffered
                            defer:NO];
    float scaleFactor = window.backingScaleFactor;
    contentRect = NSMakeRect(0, 0, INITIAL_WIDTH / scaleFactor, INITIAL_HEIGHT / scaleFactor);
    // Using -frameRectForContentRect: doesn't adjust for the title height,
    // despite that being part of the style,
    // but +frameRectForContentRect:styleMask: does work
    auto frameRect = [NSWindow frameRectForContentRect:contentRect
                                             styleMask:NSTitledWindowMask];
    [window setFrame:frameRect display:YES];
    [window cascadeTopLeftFromPoint:NSMakePoint(20,20)];
    [window setTitle: @"BitmapWindowSystem test"];

    MirrorView *view = [[MirrorView alloc]
                        initWithFrame:contentRect window:draw.get()];
    [window.contentView addSubview:view];
    view.frame = contentRect;

    [window makeKeyAndOrderFront:nil];
    [NSApp activateIgnoringOtherApps:YES];

    [NSNotificationCenter.defaultCenter addObserver:view
                                           selector:@selector(onClose:)
                                               name:NSWindowWillCloseNotification
                                             object:window];

    gWindowSystem->SetOnWindowDraw([view](gui::Window* win,
                                       std::shared_ptr<geometry::Image> frame) {
        [view setCurrentFrame:frame];
    });

    // Initialize GLFW (since Application won't be doing it), and it provides
    // a convenient way to run the native event loop incrementally.
    glfwInitHint(GLFW_COCOA_MENUBAR, GLFW_FALSE);  // no auto-create menubar
    glfwInit();

    draw.reset();
    gui::Application::EnvUnlocker noop;  // containing env is C++
    while (o3d_app.RunOneTick(noop))  // run window system event loop
        glfwWaitEventsTimeout(0.010);  // run native event loop

    glfwTerminate();
}
```

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3072)


yxlao:
Pushed a commit to fix the Windows build 16c4dbaa99f930f7cf0e34f1b5b345d64c523947. Please feel free to make additional changes if I missed something.

The following lines in `Window.cpp` is causing compilation issues and they seems unnecessary:
```cpp
#ifdef WIN32
    io.ImeWindowHandle = GetNativeDrawable();
#endif
```

Also, looks like the name `CreateWindow` collides with a macro in `WinUser.h`, so it was renamed to `CreateOSWindow`. 
```cpp
// C:\Program Files (x86)\Windows Kits\10\Include\10.0.18362.0\um\WinUser.h
...
#define CreateWindow  CreateWindowW
#else
#define CreateWindow  CreateWindowA
#endif // !UNICODE
..
```

yxlao (commented):
> Reviewed 6 of 13 files at r2, 7 of 7 files at r3.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3072)* status: all files reviewed, 2 unresolved discussions (waiting on @prewettg)

___
*[cpp/open3d/visualization/gui/CMakeLists.txt, line 49 at r3](https://reviewable.io/reviews/intel-isl/open3d/3072#-MUDFAUF4j8CcT58C2BY:-MUDFAUF4j8CcT58C2BZ:b-795liy) ([raw file](https://github.com/intel-isl/open3d/blob/16c4dbaa99f930f7cf0e34f1b5b345d64c523947/cpp/open3d/visualization/gui/CMakeLists.txt#L49)):*
> ```CMake
>     Events.h
>     FileDialog.h
>     HeadlessWindowSystem.h
> ```

Looks like we can remove all the `.h` files here. Do we have a specific reason for keeping them?
___
*[cpp/open3d/visualization/gui/HeadlessWindowSystem.cpp, line 247 at r3](https://reviewable.io/reviews/intel-isl/open3d/3072#-MUDFnp320ONRV4x7njo:-MUDFnp320ONRV4x7njp:bf72aws) ([raw file](https://github.com/intel-isl/open3d/blob/16c4dbaa99f930f7cf0e34f1b5b345d64c523947/cpp/open3d/visualization/gui/HeadlessWindowSystem.cpp#L247)):*
> ```cpp
> 
> rendering::FilamentRenderer *HeadlessWindowSystem::CreateRenderer(OSWindow w) {
>     auto *renderer = new rendering::FilamentRenderer(
> ```

Shall we call `rendering::EngineInstance::EnableHeadless();` here? 

Without this, on Linux a system without display, Filament will complain about not able to find a display. Also, shall we do something different for macOS and windows?

prewettg (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3072)* status: 5 of 21 files reviewed, 2 unresolved discussions (waiting on @errissa, @prewettg, and @yxlao)

___
*[cpp/open3d/visualization/gui/CMakeLists.txt, line 49 at r3](https://reviewable.io/reviews/intel-isl/open3d/3072#-MUDFAUF4j8CcT58C2BY:-MUFN_mx1MNRY0jUA-Ev:bybvo3f) ([raw file](https://github.com/intel-isl/open3d/blob/16c4dbaa99f930f7cf0e34f1b5b345d64c523947/cpp/open3d/visualization/gui/CMakeLists.txt#L49)):*

Looks like we can remove all the `.h` files here. Do we have a specific reason for keeping them?
</blockquote></details>

All as in everything, or all as in *WindowSystem.h? I assumed that we needed to have all header files here so that CMake would know to rebuild if one changed, is that not the case?

prewettg (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3072)* status: 5 of 21 files reviewed, 2 unresolved discussions (waiting on @errissa and @yxlao)

___
*[cpp/open3d/visualization/gui/HeadlessWindowSystem.cpp, line 247 at r3](https://reviewable.io/reviews/intel-isl/open3d/3072#-MUDFnp320ONRV4x7njo:-MUFYR58CNh92d4J_5YS:bi2l82f) ([raw file](https://github.com/intel-isl/open3d/blob/16c4dbaa99f930f7cf0e34f1b5b345d64c523947/cpp/open3d/visualization/gui/HeadlessWindowSystem.cpp#L247)):*

Shall we call `rendering::EngineInstance::EnableHeadless();` here? 

Without this, on Linux a system without display, Filament will complain about not able to find a display. Also, shall we do something different for macOS and windows?
</blockquote></details>

I've renamed HeadlessWindowSystem -> BitmapWindowSystem so that it doesn't imply headless mode, and the constructor now takes an enum that lets you specify headless mode.

yxlao (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3072)* status: 2 of 21 files reviewed, 2 unresolved discussions (waiting on @errissa, @prewettg, and @yxlao)

___
*[cpp/open3d/visualization/gui/CMakeLists.txt, line 49 at r3](https://reviewable.io/reviews/intel-isl/open3d/3072#-MUDFAUF4j8CcT58C2BY:-MUF_6Ve3aOQYJI6Q6H_:b-6kgi0w) ([raw file](https://github.com/intel-isl/open3d/blob/16c4dbaa99f930f7cf0e34f1b5b345d64c523947/cpp/open3d/visualization/gui/CMakeLists.txt#L49)):*

All as in everything, or all as in *WindowSystem.h? I assumed that we needed to have all header files here so that CMake would know to rebuild if one changed, is that not the case?
</blockquote></details>

All as in all `*.h`. CMake rebuilds if one `.h` file is changed AND the `.h` file is used in a `.cpp`file. This shall be sufficient for us. We could verify this by removing headers from `add_library` and changing a header file. 

That said, from https://stackoverflow.com/a/36174691/1255535, leaving out `.h` files could affect the `install()` behavior of public headers. Currently, in the rest of Open3D, we don't include `.h` files in `add_library` and we install all headers (so we don't have a notion of public/private headers). This can be revisited at some point. For now, it should be safe to exclude `.h` files from `add_library`.

prewettg (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3072)* status: 2 of 21 files reviewed, 2 unresolved discussions (waiting on @errissa and @yxlao)

___
*[cpp/open3d/visualization/gui/CMakeLists.txt, line 49 at r3](https://reviewable.io/reviews/intel-isl/open3d/3072#-MUDFAUF4j8CcT58C2BY:-MUGGlqy3481PjQxF6tK:bd9ohdu) ([raw file](https://github.com/intel-isl/open3d/blob/16c4dbaa99f930f7cf0e34f1b5b345d64c523947/cpp/open3d/visualization/gui/CMakeLists.txt#L49)):*

All as in all `*.h`. CMake rebuilds if one `.h` file is changed AND the `.h` file is used in a `.cpp`file. This shall be sufficient for us. We could verify this by removing headers from `add_library` and changing a header file. 

That said, from https://stackoverflow.com/a/36174691/1255535, leaving out `.h` files could affect the `install()` behavior of public headers. Currently, in the rest of Open3D, we don't include `.h` files in `add_library` and we install all headers (so we don't have a notion of public/private headers). This can be revisited at some point. For now, it should be safe to exclude `.h` files from `add_library`.
</blockquote></details>

Seems to work, removed

yxlao (approved):
Tested on Ubuntu 18.04 and macOS

errissa (changes requested):
> Reviewed 18 of 19 files at r4, 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3072)* status: all files reviewed, 4 unresolved discussions (waiting on @prewettg and @yxlao)

___
*[cpp/open3d/visualization/rendering/filament/FilamentRenderer.cpp, line 238 at r5](https://reviewable.io/reviews/intel-isl/open3d/3072#-MUKhuUIDBbGA75kba0L:-MUKhuUIDBbGA75kba0M:bvbyuea) ([raw file](https://github.com/intel-isl/open3d/blob/c6c888756ec9cbe4ad1ff0c28d295141212a79fb/cpp/open3d/visualization/rendering/filament/FilamentRenderer.cpp#L238)):*
> ```cpp
> }  // namespace
> 
> void FilamentRenderer::RequestReadPixels(
> ```

What is this used for and what is the advantage of this method over the existing RenderToBuffer path? Seems confusing having two separate paths to read pixels.
___
*[cpp/open3d/visualization/gui/BitmapWindowSystem.cpp, line 145 at r5](https://reviewable.io/reviews/intel-isl/open3d/3072#-MUKg6l0Cjqxqs2elvJ2:-MUKg6l0Cjqxqs2elvJ3:b-bk93h8) ([raw file](https://github.com/intel-isl/open3d/blob/c6c888756ec9cbe4ad1ff0c28d295141212a79fb/cpp/open3d/visualization/gui/BitmapWindowSystem.cpp#L145)):*
> ```cpp
> void BitmapWindowSystem::WaitEventsTimeout(double timeout_secs) {
>     auto t0 = std::chrono::steady_clock::now();
>     std::chrono::duration<float> duration;
> ```

`timeout_secs` is a double. You might as well use a `double` duration here as well.

prewettg (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3072)* status: 20 of 21 files reviewed, 4 unresolved discussions (waiting on @errissa and @yxlao)

___
*[cpp/open3d/visualization/rendering/filament/FilamentRenderer.cpp, line 238 at r5](https://reviewable.io/reviews/intel-isl/open3d/3072#-MUKhuUIDBbGA75kba0L:-MUfZufU8pfEXceHMKLB:b4f1ozc) ([raw file](https://github.com/intel-isl/open3d/blob/c6c888756ec9cbe4ad1ff0c28d295141212a79fb/cpp/open3d/visualization/rendering/filament/FilamentRenderer.cpp#L238)):*

What is this used for and what is the advantage of this method over the existing RenderToBuffer path? Seems confusing having two separate paths to read pixels.
</blockquote></details>

RenderToBuffer renders the Scene, and it bypasses FilamentRenderer and calls Filament directly [see FilamentRenderToBuffer::Render()]. I thought about combining them, but I didn't see an obvious way, as RenderToBuffer may be requested *during* a draw of the window (because ImGUI is immediate mode), so it needs to be postponed until afterwards so that the buffer contains the entire scene. I'm assuming FilamentRenderToBuffer uses Filament directly rather than going through FilamentRenderer to avoid problems with recursion when calling back into BeginFrame(). I've never liked storing the RenderToBuffer object in the Renderer, but I also haven't come up with an obviously better way to render later. If you have a better idea, let me know and I'll refactor it.
___
*[cpp/open3d/visualization/gui/BitmapWindowSystem.cpp, line 145 at r5](https://reviewable.io/reviews/intel-isl/open3d/3072#-MUKg6l0Cjqxqs2elvJ2:-MUfZqdjBQ05v1aGyvzH:bmfcbel) ([raw file](https://github.com/intel-isl/open3d/blob/c6c888756ec9cbe4ad1ff0c28d295141212a79fb/cpp/open3d/visualization/gui/BitmapWindowSystem.cpp#L145)):*

`timeout_secs` is a double. You might as well use a `double` duration here as well.
</blockquote></details>

Good point; fixed

errissa (approved):
> Reviewed 1 of 1 files at r6.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3072)* status: all files reviewed, 3 unresolved discussions (waiting on @prewettg and @yxlao)

___
*[cpp/open3d/visualization/rendering/filament/FilamentRenderer.cpp, line 238 at r5](https://reviewable.io/reviews/intel-isl/open3d/3072#-MUKhuUIDBbGA75kba0L:-MUiWN5-3KwaPr30IWTA:bmr7e1f) ([raw file](https://github.com/intel-isl/open3d/blob/c6c888756ec9cbe4ad1ff0c28d295141212a79fb/cpp/open3d/visualization/rendering/filament/FilamentRenderer.cpp#L238)):*

RenderToBuffer renders the Scene, and it bypasses FilamentRenderer and calls Filament directly [see FilamentRenderToBuffer::Render()]. I thought about combining them, but I didn't see an obvious way, as RenderToBuffer may be requested *during* a draw of the window (because ImGUI is immediate mode), so it needs to be postponed until afterwards so that the buffer contains the entire scene. I'm assuming FilamentRenderToBuffer uses Filament directly rather than going through FilamentRenderer to avoid problems with recursion when calling back into BeginFrame(). I've never liked storing the RenderToBuffer object in the Renderer, but I also haven't come up with an obviously better way to render later. If you have a better idea, let me know and I'll refactor it.
</blockquote></details>

Nothing obviously jumps out at me. We should leave this for a future discussion.

PR #2979, opened 2021-02-02
Update faiss_build.cmake

junha-l:
Update pointer to faiss for resolving build error on RTX30xx

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/2979)

PR #3095, opened 2021-02-26
Changed t-registration unit-test to use RegistrationPermuteDevices

reyanshsolis:
In PR #2979, The unit-test for t-registrations that require Faiss were disabled if build with `WITH_FAISS=OFF`. 
However, the updated HybridSearch uses `fixed_radius_search` instead of `knn_search`, so it might not be using FAISS anymore. 
If that's the case we may change the unit test back to run with `RegistrationPermuteDevices` instead of `RegistrationPermuteDevicesWithFaiss`. As for system build with `FAISS=OFF` and `CUDA=ON`, this might be potentially harmful. 

Seeking confirmation from @junha-l.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3095)


junha-l (approved):
Yes, you are right! Now the GPU Hybrid Search is not dependent on Faiss. We can now enable Registration unit-test again.

PR #3099, opened 2021-02-27
[CUDA] Fix memory leaks.

Dudulle:
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3099)

### DIFF
--- a/cpp/open3d/core/linalg/InverseCUDA.cpp
+++ b/cpp/open3d/core/linalg/InverseCUDA.cpp
@@ -24,6 +24,7 @@
 // IN THE SOFTWARE.
 // ----------------------------------------------------------------------------
 
+#include "open3d/core/Blob.h"
 #include "open3d/core/linalg/Inverse.h"
 #include "open3d/core/linalg/LapackWrapper.h"
 #include "open3d/core/linalg/LinalgUtils.h"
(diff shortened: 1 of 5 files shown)


yxlao (approved):

cpp/open3d/core/linalg/InverseCUDA.cpp
```diff
@@ -41,31 +42,30 @@ void InverseCUDA(void* A_data,
...
 
         OPEN3D_CUSOLVER_CHECK_WITH_DINFO(
                 getrs_cuda<scalar_t>(handle, CUBLAS_OP_N, n, n,
                                      static_cast<scalar_t*>(A_data), n,
                                      static_cast<int*>(ipiv_data),
                                      static_cast<scalar_t*>(output_data), n,
-                                     dinfo),
-                "getrs failed in InverseCUDA", dinfo, device);
-
-        MemoryManager::Free(workspace, device);
```
yxlao, line 67: We do have `Free` here. So is this a memory leak, or are you trying to be more concise by using `Blob`?

Dudulle: There can actually be a leak if an exception is thrown between the allocation and the call to `Free`.
This does happen in a few unit tests, for example [here](https://github.com/intel-isl/Open3D/blob/5919ab5397cdbdf5386b097b2cfa20e75f8f4fa0/cpp/tests/core/Linalg.cpp#L358)

yxlao: :+1:

PR #3102, opened 2021-03-01
Enable builds on Fedora.

### LINKED ISSUE #3101: Build fails in Fedora
Build fails on Fedora 33. 
The target is Ubuntu, but I would like you to modify it so that the build will succeed even in Fedora.
I get the following error:
```
$ make
~~~~~~~~~
make[2]: *** No rule to make target 'libjpeg-turbo-install/lib/libturbojpeg.a', needed by 'lib/Release/libOpen3D.so'.  Stop.
make[1]: *** [CMakeFiles/Makefile2:1557: cpp/open3d/CMakeFiles/Open3D.dir/all] Error 2
make: *** [Makefile:149: all] Error 2
```
This is because the library path of libjpeg-turbo is "lib".
Fedora requires "lib64".
https://github.com/intel-isl/Open3D/blob/6b75f88c5901c0c0140302fe57c3d02cd1634797/3rdparty/find_dependencies.cmake#L494

I think it can be fixed by using **CMAKE_INSTALL_LIBDIR**.
```
# before
LIB_DIR ${CMAKE_CURRENT_BINARY_DIR}/libjpeg-turbo-install/lib 

# after
LIB_DIR ${CMAKE_CURRENT_BINARY_DIR}/libjpeg-turbo-install/${CMAKE_INSTALL_LIBDIR}
```

**Additional context**
Later, I would like to send a pull request for this issue.

NobuoTsukamoto:
Resolves #3101

Use ${CMAKE_INSTALL_LIBDIR} instead of "lib" for the 3rdparty library installation path.
This will allow you to build on a Linux distribution (e.g. Fedora) that is the path to "lib64".

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3102)


3rdparty/find_dependencies.cmake
```diff
@@ -491,7 +491,7 @@ if(NOT USE_SYSTEM_JPEG)
     import_3rdparty_library(3rdparty_jpeg
         INCLUDE_DIRS ${CMAKE_CURRENT_BINARY_DIR}/libjpeg-turbo-install/include/
         LIBRARIES ${JPEG_TURBO_LIBRARIES}
-        LIB_DIR ${CMAKE_CURRENT_BINARY_DIR}/libjpeg-turbo-install/lib
+        LIB_DIR ${CMAKE_CURRENT_BINARY_DIR}/libjpeg-turbo-install/${CMAKE_INSTALL_LIBDIR}
```
yxlao, line 494: This is a good idea. Change it to `Open3D_INSTALL_LIB_DIR` to make it work for Windows as well

See:
- https://github.com/intel-isl/Open3D/blob/103caa859fe0a6c928b898213da4a88517595135/CMakeLists.txt#L232
- https://github.com/intel-isl/Open3D/blob/103caa859fe0a6c928b898213da4a88517595135/CMakeLists.txt#L237

yxlao (changes requested):

3rdparty/find_dependencies.cmake
NobuoTsukamoto: Thank you for your advice.

Changed from `CMAKE_INSTALL_LIBDIR` to `Open3D_INSTALL_LIB_DIR`.
I also replaced `3rdparty/librealsense/librealsense.cmake` with `Open3D_INSTALL_LIB_DIR`.

yxlao (approved):

Issue #2995, opened 2021-02-04
O3DVisualizer()  => Unsupported image number of channels: 0


MrBenGriffin:
**IMPORTANT: Please use the following template to report the bug.**
_____

**Describe the bug**
instantiating visualization.O3DVisualizer() fails.

**To Reproduce**

make file test.py

```python
# test.py
import open3d as o3d
vis = o3d.visualization.O3DVisualizer()
```

invoke

```bash
python3.8 test.py
```

console generates 
```bash
FEngine (64 bits) created at 0x7fe9bb883000 (threading is enabled)
FEngine resolved backend: OpenGL
[Open3D WARNING] Read PNG failed: unable to parse header.
Traceback (most recent call last):
  File "test.py", line 2, in <module>
    vis = o3d.visualization.O3DVisualizer()
RuntimeError: [Open3D ERROR] Unsupported image number of channels: 0
```

**Expected behavior**
The instantiation and return of a visualization.O3DVisualizer

**Environment (please complete the following information):**

- Operating system: (e.g. OSX 11.2)
- Python version: (e.g. Python 3.8)
- Open3D version: (0.12.0)
- Is this remote workstation?: no
- How did you install Open3D?: (pip3.8 install open3d)

SBCV:
I think the reason for this is the missing `gui.Application.instance.initialize()` call

Full example:
```
import open3d as o3d
import open3d.visualization.gui as gui

gui.Application.instance.initialize()
vis = o3d.visualization.O3DVisualizer()

gui.Application.instance.add_window(vis)
gui.Application.instance.run()
```

MrBenGriffin:
@SBVC, thanks that worked.  it may be useful to mention this in a 'hello world'.

prewettg:
This error message is really unhelpful, so I have created #3103 to improve the error that you get.

PR #3103, opened 2021-03-02
Better error message if you don't initialize gui::Application before creating a window

prewettg:
It's easy to do something like:
```
import open3d as o3d
vis = o3d.visualization.O3DVisualizer()
...
```
You need to initialize the Application instance before creating anything UI-related, so that we can start up GLFW and find the resources. This spits out an error instead of continuing and crashing or getting unrelated-sounding errors like not being able to load images. (See #2995)

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3103)


errissa (approved):
> Reviewed 3 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3103)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3105, opened 2021-03-02
download test datasets from intel-isl/open3d_downloads

yxlao:
This PR introduces a systematic way of managing test data for Open3D.  Please check the readme of https://github.com/intel-isl/open3d_downloads for details. (more details to come here)

Migrations in future PRs:
1. `examples/python/reconstruction_system/scripts/download_dataset.py`: RealSense dataset 
2. `examples/python/open3d_tutorial.py`: open3d tutorial example meshes, e.g. bunny, eagle, armadillo, ...
3. `examples/python/utility/downloader.py`: `get_redwood_dataset()`
4. `examples/python/reconstruction_system/scripts`
    - `download_indoor_lidar_rgbd.sh`
    - `download_redwood_objects.sh`
    - `download_redwood_simulated.sh`
    - `download_stanford.sh`
    - `download_tutorial.sh`
5. `python/test/open3d_test.py`: `download_fountain_dataset()`
6. Pre-compiled binaries
    - MKL
    - Filament
    - ARM64 CMake

Let me know if I missed something in the list above. 

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3105)


examples/test_data/download_utils.py
```diff
@@ -0,0 +1,71 @@
+# ----------------------------------------------------------------------------
```
ssheorey, line 1: Can we combine `examples/python/reconstruction_system/scripts/download_dataset.py` with this file? That script supports:
- single place to specify all download URLs. Clients can access data by name instead of URL.
- Allowing users to list available datasets.
- extracting from archives.

yxlao: Yes, that is the plan. Will do that in subsequent PRs since there are several locations to migrate. I've added a list of files to migrate to this new mechanism in the PR description above.

yxlao:
@junha-l this will be needed for the NNS benchmark as well

ShubhamAgarwal12 (approved):

examples/test_data/download_utils.py
```diff
@@ -0,0 +1,71 @@
...
+            urllib.request.urlretrieve(url, save_path)
+            print(f"Downloaded {url}\n        to {str(save_path)}")
+        except Exception as e:
+            print(f"Failed to download {url}: {str(e)}")
+
+
+def download_all_files(overwrite=False):
+    for url, relative_path in map_url_to_relative_path.items():
+        save_path = _download_dir / relative_path
+        _download_file(url, save_path, overwrite)
```
ssheorey, line 71: We should be able to speed up downloads with multithreading (eg: `concurrent.futures.ThreadPoolExecutor`)

examples/test_data/download_file_list.py
```diff
@@ -0,0 +1,15 @@
...
+#
+# For example:
+# - relative download path: foo/bar/file.txt
+# - file will be saved to : Open3D/examples/test_data/open3d_downloads/foo/bar/file.txt
+#
+# See https://github.com/intel-isl/open3d_downloads for details on how to
+# manage the test data files.
+map_url_to_relative_path = {
+    f"{_repo_prefix}/RGBD/raycast_vtx_004.npy": "RGBD/raycast_vtx_004.npy",
+}
```
ssheorey, line 15: I would suggest having a mapping like this:
`{dataset_name:  struct (URL, checksum, relative_path)}`
This makes it a bit easier to refer to the datasets by name. It will usually just be the filename in the URL. We can refer to datasets by name in unit test code as well with a helper function that uses this to return paths from names.

ssheorey (commented):
Can we add checksums, especially for the large files? Here is an example:
https://gitlab.cern.ch/proteus/proteus/blob/ffecc13802515c423eaa809dd65801c3e3e79549/test/download_data.py
This can help with interrupted downloads and is also a security measure for binaries.

ssheorey (approved):

examples/test_data/download_utils.py
yxlao: Done, change it to 2 threads. We may not want to hit too hard on the server since it might trigger download protection from the server-side. Typically, the download time is negligible since it is done in parallel while we download and build 3rd-party libraries.

yxlao:
Summary for future PR:

1. Migrate remaining datasets
2. Add helper functions to refer to data files by name

examples/test_data/download_file_list.py
yxlao: - Switch to JSON with dictionary structure. 
- Added SHA256 checksum.
- Future PR: Add helper functions to refer to data files by name in both C++ and python

PR #3110, opened 2021-03-03
Delete GCloud SSH keys to keep the profile small

yxlao:
Trying to fix (e.g in https://github.com/intel-isl/Open3D/runs/2027032298):

image

Reference: https://github.com/kyma-project/test-infra/issues/93#issuecomment-457263589

This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3110)


yxlao:
The `master` branch is building correctly now. Closing.

ssheorey (approved):
This looks to be an infrequent maintenance operation. Can we move this to `gcloud-setup`? It's slightly clearer and will happen twice instead of 4 times per CI run.

PR #3019, opened 2021-02-10
t-ICP ComputeTransform major performance improvements.

reyanshsolis:
#### Parameters:
```
Max iterations: 30, Max_correspondence_distance : 0.03, Initial Correspondences: 24037
After downsampling with factor 0.01: PointCloud Size: Source: {95354}  Target: {63857}
PointToPlane: [Final Correspondences: 60834] Fitness: 0.637981  Inlier RMSE: 0.00781341
PointToPoint: [Final Correspondences: 38958] Fitness: 0.408562  Inlier RMSE: 0.0109094
```
#### This Branch:
To run `Registration` related benchmarks on `this branch`:
``` 
git checkout rishabh/tICP3.0-reduction
mkdir build && cd build
cmake -DBUILD_CUDA_MODULE=ON -DBUILD_BENCHMARKS=ON ..
make -j$(nproc) && ./bin/benchmarks --benchmark_filter=".*Registration*."
```
```
Benchmark                                                  Time             CPU   Iterations
--------------------------------------------------------------------------------------------
BenchmarkRegistrationICPLegacy/PointToPlane / CPU        735 ms          729 ms            1
BenchmarkRegistrationICPLegacy/PointToPoint / CPU       1394 ms         1389 ms            1
BenchmarkRegistrationICP/PointToPlane / CPU             1380 ms         1152 ms            1
BenchmarkRegistrationICP/PointToPlane / CUDA             218 ms          217 ms            3
BenchmarkRegistrationICP/PointToPoint / CPU             2554 ms         2224 ms            1
BenchmarkRegistrationICP/PointToPoint / CUDA             394 ms          393 ms            2
```
#### Current Master:
To run `Registration` related benchmarks on `master` [as of Mar9-2021] branch:
``` 
git checkout rishabh/benchmark_registration_master_mar9
mkdir build && cd build
cmake -DBUILD_CUDA_MODULE=ON -DBUILD_BENCHMARKS=ON ..
make -j$(nproc) && ./bin/benchmarks --benchmark_filter=".*Registration*."
```

```
BenchmarkRegistrationICPLegacy/PointToPlane / CPU          749 ms          749 ms            1
BenchmarkRegistrationICPLegacy/PointToPoint / CPU         1408 ms         1405 ms            1
BenchmarkRegistrationICP/PointToPlane / CPU               1757 ms         1537 ms            1
BenchmarkRegistrationICP/PointToPlane / CUDA               216 ms          216 ms            3
[...]

PR #3112, opened 2021-03-03
T-Registration pybinds and python-tests.

reyanshsolis:
Added: Registration and TransformationEstimation pybinds and python unit tests.

This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3112)


cpp/pybind/t/pipelines/pipelines.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+#include "pybind/open3d_pybind.h"
+#include "pybind/t/pipelines/registration/registration.h"
+
+namespace open3d {
+namespace t {
+namespace pipelines {
+
+void pybind_pipelines(py::module& m) {
+    py::module m_pipelines = m.def_submodule("pipelines");
```
yxlao, line 37: edit `docs/documented_modules.txt` to include this module, otherwise the documentation won't be generated

python/test/t/registration/test_registration.py
```diff
@@ -0,0 +1,232 @@
...
+import numpy as np
+import pytest
+
+from open3d_test import list_devices
+
+
+@pytest.mark.parametrize("device", list_devices())
+def test_icp_convergence_criteria_constructor(device):
+
+    #Constructor.
```
yxlao, line 38: `# ` (with space after), same for others

python/test/t/registration/test_registration.py
```diff
@@ -0,0 +1,232 @@
...
+    init_trans_legacy = np.eye(4)
+    init_trans_t = o3c.Tensor.eye(4, dtype, device)
+
+    evaluation_t = o3d.t.pipelines.registration.evaluate_registration(
+        source_t, target_t, max_correspondence_distance, init_trans_t)
+    evaluation_legacy = o3d.pipelines.registration.evaluate_registration(
+        source_legacy, target_legacy, max_correspondence_distance,
+        init_trans_legacy)
+
+    np.testing.assert_almost_equal(evaluation_t.inlier_rmse,
```
yxlao, line 111: What about `assert_allclose` (we typically use that in C++ unit tests as well)? What is the consideration between using `assert_almost_equal` v.s. `assert_allclose`?

yxlao (changes requested):

python/test/t/registration/test_registration.py
reyanshsolis: `assert_allclose` works! 
I found `assert_almost_equal` for single float / double value comparision, and `assert_allclose` for comparing general objects, but `assert_allclose` works. Done.

python/test/t/registration/test_registration.py
reyanshsolis: Done.

cpp/pybind/t/pipelines/pipelines.cpp
reyanshsolis: Done.

cpp/pybind/t/pipelines/registration/registration.cpp
```diff
@@ -0,0 +1,258 @@
...
+        PYBIND11_OVERLOAD_PURE(core::Tensor, TransformationEstimationBase,
+                               source, target, corres);
+    }
+};
+
+void pybind_registration_classes(py::module &m) {
+    // open3d.t.pipelines.registration.ICPConvergenceCriteria
+    py::class_<ICPConvergenceCriteria> convergence_criteria(
+            m, "ICPConvergenceCriteria",
+            "Class that defines the convergence criteria of ICP. ICP "
```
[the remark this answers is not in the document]
reyanshsolis: Done.

reyanshsolis:
Do not merge. 

As the new CorrespondeceSet definition has been added to master through PR #3019, The python tests for TransformationEstimation to be added to this PR itself.

reyanshsolis:
TransformationEstimation python unit tests according to new correspondence set definition, added.
Ready for review.

docs/documented_modules.txt
```diff
@@ -34,6 +34,7 @@ open3d.pipelines.color_map
 open3d.pipelines.integration
 open3d.pipelines.odometry
 open3d.pipelines.registration
+open3d.t.pipelines.registration
```
yxlao, line 37: I believe we need `open3d.t.pipelines`, and for consistency, sort the list such that this appears next to other `open3d.t` items.

yxlao: Currently it is not showing up:
image

cpp/pybind/t/pipelines/registration/registration.cpp
```diff
@@ -0,0 +1,256 @@
...
+    py::class_<ICPConvergenceCriteria> convergence_criteria(
+            m, "ICPConvergenceCriteria",
+            "Class that defines the convergence criteria of ICP. "
+            "ICP algorithm stops if the relative change of fitness and rmse "
+            "hit ``relative_fitness`` and ``relative_rmse`` individually, "
+            "or the iteration number exceeds ``max_iteration``.");
+    py::detail::bind_copy_functions<ICPConvergenceCriteria>(
+            convergence_criteria);
+    convergence_criteria
+            .def(py::init([](double fitness, double rmse, int itr) {
```
yxlao, line 76: If the constructor signature is exactly the same across python and c++ (e.g. we don't need any conversions), it should be possible to use `py::init<xxx, xxx>()` with the types directly. In this case, it will look like
```cpp
.def(py::init<double, double, int>(), "relative_fitness"_a = 1e-6,
     "relative_rmse"_a = 1e-6, "max_iteration"_a = 30)
```
Could you check if it works properly and apply to the rest of the constructors in this PR?

yxlao (changes requested):

yxlao:
We can download the generated docs artifact from CI e.g. https://github.com/intel-isl/Open3D/actions/runs/698168661 and open it in the browser to check if it works properly. 

The other thing i'd like to check is whether the double backtick works properly to format monospace code.

reyanshsolis:
> We can download the generated docs artifact from CI e.g. https://github.com/intel-isl/Open3D/actions/runs/698168661 and open it in the browser to check if it works properly.
> 
> The other thing i'd like to check is whether the double backtick works properly to format monospace code.

Thanks! This helped, I didn't know this.
I also set up docs in my local system to check. The issues are resolved and I have reviewed the docs also.

cpp/pybind/t/pipelines/registration/registration.cpp
reyanshsolis: Interesting. This works, thanks. I have also used the same for the default non-parameterized constructor. 
For future reference, can we also use `vector<double>`, `core::Tensor`, `t::geometry::PointCloud`, `core::Device`, `core::Dtype` in place of `xxx` in `py::init<xxx, xxx>()` ?

docs/documented_modules.txt
reyanshsolis: Resolved.

cpp/pybind/t/pipelines/registration/registration.cpp
reyanshsolis: Done.

cpp/pybind/t/pipelines/pipelines.cpp
yxlao, line 37: image

Current the module page has no docstrings. Could you add module docstrings for `geometry`, `io` and `pipelines`?  Some thing like:

```cpp
py::module m_pipelines = m.def_submodule("pipelines", "Open3D tensor-based geometry processing pipeline module. ...");
```

Same goes for the `geometry` and `io`.

yxlao (approved):

cpp/pybind/t/pipelines/pipelines.cpp
reyanshsolis: Done.

PR #3113, opened 2021-03-04
t::PointCloud from RGBD image

ssheorey:
Unit test, Python binding.
Release GIL in RealSense code for Python multithreading speedup.
Documentation improvements.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3113)


yxlao (changes requested):
> Reviewed 9 of 10 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3113)* status: all files reviewed, 2 unresolved discussions (waiting on @ssheorey and @theNded)

___
*[cpp/open3d/t/geometry/PointCloud.cpp, line 180 at r2](https://reviewable.io/reviews/intel-isl/open3d/3113#-MVTaMFb0ljBDz2N69QU:-MVTaMFb0ljBDz2N69QV:btjmk4x) ([raw file](https://github.com/intel-isl/open3d/blob/039a5871dc79870c03c0ab0e86adcd8b3f6134db/cpp/open3d/t/geometry/PointCloud.cpp#L180)):*
> ```cpp
>         image_colors = rgbd_image.color_.AsTensor();
>     } else {
>         image_colors = rgbd_image.color_.To(core::Dtype::Float32).AsTensor();
> ```

we can do both conditions with `.To(Float32, /*copy=*/false)`
___
*[cpp/open3d/t/geometry/kernel/PointCloudImpl.h, line 59 at r2](https://reviewable.io/reviews/intel-isl/open3d/3113#-MVTbLNeBqPg0Ly0U_uT:-MVTbLNeBqPg0Ly0U_uU:bl0do23) ([raw file](https://github.com/intel-isl/open3d/blob/039a5871dc79870c03c0ab0e86adcd8b3f6134db/cpp/open3d/t/geometry/kernel/PointCloudImpl.h#L59)):*
> ```objc
>          float depth_max,
>          int64_t stride) {
>     const bool have_colors = (image_colors.NumElements() != 0);
> ```

Is it possible to use something like `utility::optional<core::Tensor>` (with `const` and `&`), it would be more clear when the colors are empty. In this way, we don't need to make empty tensors to fit the API.

ssheorey:
> > Reviewed 9 of 10 files at r1, 1 of 1 files at r2.
> > _[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3113)_ status: all files reviewed, 2 unresolved discussions (waiting on @ssheorey and @theNded)
> 
> _[cpp/open3d/t/geometry/PointCloud.cpp, line 180 at r2](https://reviewable.io/reviews/intel-isl/open3d/3113#-MVTaMFb0ljBDz2N69QU:-MVTaMFb0ljBDz2N69QV:btjmk4x) ([raw file](https://github.com/intel-isl/open3d/blob/039a5871dc79870c03c0ab0e86adcd8b3f6134db/cpp/open3d/t/geometry/PointCloud.cpp#L180)):_
> 
> > ```c++
> >         image_colors = rgbd_image.color_.AsTensor();
> >     } else {
> >         image_colors = rgbd_image.color_.To(core::Dtype::Float32).AsTensor();
> > ```
> 
> we can do both conditions with `.To(Float32, /*copy=*/false)`

Done.

> _[cpp/open3d/t/geometry/kernel/PointCloudImpl.h, line 59 at r2](https://reviewable.io/reviews/intel-isl/open3d/3113#-MVTbLNeBqPg0Ly0U_uT:-MVTbLNeBqPg0Ly0U_uU:bl0do23) ([raw file](https://github.com/intel-isl/open3d/blob/039a5871dc79870c03c0ab0e86adcd8b3f6134db/cpp/open3d/t/geometry/kernel/PointCloudImpl.h#L59)):_
> 
> > ```objc
> >          float depth_max,
> >          int64_t stride) {
> >     const bool have_colors = (image_colors.NumElements() != 0);
> > ```
> 
> Is it possible to use something like `utility::optional<core::Tensor>` (with `const` and `&`), it would be more clear when the colors are empty. In this way, we don't need to make empty tensors to fit the API.

Done, though the result is a bit awkward due to references.

theNded (approved):
> Reviewed 9 of 10 files at r1, 1 of 1 files at r2, 5 of 6 files at r3, 4 of 4 files at r4.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3113)* status: all files reviewed, 4 unresolved discussions (waiting on @ssheorey and @yxlao)

___
*[cpp/open3d/t/geometry/PointCloud.cpp, line 174 at r4](https://reviewable.io/reviews/intel-isl/open3d/3113#-MVxrYx37eNVfiDsvoN5:-MVxrYx37eNVfiDsvoN6:bx0gfij) ([raw file](https://github.com/intel-isl/open3d/blob/5fc5b7abb9284e13427b042392d927b2f3e72809/cpp/open3d/t/geometry/PointCloud.cpp#L174)):*
> ```cpp
>                                            float depth_max,
>                                            int stride) {
>     rgbd_image.depth_.AsTensor().AssertDtype(core::Dtype::UInt16);
> ```

What is the preferred format of the depth image? uint16 depth + float rgb seems unconventional...
___
*[cpp/open3d/t/geometry/kernel/PointCloudImpl.h, line 111 at r2](https://reviewable.io/reviews/intel-isl/open3d/3113#-MVN-wsQ3GQXFDaJAnxi:-MVN-wsQ3GQXFDaJAnxj:bi6ujmk) ([raw file](https://github.com/intel-isl/open3d/blob/039a5871dc79870c03c0ab0e86adcd8b3f6134db/cpp/open3d/t/geometry/kernel/PointCloudImpl.h#L111)):*
> ```objc
>                     ti.RigidTransform(x_c, y_c, z_c, vertex + 0, vertex + 1,
>                                       vertex + 2);
>                     if (have_colors) {
> ```

One potential improvement is to use a templated function with a bool parameter and call accordingly. This may reduce divergence in cuda calls. I'm planning to switch to this design pattern later, does it sound reasonable?

ssheorey (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3113)* status: all files reviewed, 4 unresolved discussions (waiting on @theNded and @yxlao)

___
*[cpp/open3d/t/geometry/PointCloud.cpp, line 174 at r4](https://reviewable.io/reviews/intel-isl/open3d/3113#-MVxrYx37eNVfiDsvoN5:-MWPdvXy0-ig-IN3910m:b-1rtxyr) ([raw file](https://github.com/intel-isl/open3d/blob/5fc5b7abb9284e13427b042392d927b2f3e72809/cpp/open3d/t/geometry/PointCloud.cpp#L174)):*

What is the preferred format of the depth image? uint16 depth + float rgb seems unconventional...
</blockquote></details>

`uint16_t` depth and *any* Dtype for rgb. The typical input will be `uint16_t` depth and `uint8_t` rgb and the output will be float for both. Since we need to convert rgb to float in any case, I didn't restrict the input Dtype. float rgb is needed to display point clouds.
___
*[cpp/open3d/t/geometry/kernel/PointCloudImpl.h, line 111 at r2](https://reviewable.io/reviews/intel-isl/open3d/3113#-MVN-wsQ3GQXFDaJAnxi:-MWPfv1n9L9kr4rwaK8W:b-pnn4mj) ([raw file](https://github.com/intel-isl/open3d/blob/039a5871dc79870c03c0ab0e86adcd8b3f6134db/cpp/open3d/t/geometry/kernel/PointCloudImpl.h#L111)):*

One potential improvement is to use a templated function with a bool parameter and call accordingly. This may reduce divergence in cuda calls. I'm planning to switch to this design pattern later, does it sound reasonable?
</blockquote></details>

`have_colors` will be `const` for a kernel (same for all threads), so it shouldn't cause divergence. Have you found a measurable speedup with this template pattern?

yxlao (approved):
> Reviewed 3 of 6 files at r3, 4 of 4 files at r4, 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3113)* status: all files reviewed, 2 unresolved discussions (waiting on @theNded)

PR #3128, opened 2021-03-06
Open3D C++ example projects as separete repos

yxlao:
See:
* [Find Pre-Installed Open3D Package in CMake](https://github.com/intel-isl/open3d-cmake-find-package)
* [Use Open3D as a CMake External Project](https://github.com/intel-isl/open3d-cmake-external-project)

This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3128)


yxlao:
This PR simplifies instructions for new C++ users quite a bit.

**open3d-cmake-external-project**
- Add Windows / MSVC example to README
    - Done
- Use job matrix for [ubuntu1804, ubuntu2040]. Can we add Windows here?
    - Will add
- We should be OK reducing the cron schedule to daily.
    - Done
- The `--skip-for-unit-test` option may confuse users. Can we use another test code that will work in CI too?
    - Since we are providing it as a CMake example, I'd like to keep the CMake file minimal. Adding a separate test script will complicate the CMake scripts. 

**open3d-cmake-find-package**
Looks good.

PR #3136, opened 2021-03-07
Optimised CopyCPU and CopyCUDA for case when dst is contiguous and src has single element. [Fill]

reyanshsolis:
Benchmark/Zeros: 
Before: 
image

After:
```
-----------------------------------------------------
Benchmark           Time             CPU   Iterations
-----------------------------------------------------
Zeros/CPU         124 ms          124 ms            6
Zeros/CUDA       10.7 ms         10.7 ms           66

```

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3136)


cpp/open3d/core/kernel/UnaryEWCPU.cpp
```diff
@@ -130,6 +130,19 @@ void CopyCPU(const Tensor& src, Tensor& dst) {
         MemoryManager::Memcpy(dst.GetDataPtr(), dst.GetDevice(),
                               src.GetDataPtr(), src.GetDevice(),
                               src_dtype.ByteSize() * shape.NumElements());
+    } else if (dst.IsContiguous() && src.NumElements() == 1) {
+        DISPATCH_DTYPE_TO_TEMPLATE_WITH_BOOL(src_dtype, [&]() {
+            using src_t = scalar_t;
+            DISPATCH_DTYPE_TO_TEMPLATE_WITH_BOOL(dst_dtype, [&]() {
+                using dst_t = scalar_t;
+                dst_t value = static_cast<dst_t>(src.ToFlatVector<src_t>()[0]);
```
yxlao, line 138: use `src.Item<src_t>` to get the source value.

yxlao (changes requested):

cpp/open3d/core/kernel/UnaryEWCPU.cpp
reyanshsolis: Done. Still, it didn't resolve the issue for `FillObject`, where dtype is initialised as `core::Dtype dtype(core::Dtype::DtypeCode::Object, byte_size, class_name);`  [Link to test](https://github.com/intel-isl/Open3D/blob/692dc22ff611388d57d3c3a7c8c040fdee4eb232/cpp/tests/core/TensorObject.cpp#L114).

Can we support this case of we need to find out a way to avoid this case, such as if there is a way to check Dtype is for objects or iterate of other standard datatypes in if statement condition?

reyanshsolis:
Linking Issue PR #3115 with this PR, and closing #3115.

cpp/open3d/core/kernel/UnaryEWCPU.cpp
yxlao: The simple fix would be:
`} else if (dst.IsContiguous() && src.NumElements() == 1 && !src_dtype.IsObject()) {`

yxlao (approved):

PR #3147, opened 2021-03-09
Add ability to set center of rotation on SceneWidget, which is necessary for copying from a camera and not glitching on rotate

prewettg:
If we set the camera directory (Camera::CopyFrom()) we need to update the center of rotation in SceneWidget or rotating will continue using the old center of rotation which will look really glitchy. Since the camera doesn't know the center of rotation, we need to provide a way for the caller to set it.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3147)


errissa (approved):
> Reviewed 5 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3147)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @prewettg)

PR #3151, opened 2021-03-09
[Windows] Fix linker errors and warnings.

Dudulle:
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3151)


yxlao (changes requested):
> Reviewed 8 of 9 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3151)* status: all files reviewed, 1 unresolved discussion (waiting on @Dudulle)

___
*[CMakeLists.txt, line 27 at r2](https://reviewable.io/reviews/intel-isl/open3d/3151#-MWSd2-C13w4hpFm2XQ5:-MWSd2-C13w4hpFm2XQ6:b-1meppa) ([raw file](https://github.com/intel-isl/open3d/blob/b602608a83fa832b5a6104e41cb958dd22c4fd68/CMakeLists.txt#L27)):*
> ```CMake
> option(ENABLE_HEADLESS_RENDERING  "Use OSMesa for headless rendering"        OFF)
> if (BUILD_SHARED_LIBS)
>     option(STATIC_WINDOWS_RUNTIME "Use static (MT/MTd) Windows runtime"      OFF)
> ```

Do we have some reference on why we prefer these default values? Can we mix arbitrary `BUILD_SHARED_LIBS` with arbitrary `STATIC_WINDOWS_RUNTIME`?

Dudulle (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3151)* status: all files reviewed, 1 unresolved discussion (waiting on @yxlao)

___
*[CMakeLists.txt, line 27 at r2](https://reviewable.io/reviews/intel-isl/open3d/3151#-MWSd2-C13w4hpFm2XQ5:-MWT1MIY9y1yJjKMNzXx:b-ojboj7) ([raw file](https://github.com/intel-isl/open3d/blob/b602608a83fa832b5a6104e41cb958dd22c4fd68/CMakeLists.txt#L27)):*

Do we have some reference on why we prefer these default values? Can we mix arbitrary `BUILD_SHARED_LIBS` with arbitrary `STATIC_WINDOWS_RUNTIME`?
</blockquote></details>

We can mix them under some conditions.
When using the static runtime in a dynamic library, the runtime will be "unique" and will not share data with the calling application/library's runtime. This includes the heap manager.

This is causing issues because a lot of classes in Open3D use stl classes in methods or in class definitions. In debug mode there are asserts everywhere.

For reference: https://docs.microsoft.com/en-us/cpp/c-runtime-library/potential-errors-passing-crt-objects-across-dll-boundaries?view=msvc-160

yxlao (approved):
thanks @Dudulle

PR #3160, opened 2021-03-10
fix docstrings for ML modules

yxlao:
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3160)


yxlao:
### Known issue
`open3d.ml.tf.modules` and `open3d.ml.torch.modules` since it lacks the importer file, i.e. we shall have
- `Open3D/python/open3d/ml/tf/modules.py`
- `Open3D/python/open3d/ml/torch/modules.py`

This will be addressed in a future PR

### Outputs
image
image

PR #3164, opened 2021-03-11
Remove use of CMAKE_BUILD_TYPE.

Dudulle:
It is best to avoid using `CMAKE_BUILD_TYPE`'s value as it is not set when using multi-config generators.

It is currently only used in `_build_config.py` and external projects.

External projects can keep using it as it will be used when using single-config generators and multi-config generators will automatically set the `--config` argument with the correct configuration.

As replacing the build type in `_build_config.py` required enabling generator expressions for `$<CONFIG>`, I took the opportunity to add support for values other than `ON/OFF` in build options (second commit)

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3164)


yxlao:
Thanks @Dudulle , i like the idea in general as it supports Visual Studio better. 

However for single-config generators, it could be helpful to have a reasonable default build type, as we found that the build without optimization flags results in unbearably slow performance for some functions. 

Following https://blog.kitware.com/cmake-and-the-default-build-type/, i added back the support for default build type for single-config generators, while keeping the rest of the PR the unchanged.

yxlao (approved):

PR #3170, opened 2021-03-12
Add support for textures in TriangleMesh::operator+=

scimad:
Add support for textures in TriangleMesh::operator+=

`TriangleMesh::operator+=` previously didn't support adding meshes with textures. This PR fixes that. Note that if either one of the meshes doesn't have texture, the resulting mesh won't have textures.

Tested and verified compliance with `WriteTriangleMeshToOBJ` as well. Let me know if anything.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3170)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * RealSense sensor configuration, live capture and recording (with example and tutorial) (PR #2748)
 * Add mouselook for the legacy visualizer (PR #2551)
 * Add function to randomly downsample pointcloud (PR #3050)
+* Allow TriangleMesh with textures to be added (PR #3170)
 
 ## 0.11
 
--- a/cpp/open3d/geometry/TriangleMesh.cpp
+++ b/cpp/open3d/geometry/TriangleMesh.cpp
@@ -113,7 +113,7 @@ TriangleMesh &TriangleMesh::operator+=(const TriangleMesh &mesh) {
                                       mesh.triangle_material_ids_.size());
         for (size_t i = 0; i < mesh.triangle_material_ids_.size(); i++) {
             triangle_material_ids_[old_mat_id_num + i] =
-                    mesh.triangle_material_ids_[i] + old_tex_num;
+                    mesh.triangle_material_ids_[i] + (int)old_tex_num;
         }
     } else {
         triangle_uvs_.clear();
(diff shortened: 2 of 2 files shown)


scimad:
Requesting for review @errissa @yxlao .

errissa (changes requested):
I reviewed `WriteTriangleMeshToOBJ` and found that because of the way it exports the MTL file it is not currently possible to have a mesh that has some textured triangles and some un-textured. For now, your solution of clearing the textures if you `operator+=` a textured mesh with an untextured one is the best solution.

> Reviewed 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3170#-:-MVqgbdn0UIVCEiHdQnH:b-syq639)* status: all files reviewed, 1 unresolved discussion (waiting on @scimad)

___
*[cpp/open3d/geometry/TriangleMesh.cpp, line 119 at r2](https://reviewable.io/reviews/intel-isl/open3d/3170#-MVqgH8bDaN-M4RDRkpo:-MVqgH8bDaN-M4RDRkpp:btfs6pd) ([raw file](https://github.com/intel-isl/open3d/blob/b43f4282cda9c9ea48e5d35f67e38dee82df55a7/cpp/open3d/geometry/TriangleMesh.cpp#L119)):*
> ```cpp
>         }
>     } else {
>         triangle_uvs_.clear();
> ```

I don't think it's necessary to clear the UVs in this case. They may still be useful even if there are no texture images.

scimad:
@errissa ,I still think it is good to perform `triangle_uvs_.clear();` because if one of the meshes doesn't have `uvs`, `triangles_.size()` would be changed once we perform addition of meshes but `triangle_uvs_` won't, and hence `triangle_uvs_.size() == 3 * triangles_.size()` would be false.

This means, `HasTriangleUvs()` would return false and from an API level, there will be little point on still having `triangle_uvs_`.

Kindly let me know if I am missing something.

errissa (approved):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3170)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @scimad)

scimad (commented):
> Reviewed 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3170)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @scimad)

PR #3177, opened 2021-03-15
Fix incorrect coordinates for mouse events on Linux

prewettg:
On high-DPI monitors, glfwGetWindowContentScale() returns > 1, which triggers problems with the new Window System code. The result is that text is properly sized, but the mouse coordinates are reduced by half, making clicking on UI elements not working (or, rather, mapped to the upper left quarter of the window).

Tested on Linux, macOS, and Windows. (Windows works for 100%, 150%, 200%)

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3177)


errissa (approved):
> Reviewed 6 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3177)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @prewettg)

PR #3178, opened 2021-03-15
Expose rendering to a depth image to Python

prewettg:
Also:
- makes resulting depth iage range from 0 (near plane) to 1 (far plane) instead of simply returning the raw values from Filament's buffer (which are reversed)
- fixes widgets that aren't SceneWidget or a Layout not working if they are a top-level child. (If you created a window and then called `window->AddChild(a_label)` (and properly sized it) the label would not draw, but if you put it in a layout and made the layout top-level, it worked fine.)

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3178)


errissa (approved):
> Reviewed 8 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3178)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @prewettg)

PR #3193, opened 2021-03-19
Added pybinds for color grading enums

### LINKED ISSUE #2918: Setting color grading from python
Hi,
I'm having hard time figuring out how I am supposed to set the color grading using the rendering api.

I create a offscreen renderer 
`render = o3d.visualization.rendering.OffscreenRenderer(1920, 1080)`

and then want to set the tone mapping using
`render.scene.get_view.set_color_grading`

but cant figure out how to set the parameters of this function to use linear tone mapping.

Thanks a lot in advance for any help!

prewettg:
Also changes open3d.visualization.rendering.Open3DScene.`get_view` to `view` (it is a property, not a function). This should have minor impact on users, since the only thing that can be done with a view is setting the color grading, but that turns out to be impossible because the enums did not exist. Hence, we should have no users using `get_view`.

Resolves #2918.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3193)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * Add mouselook for the legacy visualizer (PR #2551)
 * Add function to randomly downsample pointcloud (PR #3050)
 * Allow TriangleMesh with textures to be added (PR #3170)
+* Python property of open3d.visualization.rendering.Open3DScene `get_view` has been renamed to `view`.
 
 ## 0.11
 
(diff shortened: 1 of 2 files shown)


errissa (approved):
> Reviewed 2 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3193)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @prewettg)

PR #3199, opened 2021-03-22
upgrade scipy and cmake version

yxlao:
- CMake: uses official arm64 release binary
- Scipy: upgrade to 1.5.4 for native arm64 support (reduces arm64 compile time)
  - 1.6+ scipy releases do not have native support for Python 3.6

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3199)


ssheorey (approved):
Can we use version variables (eg `SCIPY_VER`) for `arm64.yml` (similar to `windows.yml`) for easier maintenance?
We can add `CMAKE_VER` as well.
Looks like `scipy.spatial.cKDTree` has become more efficient from v1.4.0 to v1.5.4

PR #3201, opened 2021-03-22
handle the invalid tetra warning - fix #1829

### LINKED ISSUE #1829: Invalid tetra in TetraMesh when creating TriangleMesh with alpha shapes
When using #1621  work around for creating TraingleMesh from a point cloud with 'TriangleMesh.create_from_point_cloud_alpha_shape()', I get the following error: 
`[Open3D ERROR] [CreateFromPointCloudAlphaShape] invalid tetra in TetraMesh`

**To Reproduce**
1. Create point cloud by manually creating PointCloud() object, then adding in points from a Vector3dVector()
2. run:
` alpha = .03`
`tetra_mesh, pt_map = o3d.geometry.TetraMesh.create_from_point_cloud(pcd)`
`mesh = o3d.geometry.TriangleMesh.create_from_point_cloud_alpha_shape(pcd, alpha, tetra_mesh, pt_map)`
3. See error

 - OS: [Windows 10.0]
 - Python version: 3.6.7
 - Open3D version: 0.9.0.0
 - Is this remote workstation?:  no
 - How did you install Open3D?: conda 

````

### PR DESCRIPTION
**Pull request by runette**
fixes #1829 

Change the handling of "invalid tetra" from error to warning. If the tetra is invalid - do not perform the calculation on the tetra but cycle to the next tetra to continue building the mesh.

This does not prevent the original incorrect Tetra but does mean that the mesh is created although there may be a missing vertex.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3201)

### DIFF
--- a/cpp/open3d/geometry/SurfaceReconstructionAlphaShape.cpp
+++ b/cpp/open3d/geometry/SurfaceReconstructionAlphaShape.cpp
@@ -113,22 +113,23 @@ std::shared_ptr<TriangleMesh> TriangleMesh::CreateFromPointCloudAlphaShape(
         double dz = tmp.determinant();
         // clang-format on
         if (a == 0) {
-            utility::LogError(
+            utility::LogWarning(
                     "[CreateFromPointCloudAlphaShape] invalid tetra in "
                     "TetraMesh");
-        }
-        double r = std::sqrt(dx * dx + dy * dy + dz * dz - 4 * a * c) /
-                   (2 * std::abs(a));
+        } else {
+            double r = std::sqrt(dx * dx + dy * dy + dz * dz - 4 * a * c) /
+                       (2 * std::abs(a));
 
-        if (r <= alpha) {
-            mesh->triangles_.push_back(TriangleMesh::GetOrderedTriangle(
-                    tetra(0), tetra(1), tetra(2)));
-            mesh->triangles_.push_back(TriangleMesh::GetOrderedTriangle(
-                    tetra(0), tetra(1), tetra(3)));
-            mesh->triangles_.push_back(TriangleMesh::GetOrderedTriangle(
-                    tetra(0), tetra(2), tetra(3)));
-            mesh->triangles_.push_back(TriangleMesh::GetOrderedTriangle(
-                    tetra(1), tetra(2), tetra(3)));
+            if (r <= alpha) {
+                mesh->triangles_.push_back(TriangleMesh::GetOrderedTriangle(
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by runette** [comment]
The Ubuntu OpenBLAS error does not look like it is anything to do with this PR so I will not fix that

**Comment by benjaminum** [comment]
Hi @runette ,thank you for the PR.
Do you see any problems with the generated meshes with these changes (holes, invalid triangles, etc.)?

**Comment by runette** [comment]
The fact that you get the warning means that you probably do not have exactly the mesh that you wanted - there is "probably" going to be a missing vertex.

The problem at the moment is that the algorithm is failing on perfectly acceptable point cloud input. I don't have enough maths to work out why the algorithm is failing - why the determinant is zero. When it fails at the moment - you get no output and thus nothing to work with.

All this change does is go from "hard fail" to "soft fail". It warns and attempts to continue mesh generation.

In my use case, the meshes it has created are perfectly acceptable and it allows the application to complete and provide the expected result.

Even if there are cases where the mesh is invalid (as opposed to incomplete) - at least you have some output that can be used to debug.

**Comment by benjaminum** [comment]
I think the "soft fail" should be the preferred behavior, too.
It would be still good to be able to react to the problem for use cases where we want to have the "hard fail".
I think something like this should work
```python
m = TriangleMesh.create_from_point_cloud_alpha_shape(...)
if not m.is_watertight(): raise Exception()
```
Let me know if this assumption is wrong.

benjaminum (approved):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3201)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @runette)

runette:
> I think the "soft fail" should be the preferred behavior, too.
> It would be still good to be able to react to the problem for use cases where we want to have the "hard fail".
> I think something like this should work
> 
> ```python
> m = TriangleMesh.create_from_point_cloud_alpha_shape(...)
> if not m.is_watertight(): raise Exception()
> ```
> 
> Let me know if this assumption is wrong.

@benjaminum  As far as I understand, and my experience with using this PR, is that it is not true that the mesh created is NOT watertight. The algo will still try to create a watertight mesh. The "error" will be that there is one (or more in the case of multiple warnings) point that should be on the surface of the mesh that will be inside the mesh - i.e. that the mesh will not follow the expected surface as closely as it theoretically should.

benjaminum:
@runette Thanks for clarifying!

PR #3205, opened 2021-03-23
speed up arm64 build with docker and ccache

yxlao:
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3205)


.github/workflows/openblas.yml
```diff
@@ -0,0 +1,144 @@
...
+        run: |
+          docker run --rm open3d-openblas-ci:latest \
+            pytest python/test/ --ignore python/test/ml_ops/ --ignore python/test/t/io/test_realsense.py
+
+      # https://stackoverflow.com/a/34093828/1255535
+      - name: Copy ccache out from the container
+        run: |
+          docker run --rm --entrypoint \
+            cat open3d-openblas-ci:latest /${{ env.CCACHE_TAR_NAME }}.tar.gz \
+            > ${GITHUB_WORKSPACE}/${{ env.CCACHE_TAR_NAME }}.tar.gz
```
ssheorey, line 64: Using a mounted volume should be faster than `cat`:
https://stackoverflow.com/a/49357803

.github/workflows/openblas.yml
```diff
@@ -0,0 +1,144 @@
...
+          docker build --build-arg BASE_IMAGE=${{ env.BASE_IMAGE }} \
+                       --build-arg CMAKE_VER=${{ env.CMAKE_VER }} \
+                       --build-arg CCACHE_TAR_NAME=${{ env.CCACHE_TAR_NAME }} \
+                       -t open3d-openblas-ci:latest \
+                       -f .github/workflows/Dockerfile.openblas .
+
+      - name: C++ tests
+        run: |
+          docker run --rm open3d-openblas-ci:latest build/bin/tests \
+            --gtest_filter="-*CreateFromRGBDImage*"
```
ssheorey, line 120: `CreateFromRGBDImage` should work after PR #3207

.github/workflows/openblas.yml
```diff
@@ -0,0 +1,144 @@
...
+        run: |
+          docker run --rm open3d-openblas-ci:latest \
+            pytest python/test/ --ignore python/test/ml_ops/ --ignore python/test/t/io/test_realsense.py
+
+      # https://stackoverflow.com/a/34093828/1255535
+      - name: Copy ccache out from the container
+        run: |
+          docker run --rm --entrypoint \
+            cat open3d-openblas-ci:latest /${{ env.CCACHE_TAR_NAME }}.tar.gz \
+            > ${GITHUB_WORKSPACE}/${{ env.CCACHE_TAR_NAME }}.tar.gz
```
ssheorey, line 132: Use mounted volume instead (see SO answer)

.github/workflows/openblas.yml
```diff
@@ -0,0 +1,144 @@
...
+
+      - name: Checkout source code
+        uses: actions/checkout@v2
+        with:
+          submodules: true
+
+      - name: Maximize build space
+        run: |
+          source util/ci_utils.sh
+          maximize_ubuntu_github_actions_build_space
```
ssheorey, line 101: Looks like we have about 17G of space before this. Can we skip this step and save about 3 mins?

ssheorey (approved):
Is this OK to ignore?

> Warning: rning] The requested image's platform (linux/arm64) does not match the detected host platform (linux/amd64) and no specific platform was requested

.github/workflows/openblas.yml
yxlao: Done

.github/workflows/openblas.yml
yxlao: Done

.github/workflows/openblas.yml
yxlao: Done

PR #3207, opened 2021-03-23
Fix for CreateFromRGBDImage test

ssheorey:
Add special case for Image::To when there is no dtype conversion / scaling / offset

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3207)


yxlao (approved):

yxlao:
LGTM, let's see if the CI passes

PR #3213, opened 2021-03-25
Optimising Tensor RGBD Odometry.

reyanshsolis:
Before:
```
-----------------------------------------------------------------------
Benchmark                             Time             CPU   Iterations
-----------------------------------------------------------------------
ComputePosePointToPlane/CPU         1145 ms         1089 ms           1
ComputePosePointToPlane/CUDA         171 ms          171 ms           4
```
After: 
```
-----------------------------------------------------------------------
Benchmark                             Time             CPU   Iterations
-----------------------------------------------------------------------
ComputePosePointToPlane/CPU        64.6 ms         64.5 ms           11
ComputePosePointToPlane/CUDA       26.3 ms         26.3 ms           27
```
[TBB takes twice as long on CPU, with the current implementation. Until optimized, a hybrid design is used to run TBB on windows and OpenMP on rest].

____________________________________

This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3213)


theNded (approved):
> Reviewed 4 of 8 files at r9, 2 of 2 files at r10.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3213)* status: all files reviewed, 2 unresolved discussions (waiting on @reyanshsolis)

___
*[cpp/open3d/t/pipelines/kernel/RGBDOdometryCPU.cpp, line 42 at r10](https://reviewable.io/reviews/intel-isl/open3d/3213#-MWtWvgD8uJtBlWVEWOa:-MWtWvgD8uJtBlWVEWOb:bep2aak) ([raw file](https://github.com/intel-isl/open3d/blob/0854c0cb2532997d8216303a1aca012c3b953d05/cpp/open3d/t/pipelines/kernel/RGBDOdometryCPU.cpp#L42)):*
> ```cpp
> namespace odometry {
> 
> void CreateVertexMapCPU(const core::Tensor& depth_map,
> ```

I would still prefer to keep all these in the Impl header until we wrap up the entire vo system and it is stable. This looks good in a build system, but can be hard for debugging.
___
*[cpp/open3d/t/pipelines/kernel/RGBDOdometryCUDA.cu, line 183 at r10](https://reviewable.io/reviews/intel-isl/open3d/3213#-MWtWDkBDqIndF6ILP9b:-MWtWDkBDqIndF6ILP9c:b-ntwbzp) ([raw file](https://github.com/intel-isl/open3d/blob/0854c0cb2532997d8216303a1aca012c3b953d05/cpp/open3d/t/pipelines/kernel/RGBDOdometryCUDA.cu#L183)):*
> ```Cuda
>                         r);
> 
>                 if (valid) {
> ```

Can we perform a reduction inside the kernel? That's my initial idea, and that's what the reference implementations are doing and why they are fast: reduce local warps into shared memory, and simply atomic add to 29x1 global entry. Then we don't need 29xn matrices and can finish reduction in one pass.
This can be a future work if the current implementation is fast enough.

PR #3234, opened 2021-04-01
marge

sxcblh:
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3234)


yxlao:
Not sure what this is. We're using CMake to generate make, not using make directly.

PR #3265, opened 2021-04-08
Update stdgpu dependency to latest upstream

stotko:
Currently, a patched version of stdgpu is used since its CUDA backend automatically detects the machine's GPU and sets the architecture flags accordingly. In case no GPU can be found, an error is thrown. The latest upstream version of stdgpu relaxes this constraint such that the patched version is not needed anymore.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3265)


yxlao (approved):

yxlao:
Sometimes the GPU CI can be unstable. It's good if most of the GPU CIs are passing.

PR #3281, opened 2021-04-13
Fix memory leak in stdgpu hashmap backend

stotko:
The new default hashmap backend based on `stdgpu` can leak memory in its `Rehash` function when re-allocating memory to increase its capacity. Tested internally by intentionally producing a double-free error (calling `Free()` twice) to verify the first call releases the memory.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3281)


theNded (approved):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3281)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @yxlao)

PR #3298, opened 2021-04-16
Added italic, bold, and bold italic styles

prewettg:
Can only be set for Label and CollapsableVert. For example:
```label.font_style = gui.FontStyle.ITALIC```
```cvert.font_style = gui.FontStyle.BOLD```

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3298)


errissa (changes requested):
> Reviewed 68 of 68 files at r1.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3298)* status: all files reviewed, 2 unresolved discussions (waiting on @prewettg)

___
*[cpp/open3d/visualization/gui/Util.cpp, line 172 at r1](https://reviewable.io/reviews/intel-isl/open3d/3298#-MYgRVvo2Pt2zaduE5dk:-MYgRVvpBKuHn1CrkHG8:bf6wa94) ([raw file](https://github.com/intel-isl/open3d/blob/852356668afa68a677d6bcfd94fe306d0f878157/cpp/open3d/visualization/gui/Util.cpp#L172)):*
> ```cpp
>             }
>             // Otherwise fail
>             return "";
> ```

Wouldn't it be better to warn that the bold and/or italic font file couldn't be found and return the 'normal' font file that was passed in?
___
*[cpp/open3d/visualization/gui/Widget.h, line 53 at r1](https://reviewable.io/reviews/intel-isl/open3d/3298#-MYg5ydE4erP7W_DzXsV:-MYg5ydE4erP7W_DzXsW:b-oyxo19) ([raw file](https://github.com/intel-isl/open3d/blob/852356668afa68a677d6bcfd94fe306d0f878157/cpp/open3d/visualization/gui/Widget.h#L53)):*
> ```objc
> struct LayoutContext {
>     const Theme& theme;
>     FontContext& fonts;
> ```

Why a LayoutContext? Couldn't the FontContext be part of the Theme?

prewettg (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3298)* status: all files reviewed, 2 unresolved discussions (waiting on @errissa)

___
*[cpp/open3d/visualization/gui/Util.cpp, line 172 at r1](https://reviewable.io/reviews/intel-isl/open3d/3298#-MYgRVvo2Pt2zaduE5dk:-MYk1MNE11pZ9Ll_ZQ_Z:bigyj3h) ([raw file](https://github.com/intel-isl/open3d/blob/852356668afa68a677d6bcfd94fe306d0f878157/cpp/open3d/visualization/gui/Util.cpp#L172)):*

Wouldn't it be better to warn that the bold and/or italic font file couldn't be found and return the 'normal' font file that was passed in?
</blockquote></details>

This matches the previous behavior (the function was in Application.cpp). This way we can tell that the font file was not found. The case where it can't find the bold/italic is handled by the caller, which substitutes the normal font. Actually it's a bit more efficient, because it just assigns the normal font, rather than adding more glyphs to the atlas
___
*[cpp/open3d/visualization/gui/Widget.h, line 53 at r1](https://reviewable.io/reviews/intel-isl/open3d/3298#-MYg5ydE4erP7W_DzXsV:-MYkCUJE8inMWOI8L5ka:bg6dlbz) ([raw file](https://github.com/intel-isl/open3d/blob/852356668afa68a677d6bcfd94fe306d0f878157/cpp/open3d/visualization/gui/Widget.h#L53)):*

Why a LayoutContext? Couldn't the FontContext be part of the Theme?
</blockquote></details>

I've been thinking of the Theme as the description (so to speak) and a Context as all the internal stuff required for drawing (or layout in this case). Since the FontContext (currently) returns ImGUI objects, it isn't useful outside of the drawing path. But one could make a case that the Theme is only useful outside the drawing path for creation of items, so why not put the FontContext in the Theme. I'm not opposed to changing it. The current way has the advantage that it makes it easier to change the context later, as we've taken the pain of changing Theme -> LayoutContext now, but I would entertain arguments otherwise.

errissa (approved):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3298)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @prewettg)

PR #3313, opened 2021-04-20
Added SetFullscreen functions to gui::Window class

yankailab:
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/3313)

### DIFF
--- a/cpp/open3d/visualization/gui/Window.cpp
+++ b/cpp/open3d/visualization/gui/Window.cpp
@@ -193,6 +193,9 @@ struct Window::Impl {
     bool needs_redraw_ = true;  // set by PostRedraw to defer if already drawing
     bool is_resizing_ = false;
     bool is_drawing_ = false;
+    
+    Eigen::Vector2i saved_window_size_ = Eigen::Vector2i::Zero();
+    Eigen::Vector2i saved_window_pos_ = Eigen::Vector2i::Zero();
 };
 
 Window::Window(const std::string& title, int flags /*= 0*/)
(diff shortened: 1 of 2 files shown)


prewettg (changes requested):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3313)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @yankailab)

*[a discussion](https://reviewable.io/reviews/intel-isl/open3d/3313#-MYqMvcJ7dEfGS-7Gwwl:-MYqMvcJ7dEfGS-7Gwwm:b-ff8yeh) (no related file):*
Thanks for submitting this PR! I'd like to merge in this PR, but need to request a few changes first:
1) it looks like this PR sets qhull, tinygltf, and tinyobjloader to new commits. Since this PR does not use them, please remove these changes,
2) ToggleFullscreen() seems redundant, as the user could just call SetFullscreen(!IsFullscreen()), so I think it would be better to remove it.
3) All the GUI stuff needs python bindings, which you can find in `cpp/pybind/visualization/gui/gui.cpp`. I think following the pattern of visibility is probably best:  a `set_fullscreen` function and a `is_fullscreen` read-only property. If you could add the same to `cpp/pybind/visualization/o3dvisualizer.cpp`, that would be appreciated. You can build the Python package with `make pip-package`, with the resulting package in `<build>/lib/python-package/pip-package/open3d...` that you can install for development/testing.
4) We cannot merge until all the continuous integration tests pass. The style check is easy, but requires a manual `make apply-style`.

---
___
*[cpp/open3d/visualization/gui/Window.cpp, line 197 at r1](https://reviewable.io/reviews/intel-isl/open3d/3313#-MYqQYexDqpKsfeMwYft:-MYqQYexDqpKsfeMwYfu:b-eso94k) ([raw file](https://github.com/intel-isl/open3d/blob/dbc6fdac7b04094d1bc36bae6ee971645a63247f/cpp/open3d/visualization/gui/Window.cpp#L197)):*
> ```
>     Eigen::Vector2i saved_window_size_ = Eigen::Vector2i::Zero();
>     Eigen::Vector2i saved_window_pos_ = Eigen::Vector2i::Zero();
> ```
It would be more consistent with our current code to use `gui::Rect` for `saved_window_frame`.

PR #3346, opened 2021-04-27
Improve performance of stdgpu hashmap backend

stotko:
- Update to most recent upstream version of `stdgpu` which includes significant performance improvements for `clear()`
- Fix asynchronous `cudaMemset` calls leading to wrong benchmark results
- Increase threads per block in `stdgpu` backend to `128` for better performance
- Fuse erase kernels for better performance which is safe because `stdgpu` only invalidates the iterator in `erase()` that gets removed while keeping all other valid

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3346)

### DIFF
--- a/3rdparty/stdgpu/stdgpu.cmake
+++ b/3rdparty/stdgpu/stdgpu.cmake
@@ -8,7 +8,7 @@ ExternalProject_Add(
     ext_stdgpu
     PREFIX stdgpu
     GIT_REPOSITORY https://github.com/stotko/stdgpu.git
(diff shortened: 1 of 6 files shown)


theNded (approved):
> Reviewed 4 of 6 files at r1, 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3346)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @stotko)

PR #3350, opened 2021-04-28
Properly scale UInt8 colors when converting to Legacy Point Cloud

errissa:
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3350)

### DIFF
--- a/cpp/open3d/t/geometry/PointCloud.cpp
+++ b/cpp/open3d/t/geometry/PointCloud.cpp
@@ -27,6 +27,7 @@
 #include "open3d/t/geometry/PointCloud.h"
 
 #include <Eigen/Core>
+#include <limits>
 #include <string>
 #include <unordered_map>
 
@@ -259,13 +260,43 @@ open3d::geometry::PointCloud PointCloud::ToLegacyPointCloud() const {
                 core::eigen_converter::TensorToEigenVector3dVector(GetPoints());
     }
     if (HasPointColors()) {
-        auto colors = GetPointColors();
-        if (colors.GetDtype() == core::Dtype::UInt8) {
-            colors = colors.To(core::Dtype::Float64) / 255.0;
+        bool dtype_is_supported_for_conversion = true;
+        double normalization_factor = 1.0;
+        core::Dtype point_color_dtype = GetPointColors().GetDtype();
+
+        if (point_color_dtype == core::Dtype::UInt8) {
+            normalization_factor =
+                    1.0 /
+                    static_cast<double>(std::numeric_limits<uint8_t>::max());
+        } else if (point_color_dtype == core::Dtype::UInt16) {
+            normalization_factor =
+                    1.0 /
+                    static_cast<double>(std::numeric_limits<uint16_t>::max());
+        } else if (point_color_dtype != core::Dtype::Float32 &&
+                   point_color_dtype != core::Dtype::Float64) {
+            utility::LogWarning(
+                    "Dtype {} of color attribute is not supported for "
+                    "conversion to LegacyPointCloud and will be skipped. "
+                    "Supported dtypes include UInt8, UIn16, Float32, and "
+                    "Float64",
+                    point_color_dtype.ToString());
+            dtype_is_supported_for_conversion = false;
         }
 
-        pcd_legacy.colors_ =
-                core::eigen_converter::TensorToEigenVector3dVector(colors);
+        if (dtype_is_supported_for_conversion) {
+            if (normalization_factor != 1.0) {
+                core::Tensor rescaled_colors =
+                        GetPointColors().To(core::Dtype::Float64, true) *
+                        normalization_factor;
+                pcd_legacy.colors_ =
+                        core::eigen_converter::TensorToEigenVector3dVector(
+                                rescaled_colors);
+            } else {
+                pcd_legacy.colors_ =
+                        core::eigen_converter::TensorToEigenVector3dVector(
+                                GetPointColors());
+            }
+        }
     }
     if (HasPointNormals()) {
         pcd_legacy.normals_ =
--- a/cpp/open3d/t/geometry/PointCloud.cpp
+++ b/cpp/open3d/t/geometry/PointCloud.cpp
@@ -259,8 +259,13 @@ open3d::geometry::PointCloud PointCloud::ToLegacyPointCloud() const {
                 core::eigen_converter::TensorToEigenVector3dVector(GetPoints());
     }
     if (HasPointColors()) {
-        pcd_legacy.colors_ = core::eigen_converter::TensorToEigenVector3dVector(
-                GetPointColors());
+        auto colors = GetPointColors();
+        if (colors.GetDtype() == core::Dtype::UInt8) {
+            colors = colors.To(core::Dtype::Float64) / 255.0;
+        }
+
+        pcd_legacy.colors_ =
+                core::eigen_converter::TensorToEigenVector3dVector(colors);
     }
     if (HasPointNormals()) {
         pcd_legacy.normals_ =
--- a/cpp/open3d/t/geometry/PointCloud.cpp
+++ b/cpp/open3d/t/geometry/PointCloud.cpp
@@ -286,7 +286,7 @@ open3d::geometry::PointCloud PointCloud::ToLegacyPointCloud() const {
         if (dtype_is_supported_for_conversion) {
             if (normalization_factor != 1.0) {
                 core::Tensor rescaled_colors =
-                        GetPointColors().To(core::Dtype::Float64, true) *
+                        GetPointColors().To(core::Dtype::Float64) *
                         normalization_factor;
                 pcd_legacy.colors_ =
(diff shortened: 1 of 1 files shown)


cpp/open3d/t/geometry/PointCloud.cpp
```diff
@@ -259,8 +259,13 @@ open3d::geometry::PointCloud PointCloud::ToLegacyPointCloud() const {
                 core::eigen_converter::TensorToEigenVector3dVector(GetPoints());
     }
     if (HasPointColors()) {
-        pcd_legacy.colors_ = core::eigen_converter::TensorToEigenVector3dVector(
-                GetPointColors());
+        auto colors = GetPointColors();
+        if (colors.GetDtype() == core::Dtype::UInt8) {
+            colors = colors.To(core::Dtype::Float64) / 255.0;
+        }
+
```
reyanshsolis, line 266: Do we need to check and convert for `UInt16` also ?

cpp/open3d/t/geometry/PointCloud.cpp
```diff
@@ -259,8 +259,13 @@ open3d::geometry::PointCloud PointCloud::ToLegacyPointCloud() const {
                 core::eigen_converter::TensorToEigenVector3dVector(GetPoints());
     }
     if (HasPointColors()) {
-        pcd_legacy.colors_ = core::eigen_converter::TensorToEigenVector3dVector(
-                GetPointColors());
+        auto colors = GetPointColors();
```
reyanshsolis, line 262: We may use `core::Tensor` instead of `auto` here.

reyanshsolis: `colors = GetPointColors()` the underlying memory blob is shared. 
`colors = colors.To(core::Dtype::Float64) / 255.0` <- might modify the memory blob associated with the pointcloud. 
We may use `colors = GetPointColors().Clone()` for safety and simplicity. 

If we want to avoid copy in case we don't need to convert. We may do this:
```
core::Tensor colors;
if(colors.GetDtype() == core::Dtype::UInt8) {
    colors = GetPointColors().To(core::Dtype::Float64) / 255.0;
} else if ( ) {

} else {
    colors = GetPointColors();
}
pcd_legacy.colors_ =
                core::eigen_converter::TensorToEigenVector3dVector(colors);
```

reyanshsolis (changes requested):

cpp/open3d/t/geometry/PointCloud.cpp
yxlao: Yes, the alternative to `Clone()` is `To(core::Dtype::Float64, /*copy=*/true)`, this will also copy the original tensor contents.

cpp/open3d/t/geometry/PointCloud.cpp
stotko: We probably also do not want to hard-code the constant `255.0`. So something like `static_cast<double>(std::numeric_limits<uint8_t>::max())` would be much better to understand. This would also be more generic with respect to other types like `UInt16`, etc.

reyanshsolis:
Considering there is currently no restriction on t::PointCloud side, regarding Dtype for color attribute. How about this: 
```cpp
core::Tensor colors;
double normalization_factor = 0;
bool is_dtype_supported_for_conversion = true;

if(colors.GetDtype() == core::Dtype::UInt8) {
    normalization_factor = static_cast<double>(std::numeric_limits<uint8_t>::max())
} else if (colors.GetDtype() == core::Dtype::UInt16) {
    normalization_factor = static_cast<double>(std::numeric_limits<uint16_t>::max())
} else if (colors.GetDtype() == core::Dtype::Float32 || colors.GetDtype() == core::Dtype::Float64){
    normalization_factor = 1.0;
} else {
    is_dtype_supported_for_conversion = false;
    utility::LogWarning("Dtype {}, for color attribute, is not supported for conversion to LegacyPointCloud, and will be skipped. Supported dtypes include UInt8, UInt16, Float32, Float64.", colors.GetDtype());
}
if(is_dtype_supported_for_conversion) {
     colors = GetPointColors().To(core::Dtype::Float64) / 255.0;
     pcd_legacy.colors_ =
                core::eigen_converter::TensorToEigenVector3dVector(colors);
}
```

prewettg (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3350)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @reyanshsolis and @yxlao)

___
*[cpp/open3d/t/geometry/PointCloud.cpp, line 266 at r1](https://reviewable.io/reviews/intel-isl/open3d/3350#-MZKo0GeHuqnIp6Yuxb7-r1-266:-MZOsXGR1OLbbTym7SvQ:bukbwez) ([raw file](https://github.com/intel-isl/open3d/blob/cd3c79bb42c91c13cb0da7f12b048724911df165/cpp/open3d/t/geometry/PointCloud.cpp#L266)):*

We probably also do not want to hard-code the constant `255.0`. So something like `static_cast<double>(std::numeric_limits<uint8_t>::max())` would be much better to understand. This would also be more generic with respect to other types like `UInt16`, etc.
</blockquote></details>

As an _amicus curae_ advisory brief, I personally find `static_cast<double>(std::numeric_limits<uint8_t>::max())` to be an example of some of the worst of C++ readability. Compare that to `UCHAR_MAX` from <climits>, for instance, which is more readable? I get why everything is there, but you have to actually read it and think about what it all is, while `UCHAR_MAX` is pretty obvious what it does. And `val / 255.0` in the context of color conversions is pretty obvious once you've written any color conversion more than once. Unless you're going template it, I implore at least doing `const double uint8_max = ...` so readers like me don't have to parse that mess.

stotko (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3350)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @reyanshsolis and @yxlao)

___
*[cpp/open3d/t/geometry/PointCloud.cpp, line 266 at r1](https://reviewable.io/reviews/intel-isl/open3d/3350#-MZKo0GeHuqnIp6Yuxb7-r1-266:-MZRU6Ed7qbwcp8wtId_:bu2bhlm) ([raw file](https://github.com/intel-isl/open3d/blob/cd3c79bb42c91c13cb0da7f12b048724911df165/cpp/open3d/t/geometry/PointCloud.cpp#L266)):*

As an _amicus curae_ advisory brief, I personally find `static_cast<double>(std::numeric_limits<uint8_t>::max())` to be an example of some of the worst of C++ readability. Compare that to `UCHAR_MAX` from <climits>, for instance, which is more readable? I get why everything is there, but you have to actually read it and think about what it all is, while `UCHAR_MAX` is pretty obvious what it does. And `val / 255.0` in the context of color conversions is pretty obvious once you've written any color conversion more than once. Unless you're going template it, I implore at least doing `const double uint8_max = ...` so readers like me don't have to parse that mess.
</blockquote></details>

Yes, I agree that my suggestion is quite verbose, but the intent is clear (need max value and have to cast to double). I would still argue that it is better than a magic number, but not necessarily the "best" solution. I think @reyanshsolis came up with a good compromise that also covers your concerns.

cpp/open3d/t/geometry/PointCloud.cpp
```diff
@@ -259,8 +260,43 @@ open3d::geometry::PointCloud PointCloud::ToLegacyPointCloud() const {
...
+                    "Supported dtypes include UInt8, UIn16, Float32, and "
+                    "Float64",
+                    point_color_dtype.ToString());
+            dtype_is_supported_for_conversion = false;
+        }
+
+        if (dtype_is_supported_for_conversion) {
+            if (normalization_factor != 1.0) {
+                core::Tensor rescaled_colors =
+                        GetPointColors().To(core::Dtype::Float64, true) *
```
reyanshsolis, line 289: `/*copy=*/ true` might not be required here, as if `GetPointColors` is not `Float64` it will return a copy, and anyways `/` op will return a copy. So, using `true` here will make a double copy.

reyanshsolis: * `* op`

reyanshsolis (approved):

yxlao (approved):

cpp/open3d/t/geometry/PointCloud.cpp
errissa: @reyanshsolis That is a good point. Tensor::Mul creates a new tensor. I reverted to `To` without the copy.

errissa: @stotko Although using the `std::numeric_limits` makes the intent clear, it's also very verbose and I lean toward agreeing with @prewettg that there are other ways of making the intent clear without the verbosity. Moreover, I think of `std::numeric_limits` as a tool for generic programming which is not happening here. Nonetheless, your suggestion is more consistent with the code in t/geometry and core so I'm going with `std::numeric_limits` for getting the max value.

PR #3362, opened 2021-04-29
fix scalar boolean handling in pybind

yxlao:
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3362)


ssheorey (approved):
Thanks Yixing!

PR #3380, opened 2021-05-03
Fix CUDA builds on Windows.

### LINKED ISSUE #3231: Open3D Master VS2019 with CUDA 10.2 build issue
I'm using VS2019 on Windows 10 with CUDA 10.2.

Thank you!

image
image

### LINKED ISSUE #3331: Build fails on VS 2019 with CUDA enabled
**Describe the bug**
Build fails on VS 2019 with CUDA enabled 

This error can easily fixed by using roundf:
Error		calling a __host__ function("__roundf") from a __host__ __device__ function("open3d::t::pipelines::kernel::odometry::GetJacobianPointToPlane") is not allowed	tpipelines	H:\Source\C++\open3d\Open3D\cpp\open3d\t\pipelines\kernel\RGBDOdometryJacobianImpl.h	67	

However, how to fix these ?

Error		__declspec attributes ignored	tgeometry	H:\Source\C++\open3d\Open3D\cpp\open3d\core\hashmap\Dispatch.h	94	

- Open3D version: Current 
- Compiler version VS 2019

Dudulle:
There are a few different issues when building with CUDA.

The first issue is due to Visual Studio 16.9's math functions (see https://github.com/intel-isl/Open3D/issues/3231#issuecomment-811005721)
Those functions were replaced with their float overloads.

The `Block` class in `Dispatch.h` returns a reference to a member variable. This does not seem to be allowed with CUDA on Windows and the `__host__` and `__device__` keywords are ignored because of this.
This was fixed by replacing the `(size_t)` operator with `Get/Set` methods.

There was also a mismatch with the CRT used in stdgpu and dependencies of librealsense.
This caused issues when building with `STATIC_WINDOWS_RUNTIME=ON`.
For librealsense I have added a partial patch from https://github.com/IntelRealSense/librealsense/pull/8582 to fix the compilation as the other fixes conflict with the other patches already present and only fix warnings. 

Using `STATIC_WINDOWS_RUNTIME=ON` is not allowed anymore when using `BUILD_SHARED_LIBS=ON`.
This is because there doesn't seem to be any real use case possible without causing memory leaks or triggering asserts in debug mode.
See https://github.com/intel-isl/Open3D/pull/3151#pullrequestreview-618301875 for more details.

The Windows CI now contains configurations for CUDA so Windows-specific errors will be easier to catch before being introduced.

Fixes #3231
Fixes #3331

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3380)


stotko (changes requested):
> Reviewed 11 of 11 files at r1.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3380)* status: all files reviewed, 5 unresolved discussions (waiting on @Dudulle)

___
*[3rdparty/librealsense/fix-cudacrt.patch, line 25 at r1](https://reviewable.io/reviews/intel-isl/open3d/3380#-MZqRJgjDH9f6TBsXT62:-MZqRJgjDH9f6TBsXT63:b-bjutqm) ([raw file](https://github.com/intel-isl/open3d/blob/41e5a54d299625f72cc82b67e2b3927ecf8f34b8/3rdparty/librealsense/fix-cudacrt.patch#L25)):*
> ```Diff
> +        if(${flag_var} MATCHES "-MD")
> +            string(REGEX REPLACE "-MD" "-MT" ${flag_var} "${${flag_var}}")
> +        endif(${flag_var} MATCHES "-MD")
> ```

nit: Repetition of argument in `endif` has no effect (only for backward compatibility with very old CMake versions). Removing it makes the condition a bit cleaner and easier to read.
___
*[3rdparty/librealsense/fix-cudacrt.patch, line 27 at r1](https://reviewable.io/reviews/intel-isl/open3d/3380#-MZqRuP645CMNvQMSzfT:-MZqRuP645CMNvQMSzfU:buo9nmt) ([raw file](https://github.com/intel-isl/open3d/blob/41e5a54d299625f72cc82b67e2b3927ecf8f34b8/3rdparty/librealsense/fix-cudacrt.patch#L27)):*
> ```Diff
> +        endif(${flag_var} MATCHES "-MD")
> +    endforeach(flag_var)
> +endif(WIN32 AND BUILD_WITH_STATIC_CRT)
> ```

Same here
___
*[cpp/benchmarks/CMakeLists.txt, line 1 at r1](https://reviewable.io/reviews/intel-isl/open3d/3380#-MZqSE6y02kFMPTn9iYb:-MZqSE6y02kFMPTn9iYc:bqg4udn) ([raw file](https://github.com/intel-isl/open3d/blob/41e5a54d299625f72cc82b67e2b3927ecf8f34b8/cpp/benchmarks/CMakeLists.txt#L1)):*
> ```CMake
> cmake_minimum_required(VERSION 3.0...3.18)
> ```

This can be completely removed as the top-level `CMakeLists.txt` already sets the policy level as well as the minimum version to `3.18`.
___
*[cpp/open3d/core/hashmap/Dispatch.h, line 94 at r1](https://reviewable.io/reviews/intel-isl/open3d/3380#-MZqSepR20aYBnc1uzws:-MZqSepS7msB99nvm3JU:b1x7l9h) ([raw file](https://github.com/intel-isl/open3d/blob/41e5a54d299625f72cc82b67e2b3927ecf8f34b8/cpp/open3d/core/hashmap/Dispatch.h#L94)):*
> ```objc
>     }
> 
>     T OPEN3D_HOST_DEVICE Get(size_t i) const { return data_[i]; }
> ```

Can you explain what exact error you are facing here? Is this a limitation/bug in Windows + CUDA?
___
*[cpp/tests/CMakeLists.txt, line 1 at r1](https://reviewable.io/reviews/intel-isl/open3d/3380#-MZqS_PX-HD-Uw59VA_C:-MZqS_PX-HD-Uw59VA_D:buo9nmt) ([raw file](https://github.com/intel-isl/open3d/blob/41e5a54d299625f72cc82b67e2b3927ecf8f34b8/cpp/tests/CMakeLists.txt#L1)):*
> ```CMake
> cmake_minimum_required(VERSION 3.0...3.18)
> ```

Same here

stotko:
Thanks for contributing the Windows + CUDA fixes. All the changes including the updates to CI are highly appreciated. I just have a few minor comments that should be pretty easy to address.

Dudulle (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3380)* status: all files reviewed, 4 unresolved discussions (waiting on @stotko)

___
*[3rdparty/librealsense/fix-cudacrt.patch, line 27 at r1](https://reviewable.io/reviews/intel-isl/open3d/3380#-MZqRuP645CMNvQMSzfT:-MZsDxfnAblKm9O4wDpL:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/41e5a54d299625f72cc82b67e2b3927ecf8f34b8/3rdparty/librealsense/fix-cudacrt.patch#L27)):*

Same here
</blockquote></details>

Done.
___
*[cpp/benchmarks/CMakeLists.txt, line 1 at r1](https://reviewable.io/reviews/intel-isl/open3d/3380#-MZqSE6y02kFMPTn9iYb:-MZrjoPb40rJiXwfz-V5:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/41e5a54d299625f72cc82b67e2b3927ecf8f34b8/cpp/benchmarks/CMakeLists.txt#L1)):*

This can be completely removed as the top-level `CMakeLists.txt` already sets the policy level as well as the minimum version to `3.18`.
</blockquote></details>

Done.
___
*[cpp/open3d/core/hashmap/Dispatch.h, line 94 at r1](https://reviewable.io/reviews/intel-isl/open3d/3380#-MZqSepR20aYBnc1uzws:-MZrjs99C72DDcuDU-H2:b-ctre39) ([raw file](https://github.com/intel-isl/open3d/blob/41e5a54d299625f72cc82b67e2b3927ecf8f34b8/cpp/open3d/core/hashmap/Dispatch.h#L94)):*

Can you explain what exact error you are facing here? Is this a limitation/bug in Windows + CUDA?
</blockquote></details>

Yes this is specific to Windows.
Returning a reference/pointer here causes the compiler to ignore the '\_\_host__ \_\_device__' keywords so the method ends up being host only.
This causes the error specified in #3331
___
*[cpp/tests/CMakeLists.txt, line 1 at r1](https://reviewable.io/reviews/intel-isl/open3d/3380#-MZqS_PX-HD-Uw59VA_C:-MZrjqNN2rch4R67DsMw:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/41e5a54d299625f72cc82b67e2b3927ecf8f34b8/cpp/tests/CMakeLists.txt#L1)):*

Same here
</blockquote></details>

Done.

stotko (approved):
> Reviewed 7 of 8 files at r3.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3380)* status: 10 of 11 files reviewed, 2 unresolved discussions (waiting on @Dudulle and @stotko)

___
*[3rdparty/librealsense/fix-cudacrt.patch, line 37 at r3](https://reviewable.io/reviews/intel-isl/open3d/3380#-MZsMyvo0gCrLu1eajB-:-MZsMyvo0gCrLu1eajB0:bj5qxcl) ([raw file](https://github.com/intel-isl/open3d/blob/2b98ba5b2b422a1c5d1c0209426d98048e832d18/3rdparty/librealsense/fix-cudacrt.patch#L37)):*
> ```Diff
>              if(${flag_var} MATCHES "/MD")
>                  string(REGEX REPLACE "/MD" "/MT" ${flag_var} "${${flag_var}}")
> -            endif(${flag_var} MATCHES "/MD")
> ```

After looking at this again, these changes are just noise. So ideally, `CMake/lrs_macros.cmake` should not be touched by this patch.

Dudulle (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3380)* status: 10 of 11 files reviewed, 2 unresolved discussions (waiting on @ssheorey, @stotko, and @yxlao)

___
*[3rdparty/librealsense/fix-cudacrt.patch, line 37 at r3](https://reviewable.io/reviews/intel-isl/open3d/3380#-MZsMyvo0gCrLu1eajB-:-MZsWQx7-jFWO6Ig5Bnt:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/2b98ba5b2b422a1c5d1c0209426d98048e832d18/3rdparty/librealsense/fix-cudacrt.patch#L37)):*

After looking at this again, these changes are just noise. So ideally, `CMake/lrs_macros.cmake` should not be touched by this patch.
</blockquote></details>

Done.

ssheorey (changes requested):
Thanks @dudulle for keeping Windows CUDA up to date, and especially for adding the CUDA build to the CI!

> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3380#-:-MZtAMw_94KzGQ3HUqbl:b-aziaq0)* status: 10 of 11 files reviewed, 5 unresolved discussions (waiting on @Dudulle, @stotko, and @yxlao)

___
*[.github/workflows/windows.yml, line 56 at r4](https://reviewable.io/reviews/intel-isl/open3d/3380#-MZt2XXqD1w-gCVqbFJd:-MZt2XXr4MHgtQ3wLyDX:bo7oox0) ([raw file](https://github.com/intel-isl/open3d/blob/dd397f3d4f969326769b374d3c2312cc7a916919/.github/workflows/windows.yml#L56)):*
> ```YAML
>         run: |
>           # Download and install CUDA 11.0
>           Invoke-WebRequest http://developer.download.nvidia.com/compute/cuda/11.0.3/network_installers/cuda_11.0.3_win10_network.exe -OutFile cuda.exe
> ```

Save CUDA_VER: "11.0" / CUDA_VER_FULL: "11.0.3", as an environment variable at the top to allow easy maintenance.
___
*[.github/workflows/windows.yml, line 127 at r4](https://reviewable.io/reviews/intel-isl/open3d/3380#-MZt4_lwC7dMK5CjiNSQ:-MZt4_lwC7dMK5CjiNSR:bjo3u94) ([raw file](https://github.com/intel-isl/open3d/blob/dd397f3d4f969326769b374d3c2312cc7a916919/.github/workflows/windows.yml#L127)):*
> ```
>  && matrix.BUILD_CUDA == 'OFF'
> ```
Let's skip this. The python package should load and work correctly when built with CUDA on a machine without a CUDA GPU.
___
*[cpp/CMakeLists.txt, line 46 at r4](https://reviewable.io/reviews/intel-isl/open3d/3380#-MZt8b5y-RKI-HEFhhpX:-MZt8b5y-RKI-HEFhhpY:bgmthgo) ([raw file](https://github.com/intel-isl/open3d/blob/dd397f3d4f969326769b374d3c2312cc7a916919/cpp/CMakeLists.txt#L46)):*
> ```
> list(APPEND CUDA_DISABLE_MSVC_WARNINGS /wd4700) # uninitialized local variable used (thrust)
> ```
This warning usually means a bug. We shouldn't disable if for all of Open3D. Can you use `#pragma warning pop/push` to disable it locally for thrust?

yxlao (approved):
> Reviewed 3 of 11 files at r1, 5 of 8 files at r3, 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3380)* status: all files reviewed, 5 unresolved discussions (waiting on @Dudulle and @stotko)

Dudulle (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3380)* status: 7 of 11 files reviewed, 5 unresolved discussions (waiting on @ssheorey, @stotko, and @yxlao)

___
*[.github/workflows/windows.yml, line 56 at r4](https://reviewable.io/reviews/intel-isl/open3d/3380#-MZt2XXqD1w-gCVqbFJd:-MZto0tZ3UfO1CkzmTca:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/dd397f3d4f969326769b374d3c2312cc7a916919/.github/workflows/windows.yml#L56)):*

Save CUDA_VER: "11.0" / CUDA_VER_FULL: "11.0.3", as an environment variable at the top to allow easy maintenance.
</blockquote></details>

Done.
___
*[.github/workflows/windows.yml, line 127 at r4](https://reviewable.io/reviews/intel-isl/open3d/3380#-MZt4_lwC7dMK5CjiNSQ:-MZtDlpE13b6O6dtU5oM:bx44a47) ([raw file](https://github.com/intel-isl/open3d/blob/dd397f3d4f969326769b374d3c2312cc7a916919/.github/workflows/windows.yml#L127)):*

> ```
>  && matrix.BUILD_CUDA == 'OFF'
> ```
Let's skip this. The python package should load and work correctly when built with CUDA on a machine without a CUDA GPU.
</blockquote></details>

The python package is actually still not fixed.
I've tried with your previous patch from https://github.com/intel-isl/Open3D/pull/3036#issuecomment-791814077 and it does seem to work but this step still fails because it tries to load the cpu module if there is no CUDA device (see https://github.com/Dudulle/Open3D/runs/2505828836?check_suite_focus=true)

Besides that the patch seems to be working, should I include it in this PR?
___
*[cpp/CMakeLists.txt, line 46 at r4](https://reviewable.io/reviews/intel-isl/open3d/3380#-MZt8b5y-RKI-HEFhhpX:-MZtD-gBBe5cYM9eN665:bz6k9ym) ([raw file](https://github.com/intel-isl/open3d/blob/dd397f3d4f969326769b374d3c2312cc7a916919/cpp/CMakeLists.txt#L46)):*

> ```
> list(APPEND CUDA_DISABLE_MSVC_WARNINGS /wd4700) # uninitialized local variable used (thrust)
> ```
This warning usually means a bug. We shouldn't disable if for all of Open3D. Can you use `#pragma warning pop/push` to disable it locally for thrust?
</blockquote></details>

This warning comes from headers in the CUDA toolkit so it would not be convenient to do this.
The suppression of this warning is limited to CUDA 11.0 and lower, and only for code compiled with nvcc so bugs can still be catched easily.

Dudulle:
The `<cmath>` header in Visual Studio has been properly fixed a few days ago (https://github.com/microsoft/STL/pull/1886)
I believe it would be best to wait for the update and remove f66b90e4c3c8f07aabcd71f9b97047a1be8d7a5b from this PR as it is no longer needed.

I've also added an extra commit to remove a few unused function parameters causing issues with CUDA 11.3

stotko:
Thanks for further testing this PR and keeping it up-to-date. It's also good to hear that Microsoft fixed the issues in `<cmath>` upstream in their STL implementation.

I think we should keep your fixes to maintain compatibility with users that have not updated their Visual Studio version. Perhaps adding a comment with a reference to the fix (and the expected version that will contain it) would be helpful to indicate that this can cleaned up somewhere in the future.

Dudulle:
@stotko Yes, I was expecting something like this to be included in the next revision or at worse 16.10 but I just saw [here](https://github.com/microsoft/STL/wiki/Changelog#expected-in-vs-2022-170-preview-2) that it might not be fixed until the next major release.

Where would you recommend to include the reference?
Those methods are currently in multiple places in two different files, and we can probably expect more in future.

ssheorey (changes requested):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3380)* status: 7 of 12 files reviewed, 3 unresolved discussions (waiting on @Dudulle, @stotko, and @yxlao)

___
*[.github/workflows/windows.yml, line 127 at r4](https://reviewable.io/reviews/intel-isl/open3d/3380#-MZt4_lwC7dMK5CjiNSQ:-M_1pwrH4z8D8ZDzKYkQ:b3i21jl) ([raw file](https://github.com/intel-isl/open3d/blob/dd397f3d4f969326769b374d3c2312cc7a916919/.github/workflows/windows.yml#L127)):*
<details><summary><i>Previously, Dudulle (Cyril Kaiser) wrote…</i></summary><blockquote>

The python package is actually still not fixed.
I've tried with your previous patch from https://github.com/intel-isl/Open3D/pull/3036#issuecomment-791814077 and it does seem to work but this step still fails because it tries to load the cpu module if there is no CUDA device (see https://github.com/Dudulle/Open3D/runs/2505828836?check_suite_focus=true)

Besides that the patch seems to be working, should I include it in this PR?
</blockquote></details>

You are correct that it will not work on a non-CUDA machine. We have two separate code paths for CPU and CUDA support and these are built in 2 separate DLLs. So if we build with CUDA, only the CUDA DLL gets built which doesn't work on a non-CUDA machine. Let's not test the CUDA wheel for now.

Please go ahead and include the patch. I was able to run the Python unit tests with the patch on a CUDA machine. The Python wheel will not work on a CUDA machine without that patch. I can approve the PR after that.
___
*[cpp/CMakeLists.txt, line 46 at r4](https://reviewable.io/reviews/intel-isl/open3d/3380#-MZt8b5y-RKI-HEFhhpX:-M_1seZY7oJgPLQliMyY:b-jiciyl) ([raw file](https://github.com/intel-isl/open3d/blob/dd397f3d4f969326769b374d3c2312cc7a916919/cpp/CMakeLists.txt#L46)):*
<details><summary><i>Previously, Dudulle (Cyril Kaiser) wrote…</i></summary><blockquote>

This warning comes from headers in the CUDA toolkit so it would not be convenient to do this.
The suppression of this warning is limited to CUDA 11.0 and lower, and only for code compiled with nvcc so bugs can still be catched easily.
</blockquote></details>

OK, let's keep this.

Dudulle (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3380)* status: 4 of 17 files reviewed, 3 unresolved discussions (waiting on @ssheorey, @stotko, and @yxlao)

___
*[cpp/CMakeLists.txt, line 46 at r4](https://reviewable.io/reviews/intel-isl/open3d/3380#-MZt8b5y-RKI-HEFhhpX:-M_HkzO_7dhqWf_34jET:bwrjm3p) ([raw file](https://github.com/intel-isl/open3d/blob/dd397f3d4f969326769b374d3c2312cc7a916919/cpp/CMakeLists.txt#L46)):*

OK, let's keep this.
</blockquote></details>

This is actually still required with CUDA 11.1
I've made changes it to include it, and to only disable the warning in debug mode as it doesn't appear in other configurations.

Dudulle:
Added one last fix for `zeromq` in `RelWithDebInfo` and `MinSizeRel` builds (CRT mismatch)

I've ran a [more extensive CI](https://github.com/Dudulle/Open3D/actions/runs/826711450) with all minor versions of CUDA from 10.1 to 11.2. All of them build without error now.
There's just one error that appears randomly with CUDA 10.2 in the CI, but I couldn't reproduce it locally.

PR #3389, opened 2021-05-05
fix index set for 0-D tensor

ShubhamAgarwal12:
Fixed IndexSet() for 0-D tensor (Advanced indexing cannot handle 0-D tensor).

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3389)

### DIFF
--- a/cpp/open3d/core/Tensor.cpp
+++ b/cpp/open3d/core/Tensor.cpp
@@ -702,8 +702,34 @@ Tensor Tensor::Slice(int64_t dim,
 }
 
 Tensor Tensor::IndexGet(const std::vector<Tensor>& index_tensors) const {
+    if (NumDims() == 0) {
+        const std::string error_prefix =
+                "A 0-D tensor can only be indexed by a 0-D boolean tensor";
+        if (index_tensors.size() != 1) {
+            utility::LogError("{}, but got {} index tensors.", error_prefix,
+                              index_tensors.size());
+        }
+        Tensor index_tensor = index_tensors[0];
+        index_tensor.AssertShape(
+                {}, fmt::format("{}, but got shape {}.", error_prefix,
+                                index_tensor.GetShape().ToString()));
+        index_tensor.AssertDtype(
+                Dtype::Bool, fmt::format("{}, but got dtype {}.", error_prefix,
+                                         index_tensor.GetDtype().ToString()));
+
+        if (index_tensor.IsNonZero()) {
+            // E.g. np.array(5)[np.array(True)].
+            return *this;
+        } else {
+            // E.g. np.array(5)[np.array(False)].
+            // The output tensor becomes 1D of 0 element.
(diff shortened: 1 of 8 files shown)


cpp/open3d/core/Tensor.cpp
```diff
@@ -712,8 +738,58 @@ Tensor Tensor::IndexGet(const std::vector<Tensor>& index_tensors) const {
...
+                Dtype::Bool, fmt::format("{}, but got dtype {}.",
+                                         index_tensor.GetDtype().ToString()));
+
+        // E.g.
+        // a = np.array(5)
+        // a[np.array(True)]  = 10                 // Works, assigned
+        // a[np.array(True)]  = np.array(10)       // Works, assigned
+        // a[np.array(True)]  = np.array([10])     // Works, assigned
+        // a[np.array(True)]  = np.array([[10]])   // Cannot assign 2D
+        // a[np.array(True)]  = np.array([10, 11]) // Cannot assign 1+ values
```
yxlao, line 762: Add unit test for this case. Previously the shape of `src_tensor` it is not checked.

cpp/open3d/core/Tensor.cpp
```diff
@@ -702,8 +702,34 @@ Tensor Tensor::Slice(int64_t dim,
...
+                Dtype::Bool, fmt::format("{}, but got dtype {}.",
+                                         index_tensor.GetDtype().ToString()));
+
+        if (index_tensor.IsNonZero()) {
+            // E.g. np.array(5)[np.array(True)].
+            return *this;
+        } else {
+            // E.g. np.array(5)[np.array(False)].
+            // The output tensor becomes 1D of 0 element.
+            return Tensor(/*shape=*/{0}, GetDtype(), GetDevice());
```
yxlao, line 726: Previously the dtype hard-coded to int64_t. We should keep the same type. Add unit test for this.

python/test/core/test_core.py
```diff
@@ -821,6 +821,55 @@ def test_boolean_advanced_indexing(device):
     o3_y = o3_x[o3_row_sum <= 2, :]
     np.testing.assert_equal(np_y, o3_y.cpu().numpy())
 
+    o3_a = o3d.core.Tensor(5.2, device=device)
```
yxlao, line 824: Writing it as one line is more readable.

Index get:
```python
np_result = np.array(5)[np.array(False)]
o3_result = ...
```

Index set:
```python
np_src = np.array(5)
np_src[np.array(True)] = 10

np_src = np.array(5)
np_src[np.array(True)] = np.array(10)

np_src = np.array(5)
np_src[np.array(True)] = np.array([10])

np_src = np.array(5)
np_src[np.array(True)] = np.array([[10]])  # assert throw exception

np_src = np.array(5)
np_src[np.array(True)] = np.array([10, 11]) # assert throw exception
...

yxlao (changes requested):

yxlao (approved):

PR #3392, opened 2021-05-05
webrtc-based web visualizer and jupyter support

yxlao:
## Building 
- Standalone mode (no jupyter)
    Set  `-DBUILD_JUPYTER_EXTENSION=OFF`. WebRTC shall work by default on Ubuntu.
- Jupyter mode
    1. Be on Ubuntu
    2. Install `npm`, `yarn`
    3. `pip install ipywidgets jupyter_packaging jupyterlab`
    4. `-DBUILD_JUPYTER_EXTENSION=ON` and `make install-pip-package -j`

## Usage
- Standalone mode
    - C++: Insert `visualization::webrtc_server::WebRTCServer::GetInstance().EnableWebRTC();` before all GUI calls. See `examples/cpp/DrawWebRTC.cpp`.
    - Python: Insert `o3d.visualization.webrtc_server.WebRTCServer.instance.enable_webrtc()` before all GUI calls. See `examples/python/gui/draw_webrtc.py`.
    - Default web server address is `localhost:8888`. For remote server, you might need to set `WEBRTC_IP` and `WEBRTC_PORT` environment variables.
- Jupyter mode
    - See `examples/python/visualization/jupyter_visualization.ipynb`. You'll only need the Jupyter webserver. WebRTC-visualizer does not take any additional port or ip.

This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3392)


stotko (changes requested):
> Reviewed 41 of 79 files at r1, 5 of 8 files at r2.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3392)* status: 46 of 80 files reviewed, 6 unresolved discussions (waiting on @errissa, @prewettg, @ssheorey, and @yxlao)

___
*[cpp/open3d/visualization/gui/Application.cpp, line 418 at r2](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_0oyT4BLzu5swAqrgx:-M_0oyT4BLzu5swAqrgy:bbyhist) ([raw file](https://github.com/intel-isl/open3d/blob/862f0fd9211eb3f26cad7a8ab792c5428285b919/cpp/open3d/visualization/gui/Application.cpp#L418)):*
> ```cpp
>                 std::dynamic_pointer_cast<webrtc_server::WebRTCWindowSystem>(
>                         impl_->window_system_)) {
>         // Start the WebRTC server at the first run. No-op of the server has
> ```

Typo: No-op if ...
___
*[cpp/open3d/visualization/gui/Events.cpp, line 105 at r2](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_0pyc6ETzWqHxK7rTw:-M_0pyc6ETzWqHxK7rTx:b-yz1lr8) ([raw file](https://github.com/intel-isl/open3d/blob/862f0fd9211eb3f26cad7a8ab792c5428285b919/cpp/open3d/visualization/gui/Events.cpp#L105)):*
> ```cpp
> 
>     std::string type_name = value.get("type", "").asString();
>     if (type_name == "MOVE") {
> ```

Does this work without the prefix?  For instance`ToString()` sets `Type::MOVE` but equality to `MOVE` is checked here.
___
*[cpp/open3d/visualization/gui/Events.cpp, line 130 at r2](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_0qrO75DDxK7wQ26BZ:-M_0qrO75DDxK7wQ26B_:brhumzt) ([raw file](https://github.com/intel-isl/open3d/blob/862f0fd9211eb3f26cad7a8ab792c5428285b919/cpp/open3d/visualization/gui/Events.cpp#L130)):*
> ```cpp
>                this->type == Type::BUTTON_UP) {
>         std::string button_name = value["button"].get("button", "").asString();
>         if (button_name == "NONE") {
> ```

Same here.
___
*[cpp/open3d/visualization/rendering/filament/FilamentRenderer.h, line 46 at r2](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_0ra1z3pp9xTSteM5l:-M_0ra1z3pp9xTSteM5m:b-69sqz8) ([raw file](https://github.com/intel-isl/open3d/blob/862f0fd9211eb3f26cad7a8ab792c5428285b919/cpp/open3d/visualization/rendering/filament/FilamentRenderer.h#L46)):*
> ```objc
> 
> namespace open3d {
> ```

Unnecessary whitespace
___
*[cpp/open3d/visualization/webrtc_server/CMakeLists.txt, line 18 at r2](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_0srT23r8Zdj-4fD1Y:-M_0srT23r8Zdj-4fD1Z:b-czmcg1) ([raw file](https://github.com/intel-isl/open3d/blob/862f0fd9211eb3f26cad7a8ab792c5428285b919/cpp/open3d/visualization/webrtc_server/CMakeLists.txt#L18)):*
> ```CMake
>     ${WEBRTC_STREAMER_SRC}
> )
> target_include_directories(webrtc_server SYSTEM PRIVATE
> ```

I do not see any places in the code where this is needed. Is this a workaround to temporally suppress warnings for this module?
___
*[cpp/open3d/visualization/webrtc_server/HttpServerRequestHandler.cpp, line 48 at r2](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_0v7na-sPIsXyG1KOp:-M_0v7na-sPIsXyG1KOq:bw8opgf) ([raw file](https://github.com/intel-isl/open3d/blob/862f0fd9211eb3f26cad7a8ab792c5428285b919/cpp/open3d/visualization/webrtc_server/HttpServerRequestHandler.cpp#L48)):*
> ```cpp
> 
> static int LogMessage(const struct mg_connection *conn, const char *message) {
>     fprintf(stderr, "%s\n", message);
> ```

Do we want to log using our own logging framework?

ssheorey (commented):
> Reviewed 24 of 79 files at r1, 5 of 8 files at r2, 3 of 4 files at r3.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3392)* status: 56 of 82 files reviewed, 15 unresolved discussions (waiting on @errissa, @prewettg, and @yxlao)

___
*[CMakeLists.txt, line 182 at r2](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_2O-N_0woncBAojSMr:-M_2O-N_0woncBAojSMs:bspl637) ([raw file](https://github.com/intel-isl/open3d/blob/862f0fd9211eb3f26cad7a8ab792c5428285b919/CMakeLists.txt#L182)):*
> ```CMake
> endif()
> if(BUILD_WEBRTC AND (APPLE OR WIN32 OR LINUX_AARCH64))
>     message(WARNING "WebRTC visualizer is only supported on Linux (x86), setting BUILD_WEBRTC=OFF")
> ```

Sometimes a cmake warning gets missed by a user. Let's make this an error and ensure the user realizes this is not supported.
___
*[CMakeLists.txt, line 191 at r2](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_2OzEe0xEluaHxQ5U6:-M_2OzEe0xEluaHxQ5U7:b-euv37s) ([raw file](https://github.com/intel-isl/open3d/blob/862f0fd9211eb3f26cad7a8ab792c5428285b919/CMakeLists.txt#L191)):*
> ```CMake
> if(BUILD_JUPYTER_EXTENSION AND NOT BUILD_WEBRTC)
>     # BUILD_JUPYTER_EXTENSION transitively depends on BUILD_GUI
>     message(WARNING "Jupyter extention requires BUILD_WEBRTC=ON, setting BUILD_JUPYTER_EXTENSION=OFF")
> ```

Same. change to error.
___
*[CMakeLists.txt, line 195 at r2](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_2Oa95-pt0BXlIDCf_:-M_2Oa95-pt0BXlIDCfa:buli1nx) ([raw file](https://github.com/intel-isl/open3d/blob/862f0fd9211eb3f26cad7a8ab792c5428285b919/CMakeLists.txt#L195)):*
Same. change to error.
> ```
> extention
> ```
typo: extension
also in the previous `message()`
___
*[cpp/open3d/utility/IJsonConvertible.h, line 44 at r2](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_2Q8WE3bUE-o3hWJDa:-M_2Q8WE3bUE-o3hWJDb:b57gzcx) ([raw file](https://github.com/intel-isl/open3d/blob/862f0fd9211eb3f26cad7a8ab792c5428285b919/cpp/open3d/utility/IJsonConvertible.h#L44)):*
> ```
> Json::Value StringToJson(const std::string &json_str);
> std::string JsonToString(const Json::Value json);
> ```
Add docs. to both.
___
*[cpp/open3d/visualization/gui/Window.cpp, line 275 at r2](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_2W49Q-BLtSrUt5NbL:-M_2W49Q-BLtSrUt5NbM:bthr401) ([raw file](https://github.com/intel-isl/open3d/blob/862f0fd9211eb3f26cad7a8ab792c5428285b919/cpp/open3d/visualization/gui/Window.cpp#L275)):*
> ```cpp
> private:
>     std::string GenerateUID() {
>         static size_t count = 0;
> ```

Does this need to be atomic for thread safety?
___
*[cpp/pybind/visualization/webrtc_server/webrtc_server.cpp, line 40 at r3](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_2ZBlKAJNXIq1c9pcJ:-M_2ZBlKAJNXIq1c9pcK:b-vno21y) ([raw file](https://github.com/intel-isl/open3d/blob/3903d6a3d75f57940d97fab54045edf9f3f3b594/cpp/pybind/visualization/webrtc_server/webrtc_server.cpp#L40)):*
> ```
> "Global WebRTCServer instance."
> ```
"Global Open3D WebRTCServer instance."
___
*[cpp/pybind/visualization/webrtc_server/webrtc_server.cpp, line 50 at r3](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_2Zd7t-Leu056LDadg:-M_2Zd7t-Leu056LDadh:b-1lqtmz) ([raw file](https://github.com/intel-isl/open3d/blob/3903d6a3d75f57940d97fab54045edf9f3f3b594/cpp/pybind/visualization/webrtc_server/webrtc_server.cpp#L50)):*
> ```
>     webrtc_server.def("call_http_api", &WebRTCServer::CallHttpAPI,
>                       "entry_point"_a, "query_string"_a = "", "data"_a = "");
> ```
Add documentation.
___
*[cpp/pybind/visualization/webrtc_server/webrtc_server.cpp, line 54 at r3](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_2ZfAGE9KqRg2AO5zg:-M_2ZfAGE9KqRg2AO5zh:b-a9n2j2) ([raw file](https://github.com/intel-isl/open3d/blob/3903d6a3d75f57940d97fab54045edf9f3f3b594/cpp/pybind/visualization/webrtc_server/webrtc_server.cpp#L54)):*
> ```
>     webrtc_server.def("disable_http_handshake",
>                       &WebRTCServer::DisableHttpHandshake);
> ```
Add documentation.
___
*[util/ci_utils.sh, line 497 at r3](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_2Xh9x3KZMNG5JTV1X:-M_2Xh9x3KZMNG5JTV1Y:b-bby7y4) ([raw file](https://github.com/intel-isl/open3d/blob/3903d6a3d75f57940d97fab54045edf9f3f3b594/util/ci_utils.sh#L497)):*
> ```Shell
>     cmakeOptions=("-DDEVELOPER_BUILD=$DEVELOPER_BUILD"
>         "-DCMAKE_BUILD_TYPE=Release"
>         "-DBUILD_JUPYTER_EXTENSION=ON" # Only turned on in the second build when BUILD_GUI=ON
> ```

Better to make this explicit OFF and ON in the two cases.

ssheorey (commented):
We are removing the current Jupyter integration, right? After this, would Windows and macOS users have a way to use Jupyter locally?

> Reviewed 25 of 79 files at r1, 3 of 8 files at r2, 1 of 4 files at r3.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3392#-:-M_3tDU11g8BLrxCwpt4:b-fjgxmo)* status: 65 of 83 files reviewed, 26 unresolved discussions (waiting on @errissa, @prewettg, @ssheorey, @stotko, and @yxlao)

___
*[cpp/open3d/visualization/webrtc_server/BitmapTrackSource.h, line 40 at r3](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_3c6i59rOvAnGlLm0t:-M_3c6i59rOvAnGlLm0u:bsfm5n) ([raw file](https://github.com/intel-isl/open3d/blob/3903d6a3d75f57940d97fab54045edf9f3f3b594/cpp/open3d/visualization/webrtc_server/BitmapTrackSource.h#L40)):*
> ```objc
> // This is a private header. It shall be hidden from Open3D's public API. Do not
> // put this in Open3D.h.in.
> ```

Can you document the classes / functions in this private header (similar to our public docs), for maintenance?
___
*[cpp/open3d/visualization/webrtc_server/BitmapTrackSource.h, line 100 at r3](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_3b5T70YUQXArvxagS:-M_3b5T70YUQXArvxagT:b-aters3) ([raw file](https://github.com/intel-isl/open3d/blob/3903d6a3d75f57940d97fab54045edf9f3f3b594/cpp/open3d/visualization/webrtc_server/BitmapTrackSource.h#L100)):*
> ```objc
>     // BitmapWindowSystem when there's a new frame.
>     virtual void OnFrame(const std::shared_ptr<core::Tensor>& frame) override {
>         utility::LogInfo("BitmapTrackSource::OnFrame called");
> ```

If it's an error (logic / runtime) to call this, convert to `LogError`.
___
*[examples/cpp/DrawWebRTC.cpp, line 73 at r3](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_3YUlhDVyGxq5j0czG:-M_3YUlhDVyGxq5j0czH:b-voqlqm) ([raw file](https://github.com/intel-isl/open3d/blob/3903d6a3d75f57940d97fab54045edf9f3f3b594/examples/cpp/DrawWebRTC.cpp#L73)):*
> ```cpp
>     draw->ResetCameraToDefault();
>     visualization::gui::Application::GetInstance().AddWindow(draw);
>     draw.reset();  // so we don't hold onto the pointer after Run() cleans up
> ```

Wouldn't `draw`'s destructor take care of this?
___
*[examples/cpp/DrawWebRTC.cpp, line 76 at r3](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_3_FhbB3Ka-ZxT6cAP:-M_3_FhbB3Ka-ZxT6cAQ:b-nspfw5) ([raw file](https://github.com/intel-isl/open3d/blob/3903d6a3d75f57940d97fab54045edf9f3f3b594/examples/cpp/DrawWebRTC.cpp#L76)):*
> ```cpp
> }
> 
> void EmptyBox() {
> ```

Can you add some explanatory comments here? It looks like we add one window with a box and that triggers loading a point cloud and displaying it in another window.
___
*[python/requirements.txt, line 3 at r3](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_3GsEsEuWmg2iuTK3k:-M_3GsEsEuWmg2iuTK3l:b-n44j6a) ([raw file](https://github.com/intel-isl/open3d/blob/3903d6a3d75f57940d97fab54045edf9f3f3b594/python/requirements.txt#L3)):*
> ```vimhelp
> ipywidgets>=7.6.0
> jupyter_packaging~=0.9
> jupyterlab>=3.0.0,==3.*
> ```

Is this a strict requirement for us, or can we make it optional?
Also, since we only support web visualizer on Linux, can we skip these on Windows and macOS?
___
*[python/requirements.txt, line 6 at r3](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_33dKV4PQRVXPJAAa4:-M_33dKV4PQRVXPJAAa5:b418mkn) ([raw file](https://github.com/intel-isl/open3d/blob/3903d6a3d75f57940d97fab54045edf9f3f3b594/python/requirements.txt#L6)):*
> ```vimhelp
> setuptools>=40.8.0
> wheel>=0.36.0
> numpy>=0.18.0
> ```

Do you mean 1.18? 0.18 is very old.
___
*[python/setup.py, line 27 at r3](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_38Iug8dQgjfpcZ-yA:-M_38Iug8dQgjfpcZ-yB:b-d8ljwn) ([raw file](https://github.com/intel-isl/open3d/blob/3903d6a3d75f57940d97fab54045edf9f3f3b594/python/setup.py#L27)):*
> ```Python
> # ----------------------------------------------------------------------------
> 
> from __future__ import print_function
> ```

Do we need Python 2 compatibility?
___
*[python/open3d/__init__.py, line 119 at r3](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_3K-_-BB6et4HLy80I:-M_3K-_-BB6et4HLy80J:b-75l343) ([raw file](https://github.com/intel-isl/open3d/blob/3903d6a3d75f57940d97fab54045edf9f3f3b594/python/open3d/__init__.py#L119)):*
> ```Python
> 
>     Returns
>     =======
> ```

Switch to Google style docstring.
___
*[python/open3d/__init__.py, line 138 at r3](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_3K7YxEDP7REm-khqY:-M_3K7YxEDP7REm-khqZ:b-75l343) ([raw file](https://github.com/intel-isl/open3d/blob/3903d6a3d75f57940d97fab54045edf9f3f3b594/python/open3d/__init__.py#L138)):*
> ```Python
> 
>     Returns
>     =======
> ```

Switch to Google style docstring.
___
*[python/open3d/web_visualizer.py, line 12 at r3](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_3KxKU64B5KapC5nAk:-M_3KxKU64B5KapC5nAl:b-2mm609) ([raw file](https://github.com/intel-isl/open3d/blob/3903d6a3d75f57940d97fab54045edf9f3f3b594/python/open3d/web_visualizer.py#L12)):*
> ```Python
> @ipywidgets.register
> class WebVisualizer(ipywidgets.DOMWidget):
>     """Open3D Web Visualizer based on WebRTC."""
> ```

There is some commented print statements and a lot of validation with technical error messages. We should remove the comments and make the validation errors more user friendly or reduce them if not required.
___
*[python/open3d/web_visualizer.py, line 113 at r3](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_3LgAR7pImxTD--Jy6:-M_3LgAR7pImxTD--Jy7:b-hfzdzd) ([raw file](https://github.com/intel-isl/open3d/blob/3903d6a3d75f57940d97fab54045edf9f3f3b594/python/open3d/web_visualizer.py#L113)):*
> ```Python
>         # causes deadlock when AsyncEventLoop is used. By calling
>         # reset_print_function(), all C++ prints will be directed to the
>         # terminal while python print will still remain in the cell.
> ```

I think this is done, right?

stotko (commented):
> Reviewed 8 of 12 files at r5.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3392)* status: 73 of 83 files reviewed, 21 unresolved discussions (waiting on @errissa, @prewettg, @ssheorey, @stotko, and @yxlao)

prewettg (changes requested):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3392)* status: 72 of 83 files reviewed, 25 unresolved discussions (waiting on @errissa, @ssheorey, @stotko, and @yxlao)

___
*[cpp/open3d/visualization/gui/Application.cpp, line 420 at r6](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_6fENV4ptPZvxsxvls:-M_6fENV4ptPZvxsxvlt:bwcfbgd) ([raw file](https://github.com/intel-isl/open3d/blob/d4ac6740e9ee9a7928f461ef15f3ca2a316535ef/cpp/open3d/visualization/gui/Application.cpp#L420)):*
> ```cpp
>         // Start the WebRTC server at the first run. No-op if the server has
>         // already been started.
>         webrtc_window_system->StartWebRTCServer();
> ```

This would be better in WebRTCWindowSystem::CreateOSWindow()
___
*[cpp/open3d/visualization/gui/Application.cpp, line 430 at r6](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_6e83uDb-VJKywnya1:-M_6e83uDb-VJKywnya2:bytnxww) ([raw file](https://github.com/intel-isl/open3d/blob/d4ac6740e9ee9a7928f461ef15f3ca2a316535ef/cpp/open3d/visualization/gui/Application.cpp#L430)):*
> ```cpp
>             return true;  // Should close Window.
>         };
>         window->SetOnClose(close_window_callback);
> ```

If the owner of the window has set a close callback this will overwrite it (and vice-versa if the owner sets one later). It would be better for WebRTCWindowSystem::DestroyWindow() to call CloseWindowConnections
___
*[cpp/open3d/visualization/gui/Window.cpp, line 276 at r6](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_6hfM86F_qW6D8clsu:-M_6hfM86F_qW6D8clsv:b-wm2feu) ([raw file](https://github.com/intel-isl/open3d/blob/d4ac6740e9ee9a7928f461ef15f3ca2a316535ef/cpp/open3d/visualization/gui/Window.cpp#L276)):*
> ```cpp
>     std::string GenerateUID() {
>         static std::atomic<size_t> count{0};
>         return "window_" + std::to_string(count++);
> ```

If the ID is essentially a unique int, why not make uid_ an int? (The rest of the GUI uses string identifiers because ImGUI uses the widget text/name to reference its hidden retained-mode state for the widget.) I'm not sure I understand why this is necessary, though. I'd expect that the WebRTC server would deliver events to the WebRTCWindowSystem, which would presumably know the connection/ID to get the appropriate Window, so any ID that is necessary could stay within WebRTCWindowSystem. (Not that there is anything wrong with the Window having an ID, it just seems unnecessary.)
___
*[cpp/open3d/visualization/webrtc_server/WebRTCWindowSystem.cpp, line 74 at r6](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_7Bvb_36kMTW3lzI7w:-M_7Bvb_36kMTW3lzI7x:bo48xtg) ([raw file](https://github.com/intel-isl/open3d/blob/d4ac6740e9ee9a7928f461ef15f3ca2a316535ef/cpp/open3d/visualization/webrtc_server/WebRTCWindowSystem.cpp#L74)):*
> ```cpp
>     // Client -> server message can trigger a mouse event and
>     // mouse_event_callback will be called.
>     auto mouse_event_callback = [this](const std::string &window_uid,
> ```

I don't understand the need for the callback functions. The Post*Event(), etc. calls were intended to do this, so the server would just call Post*Event(...) and continue. WebRTCWindowSystem would override any of them that it needs. Although it looks like BitmapWindowSystem needs to have the methods marked virtual. I think I had a failure of imagination not expecting a need to inherit from BitmapWindowSystem, or I just didn't notice, since PostRedrawEvent is virtual. Hmm, so having looked into this over the course of writing this comment, I suspect that the right solution is to make Post*Event virtual in BitmapWindowSystem, and to override the methods with the contents of the callbacks here.

yxlao (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3392)* status: 51 of 84 files reviewed, 25 unresolved discussions (waiting on @errissa, @prewettg, @ssheorey, @stotko, and @yxlao)

___
*[CMakeLists.txt, line 182 at r2](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_2O-N_0woncBAojSMr:-M_3TpXr1o3aCyufO_ls:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/862f0fd9211eb3f26cad7a8ab792c5428285b919/CMakeLists.txt#L182)):*

Sometimes a cmake warning gets missed by a user. Let's make this an error and ensure the user realizes this is not supported.
</blockquote></details>

Done.
___
*[CMakeLists.txt, line 191 at r2](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_2OzEe0xEluaHxQ5U6:-M_3TpuQ1DhgSm8Nr6zY:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/862f0fd9211eb3f26cad7a8ab792c5428285b919/CMakeLists.txt#L191)):*

Same. change to error.
</blockquote></details>

Done.
___
*[CMakeLists.txt, line 195 at r2](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_2Oa95-pt0BXlIDCf_:-M_3TqCzCfNC8ydHKOXW:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/862f0fd9211eb3f26cad7a8ab792c5428285b919/CMakeLists.txt#L195)):*

Same. change to error.
> ```
> extention
> ```
typo: extension
also in the previous `message()`
</blockquote></details>

Done.
___
*[cpp/open3d/utility/IJsonConvertible.h, line 44 at r2](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_2Q8WE3bUE-o3hWJDa:-M_3V3QPCJgsJO-ia2Y4:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/862f0fd9211eb3f26cad7a8ab792c5428285b919/cpp/open3d/utility/IJsonConvertible.h#L44)):*

> ```
> Json::Value StringToJson(const std::string &json_str);
> std::string JsonToString(const Json::Value json);
> ```
Add docs. to both.
</blockquote></details>

Done.
___
*[cpp/open3d/visualization/gui/Application.cpp, line 418 at r2](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_0oyT4BLzu5swAqrgx:-M_3V7yE6PLyzudUvNZ9:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/862f0fd9211eb3f26cad7a8ab792c5428285b919/cpp/open3d/visualization/gui/Application.cpp#L418)):*

Typo: No-op if ...
</blockquote></details>

Done.
___
*[cpp/open3d/visualization/gui/Application.cpp, line 420 at r6](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_6fENV4ptPZvxsxvls:-M_8IPro-LmxPCwALmgv:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/d4ac6740e9ee9a7928f461ef15f3ca2a316535ef/cpp/open3d/visualization/gui/Application.cpp#L420)):*

This would be better in WebRTCWindowSystem::CreateOSWindow()
</blockquote></details>

Done.
___
*[cpp/open3d/visualization/gui/Application.cpp, line 430 at r6](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_6e83uDb-VJKywnya1:-M_Lhp2zDYfVXF4oLFDk:bxyanmy) ([raw file](https://github.com/intel-isl/open3d/blob/d4ac6740e9ee9a7928f461ef15f3ca2a316535ef/cpp/open3d/visualization/gui/Application.cpp#L430)):*

If the owner of the window has set a close callback this will overwrite it (and vice-versa if the owner sets one later). It would be better for WebRTCWindowSystem::DestroyWindow() to call CloseWindowConnections
</blockquote></details>

Done. Moved to `WebRTCWindowSystem::DestroyWindow()`. 
Now, `Application.cpp` does not know anything about WebRTC.
___
*[cpp/open3d/visualization/gui/Events.cpp, line 105 at r2](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_0pyc6ETzWqHxK7rTw:-M_3VgRW7an-UbH8VaIm:bgc3hu8) ([raw file](https://github.com/intel-isl/open3d/blob/862f0fd9211eb3f26cad7a8ab792c5428285b919/cpp/open3d/visualization/gui/Events.cpp#L105)):*

Does this work without the prefix?  For instance`ToString()` sets `Type::MOVE` but equality to `MOVE` is checked here.
</blockquote></details>

Good catch. 

- `MouseEvent::ToString()`: is only used for printing out human-readable value.
- `MouseEvent::FromJson(const Json::Value& value)`: converts the Json value sent from the JavaScript client. The JavaScript client's Json format is consistent with this.

These two functions are independent so that's why they work. But they are inconsistent. 

Updated `MouseEvent::ToString()`.
___
*[cpp/open3d/visualization/gui/Events.cpp, line 130 at r2](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_0qrO75DDxK7wQ26BZ:-M_3Ww2l5TsdqWboYXxb:b-70iye2) ([raw file](https://github.com/intel-isl/open3d/blob/862f0fd9211eb3f26cad7a8ab792c5428285b919/cpp/open3d/visualization/gui/Events.cpp#L130)):*

Same here.
</blockquote></details>

Updated `MouseEvent::ToString()`.
___
*[cpp/open3d/visualization/gui/Window.cpp, line 275 at r2](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_2W49Q-BLtSrUt5NbL:-M_3YS-i6_YjNB9GSy6E:boj1qgt) ([raw file](https://github.com/intel-isl/open3d/blob/862f0fd9211eb3f26cad7a8ab792c5428285b919/cpp/open3d/visualization/gui/Window.cpp#L275)):*

Does this need to be atomic for thread safety?
</blockquote></details>

Good catch. Done.
___
*[cpp/open3d/visualization/gui/Window.cpp, line 276 at r6](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_6hfM86F_qW6D8clsu:-M_Li5020uRBxPnredhy:b47husb) ([raw file](https://github.com/intel-isl/open3d/blob/d4ac6740e9ee9a7928f461ef15f3ca2a316535ef/cpp/open3d/visualization/gui/Window.cpp#L276)):*
> I'd expect that the WebRTC server would deliver events to the WebRTCWindowSystem, which would presumably know the connection/ID to get the appropriate Window, so any ID that is necessary could stay within WebRTCWindowSystem. 

Updated. Now, `Window` no longer owns its UID. 

However, in the WebRTC case we still have a problem: if we create a new window (e.g. in python `w = O3DVisualizer(title, width, height)`), how do we know the UID of the Window we just created? This UID is required by the JavaScript side to initiate a connection. My current solution is to add `Window::GetWebRTCUID()`, which only returns a meaningful value when `WebRTCWindowSystem` is in use. This is cleaner than before since the Window itself does not own the uid. Let me know if you have other suggestions.

 > If the ID is essentially a unique int, why not make uid_ an int?

That's doable, and it will be cleaner. String type was used here due to a historical reason that I also use the same string as the id for HTML elements.  I will do this refactoring in a separate PR. 
___
*[cpp/open3d/visualization/rendering/filament/FilamentRenderer.h, line 46 at r2](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_0ra1z3pp9xTSteM5l:-M_3YVuW7CgoRfhw56JO:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/862f0fd9211eb3f26cad7a8ab792c5428285b919/cpp/open3d/visualization/rendering/filament/FilamentRenderer.h#L46)):*

Unnecessary whitespace
</blockquote></details>

Done.
___
*[cpp/open3d/visualization/webrtc_server/BitmapTrackSource.h, line 40 at r3](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_3c6i59rOvAnGlLm0t:-M_M2wb-AX59KuB9OMkk:b-og9rsm) ([raw file](https://github.com/intel-isl/open3d/blob/3903d6a3d75f57940d97fab54045edf9f3f3b594/cpp/open3d/visualization/webrtc_server/BitmapTrackSource.h#L40)):*

Can you document the classes / functions in this private header (similar to our public docs), for maintenance?
</blockquote></details>

I added some documents in the code above to describe the related classes and class hierarchy. 

```cpp
/// [Related classes]
/// - VideoTrackSourceInterface: WebRTC expects a custom implementation
///   of this class.
/// - BitmapTrackSourceInterface: Used as the primary interface in
///   PeerConnectionManager. It is almost the same as VideoTrackSourceInterface
///   with an additional OnFrame function for triggering frame handlers.
/// - BitmapTrackSource: Abstract class for bit map tracks.
/// - ImageTrackSource: Captures frames from Open3D visualizer.
/// - VideoFilter: Video frame processing, e.g. scaling.
///
/// [Class hierarchy]
/// BitmapTrackSourceInterface --inherits--> webrtc::VideoTrackSourceInterface
/// BitmapTrackSource --inherits--> webrtc::Notifier<BitmapTrackSourceInterface>
/// ImageTrackSource  --inherits--> BitmapTrackSource
/// ImageCapturer     --owned by--> ImageTrackSource
/// VideoFilter       --inherits--> BitmapTrackSource
```

`BitmapTrackSource` mainly contains some boilerplate code and most of its functions are inherited. I am not familiar enough to document all functions in this class, but users can some of the documentation in the parent class inside WebRTC codebase.
___
*[cpp/open3d/visualization/webrtc_server/BitmapTrackSource.h, line 100 at r3](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_3b5T70YUQXArvxagS:-M_M2tvk9zCkYgip9lRj:b-abv3h2) ([raw file](https://github.com/intel-isl/open3d/blob/3903d6a3d75f57940d97fab54045edf9f3f3b594/cpp/open3d/visualization/webrtc_server/BitmapTrackSource.h#L100)):*

If it's an error (logic / runtime) to call this, convert to `LogError`.
</blockquote></details>

Done. Changed to LogError.
___
*[cpp/open3d/visualization/webrtc_server/CMakeLists.txt, line 18 at r2](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_0srT23r8Zdj-4fD1Y:-M_3YqSj2ejkCMWKdALM:bss4a2b) ([raw file](https://github.com/intel-isl/open3d/blob/862f0fd9211eb3f26cad7a8ab792c5428285b919/cpp/open3d/visualization/webrtc_server/CMakeLists.txt#L18)):*

I do not see any places in the code where this is needed. Is this a workaround to temporally suppress warnings for this module?
</blockquote></details>

Removed.
___
*[cpp/open3d/visualization/webrtc_server/WebRTCWindowSystem.cpp, line 74 at r6](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_7Bvb_36kMTW3lzI7w:-M_LlU4v4A6wIALwihKW:b-cra1el) ([raw file](https://github.com/intel-isl/open3d/blob/d4ac6740e9ee9a7928f461ef15f3ca2a316535ef/cpp/open3d/visualization/webrtc_server/WebRTCWindowSystem.cpp#L74)):*

I don't understand the need for the callback functions. The Post*Event(), etc. calls were intended to do this, so the server would just call Post*Event(...) and continue. WebRTCWindowSystem would override any of them that it needs. Although it looks like BitmapWindowSystem needs to have the methods marked virtual. I think I had a failure of imagination not expecting a need to inherit from BitmapWindowSystem, or I just didn't notice, since PostRedrawEvent is virtual. Hmm, so having looked into this over the course of writing this comment, I suspect that the right solution is to make Post*Event virtual in BitmapWindowSystem, and to override the methods with the contents of the callbacks here.
</blockquote></details>

You're right. This is essentially an artifact that `WebRTCServer` and `WebRTCWindowSystem` shall be merged. Now they are merged. We don't need to register any new callbacks and the PostXXXEvent works just fine.
___
*[examples/cpp/DrawWebRTC.cpp, line 73 at r3](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_3YUlhDVyGxq5j0czG:-M_M4CrS98RDtUeI83Cw:b-1nji56) ([raw file](https://github.com/intel-isl/open3d/blob/3903d6a3d75f57940d97fab54045edf9f3f3b594/examples/cpp/DrawWebRTC.cpp#L73)):*

Wouldn't `draw`'s destructor take care of this?
</blockquote></details>

We're doing an equivalent implementation of https://github.com/intel-isl/Open3D/blob/2ba60d5ef840ce8c1cc2c046b0ed2613e138e088/cpp/open3d/visualization/utility/Draw.cpp#L109

We need this for a graceful exit -- otherwise, there will be a segfault after all windows are closed. 
___
*[examples/cpp/DrawWebRTC.cpp, line 76 at r3](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_3_FhbB3Ka-ZxT6cAP:-M_M5jmu0U34cpZZ69gE:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/3903d6a3d75f57940d97fab54045edf9f3f3b594/examples/cpp/DrawWebRTC.cpp#L76)):*

Can you add some explanatory comments here? It looks like we add one window with a box and that triggers loading a point cloud and displaying it in another window.
</blockquote></details>

Done.
___
*[python/requirements.txt, line 6 at r3](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_33dKV4PQRVXPJAAa4:-M_M7m4606yaNO9A7Z9x:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/3903d6a3d75f57940d97fab54045edf9f3f3b594/python/requirements.txt#L6)):*

Do you mean 1.18? 0.18 is very old.
</blockquote></details>

Done.
___
*[python/setup.py, line 27 at r3](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_38Iug8dQgjfpcZ-yA:-M_M7p11AsigeX-olprM:b-hfpv0l) ([raw file](https://github.com/intel-isl/open3d/blob/3903d6a3d75f57940d97fab54045edf9f3f3b594/python/setup.py#L27)):*

Do we need Python 2 compatibility?
</blockquote></details>

removed
___
*[python/open3d/__init__.py, line 119 at r3](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_3K-_-BB6et4HLy80I:-M_MN2lWCZadHe5sw46a:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/3903d6a3d75f57940d97fab54045edf9f3f3b594/python/open3d/__init__.py#L119)):*

Switch to Google style docstring.
</blockquote></details>

Done.
___
*[python/open3d/__init__.py, line 138 at r3](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_3K7YxEDP7REm-khqY:-M_MN3Bp-TmMXmIcsGqg:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/3903d6a3d75f57940d97fab54045edf9f3f3b594/python/open3d/__init__.py#L138)):*

Switch to Google style docstring.
</blockquote></details>

Done.
___
*[python/open3d/web_visualizer.py, line 12 at r3](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_3KxKU64B5KapC5nAk:-M_MNYZ92gdh1HmWQUn_:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/3903d6a3d75f57940d97fab54045edf9f3f3b594/python/open3d/web_visualizer.py#L12)):*

There is some commented print statements and a lot of validation with technical error messages. We should remove the comments and make the validation errors more user friendly or reduce them if not required.
</blockquote></details>

Done.
___
*[python/open3d/web_visualizer.py, line 113 at r3](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_3LgAR7pImxTD--Jy6:-M_MNZqA7dJ5I5dTiO2L:bip92m4) ([raw file](https://github.com/intel-isl/open3d/blob/3903d6a3d75f57940d97fab54045edf9f3f3b594/python/open3d/web_visualizer.py#L113)):*

I think this is done, right?
</blockquote></details>

`o3d.utility.reset_print_function()` is the workaround. 

The default logger aquires GIL, which causes issue with the `AsyncEventLoop` currently.
```cpp
    utility::Logger::GetInstance().SetPrintFunction([](const std::string& msg) {
        py::gil_scoped_acquire acquire;
        py::print(msg);
    });
```
___
*[util/ci_utils.sh, line 497 at r3](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_2Xh9x3KZMNG5JTV1X:-M_3ZKXpEZYzfUKe-heF:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/3903d6a3d75f57940d97fab54045edf9f3f3b594/util/ci_utils.sh#L497)):*

Better to make this explicit OFF and ON in the two cases.
</blockquote></details>

Done.
___
*[cpp/pybind/visualization/webrtc_server/webrtc_server.cpp, line 40 at r3](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_2ZBlKAJNXIq1c9pcJ:-M_3ZHUk8Fpf3GRUnijV:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/3903d6a3d75f57940d97fab54045edf9f3f3b594/cpp/pybind/visualization/webrtc_server/webrtc_server.cpp#L40)):*

> ```
> "Global WebRTCServer instance."
> ```
"Global Open3D WebRTCServer instance."
</blockquote></details>

Done.
___
*[cpp/pybind/visualization/webrtc_server/webrtc_server.cpp, line 50 at r3](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_2Zd7t-Leu056LDadg:-M_3_qbe3Y2z8eYZrARl:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/3903d6a3d75f57940d97fab54045edf9f3f3b594/cpp/pybind/visualization/webrtc_server/webrtc_server.cpp#L50)):*

> ```
>     webrtc_server.def("call_http_api", &WebRTCServer::CallHttpAPI,
>                       "entry_point"_a, "query_string"_a = "", "data"_a = "");
> ```
Add documentation.
</blockquote></details>

Done.
___
*[cpp/pybind/visualization/webrtc_server/webrtc_server.cpp, line 54 at r3](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_2ZfAGE9KqRg2AO5zg:-M_3_qou7kWVK2fGutvd:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/3903d6a3d75f57940d97fab54045edf9f3f3b594/cpp/pybind/visualization/webrtc_server/webrtc_server.cpp#L54)):*

> ```
>     webrtc_server.def("disable_http_handshake",
>                       &WebRTCServer::DisableHttpHandshake);
> ```
Add documentation.
</blockquote></details>

Done.

prewettg (changes requested):
> Reviewed 7 of 24 files at r7.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3392)* status: 58 of 84 files reviewed, 23 unresolved discussions (waiting on @errissa, @ssheorey, @stotko, and @yxlao)

___
*[cpp/open3d/visualization/gui/Window.cpp, line 276 at r6](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_6hfM86F_qW6D8clsu:-M_MnnGhDdZmMFvot4Pz:b-g1zc0w) ([raw file](https://github.com/intel-isl/open3d/blob/d4ac6740e9ee9a7928f461ef15f3ca2a316535ef/cpp/open3d/visualization/gui/Window.cpp#L276)):*

> I'd expect that the WebRTC server would deliver events to the WebRTCWindowSystem, which would presumably know the connection/ID to get the appropriate Window, so any ID that is necessary could stay within WebRTCWindowSystem. 

Updated. Now, `Window` no longer owns its UID. 

However, in the WebRTC case we still have a problem: if we create a new window (e.g. in python `w = O3DVisualizer(title, width, height)`), how do we know the UID of the Window we just created? This UID is required by the JavaScript side to initiate a connection. My current solution is to add `Window::GetWebRTCUID()`, which only returns a meaningful value when `WebRTCWindowSystem` is in use. This is cleaner than before since the Window itself does not own the uid. Let me know if you have other suggestions.

 > If the ID is essentially a unique int, why not make uid_ an int?

That's doable, and it will be cleaner. String type was used here due to a historical reason that I also use the same string as the id for HTML elements.  I will do this refactoring in a separate PR. 
</blockquote></details>

I would assume that the JS side would only ever know the WebRTC id, so it would never need to get it from Window. The server already knows the id, although if it doesn't, I think calling Application::GetWindowSystem(), dynamically casting to WebRTCWindowSystem (which we know it has to be), and then getting the id from WebRTCWindowSystem would be cleaner.
___
*[examples/cpp/DrawWebRTC.cpp, line 73 at r3](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_3YUlhDVyGxq5j0czG:-M_MpomzAxGD0yID5HWm:b-jxjg32) ([raw file](https://github.com/intel-isl/open3d/blob/3903d6a3d75f57940d97fab54045edf9f3f3b594/examples/cpp/DrawWebRTC.cpp#L73)):*

We're doing an equivalent implementation of https://github.com/intel-isl/Open3D/blob/2ba60d5ef840ce8c1cc2c046b0ed2613e138e088/cpp/open3d/visualization/utility/Draw.cpp#L109

We need this for a graceful exit -- otherwise, there will be a segfault after all windows are closed. 
</blockquote></details>

(just fyi) I'm beginning to think using shared_ptr in the GUI API was a mistake, especially because of Filament's cleanup requirements. In newer code I've been declaring
```
template <typename T>
std::shared_ptr<T> GiveOwnership(T *ptr) {
    return std::shared_ptr<T>(ptr);
}
```
creating everything with `new` and then doing, for instance, `app.AddWindow(GiveOwnership(newed_ptr));` It seems like it might be clearer than my old way of doing it with reset() and a comment, although maybe it just begs a related question. Just fyi in case it's a helpful idea.
___
*[python/open3d/visualization/draw.py, line 79 at r7](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_Ms9Og1vJF74owFjJ8:-M_Ms9Og1vJF74owFjJ9:bff13d3) ([raw file](https://github.com/intel-isl/open3d/blob/fdcd0f32c9bcdee6a79bc04af02fcd4c0ffbdb78/python/open3d/visualization/draw.py#L79)):*
> ```Python
>     gui.Application.instance.add_window(w)
>     if non_blocking_and_return_uid:
>         return w.uid
> ```

Is this the only place where the UID is used? It seems like it would be better to have AddWindow--or really, WebRTCWindowSystem::CreateOSWindow()--notify whoever needs to know that there is a new window that needs to be displayed. As it is, I think only the draw() function would work in Jupyter. If you created an O3DVisualizer object manually (e.g. the new text3d.py example, in `high_level()`) the window would not be displayed; neither would creating the Window using `app.create_window()`, which calls AddWindow() internally.

ssheorey (changes requested):
> Reviewed 8 of 79 files at r1, 1 of 2 files at r4, 8 of 12 files at r5, 1 of 1 files at r6, 23 of 24 files at r7.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3392)* status: 81 of 84 files reviewed, 25 unresolved discussions (waiting on @errissa, @stotko, and @yxlao)

___
*[cpp/open3d/visualization/visualizer/O3DVisualizer.cpp, line 1783 at r7](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_NAB6G6hLGVek3Qevf:-M_NAB6G6hLGVek3Qevg:bvl43z8) ([raw file](https://github.com/intel-isl/open3d/blob/fdcd0f32c9bcdee6a79bc04af02fcd4c0ffbdb78/cpp/open3d/visualization/visualizer/O3DVisualizer.cpp#L1783)):*
> ```cpp
>     // TODO: debug only, enabled the check later.
>     // if (Application::GetInstance().UsingNativeWindows()) {
>     if (true) {
> ```

Re-enable check after debugging.
___
*[cpp/open3d/visualization/webrtc_server/ImageCapturer.cpp, line 40 at r7](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_NnzanDoutVZygHCBk:-M_NnzanDoutVZygHCBl:b-7ksjx8) ([raw file](https://github.com/intel-isl/open3d/blob/fdcd0f32c9bcdee6a79bc04af02fcd4c0ffbdb78/cpp/open3d/visualization/webrtc_server/ImageCapturer.cpp#L40)):*
> ```cpp
> #include <chrono>
> #include <memory>
> #include <thread>
> ```

libstdc++ headers before other library headers.
___
*[cpp/open3d/visualization/webrtc_server/PeerConnectionManager.h, line 38 at r7](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_Nlg2X4LjPIZIUsBBC:-M_Nlg2X4LjPIZIUsBBD:b-ukjcq2) ([raw file](https://github.com/intel-isl/open3d/blob/fdcd0f32c9bcdee6a79bc04af02fcd4c0ffbdb78/cpp/open3d/visualization/webrtc_server/PeerConnectionManager.h#L38)):*
> ```objc
> // put this in Open3D.h.in.
> 
> #pragma once
> ```

Add documentation about how messages are constructed and exchanged between the client and server:
- Sequence diagram for connnection setup / messaging / teardown.
- What is the format of the messages? What are the message types?
- What are possible errors?
___
*[cpp/open3d/visualization/webrtc_server/PeerConnectionManager.cpp, line 377 at r7](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_NlLbsCFiAOerwM3sZ:-M_NlLbsCFiAOerwM3s_:bl8t4eb) ([raw file](https://github.com/intel-isl/open3d/blob/fdcd0f32c9bcdee6a79bc04af02fcd4c0ffbdb78/cpp/open3d/visualization/webrtc_server/PeerConnectionManager.cpp#L377)):*
> ```cpp
>                     utility::LogDebug("remote_description is ready.");
>                 } else {
>                     utility::LogWarning("remote_description is nullptr.");
> ```

This message needs clarification.
___
*[cpp/open3d/visualization/webrtc_server/VideoFilter.h, line 34 at r7](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_NYsbTB4H_tCm1EiJj:-M_NYsbU8vzdNBAAHBnY:b-ewuwzj) ([raw file](https://github.com/intel-isl/open3d/blob/fdcd0f32c9bcdee6a79bc04af02fcd4c0ffbdb78/cpp/open3d/visualization/webrtc_server/VideoFilter.h#L34)):*
> ```objc
> // any purpose.
> // ----------------------------------------------------------------------------
> ```

Private header?
//
// This is a private header. It shall be hidden from Open3D's public API. Do not
// put this in Open3D.h.in.
___
*[cpp/open3d/visualization/webrtc_server/VideoFilter.h, line 46 at r7](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_NYdrp-NSgDD5taRmD:-M_NYdrp-NSgDD5taRmE:b-5qsy9d) ([raw file](https://github.com/intel-isl/open3d/blob/fdcd0f32c9bcdee6a79bc04af02fcd4c0ffbdb78/cpp/open3d/visualization/webrtc_server/VideoFilter.h#L46)):*
> ```objc
> 
> template <class T>
> class VideoFilter : public BitmapTrackSource {
> ```

Add docs.
___
*[cpp/open3d/visualization/webrtc_server/WebRTCWindowSystem.h, line 39 at r7](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_NCfG_2X8cpx2wXsmJ:-M_NCfG_2X8cpx2wXsmK:b-tct4zi) ([raw file](https://github.com/intel-isl/open3d/blob/fdcd0f32c9bcdee6a79bc04af02fcd4c0ffbdb78/cpp/open3d/visualization/webrtc_server/WebRTCWindowSystem.h#L39)):*
> ```objc
> namespace webrtc_server {
> 
> class WebRTCWindowSystem : public gui::BitmapWindowSystem {
> ```

Is this the main WebRTC header file? Could you add `\file` documentation with a summary of how the WebRTC visualizer works?
___
*[cpp/open3d/visualization/webrtc_server/WebRTCWindowSystem.h, line 69 at r7](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_NJnh56rEUjDNDT9Ng:-M_NJnh56rEUjDNDT9Nh:bj36xin) ([raw file](https://github.com/intel-isl/open3d/blob/fdcd0f32c9bcdee6a79bc04af02fcd4c0ffbdb78/cpp/open3d/visualization/webrtc_server/WebRTCWindowSystem.h#L69)):*
> ```objc
>     void SendInitFrames(const std::string& window_uid);
> 
>     /// Call PeerConnectionManager's web request API.
> ```

\\\ \brief Call PeerConnectionManager's web request API.
\\\
\\\ This function...
___
*[cpp/open3d/visualization/webrtc_server/WebRTCWindowSystem.cpp, line 33 at r7](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_NMFlCCK_UwnuwAHnC:-M_NMFlCCK_UwnuwAHnD:b-sp6c4k) ([raw file](https://github.com/intel-isl/open3d/blob/fdcd0f32c9bcdee6a79bc04af02fcd4c0ffbdb78/cpp/open3d/visualization/webrtc_server/WebRTCWindowSystem.cpp#L33)):*
<details><summary><i alt="summary">Quoted 5 lines of code…</i></summary>

> ```
> #include <p2p/base/basic_packet_socket_factory.h>
> #include <p2p/base/stun_server.h>
> #include <p2p/base/turn_server.h>
> #include <rtc_base/ssl_adapter.h>
> #include <rtc_base/thread.h>
> ```
</details>

Include after standard lib headers.
___
*[cpp/open3d/visualization/webrtc_server/WebRTCWindowSystem.cpp, line 59 at r7](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_NSouyA9BBZVbf7ioV:-M_NSouyA9BBZVbf7ioW:b-i2dota) ([raw file](https://github.com/intel-isl/open3d/blob/fdcd0f32c9bcdee6a79bc04af02fcd4c0ffbdb78/cpp/open3d/visualization/webrtc_server/WebRTCWindowSystem.cpp#L59)):*
> ```
> WEBRTC_IP
> ```
Does this need a sanity check to ensure that the IP actually belongs to the machine? e.g. What would happen if:
machine IP is: `192.168.1.22` and this is set to `192.168.1.23` by mistake?
___
*[cpp/open3d/visualization/webrtc_server/WebRTCWindowSystem.cpp, line 63 at r7](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_NTWN02WKMB9_NhtPn:-M_NTWN02WKMB9_NhtPo:blsl36h) ([raw file](https://github.com/intel-isl/open3d/blob/fdcd0f32c9bcdee6a79bc04af02fcd4c0ffbdb78/cpp/open3d/visualization/webrtc_server/WebRTCWindowSystem.cpp#L63)):*
> ```
> WEBRTC_PORT
> ```
Sanity check for this input. Must be integer in the range [1024, 65535].
___
*[cpp/open3d/visualization/webrtc_server/WebRTCWindowSystem.cpp, line 70 at r7](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_NUVCNAakVfLP0X7YV:-M_NUVCNAakVfLP0X7YW:b-k0ebtl) ([raw file](https://github.com/intel-isl/open3d/blob/fdcd0f32c9bcdee6a79bc04af02fcd4c0ffbdb78/cpp/open3d/visualization/webrtc_server/WebRTCWindowSystem.cpp#L70)):*
> ```cpp
> static std::string GetEnvWebRTCWebRoot() {
>     if (const char *env_p = std::getenv("WEBRTC_WEB_ROOT")) {
>         return std::string(env_p);
> ```

Add input check. Does this need to just start with '/' or be an existing relative path?
___
*[cpp/open3d/visualization/webrtc_server/WebRTCWindowSystem.cpp, line 193 at r7](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_NNiHq8yR1IoUrRxZs:-M_NNiHq8yR1IoUrRxZt:b-xcvhty) ([raw file](https://github.com/intel-isl/open3d/blob/fdcd0f32c9bcdee6a79bc04af02fcd4c0ffbdb78/cpp/open3d/visualization/webrtc_server/WebRTCWindowSystem.cpp#L193)):*
> ```cpp
>             // Logging settings.
>             // src/rtc_base/logging.h: LS_VERBOSE, LS_ERROR
>             rtc::LogMessage::LogToDebug((rtc::LoggingSeverity)rtc::LS_ERROR);
> ```

We can sync this with the Open3D logging level to make it easier to debug webrtc issues in the future.
___
*[cpp/open3d/visualization/webrtc_server/WebRTCWindowSystem.cpp, line 201 at r7](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_NOBqz7Y8oRRw_FNEd:-M_NOBqz7Y8oRRw_FNEe:b-kgp3xh) ([raw file](https://github.com/intel-isl/open3d/blob/fdcd0f32c9bcdee6a79bc04af02fcd4c0ffbdb78/cpp/open3d/visualization/webrtc_server/WebRTCWindowSystem.cpp#L201)):*
> ```
> stun:stun.l.google.com:19302"
> ```
Let's make this a constant in the header. It will be easier to update if for example the default server goes down / is inaccessible for specific users or is renamed by Google.
___
*[cpp/open3d/visualization/webrtc_server/WebRTCWindowSystem.cpp, line 228 at r7](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_NR_z5-1ZaoCUyN1XC:-M_NR_z6DRjGEQ8SV9Mf:b-n4qe6g) ([raw file](https://github.com/intel-isl/open3d/blob/fdcd0f32c9bcdee6a79bc04af02fcd4c0ffbdb78/cpp/open3d/visualization/webrtc_server/WebRTCWindowSystem.cpp#L228)):*
<details><summary><i alt="summary">Quoted 17 lines of code…</i></summary>

> ```
>                 std::vector<std::string> options;
>                 options.push_back("document_root");
>                 options.push_back(impl_->web_root_);
>                 options.push_back("enable_directory_listing");
>                 options.push_back("no");
>                 options.push_back("additional_header");
>                 options.push_back("X-Frame-Options: SAMEORIGIN");
>                 options.push_back("access_control_allow_origin");
>                 options.push_back("*");
>                 options.push_back("listening_ports");
>                 options.push_back(impl_->http_address_);
>                 options.push_back("enable_keep_alive");
>                 options.push_back("yes");
>                 options.push_back("keep_alive_timeout_ms");
>                 options.push_back("1000");
>                 options.push_back("decode_url");
>                 options.push_back("no");
> ```
</details>

Use initializer list for options.
___
*[cpp/open3d/visualization/webrtc_server/WebRTCWindowSystem.cpp, line 274 at r7](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_NWTtxEzjGaHCu1_Y-:-M_NWTtxEzjGaHCu1_Y0:b-h2sfjc) ([raw file](https://github.com/intel-isl/open3d/blob/fdcd0f32c9bcdee6a79bc04af02fcd4c0ffbdb78/cpp/open3d/visualization/webrtc_server/WebRTCWindowSystem.cpp#L274)):*
<details><summary><i alt="summary">Quoted 4 lines of code…</i></summary>

> ```
>             value.get("window_uid", "").asString() != "" &&
...
>             const std::string window_uid =
>                     value.get("window_uid", "").asString();
> ```
</details>

These two operations are common in the `if / else` branches and can be done before `if.
___
*[cpp/open3d/visualization/webrtc_server/WebRTCWindowSystem.cpp, line 291 at r7](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_NWlBX8K6QcooH_4qP:-M_NWlBX8K6QcooH_4qQ:b2mmcf) ([raw file](https://github.com/intel-isl/open3d/blob/fdcd0f32c9bcdee6a79bc04af02fcd4c0ffbdb78/cpp/open3d/visualization/webrtc_server/WebRTCWindowSystem.cpp#L291)):*
> ```cpp
>             SetWindowSize(GetOSWindowByUID(window_uid), width, height);
>         }
>     } catch (...) {
> ```

Can we make this `catch` more specific? Else it will also catch errors from the GUI calls. (which may be OK, but needs a different message).
___
*[cpp/open3d/visualization/webrtc_server/WebRTCWindowSystem.cpp, line 317 at r7](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_NXAbr7Rc_KreynVF6:-M_NXAbr7Rc_KreynVF7:bwnfmp0) ([raw file](https://github.com/intel-isl/open3d/blob/fdcd0f32c9bcdee6a79bc04af02fcd4c0ffbdb78/cpp/open3d/visualization/webrtc_server/WebRTCWindowSystem.cpp#L317)):*
> ```
>     std::thread thread(sender);
>     thread.join();
> ```
We launch a new thread and then just wait for it to finish? Is the thread needed in that case?
___
*[cpp/pybind/visualization/o3dvisualizer.cpp, line 151 at r7](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_NpyEH1NbjUlfDGZfc:-M_NpyEH1NbjUlfDGZfd:bwqfyf7) ([raw file](https://github.com/intel-isl/open3d/blob/fdcd0f32c9bcdee6a79bc04af02fcd4c0ffbdb78/cpp/pybind/visualization/o3dvisualizer.cpp#L151)):*
> ```cpp
>             .def_property_readonly(
>                     "uid", &O3DVisualizer::GetWebRTCUID,
>                     "Window's unique ID when WebRTCWindowSystem is use.")
> ```

    " Returns 'window_undefined' otherwise."
___
*[python/open3d/web_visualizer.py, line 113 at r3](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_3LgAR7pImxTD--Jy6:-M_NrQoOBxS6f0rfOR0h:b-n289bb) ([raw file](https://github.com/intel-isl/open3d/blob/3903d6a3d75f57940d97fab54045edf9f3f3b594/python/open3d/web_visualizer.py#L113)):*

`o3d.utility.reset_print_function()` is the workaround. 

The default logger aquires GIL, which causes issue with the `AsyncEventLoop` currently.
```cpp
    utility::Logger::GetInstance().SetPrintFunction([](const std::string& msg) {
        py::gil_scoped_acquire acquire;
        py::print(msg);
    });
```
</blockquote></details>

# TODO(Yixing): ...

yxlao (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3392)* status: 54 of 86 files reviewed, 25 unresolved discussions (waiting on @errissa, @prewettg, @ssheorey, @stotko, and @yxlao)

___
*[python/requirements.txt, line 3 at r3](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_3GsEsEuWmg2iuTK3k:-M_bOPNOCmQvoeCBVqgo:bjmunbo) ([raw file](https://github.com/intel-isl/open3d/blob/3903d6a3d75f57940d97fab54045edf9f3f3b594/python/requirements.txt#L3)):*

Refactored. 

Now we have `requirements_jupyter.txt`, which gets appended to `requirements.txt` automatically by CMake when building Jupyter.
</blockquote></details>

~~@ssheorey I reverted back this change, after several attempts, separating `requirements_jupyter.txt` and `requirements.txt` is still causing some CI errors. I'll address it in a future PR to get the current one unblocked.~~





update: I am trying to fix this again

prewettg (approved):
> Reviewed 4 of 32 files at r8.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3392)* status: 56 of 87 files reviewed, 24 unresolved discussions (waiting on @errissa, @prewettg, @ssheorey, @stotko, and @yxlao)

yxlao (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3392)* status: 53 of 87 files reviewed, 24 unresolved discussions (waiting on @errissa, @prewettg, @ssheorey, @stotko, and @yxlao)

___
*[python/requirements.txt, line 3 at r3](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_3GsEsEuWmg2iuTK3k:-M_cZfHu6MQ-ZltBrPbL:b-fx3dpb) ([raw file](https://github.com/intel-isl/open3d/blob/3903d6a3d75f57940d97fab54045edf9f3f3b594/python/requirements.txt#L3)):*

@ssheorey I reverted back this change, after several attempts, separating `requirements_jupyter.txt` and `requirements.txt` is still causing some CI errors. I'll address it in a future PR to get the current one unblocked.
</blockquote></details>

update: now it is working properly. We have `requirements_jupyter.txt` and `requirements.txt` as two files.

- At build time: we need `requirements_jupyter.txt` to generate the wheel with Jupyter on Linux.
- At install time: Linux user will get `requirements_jupyter.txt`+`requirements.txt` merged; Windows/macOS user will only get `requirements.txt`.

yxlao (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3392)* status: 53 of 87 files reviewed, 24 unresolved discussions (waiting on @errissa, @prewettg, @ssheorey, @stotko, and @yxlao)

___
*[python/requirements.txt, line 3 at r3](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_3GsEsEuWmg2iuTK3k:-M_c_9wRDU9QaUXOrMdl:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/3903d6a3d75f57940d97fab54045edf9f3f3b594/python/requirements.txt#L3)):*

update: now it is working properly. We have `requirements_jupyter.txt` and `requirements.txt` as two files.

- At build time: we need `requirements_jupyter.txt` to generate the wheel with Jupyter on Linux.
- At install time: Linux user will get `requirements_jupyter.txt`+`requirements.txt` merged; Windows/macOS user will only get `requirements.txt`.
</blockquote></details>

Done.

ssheorey (approved):
> Reviewed 7 of 32 files at r8, 2 of 4 files at r9.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3392)* status: 62 of 87 files reviewed, 4 unresolved discussions (waiting on @errissa, @prewettg, @ssheorey, @stotko, and @yxlao)

___
*[cpp/open3d/visualization/webrtc_server/ImageCapturer.cpp, line 40 at r7](https://reviewable.io/reviews/intel-isl/open3d/3392#-M_NnzanDoutVZygHCBk:-M_R9IIp-TdMQlSHtUOD:b-cfij45) ([raw file](https://github.com/intel-isl/open3d/blob/fdcd0f32c9bcdee6a79bc04af02fcd4c0ffbdb78/cpp/open3d/visualization/webrtc_server/ImageCapturer.cpp#L40)):*

The order comes from `SortIncludes: true` in our `.clang-format`.  We could add `// clang-format off` and `// clang-format on` to enforce a manual order. By reading the discussions from https://stackoverflow.com/q/2762568/1255535, it seems that there are no unified standards. I guess in this case clang-format works fine for us, we can keep it for now.

This is still a good reminder about headers. I went through the includes in the `webrtc_server` directory and removed the unnecessary ones.
</blockquote></details>

This is inconsistent with the [Google style guide](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes). But the stack-overflow accepted answer makes sense. OK to stick with clang-format output, though.

PR #3401, opened 2021-05-07
ICP GUI Demo 

reyanshsolis:
TODO:
- [x] Add dataset and download mechanism.

This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3401)


prewettg (changes requested):
> Reviewed 13 of 13 files at r1.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3401)* status: all files reviewed, 16 unresolved discussions (waiting on @errissa and @reyanshsolis)

*[a discussion](https://reviewable.io/reviews/intel-isl/open3d/3401#-M_83S8yBw1zQyb8iFsg:-M_83S8yBw1zQyb8iFsh:bqt82op) (no related file):*
In the example code it would be nice--as a user--to have some comments explaining the steps in the algorithm. It does a good job of commenting some quirks, but some explanation of the basic steps as they happen would help me learn what I need to do (and what I may need to modify for my situation). Also, since threaded programming is known to be difficult, it's probably a good idea to explain somewhere in the code what the mutex is locking, and why (namely, to ensure that the visualization thread doesn't read data that is in the middle of changing, and possibly crashing).

---
___
*[examples/cpp/TICPOdometry.cpp, line 44 at r1](https://reviewable.io/reviews/intel-isl/open3d/3401#-M_8-6Y8CKWj9I7X7IAB:-M_8-6Y8CKWj9I7X7IAC:bfkjwj) ([raw file](https://github.com/intel-isl/open3d/blob/ca55d9a8c1252c3c0892dd533712995cd7b719f2/examples/cpp/TICPOdometry.cpp#L44)):*
> ```
> const int WIDTH = 1600;
> const int HEIGHT = 900;
> ```
This might exceed some users' screen size (say, a 13" MacBook Air or something), although if our users are not likely to be able to use small screens due to the limitations of the hardware usually attached to them it would be okay.
___
*[examples/cpp/TICPOdometry.cpp, line 123 at r1](https://reviewable.io/reviews/intel-isl/open3d/3401#-M_8-_4j0tMxpOeY_osn:-M_8-_4j0tMxpOeY_oso:b-rrji8m) ([raw file](https://github.com/intel-isl/open3d/blob/ca55d9a8c1252c3c0892dd533712995cd7b719f2/examples/cpp/TICPOdometry.cpp#L123)):*
> ```cpp
> 
>         // Loads the pointcloud, converts to Float32, estimates normals if
>         // required, sets the "__visualization_scalar" parameter and it's min
> ```

cosmetic: "it's" -> "its"
___
*[examples/cpp/TICPOdometry.cpp, line 135 at r1](https://reviewable.io/reviews/intel-isl/open3d/3401#-M_8-g6A1l_0QMk0YigY:-M_8-g6A1l_0QMk0YigZ:b-2o73ki) ([raw file](https://github.com/intel-isl/open3d/blob/ca55d9a8c1252c3c0892dd533712995cd7b719f2/examples/cpp/TICPOdometry.cpp#L135)):*
> ```cpp
>         pointcloud_mat_ = GetPointCloudMaterial();
> 
>         // When window is closed, it will stop the execute of the code.
> ```

cosmetic: execute -> execution
___
*[examples/cpp/TICPOdometry.cpp, line 177 at r1](https://reviewable.io/reviews/intel-isl/open3d/3401#-M_8-pjn2h03MPa0DpY2:-M_8-pjn2h03MPa0DpY3:bq56vre) ([raw file](https://github.com/intel-isl/open3d/blob/ca55d9a8c1252c3c0892dd533712995cd7b719f2/examples/cpp/TICPOdometry.cpp#L177)):*
> ```cpp
>         // Final scale level downsampling is already performed while loading the
>         // data. -1 avoids re-downsampling for the last scale level.
>         voxel_sizes_[icp_scale_levels_ - 1] = -1;
> ```

Instead of -1 (but with comment, so that's good), doing something like:
```
namespace {  // anonymous namespace limits scope to just this file
    static const int DONT_RESAMPLE = -1;
}
```
at the top would make this (and wherever this is checked) more readable.
___
*[examples/cpp/TICPOdometry.cpp, line 195 at r1](https://reviewable.io/reviews/intel-isl/open3d/3401#-M_7y0s63lVfjlN-t9Ca:-M_7y0s63lVfjlN-t9Cb:b-d646qe) ([raw file](https://github.com/intel-isl/open3d/blob/ca55d9a8c1252c3c0892dd533712995cd7b719f2/examples/cpp/TICPOdometry.cpp#L195)):*
> ```cpp
> 
>         // --------------------- Main Function -----------------------
>         int i = 0;
> ```

Use `for (int i = 0; ...)`
___
*[examples/cpp/TICPOdometry.cpp, line 320 at r1](https://reviewable.io/reviews/intel-isl/open3d/3401#-M_7yL7o4LE3XODM7o5e:-M_7yL7o4LE3XODM7o5f:bt95fa9) ([raw file](https://github.com/intel-isl/open3d/blob/ca55d9a8c1252c3c0892dd533712995cd7b719f2/examples/cpp/TICPOdometry.cpp#L320)):*
> ```cpp
>         utility::LogInfo(" Dataset path: {}", path_dataset);
>         if (end_range_ > 500) {
>             utility::LogWarning(" Too large range. Memory might exceed.");
> ```

cosmetic: "Might exceed memory".
Also, why 500, specifically?
___
*[examples/cpp/TICPOdometry.cpp, line 420 at r1](https://reviewable.io/reviews/intel-isl/open3d/3401#-M_7weZA1PDcbTUWI91E:-M_7weZA1PDcbTUWI91F:bp3sewe) ([raw file](https://github.com/intel-isl/open3d/blob/ca55d9a8c1252c3c0892dd533712995cd7b719f2/examples/cpp/TICPOdometry.cpp#L420)):*
> ```cpp
> 
>                 // Dtype conversion to Float32. Currently only Float32
>                 // pointcloud is supported.
> ```

The Float32 limitation is just the renderer, right? If so, you might update to mention that, in case people prefer to do their calculations some other way and convert to Float32 for rendering (although that would probably be a large performance hit)
___
*[examples/cpp/TICPOdometry.cpp, line 429 at r1](https://reviewable.io/reviews/intel-isl/open3d/3401#-M_7yitIAAoWUVT-Eh_Y:-M_7yitIAAoWUVT-Eh_Z:bggcdm9) ([raw file](https://github.com/intel-isl/open3d/blob/ca55d9a8c1252c3c0892dd533712995cd7b719f2/examples/cpp/TICPOdometry.cpp#L429)):*
> ```cpp
>                 }
> 
>                 pointcloud_local.SetPointAttr("__visualization_scalar",
> ```

Since `__visualization_scalar` is currently undocumented and clearly internal (hence the double underscores), this should have a comment that it is how `unlitGradient` calculates the colors.
___
*[examples/cpp/TICPOdometry.cpp, line 452 at r1](https://reviewable.io/reviews/intel-isl/open3d/3401#-M_7xEhe1c5r952rWhiT:-M_7xEhe1c5r952rWhiU:bhd7c5h) ([raw file](https://github.com/intel-isl/open3d/blob/ca55d9a8c1252c3c0892dd533712995cd7b719f2/examples/cpp/TICPOdometry.cpp#L452)):*
> ```cpp
>                 // Adding it to our vector of pointclouds.
> 
>                 pointclouds_device[i++] =
> ```

pointclouds_device.push_back() is more idiomatic. Also doesn't crash if the vector isn't big enough (which it looks like it is in this case)
___
*[examples/cpp/TICPOdometry.cpp, line 458 at r1](https://reviewable.io/reviews/intel-isl/open3d/3401#-M_7xP-81fcn9it_zWnb:-M_7xP-81fcn9it_zWnc:b-8k0soi) ([raw file](https://github.com/intel-isl/open3d/blob/ca55d9a8c1252c3c0892dd533712995cd7b719f2/examples/cpp/TICPOdometry.cpp#L458)):*
> ```cpp
>                 max_points_in_frame_ = std::max(
>                         max_points_in_frame_,
>                         pointclouds_device[i - 1].GetPoints().GetLength());
> ```

pointclouds_device.back() would eliminate the need for the `i` loop variable (along with push_back()) above)
___
*[examples/cpp/TICPOdometry.cpp, line 463 at r1](https://reviewable.io/reviews/intel-isl/open3d/3401#-M_80wRc8NKOQAQuoL_x:-M_80wRc8NKOQAQuoL_y:b-531r8i) ([raw file](https://github.com/intel-isl/open3d/blob/ca55d9a8c1252c3c0892dd533712995cd7b719f2/examples/cpp/TICPOdometry.cpp#L463)):*
<details><summary><i alt="summary">Quoted 5 lines of code…</i></summary>

> ```
>                     " Failed to read pointcloud in sequence. Ensure pointcloud "
>                     "files are present in the given dataset path in continuous "
>                     "sequence from 0 to {}. Also, in case of large range, the "
>                     "system might be going out-of-memory. ",
>                     end_range_);
> ```
</details>

cosmetic: I recommend "Also, in case the range is large, the system might have run out of memory."
As a user, errors like this are frustrating, because it does not provide enough details to fix the problem (it is a little more detailed version of "an error occurred"). It combines two errors (I/O and out of memory), and for the former, it does not include the iteration that failed. Doing `"Failed to read pointcloud {} in sequence ...", i` would provide actionable details to debug the first problem. If you expect the possibility of running out of memory enough to include it in the message, you should catch `std::bad_alloc` and then since you have two cases, you can be explicit about which one is the problem. (Unless the memory was not allocated with `new`, in which case there's not much else you can do, unless you want to try to catch signals) 
___
*[examples/cpp/TICPOdometry.cpp, line 479 at r1](https://reviewable.io/reviews/intel-isl/open3d/3401#-M_7xa1K0myhw1jqE5jH:-M_7xa1K0myhw1jqE5jI:ba3dfna) ([raw file](https://github.com/intel-isl/open3d/blob/ca55d9a8c1252c3c0892dd533712995cd7b719f2/examples/cpp/TICPOdometry.cpp#L479)):*
> ```cpp
>         pointcloud_mat.scalar_max = max_visualization_scalar_;
>         pointcloud_mat.point_size = 0.3f;
>         // pointcloud_mat.base_color =
> ```

Remove?
___
*[examples/cpp/TICPOdometry.cpp, line 498 at r1](https://reviewable.io/reviews/intel-isl/open3d/3401#-M_7vnK3447dR3KF2q0U:-M_7vnK472MrFyi9izqS:bbe678v) ([raw file](https://github.com/intel-isl/open3d/blob/ca55d9a8c1252c3c0892dd533712995cd7b719f2/examples/cpp/TICPOdometry.cpp#L498)):*
> ```cpp
> 
> private:
>     std::mutex cloud_lock_;
> ```

It's not clear from this what things cloud_lock_ locks. I usually put the members that are protected behind the mutex directly under the mutex, followed by a blank line. For an example you probably want an explicit comment.
___
*[examples/cpp/TICPReconstruction.cpp, line 289 at r1](https://reviewable.io/reviews/intel-isl/open3d/3401#-M_7zQlZ2-2BvcqTqzFd:-M_7zQlZ2-2BvcqTqzFe:b-e0y0h0) ([raw file](https://github.com/intel-isl/open3d/blob/ca55d9a8c1252c3c0892dd533712995cd7b719f2/examples/cpp/TICPReconstruction.cpp#L289)):*
> ```cpp
> 
>             for (int j = 0; j < criterias[i].max_iteration_; j++) {
>                 std::this_thread::sleep_for(std::chrono::milliseconds(100));
> ```

If you sleep at the end of the loop you'll get the first update to the screen faster
___
*[examples/cpp/TICPReconstruction.cpp, line 617 at r1](https://reviewable.io/reviews/intel-isl/open3d/3401#-M_7zo4wEHD8KMbGqFr-:-M_7zo4wEHD8KMbGqFr0:bjadkvn) ([raw file](https://github.com/intel-isl/open3d/blob/ca55d9a8c1252c3c0892dd533712995cd7b719f2/examples/cpp/TICPReconstruction.cpp#L617)):*
> ```cpp
> 
> private:
>     std::mutex cloud_lock_;
> ```

Again, not clear what this locks

errissa (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3401)* status: 8 of 13 files reviewed, 16 unresolved discussions (waiting on @prewettg and @reyanshsolis)

___
*[examples/cpp/TICPOdometry.cpp, line 44 at r1](https://reviewable.io/reviews/intel-isl/open3d/3401#-M_8-6Y8CKWj9I7X7IAB:-M_MHvjc2FuuxgLoPifl:b-9k55vd) ([raw file](https://github.com/intel-isl/open3d/blob/ca55d9a8c1252c3c0892dd533712995cd7b719f2/examples/cpp/TICPOdometry.cpp#L44)):*

Is 1024x768 fine? Should I also change this in the `VoxelHashingGUI` example?
</blockquote></details>

@reyanshsolis The demos should probably be consistent so I would say yes. I run my MBP in "More Space" mode which is the equivalent of 1680x1050. In "Default" mode it's 1440x900. I personally think the layout of the demo works well with the aspect ratio of 1600x900. 1024x768 is too square. Maybe 1280x800 is a good compromise?

reyanshsolis (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3401)* status: 7 of 13 files reviewed, 16 unresolved discussions (waiting on @prewettg and @reyanshsolis)

___
*[examples/cpp/TICPOdometry.cpp, line 44 at r1](https://reviewable.io/reviews/intel-isl/open3d/3401#-M_8-6Y8CKWj9I7X7IAB:-M_Mty9DB-F3NBKfTY-L:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/ca55d9a8c1252c3c0892dd533712995cd7b719f2/examples/cpp/TICPOdometry.cpp#L44)):*

@reyanshsolis The demos should probably be consistent so I would say yes. I run my MBP in "More Space" mode which is the equivalent of 1680x1050. In "Default" mode it's 1440x900. I personally think the layout of the demo works well with the aspect ratio of 1600x900. 1024x768 is too square. Maybe 1280x800 is a good compromise? 
</blockquote></details>

Done.

prewettg (changes requested):
> Reviewed 3 of 6 files at r2.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3401)* status: 10 of 13 files reviewed, 3 unresolved discussions (waiting on @prewettg and @reyanshsolis)

___
*[examples/cpp/TICPOdometry.cpp, line 458 at r1](https://reviewable.io/reviews/intel-isl/open3d/3401#-M_7xP-81fcn9it_zWnb:-M_MvA9gEBz_fX40zcMj:b-6gskee) ([raw file](https://github.com/intel-isl/open3d/blob/ca55d9a8c1252c3c0892dd533712995cd7b719f2/examples/cpp/TICPOdometry.cpp#L458)):*

pointclouds_device.back() would eliminate the need for the `i` loop variable (along with push_back()) above)
</blockquote></details>

I'd like to mark this resolved, as I won't block based on this, but that is no longer an option (probably unless you reply), so you can click resolved yourself.
___
*[examples/cpp/TICPOdometry.cpp, line 498 at r1](https://reviewable.io/reviews/intel-isl/open3d/3401#-M_7vnK3447dR3KF2q0U:-M_MvYKiDIn4s9Hk9m_-:bxrixne) ([raw file](https://github.com/intel-isl/open3d/blob/ca55d9a8c1252c3c0892dd533712995cd7b719f2/examples/cpp/TICPOdometry.cpp#L498)):*

Done.
</blockquote></details>

There's a comment above, but I still think that either moving items here or a comment here would be a good idea.
___
*[examples/cpp/TICPReconstruction.cpp, line 617 at r1](https://reviewable.io/reviews/intel-isl/open3d/3401#-M_7zo4wEHD8KMbGqFr-:-M_Mvk2B5ZMYaSY5YXG2:bw78rpa) ([raw file](https://github.com/intel-isl/open3d/blob/ca55d9a8c1252c3c0892dd533712995cd7b719f2/examples/cpp/TICPReconstruction.cpp#L617)):*

Done.
</blockquote></details>

Same as above, it's not clear in this section of the code what the lock locks.

reyanshsolis (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3401)* status: 10 of 13 files reviewed, 3 unresolved discussions (waiting on @prewettg)

___
*[examples/cpp/TICPOdometry.cpp, line 452 at r1](https://reviewable.io/reviews/intel-isl/open3d/3401#-M_7xEhe1c5r952rWhiT:-M_NE2j82vW8UtLmboRr:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/ca55d9a8c1252c3c0892dd533712995cd7b719f2/examples/cpp/TICPOdometry.cpp#L452)):*

Ok, although I personally think:
```
vec.reserve(size);
for (...) {
    vec.push_back(...);
}
is clearer.
</blockquote></details>

Done.
___
*[examples/cpp/TICPOdometry.cpp, line 458 at r1](https://reviewable.io/reviews/intel-isl/open3d/3401#-M_7xP-81fcn9it_zWnb:-M_NE3VzDKRoa4VxOv_c:bn6prrr) ([raw file](https://github.com/intel-isl/open3d/blob/ca55d9a8c1252c3c0892dd533712995cd7b719f2/examples/cpp/TICPOdometry.cpp#L458)):*

I'd like to mark this resolved, as I won't block based on this, but that is no longer an option (probably unless you reply), so you can click resolved yourself.
</blockquote></details>

Since it was related to the above discussion, I was waiting for it to be resolved first :(. 

Anyways, we need `i / count` for displaying the `loading progress`. 
Also, in the first iteration, I thought of using the `reserve` design, but I wanted to make use of the `device` constructor for each of the pointcloud, that's why I had used 
```
std::vector<t::geometry::PointCloud> pointclouds_device(
                filenames_.size(), t::geometry::PointCloud(device_));
``` this design earlier. 
However, I have changed it according to your suggestions now. 
```

___
*[examples/cpp/TICPOdometry.cpp, line 498 at r1](https://reviewable.io/reviews/intel-isl/open3d/3401#-M_7vnK3447dR3KF2q0U:-M_NDwz58WJjpl0l3iAI:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/ca55d9a8c1252c3c0892dd533712995cd7b719f2/examples/cpp/TICPOdometry.cpp#L498)):*

There's a comment above, but I still think that either moving items here or a comment here would be a good idea.
</blockquote></details>

Done.
___
*[examples/cpp/TICPReconstruction.cpp, line 617 at r1](https://reviewable.io/reviews/intel-isl/open3d/3401#-M_7zo4wEHD8KMbGqFr-:-M_N61apDnFmeTC6uKKr:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/ca55d9a8c1252c3c0892dd533712995cd7b719f2/examples/cpp/TICPReconstruction.cpp#L617)):*

Same as above, it's not clear in this section of the code what the lock locks.
</blockquote></details>

Done.

prewettg (approved):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3401)* status: 6 of 14 files reviewed, all discussions resolved (waiting on @prewettg)

PR #3403, opened 2021-05-07
Fix O3DVisualizer not accepting t::geometry::PointCloud

prewettg:
O3DVisualizer did not accept TPointCloud if a material is specified, because `valid_pcd` was only set in the no-material case


errissa (approved):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3403)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @prewettg)

PR #3404, opened 2021-05-08
PEP600 Linux wheel naming

ssheorey:
Updates Linux x86_64 wheel platform tag from `linux_x86_64` to `manylinux_X_Y_x86_64` where (X,Y) is the `glibc` (major,minor) version on the build OS. This follows the perennial manylinux convention [PEP600](https://www.python.org/dev/peps/pep-0600/).

**eg:** Current name:
`open3d-0.12.0+9e7e99cc9-cp38-cp38-linux_x86_64.whl`
new name:
`open3d-0.12.0+9e7e99cc9-cp38-cp38-manylinux_2_27_x86_64.whl`

PEP600 is supported by pip version >=20.3. If the PyPA index has both `manylinux_X_Y` and `linux` wheels, the 
tested behavior during install:

- pip>=20.3: Pick `manylinux_X_Y` if (X,Y) <= OS glibc version, else pick `linux` wheel.
- pip<20.3: Does not recognize `manylinux_X_Y`. Picks `linux` wheel.

For development wheels, we will only upload the PEP600 wheels. Advanced users will now need a recent `pip` to install them. pip will refuse to install wheels on too old platforms.

For release wheels, we will upload both `manylinux2014` and `manylinux_2_27` (Ubuntu 18.04 based) wheels. `pip`>=20.3 will pick the manylinux_2_27 wheels. Since the manylinux2014 wheels are also present, incompatible installs will still happen in the transition period. As PEP600 gets more widespread, we will discontinue manylinux2014 wheels to prevent incompatible installs.

**TODO:** The new manylinux docker images are based on Debian 9 (glibc 2.23). It should be easier to port wheel creation to them since package names should be the same. We can evaluate the need for this based on user feedback.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3404)


keithahern:
NOTE the links to the Linux downloads under "Development Version (pip)" on the [Getting Started ](http://www.open3d.org/docs/latest/getting_started.html) page are broken since this change.  Presumably the link template needs to be aware of a new file pattern after the activation date of this change.

ssheorey:
> NOTE the links to the Linux downloads under "Development Version (pip)" on the [Getting Started ](http://www.open3d.org/docs/latest/getting_started.html) page are broken since this change. Presumably the link template needs to be aware of a new file pattern after the activation date of this change.

Thanks for checking! We just updated the links. Can you try again?

keithahern:
Still not working for me

```
<Error>
<Code>NoSuchKey</Code>
<Message>The specified key does not exist.</Message>
<Details>
No such object: open3d-releases-master/python-wheels/open3d-0.12.0+7e813a8-cp38-cp38-manylinux_2_27_x86_64.whl
</Details>
</Error>
```

ssheorey:
The template is now correct (manylinux_2_27).

You are seeing this error since there are two separate processes that update the documentation and build and upload the wheel. They are not synchronized, so stale links are possible during updates. Please check back in some time. If you need the latest wheel immediately, you can get it directly from the CI:

- https://github.com/intel-isl/Open3D/actions?query=branch%3Amaster
- Click on your OS on the left (Ubuntu CI, macOS CI, Windows CI)
- Click on the latest successful run
- Artifacts -> open3d_*_wheels

keithahern:
Confirmed working now. 

Thank you

ajinkyakhoche:
Could you add pip>=20.3 as an installation requirement? It didn't seem to work till i updated pip.

PR #3426, opened 2021-05-11
Extended TriangleMesh properties with IsOriented. Simplified OrientTriangleHelper. Fixed checks for GetVolume

JoseLlorensRipolles:
The TriangleMesh::GetVolume method checks if the mesh is **orientable**, but what the authors of the paper referenced in the method (Zhang and Chen, "Efficient feature extraction for 2D/3D objects in representation", 2001) assume is that the mesh is **oriented**. So I extended the mesh properties to add IsOriented to them. Then I changed the GetVolume to check if the mesh is oriented, and suggest using OrientTriangles if it is not.

To check if a TriangleMesh is oriented, we can just check that no directed edge repeat twice or more. A similar approach was being used for the OrientTriangles and the IsOrientable methods. 
The methods used a map from sorted edges (minor vertex first, major vertex second) to their orientation (The directed edge). Then a sequential check was done for all directed edges to either not being on the map keys as a sorted edge (Meaning this directed edge has not been seen before), or being different to the map value for them as a sorted edge (Meaning this directed edge has not been seen before). This was a bit difficult to understand and archived the same than just keeping a set and checking that the directed edge is not in it, but with more steps. I unified all three methods to use the set approach.

Thank you in advance, let me know if you want me to add something to the changelog/docs.

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/3426)


JoseLlorensRipolles:
Hi @germanros1987 ,

Would it be possible to get some reviewers assigned / workflows approved?

Thank you in advance and thank you very much for the awesome work in Open3D!

JoseLlorensRipolles:
Hi @yxlao @reyanshsolis @theNded ,

Is it possible that this PR has passed unnoticed? If so, would it be possible to get some reviewers or workflows approved? If not, could it be closed?

Thank you very much in advance!

reyanshsolis:
Hi @JoseLlorensRipolles, sorry for the delayed response, we appreciate your valuable contribution. 
I will review this PR soon and share my suggestions. 

Thanks for your patience

JoseLlorensRipolles:
Hey @reyanshsolis ,

Absolutely no problem. Thank **you** for your awesome open source work!

reyanshsolis:
I may take some time to verify the changes. Meanwhile, may you please
1. Add python unit-test.
2. merge the latest master to this branch (might resolve the documentation issue).
3. fix formatting (style check CI). [If you are on Mac or Linux, you may do `make apply-style` from inside your build directory, or refer [style guide](http://www.open3d.org/docs/release/contribute/styleguide.html)].

Thanks

JoseLlorensRipolles:
Hi,

I've been thinking about this a bit more. I reckon that unittests for IsOrientable and OrientTriangles are needed.

Thinking about the test for IsOrientable I thought that orientability is just and alias for being edge manifold with allowed boundaries. Isn't it?. However, I found the documentation example for a [non-manifold-edge](http://www.open3d.org/docs/release/tutorial/geometry/mesh.html#Mesh-properties) mesh. That mesh is not considered edge-manifold-boundary, but it is considered orientable. I don't see how that mesh could be orientable, as 3 triangles share the same edge. Maybe I am missing something, but could it be a bug? As an example I've prepared a [gist](https://gist.github.com/JoseLlorensRipolles/2d313b34d5c3a3e6b4db5f4f15302f44) (needs to be run inside the examples/python/geometry folder)  with the visualization of the mesh before and after orienting it. I will be able to take a closer look in (hopefully) a few days 

The tests for OrientTriangles seem a bit more tricky. We would need to compare two mesh, without relying on triangle orders and absolute vertex orders but checking relative vertex order. I have a few ideas and would like to discuss them if you think that the tests are needed. 

What are your thoughts on this? Thank you in advance!

JoseLlorensRipolles:
I've checked and IsOrientable is not an alias for being edge manifold with allowed boundaries. For example the Möbius strip is not orientable but it is a 2D manifold. 

About the docs example of an orientable, non-manifold edge mesh, I am not sure orientability can even be applied to non manifold meshes, as orientability is a property of surfaces, which are 2D manifolds. 

In the [paper](https://ieeexplore.ieee.org/document/958278) cited above, they state :

> The consistent condition is very easy to satisfy. For two neighboring triangles, if the common edge has different directions, then  the normals of the two triangles are consistent.

That consistent condition is the mesh being oriented. Therefore for three neighboring triangles, as in the non-manifold edge mesh, it would be impossible to satisfy that consistency. 

I am trying to find proper references, but [here](http://graphics.stanford.edu/courses/cs468-10-fall/LectureSlides/02_Basics.pdf) we can find orientability definitions that would contradict the non-manifold edge example of the docs being orientable. 

Assuming it is not the desired outcome, this is why the code has this bug:

Given 3 triangles that share an edge (Let's say A, B and C), the edge A will be kept in the "edge_to_orientation" map. The edge B will be compared with the map and allowed to flip so that A and B are oriented. Once the comparison is done, the map does not change and no information about B is kept. When the edge C is visited, it is compared with the map again, and allowed to flip to be oriented with A, however it is not compared with B. So, B will be oriented with A and C will be oriented with A. But B is not oriented with C and the other way around.

The refactor I propose here fixes this bug, as all directed edges are kept in the set, so the triangle C will need to be oriented with A **and** with B.

reyanshsolis:
Hi @JoseLlorensRipolles, thanks for your efforts. Sorry it is taking some time for us to review this PR, as the team is out of bandwidth, working on the upcoming release. We will get back to this soon. The contribution is appreciated. 
Thanks.

JoseLlorensRipolles:
Hi @reyanshsolis 

No worries, really. Best of luck with the release, looking forward to it!

agrodet:
Looking for a faster way to do `OrientTriangles()`, I tried your pull request with my own data. I tried with a mesh `geometry::TriangleMesh: 682488 triangles and 341248 vertices`, and unfortunately it took me 30% longer than the previous version. However, I must say that I didn't conduct any more detailed benchmark.

Also, I think it would be better to replace 
```
std::unordered_map<Eigen::Vector2i, std::unordered_set<int>,
                       utility::hash_eigen<Eigen::Vector2i>>
            adjacent_triangles;
```
by
```
std::unordered_map<Eigen::Vector2i, std::vector<int>,
                       utility::hash_eigen<Eigen::Vector2i>>
            adjacent_triangles;
```
similarly to what is done in [`GetEdgeToTrianglesMap()`](https://github.com/intel-isl/Open3D/blob/3ec7a1919e1e27ca4c3ed14db41f4c96cbdfb316/cpp/open3d/geometry/TriangleMesh.cpp#L1139-L1142). I don't think we really need to use a set here.

JoseLlorensRipolles:
Hi @agrodet,

Thank you for bringing this up! I've done some benchmarking but I'm only getting ~7% of a slowdown, could you give me some more info about your mesh? Is it mostly already oriented etc etc? Thank you in advance.

Meanwhile keep in mind the bug that this PR is fixing, in case it is something that applies to your proyect.

About finding faster ways to do it, I will think about it.

JoseLlorensRipolles:
Hi, 

I'm closing this PR. I understand how tight on time/resources open source teams are, no worries. If at any point you feel like it is worth-it to fix it feel complitely free to reopen or reach me out for any clarifications.

Thank you!

PR #3431, opened 2021-05-13
Combine O3DVisualizer and GuiVisualizer

prewettg:
- bug?: the ground plane seems to be related to the rotation of the sun or the IBL
- note: Mobileye will need to update their script as `Open3DScene.show_ground_plane()` only takes one argument; the other is now the `model_up` property.
- test: passing results of read_triangle_model() results in lit model (I believe I resolved the merge conflict correctly, but should check)
- test: lighting changed, and might need adjustments. Also, since the app and O3DVisualizer both use O3DVisualizer, the app might need lighting adjustments.

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/3431)


benjaminum:
The documentation of the visualizer classes need some updates. I think this would fit well into this PR or a new PR after merging this. @errissa do you know what the status of this PR is?

Issue #2837, opened 2020-12-30
Bug with RealSense bag reader


shiyoung77:
I was trying to use the sample code from "http://www.open3d.org/docs/release/tutorial/sensor/realsense.html". 
```
import open3d as o3d
bag_reader = o3d.t.io.RSBagReader()
bag_reader.open('L515_test.bag')
while not bag_reader.is_eof():
    im_rgbd = bag_reader.next_frame()
bag_reader.close()
```
However, open3d failed to open the bag but gave me the following error instead.
```
free(): invalid pointer
Aborted
```

**Environment:**

- Operating system:Ubuntu 18.04
- Python version: Python 3.7
- Open3D version: 0.12.0
- Is this remote workstation?:  no
- How did you install Open3D?:
  - conda create -n open3d
  - conda activate open3d
  - pip install open3d-0.12.0-cp37-cp37m-manylinux2014_x86_64.whl 
- Compiler version (if built from source): N/A

colabi:
Is it known what version worked last?  I've been banging my head on this and just found this new issue.

PR #3434, opened 2021-05-13
GLIBCXX_USE_CXX11_ABI sync with librealsense

ssheorey:
- Remove patch for #2837 This provides a fix for the same issue.
- Upgrade librealsense to v2.44.
- RSUSB_BACKEND is no longer required in Linux.
- add --force-reinstall to make install-pip-package.

Based on user feedback from @NobuoTsukamoto for https://github.com/intel-isl/Open3D/pull/3191

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3434)


yxlao (changes requested):
> Reviewed 5 of 7 files at r1, 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3434)* status: all files reviewed, 1 unresolved discussion (waiting on @germanros1987 and @ssheorey)

___
*[cpp/pybind/make_install_pip_package.cmake, line 7 at r2](https://reviewable.io/reviews/intel-isl/open3d/3434#-M_bgeULE06b1VYOq8tT:-M_bgeUMASkBx3FtBISd:b-kpa9a5) ([raw file](https://github.com/intel-isl/open3d/blob/4ac743ea89cdeca4f02a216774e179089e70f010/cpp/pybind/make_install_pip_package.cmake#L7)):*
> ```CMake
> #       it is guaranteed that there is only one wheel in ${PYTHON_PACKAGE_DST_DIR}/pip_package/*.whl
> file(GLOB WHEEL_FILE "${PYTHON_PACKAGE_DST_DIR}/pip_package/*.whl")
> execute_process(COMMAND pip install --force-reinstall --upgrade ${WHEEL_FILE})
> ```

The last time I tried this, it seems like this actually reinstalls not only Open3D but also all of Open3D's dependencies, which makes the installation pretty slow. Is this the case?

The alternative method is to `pip uninstall open3d` if Open3D is already installed and `pip install -U open3d`. Since it is not directly related to other changes in this PR, we can make it in a separate one.

ssheorey (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3434)* status: 6 of 7 files reviewed, 1 unresolved discussion (waiting on @germanros1987 and @yxlao)

___
*[cpp/pybind/make_install_pip_package.cmake, line 7 at r2](https://reviewable.io/reviews/intel-isl/open3d/3434#-M_bgeULE06b1VYOq8tT:-M_bnNV2CcJSXJRRIDyl:bp3aow6) ([raw file](https://github.com/intel-isl/open3d/blob/4ac743ea89cdeca4f02a216774e179089e70f010/cpp/pybind/make_install_pip_package.cmake#L7)):*

The last time I tried this, it seems like this actually reinstalls not only Open3D but also all of Open3D's dependencies, which makes the installation pretty slow. Is this the case?

The alternative method is to `pip uninstall open3d` if Open3D is already installed and `pip install -U open3d`. Since it is not directly related to other changes in this PR, we can make it in a separate one.
</blockquote></details>

Reverted.

yxlao (approved):

PR #3507, opened 2021-05-27
Add support for GPU-based tgeometries internally

errissa:
* Automatically copies GPU-based T point clouds to CPU but outputs warning

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3507)


ssheorey (changes requested):
Not familiar with this code, so I can only give some general thoughts:
- We should cover all interfaces that a user may give us a TPointCloud, including update().
- We should also update TImage display code [also update_image()].
- CPU() will return a copy for a CUDA point cloud and return self for a CPU point cloud.
- I think an automatic display of geometry regardless of where it is, is what a user would expect, so it should be OK to skip the warning.

PR #3456, opened 2021-05-17
Modernize CMake build scripts

stotko:
- Split collecting source files into subdirectories: Only single layer (e.g. `add_subdirectory(core)`) with all further subdirectories handled inside that layer (e.g. `core/hashmap` in `core/CMakeLists.txt`)
- Use `target_sources(<target> PRIVATE ...)` to specify source files for `<target>`
- Sort source files by name and split them into groups based on their (sub)directory within the layer
- Remove `GLOB` statements and list all source files explicitly
- Avoid variables and set properties directly for better readability
- Consistent function naming convention `open3d_<function_name>(...)`
- Link against `Open3D::Open3D` instead of `${PROJECT_NAME}`

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3456)

PR #3524, opened 2021-05-31
fix using wrong pip module for installing the pip package

benjaminum:
This PR ensures that the `install-pip-package` target uses the pip module that matches the python environment.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3524)

### DIFF
--- a/cpp/pybind/CMakeLists.txt
+++ b/cpp/pybind/CMakeLists.txt
@@ -304,6 +304,7 @@ add_custom_target(pip-package
 add_custom_target(install-pip-package
     COMMAND ${CMAKE_COMMAND}
             -DPYTHON_PACKAGE_DST_DIR=${PYTHON_PACKAGE_DST_DIR}
+            -DPYTHON_EXECUTABLE=${PYTHON_EXECUTABLE}
             -P ${CMAKE_CURRENT_SOURCE_DIR}/make_install_pip_package.cmake
     DEPENDS pip-package
 )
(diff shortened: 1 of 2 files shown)


yxlao (approved):
LGTM as long as the test is passing.

There seems to be some issue with Windows + CUDA, could you double-check that?

benjaminum:
@yxlao seems to be related to the fmt library. Error messages do not make sense. Could be a duplicate definition. Probably affects other PSs too.

stotko:
> @yxlao It seems to be related to the fmt library. Error messages do not make sense. Could be a duplicate definition.
> #3456 has similar errors and also does have changes to CMake scripts. Seems like a small change to a cmake script can trigger this.

The error in #3456 occurred after rebasing to the latest `master` branch changes and addressing some minor comments. First. I thought that this could be related to the latter part, but it now looks more like a regression coming from `master`.

stotko:
After checking a few things, this seems like a regression in GitHub Actions' Windows CI image. They recently upgraded Visual Studio 2019 from 16.9 to 16.10 which seems to introduce these error when compiling with CUDA.

Reference: https://github.com/actions/virtual-environments/issues/3485

benjaminum:
> After checking a few things, this seems like a regression in GitHub Actions' Windows CI image. They recently upgraded Visual Studio 2019 from 16.9 to 16.10 which seems to introduce these error when compiling with CUDA.
> 
> Reference: [actions/virtual-environments#3485](https://github.com/actions/virtual-environments/issues/3485)

This makes sense. I did re-run the Win CI for merged PR #3521 (basically master) and we get the same errors.

PR #3542, opened 2021-06-04
Hide 3rd party library symbols from client code

ssheorey:
Currently, on Linux and macOS, Open3D exports all symbols, including those from 3rd party libraries. This can cause issues:
- Increased loading time of the Open3D (or pybind) DSO since all symbols need to be relocated by the dynamic linker at program startup.
- Possible symbol collision with other user libraries with the same dependencies as Open3D, possibly of different versions. The actual symbol that gets used depends on the user's loading order. This can cause bugs or crashes.
- Increased run time, since each exported symbol needs a PLT lookup, instead of direct access. This very slightly increases the time of each function call.
- The exported symbols also take up a lot of space making the Open3D DSO larger.
- There's a small chance of user code calling Open3D 3rdparty library code (un)intentionally.

For more information, see:
[How to write shared libraries - Ulrich Drepper](https://akkadia.org/drepper/dsohowto.pdf)

This PR:

- `ExternalProject_CMAKE_ARGS` and `ExternalProject_CMAKE_ARGS_hidden` CMake variables to uniformly initialize CMake properties of 3rd party libraries that use ExternalProject.
- Builds all 3rd party libraries with `-fvisibility=hidden`. Open3D code building is not changed. Libraries built externally export symbols according to their compiler settings. IPP and MKL seem to export all symbols, but not WebRTC.
    - The Open3D pybind DLL now exports about **19K** symbols instead of the previous **50K**.
    - `python -c "import open3d"` (only CPU) time decreased from **1.4s to 0.7s** on laptop (both stripped binaries).
- Updates `OPEN3D_API` macro to ensure both MSVC and Unix have the same exports. See [GCC visiibility](https://gcc.gnu.org/wiki/Visibility).
- New `HIDDEN` option to CMake macro `open3d_set_open3d_lib_properties` to hide **all** symbols from an object library. Used for hiding symbols from `t::pipelines::kernel` and `t::geometry::kernel`
- Fixes link option `-S` -> `LINKER:-S` to strip debug symbols for smaller binaries when DEVELOPER_BUILD is OFF. This reduces the CPU pybind file size from about **420MB -> 141MB** (approx.).
- MKL, IPP (prebuilt with visible symbols) -> Hidden on Linux (GNU ld `--exclude-libs` option) and macOS (llvm ld `-load_hidden` option).

TODO (future PRs):
- filament -> declared as HEADER. Library headers belongs to the public interface, so presumably filament symbols are part of Open3D ABI. Confirm and hide if not necessary.
- Assimp symbols are leaking....
- Header only libraries: Eigen, flann, ...
- Hide remaining internal  Open3D symbols. 
- Use static or anonymous namespaces for any functions that are only used inside a file. [This is mostly correctly implemented now].

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3542)


stotko (commented):
> Reviewed 13 of 19 files at r1, 2 of 2 files at r2, 10 of 10 files at r4.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3542)* status: all files reviewed, 10 unresolved discussions (waiting on @ssheorey and @yxlao)

___
*[CMakeLists.txt, line 304 at r4](https://reviewable.io/reviews/intel-isl/open3d/3542#-Mb_V7uI1Zw7Ro2VfxlW:-Mb_V7uKDhILjudcaJ2d:b-7a3cqs) ([raw file](https://github.com/intel-isl/open3d/blob/05ffd4ebe53bf940795b4cbaedcd6567666d64dd/CMakeLists.txt#L304)):*
> ```CMake
>         -Wl,-z,relro,-z,now         # Data relocation protection
>         -pie                        # Position independent executable
>         $<$<CONFIG:Release>:LINKER:-S>
> ```

Add comment for consistency
___
*[CMakeLists.txt, line 318 at r4](https://reviewable.io/reviews/intel-isl/open3d/3542#-Mb_VB-Z62h6hlL4j0ra:-Mb_VB-Z62h6hlL4j0rb:b-34vmyd) ([raw file](https://github.com/intel-isl/open3d/blob/05ffd4ebe53bf940795b4cbaedcd6567666d64dd/CMakeLists.txt#L318)):*
> ```CMake
>     set(HARDENING_LDFLAGS  ""
>         # -pie Position independent execution is default on macOSX 10.6+
>         $<$<CONFIG:Release>:LINKER:-S>
> ```

See above
___
*[CMakeLists.txt, line 330 at r4](https://reviewable.io/reviews/intel-isl/open3d/3542#-Mb_VCCI0PqousPJJSrj:-Mb_VCCI0PqousPJJSrk:b-34vmyd) ([raw file](https://github.com/intel-isl/open3d/blob/05ffd4ebe53bf940795b4cbaedcd6567666d64dd/CMakeLists.txt#L330)):*
> ```CMake
>         -Wl,-z,relro,-z,now  # Data relocation protection
>         -pie                 # Position independent executable
>         $<$<CONFIG:Release>:LINKER:--strip-debug>
> ```

See above
___
*[3rdparty/find_dependencies.cmake, line 33 at r4](https://reviewable.io/reviews/intel-isl/open3d/3542#-Mb_VOUj1WbWZGY1Eg70:-Mb_VOUj1WbWZGY1Eg71:b-qrskoc) ([raw file](https://github.com/intel-isl/open3d/blob/05ffd4ebe53bf940795b4cbaedcd6567666d64dd/3rdparty/find_dependencies.cmake#L33)):*
> ```CMake
> 
> # Apply LANG_VISIBILITY_PRESET to static libraries and archives as well
> cmake_policy(SET CMP0063 NEW)
> ```

This should not be necessary. `cmake_minimum_required(VERSION 3.18)` in the top-level CMake file already sets all policies up to including `CMP0108` to `NEW`.
___
*[3rdparty/find_dependencies.cmake, line 66 at r4](https://reviewable.io/reviews/intel-isl/open3d/3542#-Mb_VuSO5pIhuwX6pH6I:-Mb_VuSP53gMBGzwaHsY:b6ae466) ([raw file](https://github.com/intel-isl/open3d/blob/05ffd4ebe53bf940795b4cbaedcd6567666d64dd/3rdparty/find_dependencies.cmake#L66)):*
> ```CMake
> #    LIBS <target> [<target> ...]
> #        extra link dependencies
> ```

nit: Revert for consistency with remaining code.
___
*[3rdparty/find_dependencies.cmake, line 171 at r4](https://reviewable.io/reviews/intel-isl/open3d/3542#-Mb_WfopCbfFCx36akAf:-Mb_WfopCbfFCx36akAg:b-pudagu) ([raw file](https://github.com/intel-isl/open3d/blob/05ffd4ebe53bf940795b4cbaedcd6567666d64dd/3rdparty/find_dependencies.cmake#L171)):*
> ```CMake
>     -DCMAKE_CUDA_COMPILER_LAUNCHER=${CMAKE_CUDA_COMPILER_LAUNCHER}
>     -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
>     $<$<PLATFORM_ID:Windows>:-DCMAKE_POLICY_DEFAULT_CMP0091:STRING=NEW>
> ```

`$<$<PLATFORM_ID:Windows>:...>` generator expression is not directly needed. CMake is smart enough to use the policy only when needed, so passing the value unconditionally makes the expression look simpler.
___
*[3rdparty/filament/filament_build.cmake, line 66 at r4](https://reviewable.io/reviews/intel-isl/open3d/3542#-Mb_XYcW6tJa1VucD0S2:-Mb_XYcW6tJa1VucD0S3:b-iybzkj) ([raw file](https://github.com/intel-isl/open3d/blob/05ffd4ebe53bf940795b4cbaedcd6567666d64dd/3rdparty/filament/filament_build.cmake#L66)):*
> ```CMake
>         -DFILAMENT_SKIP_SAMPLES=ON
>         -DFILAMENT_OPENGL_HANDLE_ARENA_SIZE_IN_MB=20 # to support many small entities
>         -DCMAKE_POSITION_INDEPENDENT_CODE=ON
> ```

nit: Revert this change to minimize noise
___
*[3rdparty/zlib/zlib.cmake, line 17 at r4](https://reviewable.io/reviews/intel-isl/open3d/3542#-Mb_YFFM1OTfXexzkDbd:-Mb_YFFN9F9M8gNk6nP6:br4q38o) ([raw file](https://github.com/intel-isl/open3d/blob/05ffd4ebe53bf940795b4cbaedcd6567666d64dd/3rdparty/zlib/zlib.cmake#L17)):*
> ```CMake
>     CMAKE_ARGS
>         -DCMAKE_INSTALL_PREFIX=<INSTALL_DIR>
>         ${ExternalProject_CMAKE_ARGS}  # Symbols not hidden due to examples built
> ```

Is it possible to disable building their examples such that we can use the hidden version here as well?
___
*[cpp/open3d/t/geometry/CMakeLists.txt, line 20 at r4](https://reviewable.io/reviews/intel-isl/open3d/3542#-Mb_YURZ8JnpNO_ZMWYA:-Mb_YURZ8JnpNO_ZMWYB:b-bezqam) ([raw file](https://github.com/intel-isl/open3d/blob/05ffd4ebe53bf940795b4cbaedcd6567666d64dd/cpp/open3d/t/geometry/CMakeLists.txt#L20)):*
> ```CMake
> 
> if(BUILD_CUDA_MODULE)
>   target_include_directories(tgeometry SYSTEM
> ```

nit: Indentation and line break are inconsistent now.
___
*[cpp/open3d/t/geometry/kernel/CMakeLists.txt, line 1 at r4](https://reviewable.io/reviews/intel-isl/open3d/3542#-Mb_YfG3Crf1slNw1L06:-Mb_YfG3Crf1slNw1L07:bcg3o1p) ([raw file](https://github.com/intel-isl/open3d/blob/05ffd4ebe53bf940795b4cbaedcd6567666d64dd/cpp/open3d/t/geometry/kernel/CMakeLists.txt#L1)):*
> ```CMake
> add_library(tgeometry_kernel OBJECT)
> ```

Would it make sense to revert these changes and apply setting the visibility per function via `OPEN3D_API` in a follow-up PR?

ssheorey (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3542)* status: all files reviewed, 8 unresolved discussions (waiting on @stotko and @yxlao)

___
*[CMakeLists.txt, line 304 at r4](https://reviewable.io/reviews/intel-isl/open3d/3542#-Mb_V7uI1Zw7Ro2VfxlW:-MbcWjmQ2ZNMi2PbOnEq:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/05ffd4ebe53bf940795b4cbaedcd6567666d64dd/CMakeLists.txt#L304)):*

Add comment for consistency
</blockquote></details>

Done.
___
*[CMakeLists.txt, line 318 at r4](https://reviewable.io/reviews/intel-isl/open3d/3542#-Mb_VB-Z62h6hlL4j0ra:-MbcX-l00tRHTIs9NX4P:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/05ffd4ebe53bf940795b4cbaedcd6567666d64dd/CMakeLists.txt#L318)):*

See above
</blockquote></details>

Done.
___
*[CMakeLists.txt, line 330 at r4](https://reviewable.io/reviews/intel-isl/open3d/3542#-Mb_VCCI0PqousPJJSrj:-MbcX0fL2tCGT77mJkC2:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/05ffd4ebe53bf940795b4cbaedcd6567666d64dd/CMakeLists.txt#L330)):*

See above
</blockquote></details>

Done.
___
*[3rdparty/find_dependencies.cmake, line 33 at r4](https://reviewable.io/reviews/intel-isl/open3d/3542#-Mb_VOUj1WbWZGY1Eg70:-MbcXCBF9XecuFn3WvXu:b-cnjyyx) ([raw file](https://github.com/intel-isl/open3d/blob/05ffd4ebe53bf940795b4cbaedcd6567666d64dd/3rdparty/find_dependencies.cmake#L33)):*

This should not be necessary. `cmake_minimum_required(VERSION 3.18)` in the top-level CMake file already sets all policies up to including `CMP0108` to `NEW`.
</blockquote></details>

Unfortunately, it doesn't. `OLD` is used with a warning if unset.
https://cmake.org/cmake/help/v3.18/policy/CMP0063.html?highlight=63#policy:CMP0063
___
*[3rdparty/find_dependencies.cmake, line 66 at r4](https://reviewable.io/reviews/intel-isl/open3d/3542#-Mb_VuSO5pIhuwX6pH6I:-MbcY_OC-KMm61Zs6J8e:bb6un1j) ([raw file](https://github.com/intel-isl/open3d/blob/05ffd4ebe53bf940795b4cbaedcd6567666d64dd/3rdparty/find_dependencies.cmake#L66)):*

nit: Revert for consistency with remaining code.
</blockquote></details>

Done
___
*[3rdparty/find_dependencies.cmake, line 171 at r4](https://reviewable.io/reviews/intel-isl/open3d/3542#-Mb_WfopCbfFCx36akAf:-MbcYwDM07TucysgFf-E:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/05ffd4ebe53bf940795b4cbaedcd6567666d64dd/3rdparty/find_dependencies.cmake#L171)):*

`$<$<PLATFORM_ID:Windows>:...>` generator expression is not directly needed. CMake is smart enough to use the policy only when needed, so passing the value unconditionally makes the expression look simpler.
</blockquote></details>

Done.
___
*[3rdparty/filament/filament_build.cmake, line 66 at r4](https://reviewable.io/reviews/intel-isl/open3d/3542#-Mb_XYcW6tJa1VucD0S2:-MbcZQZtDeh_qly2Gf0i:bb6un1j) ([raw file](https://github.com/intel-isl/open3d/blob/05ffd4ebe53bf940795b4cbaedcd6567666d64dd/3rdparty/filament/filament_build.cmake#L66)):*

nit: Revert this change to minimize noise
</blockquote></details>

Done
___
*[3rdparty/zlib/zlib.cmake, line 17 at r4](https://reviewable.io/reviews/intel-isl/open3d/3542#-Mb_YFFM1OTfXexzkDbd:-Mbc_1vG5tqRCRQ2m_y2:b-bi2j77) ([raw file](https://github.com/intel-isl/open3d/blob/05ffd4ebe53bf940795b4cbaedcd6567666d64dd/3rdparty/zlib/zlib.cmake#L17)):*

Is it possible to disable building their examples such that we can use the hidden version here as well?
</blockquote></details>

Possible, but messy. cmake doesn't let us control built targets in ExternalProjects directly. The work-around is to explicitly hide symbols from zlib during linking, which we need anyway for MKL and IPP. 
___
*[cpp/open3d/t/geometry/CMakeLists.txt, line 20 at r4](https://reviewable.io/reviews/intel-isl/open3d/3542#-Mb_YURZ8JnpNO_ZMWYA:-Mbca20OC-D8Ird05hHs:b-886gsx) ([raw file](https://github.com/intel-isl/open3d/blob/05ffd4ebe53bf940795b4cbaedcd6567666d64dd/cpp/open3d/t/geometry/CMakeLists.txt#L20)):*

nit: Indentation and line break are inconsistent now.
</blockquote></details>

Fixed
___
*[cpp/open3d/t/geometry/kernel/CMakeLists.txt, line 1 at r4](https://reviewable.io/reviews/intel-isl/open3d/3542#-Mb_YfG3Crf1slNw1L06:-Mbca3Vh6x5xBEKSZN4H:b9oev6c) ([raw file](https://github.com/intel-isl/open3d/blob/05ffd4ebe53bf940795b4cbaedcd6567666d64dd/cpp/open3d/t/geometry/kernel/CMakeLists.txt#L1)):*

Would it make sense to revert these changes and apply setting the visibility per function via `OPEN3D_API` in a follow-up PR?
</blockquote></details>

Yes - that's ideal. That would mean building Open3D with default hidden visibility and exposing only the declared API. Since that's a large change, we can use this method that allows us to hide specific object libraries.

stotko (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3542)* status: all files reviewed, 4 unresolved discussions (waiting on @ssheorey and @stotko)

___
*[3rdparty/find_dependencies.cmake, line 33 at r4](https://reviewable.io/reviews/intel-isl/open3d/3542#-Mb_VOUj1WbWZGY1Eg70:-MbeU1sXE8IymDFBUbI-:brob580) ([raw file](https://github.com/intel-isl/open3d/blob/05ffd4ebe53bf940795b4cbaedcd6567666d64dd/3rdparty/find_dependencies.cmake#L33)):*

Unfortunately, it doesn't. `OLD` is used with a warning if unset.
https://cmake.org/cmake/help/v3.18/policy/CMP0063.html?highlight=63#policy:CMP0063
</blockquote></details>

That is very strange. Placing
```
cmake_policy(GET CMP0063 visiblity_policy)
message(FATAL_ERROR "visiblity_policy ${visiblity_policy}")
```
right after `cmake_minimum_required(VERSION 3.18)` as well as within `find_dependencies` prints `NEW` for me (in `master` branch) which is exactly what I expected . Could you check again with a fresh build, i.e. removing all contents of the `build` directory to get rid of potentially colliding cache variables?
___
*[3rdparty/find_dependencies.cmake, line 172 at r5](https://reviewable.io/reviews/intel-isl/open3d/3542#-MbeVBiBB9hsF_PNo9C8:-MbeVBiBB9hsF_PNo9C9:b43ga60) ([raw file](https://github.com/intel-isl/open3d/blob/139041917ee93e14bf252c87dd10f0459fdc62d8/3rdparty/find_dependencies.cmake#L172)):*
> ```CMake
>     -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
>     -DCMAKE_POLICY_DEFAULT_CMP0091:STRING=NEW
>     $<$<PLATFORM_ID:Windows>:-DCMAKE_MSVC_RUNTIME_LIBRARY:STRING=${CMAKE_MSVC_RUNTIME_LIBRARY}>
> ```

I thought this will be cleaned up as well, so I have not added a comment for this line. My fault. Please also remove this conditional statement.

stotko (commented):
**Edit:** Fixed missing "not" in statement. Sorry about that.


> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3542)* status: all files reviewed, 5 unresolved discussions (waiting on @ssheorey and @stotko)

___
*[cpp/open3d/CMakeLists.txt, line 103 at r5](https://reviewable.io/reviews/intel-isl/open3d/3542#-MbehQ--0RaNVyIIpUQD:-MbehQ--0RaNVyIIpUQE:b250q4v) ([raw file](https://github.com/intel-isl/open3d/blob/139041917ee93e14bf252c87dd10f0459fdc62d8/cpp/open3d/CMakeLists.txt#L103)):*
> ```CMake
> # direct consumers of libOpen3D.a
> target_link_options(Open3D INTERFACE
>     $<$<STREQUAL:$<TARGET_PROPERTY:Open3D,TYPE>,STATIC_LIBRARY>:$<LINK_ONLY:${OPEN3D_HIDDEN_3RDPARTY_LINK_OPTIONS}>>)
> ```

After looking at this in more detail, I am wondering whether that expression might lead to potential issues in the future. CMake documents that `$<LINK_ONLY:...>` should only be used for the `INTERFACE_LINK_LIBRARIES` property (see https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#output-related-expressions). Looking at CMake's source code, it seems like they are already checking if it is used in a linking context (see https://gitlab.kitware.com/cmake/cmake/-/blob/v3.20.3/Source/cmGeneratorExpressionNode.cxx#L855), although the check seems weaker than the statement in the documentation. I am **not** sure what their plans for that generator expression are, but it is possible that it could become more restrictive to fix/prevent bugs.

If we want to hide these symbols for our consumers of `Open3D::Open3D` in our project, like `pybind`, would it be better to explicitly apply it to them instead? There are only a few places to consider, so the maintenance overhead should be quite small.
___
*[cpp/open3d/t/geometry/kernel/CMakeLists.txt, line 1 at r4](https://reviewable.io/reviews/intel-isl/open3d/3542#-Mb_YfG3Crf1slNw1L06:-MbekLFI2dUHz3H8RttX:b-t3x2g8) ([raw file](https://github.com/intel-isl/open3d/blob/05ffd4ebe53bf940795b4cbaedcd6567666d64dd/cpp/open3d/t/geometry/kernel/CMakeLists.txt#L1)):*

Yes - that's ideal. That would mean building Open3D with default hidden visibility and exposing only the declared API. Since that's a large change, we can use this method that allows us to hide specific object libraries.
</blockquote></details>

I am still wondering if we should introduce this additional layer to get a short-term benefit. It would become obsolete in the long term once the whole library properly exports its symbols. Just asking to make sure the strategy/roadmap is clear.

ssheorey (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3542)* status: 22 of 25 files reviewed, 5 unresolved discussions (waiting on @stotko and @yxlao)

___
*[3rdparty/find_dependencies.cmake, line 33 at r4](https://reviewable.io/reviews/intel-isl/open3d/3542#-Mb_VOUj1WbWZGY1Eg70:-MbhS6FJ8e0qFqWmqP52:b-nl4w0y) ([raw file](https://github.com/intel-isl/open3d/blob/05ffd4ebe53bf940795b4cbaedcd6567666d64dd/3rdparty/find_dependencies.cmake#L33)):*

That is very strange. Placing
```
cmake_policy(GET CMP0063 visiblity_policy)
message(FATAL_ERROR "visiblity_policy ${visiblity_policy}")
```
right after `cmake_minimum_required(VERSION 3.18)` as well as within `find_dependencies` prints `NEW` for me (in `master` branch) which is exactly what I expected . Could you check again with a fresh build, i.e. removing all contents of the `build` directory to get rid of potentially colliding cache variables?
</blockquote></details>

Yes, that's correct. Looks like the documentation is out of date. Policy setting removed.
___
*[3rdparty/find_dependencies.cmake, line 172 at r5](https://reviewable.io/reviews/intel-isl/open3d/3542#-MbeVBiBB9hsF_PNo9C8:-MbhRHvS772mPZzjphEs:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/139041917ee93e14bf252c87dd10f0459fdc62d8/3rdparty/find_dependencies.cmake#L172)):*

I thought this will be cleaned up as well, so I have not added a comment for this line. My fault. Please also remove this conditional statement.
</blockquote></details>

Done.
___
*[3rdparty/zlib/zlib.cmake, line 17 at r4](https://reviewable.io/reviews/intel-isl/open3d/3542#-Mb_YFFM1OTfXexzkDbd:-MbhYVvj1FrdXAFiQktm:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/05ffd4ebe53bf940795b4cbaedcd6567666d64dd/3rdparty/zlib/zlib.cmake#L17)):*

`-DCMAKE_SKIP_INSTALL_ALL_DEPENDENCY=ON` seems to do the trick. 

Something like this might work (only verified on ubuntu): 
```diff
diff --git a/3rdparty/zlib/zlib.cmake b/3rdparty/zlib/zlib.cmake
index 3412e7411..0d27e0f09 100644
--- a/3rdparty/zlib/zlib.cmake
+++ b/3rdparty/zlib/zlib.cmake
@@ -14,7 +14,9 @@ ExternalProject_Add(
     UPDATE_COMMAND ""
     CMAKE_ARGS
         -DCMAKE_INSTALL_PREFIX=<INSTALL_DIR>
-        ${ExternalProject_CMAKE_ARGS}  # Symbols not hidden due to examples built
+        -DCMAKE_SKIP_INSTALL_ALL_DEPENDENCY=ON
+        ${ExternalProject_CMAKE_ARGS_hidden}
+    BUILD_COMMAND ${CMAKE_COMMAND} --build . --config ${CMAKE_BUILD_TYPE} --target zlib zlibstatic
     BUILD_BYPRODUCTS
         <INSTALL_DIR>/lib/${CMAKE_STATIC_LIBRARY_PREFIX}${lib_name}${CMAKE_STATIC_LIBRARY_SUFFIX}
         <INSTALL_DIR>/lib/${CMAKE_STATIC_LIBRARY_PREFIX}${lib_name}d${CMAKE_STATIC_LIBRARY_SUFFIX}
```
</blockquote></details>

Done.
___
*[cpp/open3d/CMakeLists.txt, line 103 at r5](https://reviewable.io/reviews/intel-isl/open3d/3542#-MbehQ--0RaNVyIIpUQD:-MbhSKKb81RIz9-Qq9zk:b-lvhoyb) ([raw file](https://github.com/intel-isl/open3d/blob/139041917ee93e14bf252c87dd10f0459fdc62d8/cpp/open3d/CMakeLists.txt#L103)):*

After looking at this in more detail, I am wondering whether that expression might lead to potential issues in the future. CMake documents that `$<LINK_ONLY:...>` should only be used for the `INTERFACE_LINK_LIBRARIES` property (see https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#output-related-expressions). Looking at CMake's source code, it seems like they are already checking if it is used in a linking context (see https://gitlab.kitware.com/cmake/cmake/-/blob/v3.20.3/Source/cmGeneratorExpressionNode.cxx#L855), although the check seems weaker than the statement in the documentation. I am sure what their plans for that generator expression are, but it is possible that it could become more restrictive to fix/prevent bugs.

If we want to hide these symbols for our consumers of `Open3D::Open3D` in our project, like `pybind`, would it be better to explicitly apply it to them instead? There are only a few places to consider, so the maintenance overhead should be quite small.
</blockquote></details>

I wouldn't worry about this. Restricting LINK_ONLY to only allow link libraries and disallow link options would be counter-intuitive. If they do decide something like this, we will get an error and can fix it at that point. 
___
*[cpp/open3d/t/geometry/kernel/CMakeLists.txt, line 1 at r4](https://reviewable.io/reviews/intel-isl/open3d/3542#-Mb_YfG3Crf1slNw1L06:-MbhaDxr5VItds-493mf:botk5yp) ([raw file](https://github.com/intel-isl/open3d/blob/05ffd4ebe53bf940795b4cbaedcd6567666d64dd/cpp/open3d/t/geometry/kernel/CMakeLists.txt#L1)):*

I am still wondering if we should introduce this additional layer to get a short-term benefit. It would become obsolete in the long term once the whole library properly exports its symbols. Just asking to make sure the strategy/roadmap is clear.
</blockquote></details>

I think it's OK to make incremental progress in this case. Actually, both methods:
- fine grained symbol level visibility with `OPEN3D_API`
- object or file level visibility with CMake `open3d_set_open3d_lib_properties(<name> HIDDEN)`
can be used at the same time. The 2nd method is more convenient since most of our code is separated at a folder level into internal and public functions. Also this leaves the actual source code a bit cleaner.

Yixing can comment on a longer term strategy.

yxlao (approved):

stotko (approved):
> Reviewed 3 of 3 files at r6.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3542)* status: all files reviewed, 1 unresolved discussion (waiting on @yxlao)

___
*[3rdparty/zlib/zlib.cmake, line 17 at r4](https://reviewable.io/reviews/intel-isl/open3d/3542#-Mb_YFFM1OTfXexzkDbd:-MbjcrcNCwXXo6ajbkWe:b-72aewx) ([raw file](https://github.com/intel-isl/open3d/blob/05ffd4ebe53bf940795b4cbaedcd6567666d64dd/3rdparty/zlib/zlib.cmake#L17)):*

Done.
</blockquote></details>

There is one remaining build error in Ubuntu GPU CI. Otherwise, LGTM.

ssheorey (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3542)* status: all files reviewed, 1 unresolved discussion (waiting on @stotko)

___
*[3rdparty/zlib/zlib.cmake, line 17 at r4](https://reviewable.io/reviews/intel-isl/open3d/3542#-Mb_YFFM1OTfXexzkDbd:-MbljE2n9M1uCfeEb0tQ:b-ick1rv) ([raw file](https://github.com/intel-isl/open3d/blob/05ffd4ebe53bf940795b4cbaedcd6567666d64dd/3rdparty/zlib/zlib.cmake#L17)):*

There is one remaining build error in Ubuntu GPU CI. Otherwise, LGTM.
</blockquote></details>

Yes the error is:
```
/usr/bin/cc -D_LARGEFILE64_SOURCE=1 -I/root/Open3D/build/zlib/src/ext_zlib-build -I/root/Open3D/build/zlib/src/ext_zlib -O3 -DNDEBUG -fPIC -MD -MT CMakeFiles/zlibstatic.dir/adler32.o -MF CMakeFiles/zlibstatic.dir/adler32.o.d -o CMakeFiles/zlibstatic.dir/adler32.o -c /root/Open3D/build/zlib/src/ext_zlib/adler32.c
make  -f CMakeFiles/jpeg-static.dir/build.make CMakeFiles/jpeg-static.dir/depend
as: symbol lookup error: as: undefined symbol: deflate
```
I can't duplicate this on my CUDA laptop (or in Github CI).

I'm going to revert this change. We can hide the `zlib` symbols during linking instead as before.

ssheorey (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3542)* status: 23 of 25 files reviewed, 1 unresolved discussion (waiting on @stotko)

___
*[3rdparty/find_dependencies.cmake, line 33 at r4](https://reviewable.io/reviews/intel-isl/open3d/3542#-Mb_VOUj1WbWZGY1Eg70:-MbrfKXt6PghPiObuquj:bxqu846) ([raw file](https://github.com/intel-isl/open3d/blob/05ffd4ebe53bf940795b4cbaedcd6567666d64dd/3rdparty/find_dependencies.cmake#L33)):*

Yes, that's correct. Looks like the documentation is out of date. Policy setting removed.
</blockquote></details>

Actually, this doesn't work, cmake complains about the policy and the symbols are exported. Reverting.
https://github.com/intel-isl/Open3D/pull/3542/checks?check_run_id=2787133535#step:8:13664

I think the default policy settings are determined by the `cmake_minimum_version` for each project.

PR #3567, opened 2021-06-10
Real-time Reconstruction App

theNded:
- Move reconstruction system to `cpp/apps`
- Add a detailed README

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,6 +34,7 @@ option(BUILD_CUDA_MODULE          "Build the CUDA module"                    OFF
 option(BUILD_COMMON_CUDA_ARCHS    "Build for common CUDA GPUs (for release)" OFF)
 option(BUILD_CACHED_CUDA_MANAGER  "Build the cached CUDA memory manager"     ON )
 option(BUILD_GUI                  "Builds new GUI"                           ON )
+option(BUILD_SLAM_APP             "Builds RGBD Dense SLAM GUI"               ON )
 option(WITH_OPENMP                "Use OpenMP multi-threading"               ON )
 option(WITH_IPPICV                "Use Intel Performance Primitives"         ON )
 option(ENABLE_HEADLESS_RENDERING  "Use OSMesa for headless rendering"        OFF)
--- a/cpp/apps/CMakeLists.txt
+++ b/cpp/apps/CMakeLists.txt
@@ -5,7 +5,7 @@ macro(open3d_add_app SRC_DIR APP_NAME TARGET_NAME)
     file(GLOB SOURCE_FILES "${SOURCE_DIR}/*.cpp")
     file(GLOB HEADER_FILES "${SOURCE_DIR}/*.h")
 
-    if (APPLE)
+    if (APPLE AND IS_DESKTOP_APP)
         file(GLOB OBJC_FILES "${SOURCE_DIR}/*.mm")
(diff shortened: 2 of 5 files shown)


yxlao (changes requested):
> Reviewed 4 of 5 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3567)* status: all files reviewed, 3 unresolved discussions (waiting on @stotko and @theNded)

___
*[CMakeLists.txt, line 37 at r2](https://reviewable.io/reviews/intel-isl/open3d/3567#-MbtTI50440lCF-K6PoR:-MbtTI50440lCF-K6PoS:b-jnivj5) ([raw file](https://github.com/intel-isl/open3d/blob/5aa9c92e5e9026e6de4f6e9812d773aed4cc32bf/CMakeLists.txt#L37)):*
> ```CMake
> option(BUILD_CACHED_CUDA_MANAGER  "Build the cached CUDA memory manager"     ON )
> option(BUILD_GUI                  "Builds new GUI"                           ON )
> option(BUILD_SLAM_APP             "Builds RGBD Dense SLAM GUI"               ON )
> ```

I would always build it as long as `BUILD_GUI=ON` since the slam app does not require additional dependencies (i.e. we could consider removing the `BUILD_SLAM_APP` option).
___
*[cpp/apps/CMakeLists.txt, line 122 at r2](https://reviewable.io/reviews/intel-isl/open3d/3567#-MbtaJPVCVWpWsjbivEt:-MbtaJPVCVWpWsjbivEu:bblai1a) ([raw file](https://github.com/intel-isl/open3d/blob/5aa9c92e5e9026e6de4f6e9812d773aed4cc32bf/cpp/apps/CMakeLists.txt#L122)):*
> ```CMake
> 
> 
> macro(open3d_add_demo_app SRC_DIR APP_NAME)
> ```

I believe German mentioned it at one I am not sure about the motivation of moving the example. It seems to me that keeping this as an "example" is as convenient as moving this to an "app" for the users. In both cases, the key is to provide sufficient documentation such that the users can find it easily.

Plus, we may have more reconstruction examples in the future. When we shall keep them as "example" and when shall we move them to an "app"?

We can discuss this in the meetings.
___
*[cpp/apps/Open3DDenseSLAM/README.md, line 1 at r2](https://reviewable.io/reviews/intel-isl/open3d/3567#-MbtU5f_A-iFC2I0md0l:-MbtU5f_A-iFC2I0md0m:b-6lstmh) ([raw file](https://github.com/intel-isl/open3d/blob/5aa9c92e5e9026e6de4f6e9812d773aed4cc32bf/cpp/apps/Open3DDenseSLAM/README.md#L1)):*
> ```Markdown
> # Open3D Dense SLAM
> ```

I would also expose this documentation to Sphinx such that it shows up in the documentaiton website. We can: 

1. add a new `.rst`file, e.g. in `docs/tutorial/pipelines/dense_slam.rst`
2. in `dense_slam.rst`, include this markdown file with `mdinclude`, similar to https://github.com/intel-isl/Open3D/blob/master/docs/introduction.rst . Hopefully Sphinx can convert the markdown format correctly for us.
3. update `docs/tutorial/pipelines/index.rst`to add an entry for `dense_slam`. You may also look at `docs/tutorial/reconstruction_system/index.rst` as an example.

PR #3576, opened 2021-06-11
Test for -load_hidden flag

ssheorey:
Prevent build failures on older AppleClang compilers where the linker flag is absent.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3576)


ShubhamAgarwal12 (approved):
This works on my Mac.

PR #3586, opened 2021-06-15
Cleanup ssh key individually for each GCE run when deleting VM

ssheorey:
Remove deleting *all* ssh keys that may cause GCE VM login errors in other runs.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3586)


yxlao (approved):

ssheorey:
The CI may fail for delete ssh key (delete VM step) since currently running master / other branch CI runs delete **all** ssh keys. This runs successfully if another CI run doesn't delete its ssh keys.

yxlao:
> The CI may fail for delete ssh key (delete VM step) since currently running master / other branch CI runs delete **all** ssh keys. This runs successfully if another CI run doesn't delete its ssh keys.

Thanks for confirming. We can merge this PR for now and all new PRs shall merge/rebase master.

PR #3620, opened 2021-06-21
Remove Flann dependency

stotko:
Moving to C++17 in the future is currently blocked by third-party dependencies that rely on features that are removed in C++17. In particular, this affects `Flann` which makes use of `std::random_shuffle`.

- Replace `Flann` by `nanoflann` in legacy `geometry::KDTreeFlann` class
- Upgrade `nanoflann` to 1.3.2 (required for col-major data access)
- Port `nanoflann` submodule to `ExternalProject_Add` URL downloading approach to reduce dependency to git submodules
- Remove `Flann` from `3rdparty`
- Remove examples that directly use `Flann` as well as our legacy API

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3620)


yxlao:
`nanaflann` does not support native hybrid search, so we have to do the hybrid search in two steps. This could result in some performance penalties, e.g. for the (non-tensor) ICP pipeline that uses hybrid search. 

Benchmark command:
```bash
make benchmarks -j && OMP_NUM_THREADS=18 ./bin/benchmarks --benchmark_filter=".*ICPLegacy.*PointToPoint.*"
```

On `master`:
```bash
--------------------------------------------------------------------------------------------
Benchmark                                                  Time             CPU   Iterations
--------------------------------------------------------------------------------------------
BenchmarkRegistrationICPLegacy/PointToPoint / CPU       1.38 ms         1.38 ms          456
```

On `stotko/remove-flann`:
```bash
--------------------------------------------------------------------------------------------
Benchmark                                                  Time             CPU   Iterations
--------------------------------------------------------------------------------------------
BenchmarkRegistrationICPLegacy/PointToPoint / CPU       2.54 ms         2.54 ms          272
```
It is now `1.84x` slower than the original implementation.

- If we use `flann-lib/flann`, we'll have to patch it such that it supports C++17 and we'll also need to refactor `core::nns`. 
- If we proceed with `nanaoflann`, we shall be aware of the performance impact. This may be what we want to do as `nanoflann` is also used in `core::nns`.

stotko:
I switched to using KNN search instead to do hybrid search. Here are some results on my machine (Intel i7-10750H @ 6 Cores  / 12 Threads with `OMP_NUM_THREADS=6`):

On `master`:

```
BenchmarkRegistrationICPLegacy/PointToPoint / CPU       2.01 ms         2.01 ms          333
```

On `stotko/remove-flann` with Radius Search:

```
BenchmarkRegistrationICPLegacy/PointToPoint / CPU       5.05 ms         5.05 ms          137
```

On `stotko/remove-flann` with KNN Search:
```
BenchmarkRegistrationICPLegacy/PointToPoint / CPU       1.78 ms         1.78 ms          401
```

I would suggest to proceed with `nanoflann` and adjust `core::nns` if necessary. Dropping `Flann` now could reveal further issues with the new API. So we get a chance to fix them earlier.

yxlao (approved):

yxlao (approved):

PR #3656, opened 2021-06-28
Remove boost dependency

stotko:
`boost` is only used to detect endianness and provide a respective preprocessor definition. The actual use case in Open3D is to conditionally use a specific string for message types in the RPC interface.

- Implement run-time endianness check based on `memcpy`. Other approaches like `union` are avoided due to undefined behaviour. This will not introduce a performance overhead as the compiler can optimize all code to a single constant.
- Replace preprocessor definition `ENDIANNESS_STR` with proper C++ function `EndiannessStr()`.
- Remove `boost` dependency to get rid of costly recursive `git clone` operation.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3656)


ssheorey (commented):
This is good, I was wondering what we were using boost for :-)

> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3656#-:-MdJSrOx6NxoqI6Xz3iW:b-ndogz6)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @stotko and @yxlao)

___
*[cpp/open3d/io/rpc/Messages.h, line 51 at r1](https://reviewable.io/reviews/intel-isl/open3d/3656#-MdJRS7a8hffLWOTAs66:-MdJRS7b1XyethqZpEqW:bbzgk9m) ([raw file](https://github.com/intel-isl/open3d/blob/f580d5d7e5d080cfc4f91cfdc806c3793b6b18dc/cpp/open3d/io/rpc/Messages.h#L51)):*
> ```objc
> 
>     return IsLittleEndian() ? "<" : ">";
> }
> ```

Can you comment on how the lambda `IsLittleEndian()` ensures that this is executed only one?
If this is not guaranteed, can we use a `static constexpr char` variable to store ENDIANNESS_STR so that it is initialized only once?

stotko (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3656)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @ssheorey and @yxlao)

___
*[cpp/open3d/io/rpc/Messages.h, line 51 at r1](https://reviewable.io/reviews/intel-isl/open3d/3656#-MdJRS7a8hffLWOTAs66:-MdL_C2zBVfJfGq6TSJK:b-87i84d) ([raw file](https://github.com/intel-isl/open3d/blob/f580d5d7e5d080cfc4f91cfdc806c3793b6b18dc/cpp/open3d/io/rpc/Messages.h#L51)):*

Can you comment on how the lambda IsLittleEndian()` ensures that this is executed only one?
If this is not guaranteed, can we use a `static constexpr char` variable to store ENDIANNESS_STR so that it is initialized only once?
</blockquote></details>

It is not guaranteed that `IsLittleEndian()` is only called once and, unfortunately, it is also not possible to make it `constexpr` as `memcpy` is not marked `constexpr`. However, the expression inside the lambda function is very simple such that the compiler knows the value of `a` and can infer the value of `b` from `memcpy` based on that information. As a result, it can optimize the `IsLittleEndian()` call to a constant and, consequently, do the same with `EndiannessStr()`.

I have primarily defined the lambda `IsLittleEndian()` to improve readability and to make it clear why we need the magic `memcpy` and the constants in this context. Furthermore, a lambda function makes `IsLittleEndian()` an implementation detail of `EndiannessStr()`, but we can make it a proper function in the future if we want to use it in different places.

yxlao (approved):

yxlao:
I tested with 
```cpp
#include <cstdint>
#include <cstring>

int main() {
    auto IsLittleEndian = []() -> bool {
        uint32_t a = 1;
        uint8_t b;
        /// Use memcpy as a reliable way to access a single byte.
        /// Other approaches, e.g. union, often rely on undefined behaviour.
        std::memcpy(&b, &a, sizeof(uint8_t));
        return b == 1;
    };
    return int(IsLittleEndian());
}
```
Under `Release` build, `gcc` on x86 indeed performs the optimization:
```s
	.file	"endian.cpp"
	.text
	.section	.text.startup,"ax",@progbits
	.p2align 4,,15
	.globl	main
	.type	main, @function
main:
.LFB27:
	.cfi_startproc
	movl	$1, %eax
	ret
	.cfi_endproc
.LFE27:
	.size	main, .-main
	.ident	"GCC: (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0"
	.section	.note.GNU-stack,"",@progbits

```

cpp/open3d/io/rpc/Messages.h
```diff
@@ -25,26 +25,31 @@
...
 namespace open3d {
 namespace io {
 namespace rpc {
 namespace messages {
 
+inline std::string EndiannessStr() {
+    auto IsLittleEndian = []() -> bool {
+        uint32_t a = 1;
+        uint8_t b;
+        /// Use memcpy as a reliable way to access a single byte.
```
yxlao, line 44: nit: let's use `//` here as it is not a docstring

stotko (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3656)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @ssheorey)

___
*[cpp/open3d/io/rpc/Messages.h, line 44 at r1](https://reviewable.io/reviews/intel-isl/open3d/3656#-MdI1AWs657LGCN9EE6G-r1-44:-MdM_5zLAmZCX4lsTmU_:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/f580d5d7e5d080cfc4f91cfdc806c3793b6b18dc/cpp/open3d/io/rpc/Messages.h#L44)):*

nit: let's use `//` here as it is not a docstring
</blockquote></details>

Done.

ssheorey (approved):
I can confirm @yxlao 's observation on all of our compilers:
https://godbolt.org/z/d59854cd1

> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3656#-:-MdNhn9tAzIur_hK__93:b-d87nur)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @ssheorey)

PR #3666, opened 2021-06-29
Return failure exit code on leaks at program end

stotko:
Summary:
- Add `HasLeaks()` function to memory statistics to query current leak status. This simplifies some internal code.
- Add `Reset()` function to discard all recorded statistics. Used to suppress leak warning for objects managed by `pybind11`.
- Filter pointers and sizes to automatically account for previous calls to `Reset()`.
- Print list of leaking pointers to provide more information about their distribution.
- Fix underflow in balance computation by changing `size_t` to `int64_t` .

Manually checked exit code of C++ and Python unit tests with an intentional leak. Works as expected. CI must all pass as Open3D should be leak-free as of this PR.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3666)

### DIFF
--- a/cpp/open3d/core/MemoryManager.cpp
+++ b/cpp/open3d/core/MemoryManager.cpp
@@ -40,14 +40,15 @@ namespace core {
 
 void* MemoryManager::Malloc(size_t byte_size, const Device& device) {
     void* ptr = GetDeviceMemoryManager(device)->Malloc(byte_size, device);
-    MemoryManagerStatistic::GetInstance().IncrementCountMalloc(ptr, byte_size,
-                                                               device);
+    MemoryManagerStatistic::GetInstance().CountMalloc(ptr, byte_size, device);
     return ptr;
 }
 
 void MemoryManager::Free(void* ptr, const Device& device) {
+    // Update statistics before freeing the memory. This ensures a consistent
+    // order in case a subsequent Malloc requires the currently freed memory.
+    MemoryManagerStatistic::GetInstance().CountFree(ptr, device);
     GetDeviceMemoryManager(device)->Free(ptr, device);
-    MemoryManagerStatistic::GetInstance().IncrementCountFree(ptr, device);
 }
 
 void MemoryManager::Memcpy(void* dst_ptr,
--- a/cpp/open3d/core/MemoryManagerStatistic.cpp
+++ b/cpp/open3d/core/MemoryManagerStatistic.cpp
@@ -27,6 +27,7 @@
 #include "open3d/core/MemoryManagerStatistic.h"
 
 #include <algorithm>
+#include <cstdlib>
 #include <numeric>
 
 #include "open3d/utility/Logging.h"
@@ -52,6 +53,12 @@ MemoryManagerStatistic::~MemoryManagerStatistic() {
         // at this point in time.
         utility::Logger::GetInstance().ResetPrintFunction();
         Print();
+
+        // Indicate failure if possible leaks have been detected.
+        // This is useful to automatically let unit tests fail.
+        if (HasLeaks()) {
+            std::exit(EXIT_FAILURE);
+        }
     }
 }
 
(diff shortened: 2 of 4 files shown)


ssheorey (changes requested):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3666)* status: 0 of 4 files reviewed, 3 unresolved discussions (waiting on @stotko and @yxlao)

___
*[cpp/open3d/core/MemoryManagerStatistic.h, line 70 at r1](https://reviewable.io/reviews/intel-isl/open3d/3666#-MdNkpAa-uAHBxRxyOqU:-MdNkpAa-uAHBxRxyOqV:b-q9avpj) ([raw file](https://github.com/intel-isl/open3d/blob/683587e4658b196c13bb2c0eae3abe3186f5aea4/cpp/open3d/core/MemoryManagerStatistic.h#L70)):*
> ```objc
> 
>     void Reset();
> ```

Add doxygen docs to functions here.
___
*[cpp/open3d/core/MemoryManagerStatistic.h, line 87 at r1](https://reviewable.io/reviews/intel-isl/open3d/3666#-MdNlBNLBGsaIujpXebV:-MdNlBNLBGsaIujpXebW:b-5azcg8) ([raw file](https://github.com/intel-isl/open3d/blob/683587e4658b196c13bb2c0eae3abe3186f5aea4/cpp/open3d/core/MemoryManagerStatistic.h#L87)):*
> ```objc
>     // Print at the end of the program be default. If leaks are detected,
>     // the exit code will be overriden with EXIT_FAILURE.
>     bool print_at_program_end_ = true;
> ```

The default setting should be false, since this may be annoying to users who don't want to bother with fixing their memory leaks at that moment. For our tests, we can explicitly set it to true.

Also, can we simplify print settings from `level_` and `print_at_program_end_` to a single setting?
Perhaps `print_at_program_end_ = level_==PrintLevel::Unbalanced` ?
Another option to simplify this would be to use the Open3D verbosity level.
___
*[cpp/open3d/core/MemoryManagerStatistic.cpp, line 110 at r1](https://reviewable.io/reviews/intel-isl/open3d/3666#-MdNr_g7CsRCNN3ekJBZ:-MdNr_g7CsRCNN3ekJB_:bqvnbhh) ([raw file](https://github.com/intel-isl/open3d/blob/683587e4658b196c13bb2c0eae3abe3186f5aea4/cpp/open3d/core/MemoryManagerStatistic.cpp#L110)):*
> ```cpp
>                 utility::LogWarning("    {} @ {} bytes", fmt::ptr(leak.first),
>                                     leak.second);
>             }
> ```

If Open3D verbosity level is set to WARNING, then the header / footer will not be printed. Can we construct a string and print it all at once with a single logging statement?

yxlao (approved):
LGTM after addressing Sameer's comments.

> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3666#-:-MdP_Tpk4eobas_YinLu:b-cvtu0w)* status: all files reviewed, 3 unresolved discussions (waiting on @ssheorey and @stotko)

___
*[cpp/open3d/core/MemoryManagerStatistic.h, line 87 at r1](https://reviewable.io/reviews/intel-isl/open3d/3666#-MdNlBNLBGsaIujpXebV:-MdPTOBr0sBFbX4iHDxA:b-u9ib5s) ([raw file](https://github.com/intel-isl/open3d/blob/683587e4658b196c13bb2c0eae3abe3186f5aea4/cpp/open3d/core/MemoryManagerStatistic.h#L87)):*

The default setting should be false, since this may be annoying to users who don't want to bother with fixing their memory leaks at that moment. For our tests, we can explicitly set it to true.

Also, can we simplify print settings from `level_` and `print_at_program_end_` to a single setting?
Perhaps `print_at_program_end_ = level_==PrintLevel::Unbalanced` ?
Another option to simplify this would be to use the Open3D verbosity level.
</blockquote></details>

Perhaps printing memory leak info is fine. Typically the "users" the memory manager are the Open3D internal classes like the tensor and geometry classes. End-users of Open3D typically don't use the memory manager directly. Additionally, the print doesn't block the program execution.

stotko (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3666)* status: 1 of 4 files reviewed, 3 unresolved discussions (waiting on @ssheorey and @yxlao)

___
*[cpp/open3d/core/MemoryManagerStatistic.h, line 70 at r1](https://reviewable.io/reviews/intel-isl/open3d/3666#-MdNkpAa-uAHBxRxyOqU:-MdQtpq-EP6ZiZwg2qdX:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/683587e4658b196c13bb2c0eae3abe3186f5aea4/cpp/open3d/core/MemoryManagerStatistic.h#L70)):*

Add doxygen docs to functions here.
</blockquote></details>

Done.
___
*[cpp/open3d/core/MemoryManagerStatistic.h, line 87 at r1](https://reviewable.io/reviews/intel-isl/open3d/3666#-MdNlBNLBGsaIujpXebV:-MdQtrMn1fydqYTBAp4Z:baejg0c) ([raw file](https://github.com/intel-isl/open3d/blob/683587e4658b196c13bb2c0eae3abe3186f5aea4/cpp/open3d/core/MemoryManagerStatistic.h#L87)):*

Perhaps printing memory leak info is fine. Typically the "users" the memory manager are the Open3D internal classes like the tensor and geometry classes. End-users of Open3D typically don't use the memory manager directly. Additionally, the print doesn't block the program execution.
</blockquote></details>

I think printing at the end by default is the better option. Memory leaks are quite hard to find without any hints, so users will now immediately get an idea when something goes wrong. This will only happen if they actually allocate memory themselves via the memory manager and forget to free it. Any other leak that happens in Open3D is likely a bug on our side and reporting a warning by default is better to get it fixed sooner.

If that option/default is not suitable, users can still disable the check by calling `SetPrintAtProgramEnd(false)` somewhere in their program.
___
*[cpp/open3d/core/MemoryManagerStatistic.cpp, line 110 at r1](https://reviewable.io/reviews/intel-isl/open3d/3666#-MdNr_g7CsRCNN3ekJBZ:-MdQvzhw0ccvLIdBKKtb:bxsij8l) ([raw file](https://github.com/intel-isl/open3d/blob/683587e4658b196c13bb2c0eae3abe3186f5aea4/cpp/open3d/core/MemoryManagerStatistic.cpp#L110)):*

If Open3D verbosity level is set to WARNING, then the header / footer will not be printed. Can we construct a string and print it all at once with a single logging statement?
</blockquote></details>

Done. I actually went with temporally overriding the verbosity level. This allows to keep highlighting unbalanced allocations (via `LogWarning`) whereas all other information such as the header, footer and balanced allocations (for `PrintLevel::All`) are printed normally (via `LogInfo`).

ssheorey (approved):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3666)* status: 1 of 4 files reviewed, all discussions resolved (waiting on @yxlao)

PR #3691, opened 2021-07-06
EstimateNormals for TPointCloud

reyanshsolis:
```
------------------------------------------------------------------------------------------------
Benchmark                                                      Time             CPU   Iterations
------------------------------------------------------------------------------------------------
EstimateNormals/CPU F32 Hybrid[0.02 | 30 | 0.06]            22.1 ms         21.7 ms           33
EstimateNormals/CPU F64 Hybrid[0.02 | 30 | 0.06]            22.0 ms         21.6 ms           31
EstimateNormals/CPU F32 KNN[0.02 | 30]                      59.6 ms         59.5 ms           11
EstimateNormals/CPU F64 KNN[0.02 | 30]                      51.2 ms         51.2 ms           13
EstimateNormals/CUDA F32 Hybrid[0.02 | 30 | 0.06]           17.4 ms         17.4 ms           39
EstimateNormals/CUDA F64 Hybrid[0.02 | 30 | 0.06]           20.0 ms         20.0 ms           35
EstimateNormals/CUDA F32 KNN[0.02 | 30]                      280 ms          280 ms            2
EstimateNormals/CUDA F64 KNN[0.02 | 30]                      396 ms          396 ms            2
LegacyEstimateNormals/Legacy Hybrid[0.02 | 30 | 0.06]       29.8 ms         29.8 ms           24
LegacyEstimateNormals/Legacy KNN[0.02 | 30]                 30.9 ms         30.8 ms           24

```

```
pcd.estimate_normals() -> uses KNN search with max_nn = 30.
pcd.estimate_normals(20) -> uses KNN search with max_nn = 20.
pcd.estimate_normals(30, 0.04) -> uses Hybrid search with max_nn = 30, search_radius = 0.04.
```
-------
### Open3D Legacy v/s Tensor CPU v/s Meshlab
Search Param = KNN(30) | Points | Open3D (ms) Legacy | Open3D (ms) Tensor CPU | Meshlab (ms)
-- | -- | -- | -- | --
apartment.ply | 22,763,360 | 60431 | 72062 | 145000
bedroom.ply | 16,164,488 | 44725 | 51787 | 111057
boardroom.ply | 20,061,457 | 55106 | 63674 | 140273
loft.ply | 22,910,305 | 63642 | 74094 | 153144
------
### Performance breakdown
PointCloud : bedroom.ply downsampled by 0.02 voxel-size, 507943 points.

#### GTX 1050Ti m 4GB | i7-7700 4-core

bedroom.ply [507943 points] | CPU F32 | CPU F32 | CUDA F32 | CUDA F64
-- | -- | -- | -- | --
-- NNS Index Setup | 140 | 147 | 0.05 | 0.07
-- NNS Search | 1120 | 1292 | 9579 | 46656
-- Covar. Compute | 52 | 64 | 0.072 | 0.057
Total Covar. Estimation | 1318 | 1509 | 9627 | 46724
Estimating Normals from Covariances | 18 | 20 | 0.98 | 7.45
overheads | - | - | - | -
Total | 1336 | 1530 | 9628 | 46732

#### RTX 3070 m 8GB | i7 6-core

bedroom.ply[507943 points] | CPU F32 | CPU F32 | CUDA F32 | CUDA F64
-- | -- | -- | -- | --
-- NNS Index Setup | 97.4216 | 100.201 | 0.0212402 | 0.00830078
-- NNS Search | 532.207 | 593.605 | 1606.69 | 15162.5
-- Covar. Compute | 19.4456 | 28.3066 | 0.0566406 | 0.0568848
Total Covar. Estimation | 658.33 | 729.218 | 1611.09 | 15170.9
Estimating Normals from Covariances | 11.3926 | 13.7703 | 0.02 | 2.45581
overheads | - | - | - | -
Total | 669.768 | 744.48 | 1611.29 | 15173.4

------

Python Example:
image

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/3691)


ssheorey (changes requested):
OpenBLAS failing test:

`[  FAILED  ] PointCloud/PointCloudPermuteDevices.EstimateNormalsAndCovariances/0, where GetParam() = 8-byte object <00-00 00-00 00-00 00-00>`

General observation: We are writing a lot of low level code here that raises the bar for future contributions and maintenance of Open3D. Can we substitute this with library functions from Eigen / BLAS / MKL / cuBLAS. Those will almost certainly be faster than our code as well.

Will add more review comments later...

> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3691#-:-Me0rE55DjSTda8Pq30f:b-yjvde8)* status: 0 of 12 files reviewed, 4 unresolved discussions (waiting on @reyanshsolis, @stotko, and @yxlao)

___
*[cpp/open3d/t/geometry/PointCloud.h, line 333 at r2](https://reviewable.io/reviews/intel-isl/open3d/3691#-Me5ynz50ep05Mo4XFM5:-Me5ynz61sQL65_qz-1w:b-9u0uw9) ([raw file](https://github.com/intel-isl/open3d/blob/f6c999b19037d3917e8bd1b65e2ec4e0ac33042b/cpp/open3d/t/geometry/PointCloud.h#L333)):*
> ```objc
>     /// [Recommended ~1.4x voxel size].
>     /// \param max_nn NeighbourSearch max neighbours parameter [Default = 30].
>     void EstimateCovariances(const double radius, const int max_nn = 30);
> ```

This functionality looks like an intermediate step in calculating normals. This is not exposed to the user in the legacy API. Can we keep it internal / hidden in the new API as well?
___
*[cpp/open3d/t/geometry/kernel/PointCloud.cpp, line 113 at r2](https://reviewable.io/reviews/intel-isl/open3d/3691#-Me6-l8m2g1lUm2ZoesU:-Me6-l8m2g1lUm2ZoesV:b-g5e66z) ([raw file](https://github.com/intel-isl/open3d/blob/f6c999b19037d3917e8bd1b65e2ec4e0ac33042b/cpp/open3d/t/geometry/kernel/PointCloud.cpp#L113)):*
> ```cpp
>                             core::Tensor& color_gradient,
>                             const double& radius,
>                             const int64_t& max_nn) {
> ```

Why do we need this additional function call (`kernel::Estimate*`)? Can we include the body of this function directly into the user level function `Estimate*`? kernel functions should be directly CPU / CUDA.
___
*[cpp/open3d/t/geometry/kernel/PointCloud.cpp, line 128 at r2](https://reviewable.io/reviews/intel-isl/open3d/3691#-Me60VTQERBd4CdYTQhy:-Me60VTQERBd4CdYTQhz:brhumzt) ([raw file](https://github.com/intel-isl/open3d/blob/f6c999b19037d3917e8bd1b65e2ec4e0ac33042b/cpp/open3d/t/geometry/kernel/PointCloud.cpp#L128)):*
> ```cpp
> }
> 
> void EstimateCovariances(const core::Tensor& points,
> ```

Same here.
___
*[cpp/open3d/t/geometry/kernel/PointCloud.cpp, line 144 at r2](https://reviewable.io/reviews/intel-isl/open3d/3691#-Me60W_N0Rcqf92ktwnl:-Me60W_N0Rcqf92ktwnm:b68d49c) ([raw file](https://github.com/intel-isl/open3d/blob/f6c999b19037d3917e8bd1b65e2ec4e0ac33042b/cpp/open3d/t/geometry/kernel/PointCloud.cpp#L144)):*
> ```cpp
> }
> 
> void EstimateNormals(const core::Tensor& covariances,
> ```

..and here.

reyanshsolis (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3691)* status: 0 of 12 files reviewed, 4 unresolved discussions (waiting on @reyanshsolis, @ssheorey, @stotko, and @yxlao)

___
*[cpp/open3d/t/geometry/PointCloud.h, line 333 at r2](https://reviewable.io/reviews/intel-isl/open3d/3691#-Me5ynz50ep05Mo4XFM5:-MeAgbp39b91pz0TqzE1:b-vth61c) ([raw file](https://github.com/intel-isl/open3d/blob/46378ba958eeecabe8163fe8ad300ba4d59f4c7c/cpp/open3d/t/geometry/PointCloud.h#L333)):*

This functionality looks like an intermediate step in calculating normals. This is not exposed to the user in the legacy API. Can we keep it internal / hidden in the new API as well?
</blockquote></details>

This functionality will be exposed in LegacyAPI too, with GICP PR 3181. This will also be useful when GICP is implemented for t::pipelines.
___
*[cpp/open3d/t/geometry/kernel/PointCloud.cpp, line 113 at r2](https://reviewable.io/reviews/intel-isl/open3d/3691#-Me6-l8m2g1lUm2ZoesU:-MeAh3vgCtXpa0qCSZyx:b8zlbov) ([raw file](https://github.com/intel-isl/open3d/blob/46378ba958eeecabe8163fe8ad300ba4d59f4c7c/cpp/open3d/t/geometry/kernel/PointCloud.cpp#L113)):*

Why do we need this additional function call (`kernel::Estimate*`)? Can we include the body of this function directly into the user level function `Estimate*`? kernel functions should be directly CPU / CUDA.
</blockquote></details>

I have kept it for consistency. Generally, Exposed API function `X::foo` calls a kernel function `X::kernel::foo` which then pre-process the data if required and calls the function `fooCPU` and `fooCUDA` which are only defined in a file `kernel/XImpl.h` (which is not included in X.h`.  In order to expose `fooCPU` and `fooCUDA` to `X.cpp` we will have to include `XImpl.h` which is avoided in the current design generally. 
We may discuss this, in  Monday's meeting?

reyanshsolis (commented):
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3691)* status: 0 of 13 files reviewed, 4 unresolved discussions (waiting on @ssheorey, @stotko, and @yxlao)

___
*[cpp/open3d/t/geometry/PointCloud.h, line 333 at r2](https://reviewable.io/reviews/intel-isl/open3d/3691#-Me5ynz50ep05Mo4XFM5:-Mf2fmeWEUbMR-pa9mON:b2pwc79) ([raw file](https://github.com/intel-isl/open3d/blob/46378ba958eeecabe8163fe8ad300ba4d59f4c7c/cpp/open3d/t/geometry/PointCloud.h#L333)):*

This functionality will be exposed in LegacyAPI too, with GICP PR 3181. This will also be useful when GICP is implemented for t::pipelines.
</blockquote></details>

Hidden the functionality for now. Might expose in future if required.
___
*[cpp/open3d/t/geometry/kernel/PointCloud.cpp, line 113 at r2](https://reviewable.io/reviews/intel-isl/open3d/3691#-Me6-l8m2g1lUm2ZoesU:-Mf2fsG18_jtxQb7-SbJ:bpz1krn) ([raw file](https://github.com/intel-isl/open3d/blob/46378ba958eeecabe8163fe8ad300ba4d59f4c7c/cpp/open3d/t/geometry/kernel/PointCloud.cpp#L113)):*

I have kept it for consistency. Generally, Exposed API function `X::foo` calls a kernel function `X::kernel::foo` which then pre-process the data if required and calls the function `fooCPU` and `fooCUDA` which are only defined in a file `kernel/XImpl.h` (which is not included in X.h`.  In order to expose `fooCPU` and `fooCUDA` to `X.cpp` we will have to include `XImpl.h` which is avoided in the current design generally. 
We may discuss this, in  Monday's meeting?
</blockquote></details>

Removed the kernel wrapping functions. Directly calling `CPU` and `CUDA` functions.
___
*[cpp/open3d/t/geometry/kernel/PointCloud.cpp, line 128 at r2](https://reviewable.io/reviews/intel-isl/open3d/3691#-Me60VTQERBd4CdYTQhy:-Mf2fzSu0StFCskcjqGo:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/46378ba958eeecabe8163fe8ad300ba4d59f4c7c/cpp/open3d/t/geometry/kernel/PointCloud.cpp#L128)):*

Same here.
</blockquote></details>

Done.
___
*[cpp/open3d/t/geometry/kernel/PointCloud.cpp, line 144 at r2](https://reviewable.io/reviews/intel-isl/open3d/3691#-Me60W_N0Rcqf92ktwnl:-Mf2fztg3ai2Zqr8D4x_:b-896fix) ([raw file](https://github.com/intel-isl/open3d/blob/46378ba958eeecabe8163fe8ad300ba4d59f4c7c/cpp/open3d/t/geometry/kernel/PointCloud.cpp#L144)):*

..and here.
</blockquote></details>

Done.

stotko (changes requested):
> Reviewed 2 of 9 files at r2, 11 of 11 files at r4.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3691)* status: all files reviewed, 12 unresolved discussions (waiting on @reyanshsolis, @ssheorey, and @yxlao)

___
*[cpp/benchmarks/t/geometry/PointCloud.cpp, line 149 at r4](https://reviewable.io/reviews/isl-org/open3d/3691#-MfmMIBz3RwHsUnyjant:-MfmMIBz3RwHsUnyjanu:b-q1c0iy) ([raw file](https://github.com/isl-org/open3d/blob/48c3b05e2182dac0593a5de5523ebe68651af86b/cpp/benchmarks/t/geometry/PointCloud.cpp#L149)):*
> ```cpp
>     for (auto _ : state) {
>         pcd.EstimateNormals(max_nn, radius);
>         pcd.RemovePointAttr("normals");
> ```

I think this should not be timed as well. Instead only the line above should be benchmarked.
___
*[cpp/benchmarks/t/geometry/PointCloud.cpp, line 168 at r4](https://reviewable.io/reviews/isl-org/open3d/3691#-MfmM_vi-4B_SmPpg1Az:-MfmM_vi-4B_SmPpg1B-:br1zuxh) ([raw file](https://github.com/isl-org/open3d/blob/48c3b05e2182dac0593a5de5523ebe68651af86b/cpp/benchmarks/t/geometry/PointCloud.cpp#L168)):*
> ```cpp
>     for (auto _ : state) {
>         pcd_down->EstimateNormals(search_param, true);
>         pcd_down->normals_.clear();
> ```

Same here. See above.
___
*[cpp/open3d/core/linalg/kernel/SVD3x3.h, line 62 at r4](https://reviewable.io/reviews/isl-org/open3d/3691#-MfmN5Wm20GFkQom2eBo:-MfmN5Wm20GFkQom2eBp:bf8h644) ([raw file](https://github.com/isl-org/open3d/blob/7cf9c46fde188d5dbbada39aea275e0535e05655/cpp/open3d/core/linalg/kernel/SVD3x3.h#L62)):*
> ```objc
> 
> #define gcosine_pi_over_eight 1064076127
> #define gsmall_number 1.e-12f
> ```

Although I like getting rid of this define, I wonder why only this one has been selected to removal. For consistency, I would suggest to either revert that change (probably the easier approach) or port all constants away from defines.
___
*[cpp/open3d/t/geometry/PointCloud.h, line 333 at r2](https://reviewable.io/reviews/isl-org/open3d/3691#-Me5ynz50ep05Mo4XFM5:-MfmNf9X3oh7NjdywSR1:b-b3ktup) ([raw file](https://github.com/isl-org/open3d/blob/46378ba958eeecabe8163fe8ad300ba4d59f4c7c/cpp/open3d/t/geometry/PointCloud.h#L333)):*

Hidden the functionality for now. Might expose in future if required.
</blockquote></details>

This function still seems to be `public`. Is this intentional?
___
*[cpp/open3d/t/geometry/PointCloud.cpp, line 259 at r4](https://reviewable.io/reviews/isl-org/open3d/3691#-MfmOHlt2nB74ymEslqj:-MfmOHlt2nB74ymEslqk:b-bmm24e) ([raw file](https://github.com/isl-org/open3d/blob/48c3b05e2182dac0593a5de5523ebe68651af86b/cpp/open3d/t/geometry/PointCloud.cpp#L259)):*
> ```cpp
>     if (radius.has_value()) {
>         utility::LogDebug("Using Hybrid Search for computing covariances");
>         // Computes and set `covariances` attribute using Hybrid Search
> ```

nit: `set` -> `sets`
___
*[cpp/open3d/t/geometry/PointCloud.cpp, line 276 at r4](https://reviewable.io/reviews/isl-org/open3d/3691#-MfmOKR9-lYX7rAKEmyx:-MfmOKR9-lYX7rAKEmyy:b-q64pkt) ([raw file](https://github.com/isl-org/open3d/blob/48c3b05e2182dac0593a5de5523ebe68651af86b/cpp/open3d/t/geometry/PointCloud.cpp#L276)):*
> ```cpp
>     } else {
>         utility::LogDebug("Using KNN Search for computing covariances");
>         // Computes and set `covariances` attribute using KNN Search method.
> ```

See above.
___
*[cpp/open3d/t/geometry/PointCloud.cpp, line 295 at r4](https://reviewable.io/reviews/isl-org/open3d/3691#-MfmOQtY9rgoD5V8qc9f:-MfmOQtY9rgoD5V8qc9g:b-pyl6n2) ([raw file](https://github.com/isl-org/open3d/blob/48c3b05e2182dac0593a5de5523ebe68651af86b/cpp/open3d/t/geometry/PointCloud.cpp#L295)):*
> ```cpp
>     }
> 
>     // Estiamte `normal` of each point using it's `covariance` matrix.
> ```

`Estiamte` -> `Estimate`
`it's` -> `its`
___
*[cpp/open3d/t/geometry/kernel/PointCloudImpl.h, line 244 at r4](https://reviewable.io/reviews/isl-org/open3d/3691#-MfmP7vh6m1DgkAN07cX:-MfmP7vh6m1DgkAN07cY:b-q84w8x) ([raw file](https://github.com/isl-org/open3d/blob/48c3b05e2182dac0593a5de5523ebe68651af86b/cpp/open3d/t/geometry/kernel/PointCloudImpl.h#L244)):*
> ```objc
>     for (int64_t i = 0; i < indices_count; i++) {
>         int64_t idx = 3 * indices_ptr[i];
>         double x = static_cast<double>(points_ptr[idx]) - centeroid[0];
> ```

`scalar_t` as `cumulants has the same type.
___
*[cpp/open3d/t/geometry/kernel/PointCloudImpl.h, line 265 at r4](https://reviewable.io/reviews/isl-org/open3d/3691#-MfmPLJj9Nl4E6qgBwK2:-MfmPLJj9Nl4E6qgBwK3:b-x5kvqe) ([raw file](https://github.com/isl-org/open3d/blob/48c3b05e2182dac0593a5de5523ebe68651af86b/cpp/open3d/t/geometry/kernel/PointCloudImpl.h#L265)):*
> ```objc
> 
>     // Covariances(0, 0)
>     covariance_ptr[0] = static_cast<scalar_t>(cumulants[0]);
> ```

No cast needed here.

ssheorey (changes requested):
Following up on my concern about the low level code here that will be difficult to maintain (e.g.: vectorization for the 3x3 eigenvalue computations, etc.). Have we tried comparing / benchmarking alternatives?

CPU:  MKL / eigen
MKL: https://software.intel.com/content/www/us/en/develop/documentation/onemkl-developer-reference-c/top/lapack-routines/lapack-least-squares-and-eigenvalue-problem-routines/lapack-least-squares-and-eigenvalue-problem-driver-routines/symmetric-eigenvalue-problems-lapack-driver-routines/syev.html 
Eigen: https://eigen.tuxfamily.org/dox/classEigen_1_1SelfAdjointEigenSolver.html
CUDA:  cusolver: https://docs.nvidia.com/cuda/cusolver/index.html (multiple streams are recommended for solving many small problems at once).

We should go with these well tested and optimized libraries unless there is a clear advantage to implementing our own code.

> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3691#-:-Mfp7Q3e3UOn9JS8YoxT:bvmangd)* status: all files reviewed, 14 unresolved discussions (waiting on @reyanshsolis, @ssheorey, and @yxlao)

___
*[cpp/open3d/t/geometry/kernel/PointCloudImpl.h, line 159 at r4](https://reviewable.io/reviews/isl-org/open3d/3691#-Mfp-i4u9IlT-_tEZfPy:-Mfp-i4v4P4mr77cnekg:b-rgc5hk) ([raw file](https://github.com/isl-org/open3d/blob/48c3b05e2182dac0593a5de5523ebe68651af86b/cpp/open3d/t/geometry/kernel/PointCloudImpl.h#L159)):*
> ```objc
> // computation.
> template <typename scalar_t>
> OPEN3D_HOST_DEVICE void EstimatePointWiseNormalizedCovarianceKernel(
> ```

OK to remove the less robust method and just keep the more robust method.
___
*[cpp/open3d/t/geometry/kernel/PointCloudImpl.h, line 227 at r4](https://reviewable.io/reviews/isl-org/open3d/3691#-Mfp02AmBFf7qu-d17za:-Mfp02AmBFf7qu-d17zb:b-yrliqu) ([raw file](https://github.com/isl-org/open3d/blob/48c3b05e2182dac0593a5de5523ebe68651af86b/cpp/open3d/t/geometry/kernel/PointCloudImpl.h#L227)):*
> ```objc
>         const int64_t& indices_count,
>         scalar_t* covariance_ptr) {
>     double centeroid[3] = {0};
> ```

nit: `centroid`
___
*[cpp/open3d/t/geometry/kernel/PointCloudImpl.h, line 241 at r4](https://reviewable.io/reviews/isl-org/open3d/3691#-Mfp1yEW6tMq3BSF2hD1:-Mfp1yEW6tMq3BSF2hD2:b65n391) ([raw file](https://github.com/isl-org/open3d/blob/48c3b05e2182dac0593a5de5523ebe68651af86b/cpp/open3d/t/geometry/kernel/PointCloudImpl.h#L241)):*
> ```objc
> 
>     // cumulants must always be Float64 to ensure precision.
>     scalar_t cumulants[6] = {0};
> ```

Use explicit `double` instead of `scalar_t` if you need `Float64` .
___
*[cpp/open3d/t/geometry/kernel/PointCloudImpl.h, line 262 at r4](https://reviewable.io/reviews/isl-org/open3d/3691#-Mfp2XkF3fKsG4VfKruc:-Mfp2XkF3fKsG4VfKrud:b-54vmoy) ([raw file](https://github.com/isl-org/open3d/blob/48c3b05e2182dac0593a5de5523ebe68651af86b/cpp/open3d/t/geometry/kernel/PointCloudImpl.h#L262)):*
> ```objc
>     cumulants[3] /= num_indices;
>     cumulants[4] /= num_indices;
>     cumulants[5] /= num_indices;
> ```

Use Bessel's correction: divide by `(num_indices-1)`. Also need to check for the case when `num_indices==1`.
See: https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance
___
*[cpp/pybind/t/geometry/pointcloud.cpp, line 135 at r4](https://reviewable.io/reviews/isl-org/open3d/3691#-Mfp8Wb8DrX4xsLpKY46:-Mfp8Wb8DrX4xsLpKY47:bkylxzd) ([raw file](https://github.com/isl-org/open3d/blob/48c3b05e2182dac0593a5de5523ebe68651af86b/cpp/pybind/t/geometry/pointcloud.cpp#L135)):*
> ```
>                    "Normals are oriented "
>                    "with respect to the input point cloud if normals exist. It "
> ```
This sentence is not clear. Also add more detail as in the C++ interface (recommended argument  values, etc.) Document arguments with `FunctionDocInject`.

reyanshsolis (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3691)* status: 8 of 16 files reviewed, 12 unresolved discussions (waiting on @reyanshsolis, @ssheorey, @stotko, and @yxlao)

___
*[cpp/open3d/t/geometry/PointCloud.h, line 333 at r2](https://reviewable.io/reviews/isl-org/open3d/3691#-Me5ynz50ep05Mo4XFM5:-MfpEXk8DFOS8UXMhFTJ:bwjsw7p) ([raw file](https://github.com/isl-org/open3d/blob/46378ba958eeecabe8163fe8ad300ba4d59f4c7c/cpp/open3d/t/geometry/PointCloud.h#L333)):*

This function still seems to be `public`. Is this intentional?
</blockquote></details>

`EstimateNormals` is directly used by users. 
`EstimateCovariances` is hidden, for now, but might be exposed in the future if required. (it is also exposed in `legacy` and `pcl`)
___
*[cpp/open3d/t/geometry/PointCloud.cpp, line 259 at r4](https://reviewable.io/reviews/isl-org/open3d/3691#-MfmOHlt2nB74ymEslqj:-Mfp9kmXC6cGhQuUWaSU:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/48c3b05e2182dac0593a5de5523ebe68651af86b/cpp/open3d/t/geometry/PointCloud.cpp#L259)):*

nit: `set` -> `sets`
</blockquote></details>

Done.
___
*[cpp/open3d/t/geometry/PointCloud.cpp, line 276 at r4](https://reviewable.io/reviews/isl-org/open3d/3691#-MfmOKR9-lYX7rAKEmyx:-Mfp9iIsAYUNo15OETVj:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/48c3b05e2182dac0593a5de5523ebe68651af86b/cpp/open3d/t/geometry/PointCloud.cpp#L276)):*

See above.
</blockquote></details>

Done.
___
*[cpp/open3d/t/geometry/PointCloud.cpp, line 295 at r4](https://reviewable.io/reviews/isl-org/open3d/3691#-MfmOQtY9rgoD5V8qc9f:-Mfp9hoz1ku7AGWrdWkA:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/48c3b05e2182dac0593a5de5523ebe68651af86b/cpp/open3d/t/geometry/PointCloud.cpp#L295)):*

`Estiamte` -> `Estimate`
`it's` -> `its`
</blockquote></details>

Done.
___
*[cpp/open3d/t/geometry/kernel/PointCloudImpl.h, line 159 at r4](https://reviewable.io/reviews/isl-org/open3d/3691#-Mfp-i4u9IlT-_tEZfPy:-MfpFuHX6t-tSm0wbIeU:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/48c3b05e2182dac0593a5de5523ebe68651af86b/cpp/open3d/t/geometry/kernel/PointCloudImpl.h#L159)):*

OK to remove the less robust method and just keep the more robust method.
</blockquote></details>

Done.
___
*[cpp/open3d/t/geometry/kernel/PointCloudImpl.h, line 241 at r4](https://reviewable.io/reviews/isl-org/open3d/3691#-Mfp1yEW6tMq3BSF2hD1:-MfpFvr334bKYO31L7Wq:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/48c3b05e2182dac0593a5de5523ebe68651af86b/cpp/open3d/t/geometry/kernel/PointCloudImpl.h#L241)):*

Use explicit `double` instead of `scalar_t` if you need `Float64` .
</blockquote></details>

Done.
___
*[cpp/open3d/t/geometry/kernel/PointCloudImpl.h, line 244 at r4](https://reviewable.io/reviews/isl-org/open3d/3691#-MfmP7vh6m1DgkAN07cX:-MfpEvdc9qv9DgNMLv9d:b-8odhod) ([raw file](https://github.com/isl-org/open3d/blob/48c3b05e2182dac0593a5de5523ebe68651af86b/cpp/open3d/t/geometry/kernel/PointCloudImpl.h#L244)):*

`scalar_t` as `cumulants has the same type.
</blockquote></details>

changed `cumulants` to `double`. In experiments, it was observed, that using `cumulants` as `double` gave better results while using the `textbook` method. However the `two-pass` method is robust by itself, but still having `cumulants` as `double` might be more robust. It also doesn't hurt the performance much. 
___
*[cpp/open3d/t/geometry/kernel/PointCloudImpl.h, line 262 at r4](https://reviewable.io/reviews/isl-org/open3d/3691#-Mfp2XkF3fKsG4VfKruc:-MfpKaZW9Cp97LUGFC1e:b-jgkd8a) ([raw file](https://github.com/isl-org/open3d/blob/48c3b05e2182dac0593a5de5523ebe68651af86b/cpp/open3d/t/geometry/kernel/PointCloudImpl.h#L262)):*

Use Bessel's correction: divide by `(num_indices-1)`. Also need to check for the case when `num_indices==1`.
See: https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance
</blockquote></details>

Done. [num_indices checked performed directly for < 3 case].
___
*[cpp/open3d/t/geometry/kernel/PointCloudImpl.h, line 265 at r4](https://reviewable.io/reviews/isl-org/open3d/3691#-MfmPLJj9Nl4E6qgBwK2:-Mfp9aJJ00-3Z_BxPn3O:bqjsmsw) ([raw file](https://github.com/isl-org/open3d/blob/48c3b05e2182dac0593a5de5523ebe68651af86b/cpp/open3d/t/geometry/kernel/PointCloudImpl.h#L265)):*

No cast needed here.
</blockquote></details>

Please see above comment. 
___
*[cpp/pybind/t/geometry/pointcloud.cpp, line 135 at r4](https://reviewable.io/reviews/isl-org/open3d/3691#-Mfp8Wb8DrX4xsLpKY46:-MfpPyL0CMuerB53fe2W:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/48c3b05e2182dac0593a5de5523ebe68651af86b/cpp/pybind/t/geometry/pointcloud.cpp#L135)):*

> ```
>                    "Normals are oriented "
>                    "with respect to the input point cloud if normals exist. It "
> ```
This sentence is not clear. Also add more detail as in the C++ interface (recommended argument  values, etc.) Document arguments with `FunctionDocInject`.
</blockquote></details>

Done.
___
*[cpp/open3d/core/linalg/kernel/SVD3x3.h, line 62 at r4](https://reviewable.io/reviews/isl-org/open3d/3691#-MfmN5Wm20GFkQom2eBo:-MfpDSjY0EVJebiVkY5C:brj0csv) ([raw file](https://github.com/isl-org/open3d/blob/7cf9c46fde188d5dbbada39aea275e0535e05655/cpp/open3d/core/linalg/kernel/SVD3x3.h#L62)):*

Although I like getting rid of this define, I wonder why only this one has been selected to removal. For consistency, I would suggest to either revert that change (probably the easier approach) or port all constants away from defines.
</blockquote></details>

I agree. Using `max` and `min` functionalities, they only support `(float, float)`, `(double, double)` and throws an error for `(double, float)`, so I moved the `gsmall_number` from define to inside function. 
Since I have removed `estimate_color_gradients` functionality from this PR, I am reverting change on this file for this PR and will move all the define for consistency if the above issue comes up again.

reyanshsolis (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3691)* status: 7 of 16 files reviewed, 12 unresolved discussions (waiting on @ssheorey, @stotko, and @yxlao)

___
*[cpp/benchmarks/t/geometry/PointCloud.cpp, line 149 at r4](https://reviewable.io/reviews/isl-org/open3d/3691#-MfmMIBz3RwHsUnyjant:-Mfqln9PDhFbwxr_hbFR:b9uhzb9) ([raw file](https://github.com/isl-org/open3d/blob/48c3b05e2182dac0593a5de5523ebe68651af86b/cpp/benchmarks/t/geometry/PointCloud.cpp#L149)):*

I think this should not be timed as well. Instead only the line above should be benchmarked.
</blockquote></details>

Done. 
Well, the EstimateNormals operation re-computes the normals anyway so even `not removing` them does not change the benchmark time. However the `remove` operation is only a `unordered_map::erase()` so it does not take any significant time, so I had it for clarity that the time is for point clouds without normals.
___
*[cpp/benchmarks/t/geometry/PointCloud.cpp, line 168 at r4](https://reviewable.io/reviews/isl-org/open3d/3691#-MfmM_vi-4B_SmPpg1Az:-MfqmMubCQrHOfqQeJ32:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/48c3b05e2182dac0593a5de5523ebe68651af86b/cpp/benchmarks/t/geometry/PointCloud.cpp#L168)):*

Same here. See above.
</blockquote></details>

Done.

ssheorey (changes requested):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3691)* status: 7 of 16 files reviewed, 8 unresolved discussions (waiting on @reyanshsolis, @stotko, and @yxlao)

___
*[cpp/open3d/t/geometry/kernel/PointCloudImpl.h, line 262 at r4](https://reviewable.io/reviews/isl-org/open3d/3691#-Mfp2XkF3fKsG4VfKruc:-MfsQ51oDSky9T0JkIv8:bef7jf0) ([raw file](https://github.com/isl-org/open3d/blob/48c3b05e2182dac0593a5de5523ebe68651af86b/cpp/open3d/t/geometry/kernel/PointCloudImpl.h#L262)):*

Done. [num_indices checked performed directly for < 3 case].
</blockquote></details>

The `num_indices<3` is a user input level check. But with radius or hybrid search you may only get 0 or 1 neighbor for a specific point (`neighbor_count`) though the user requested more.

reyanshsolis (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3691)* status: 7 of 16 files reviewed, 8 unresolved discussions (waiting on @ssheorey, @stotko, and @yxlao)

___
*[cpp/open3d/t/geometry/kernel/PointCloudImpl.h, line 262 at r4](https://reviewable.io/reviews/isl-org/open3d/3691#-Mfp2XkF3fKsG4VfKruc:-MfsUOh0CdfWkP3eIo6Q:b-xgf2zt) ([raw file](https://github.com/isl-org/open3d/blob/48c3b05e2182dac0593a5de5523ebe68651af86b/cpp/open3d/t/geometry/kernel/PointCloudImpl.h#L262)):*
> ```
>  EstimatePointWiseRobustNormalizedCovarianceKernel
> ```
I have added the check for `indices_count < 3` in  `EstimatePointWiseRobustNormalizedCovarianceKernel` function, which is called by both `Hybrid` and `KNN` search methods, so this takes care of the condition (`num_indices = static_cast<double>(indices_count)`). 
If `indices_count < 3` `Covariance matrix` is set as `Identity`, otherwise it is computed.

reyanshsolis (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3691)* status: 7 of 16 files reviewed, 8 unresolved discussions (waiting on @ssheorey, @stotko, and @yxlao)

___
*[cpp/open3d/t/geometry/kernel/PointCloudImpl.h, line 262 at r4](https://reviewable.io/reviews/isl-org/open3d/3691#-Mfp2XkF3fKsG4VfKruc:-MfsVHFp-6JixzmFQhwF:bo9x52x) ([raw file](https://github.com/isl-org/open3d/blob/48c3b05e2182dac0593a5de5523ebe68651af86b/cpp/open3d/t/geometry/kernel/PointCloudImpl.h#L262)):*

> ```
>  EstimatePointWiseRobustNormalizedCovarianceKernel
> ```
I have added the check for `indices_count < 3` in  `EstimatePointWiseRobustNormalizedCovarianceKernel` function, which is called by both `Hybrid` and `KNN` search methods, so this takes care of the condition (`num_indices = static_cast<double>(indices_count)`). 
If `indices_count < 3` `Covariance matrix` is set as `Identity`, otherwise it is computed.  
</blockquote></details>

* condition for (num_indices != 1)

ssheorey:
@reyanshsolis Do you know why
```
EstimateNormals/CPU F32 KNN[0.02 | 30]                      59.6 ms         59.5 ms           11
EstimateNormals/CPU F64 KNN[0.02 | 30]                      51.2 ms         51.2 ms           13
```
is slower than:
```
LegacyEstimateNormals/Legacy KNN[0.02 | 30]                 30.9 ms         30.8 ms           24
```

ssheorey (commented):
> Reviewed 3 of 8 files at r5, 1 of 6 files at r7, 5 of 5 files at r8.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3691)* status: 15 of 16 files reviewed, 12 unresolved discussions (waiting on @reyanshsolis, @stotko, and @yxlao)

___
*[cpp/open3d/t/geometry/kernel/PointCloudImpl.h, line 185 at r8](https://reviewable.io/reviews/isl-org/open3d/3691#-Mgv2cHF8gKEJvy5rWy9:-Mgv2cHF8gKEJvy5rWyA:badajfs) ([raw file](https://github.com/isl-org/open3d/blob/53d2f0de7bcdae7cdbd614d75a2fdc55550bdedc/cpp/open3d/t/geometry/kernel/PointCloudImpl.h#L185)):*
> ```
>     double num_indices = static_cast<double>(indices_count);
> ```
nit: Unnecessary. The compiler will take care of this since we only do `double / int32`.
___
*[cpp/open3d/t/geometry/kernel/PointCloudImpl.h, line 209 at r8](https://reviewable.io/reviews/isl-org/open3d/3691#-Mgv3r6VEA9e-J_fj5ne:-Mgv3r6VEA9e-J_fj5nf:bw9t4bd) ([raw file](https://github.com/isl-org/open3d/blob/53d2f0de7bcdae7cdbd614d75a2fdc55550bdedc/cpp/open3d/t/geometry/kernel/PointCloudImpl.h#L209)):*
> ```objc
>     // Using Bessel's correction (dividing by (n - 1) instead of n).
>     // Refer: https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance
>     cumulants[0] /= (num_indices - 1.0);
> ```

nit: Use loop for clearer code. The compiler will figure out if unrolling is faster.
___
*[cpp/open3d/t/geometry/kernel/PointCloudImpl.h, line 252 at r8](https://reviewable.io/reviews/isl-org/open3d/3691#-Mgv7w5m9PEh1C839vvm:-Mgv7w5m9PEh1C839vvn:b-104980) ([raw file](https://github.com/isl-org/open3d/blob/53d2f0de7bcdae7cdbd614d75a2fdc55550bdedc/cpp/open3d/t/geometry/kernel/PointCloudImpl.h#L252)):*
> ```objc
>     if (!check) {
>         utility::LogError(
>                 "NearestNeighborSearch::FixedRadiusIndex Index is not set.");
> ```

Is this an index building failure? Clarify error message e.g.: "Building FixedRadiusIndex failed." Is there any action that the user can take to fix this?
___
*[cpp/open3d/t/geometry/kernel/PointCloudImpl.h, line 301 at r8](https://reviewable.io/reviews/isl-org/open3d/3691#-Mgv6SQOBjI6Q7zV2KsR:-Mgv6SQOBjI6Q7zV2KsS:b-ht68tn) ([raw file](https://github.com/isl-org/open3d/blob/53d2f0de7bcdae7cdbd614d75a2fdc55550bdedc/cpp/open3d/t/geometry/kernel/PointCloudImpl.h#L301)):*
> ```objc
>     bool check = tree.KnnIndex();
>     if (!check) {
>         utility::LogError("KnnIndex is not set.");
> ```

Is this an index building failure? Clarify error message e.g.: "Building KnnIndex failed." Is there any action that the user can take to fix this?
___
*[cpp/open3d/t/geometry/kernel/PointCloudImpl.h, line 313 at r8](https://reviewable.io/reviews/isl-org/open3d/3691#-Mgv7cSxCtUT4wvz8I2S:-Mgv7cSxCtUT4wvz8I2T:byorswj) ([raw file](https://github.com/isl-org/open3d/blob/53d2f0de7bcdae7cdbd614d75a2fdc55550bdedc/cpp/open3d/t/geometry/kernel/PointCloudImpl.h#L313)):*
> ```objc
>         utility::LogError(
>                 "Not enought neighbors to compute Covariances / Normals. Try "
>                 "changing the search parameter.");
> ```

Clarify error message. e.g. "Try increasing k."

ssheorey (approved):
Looks good overall. It would be good to have some sense of performance comparison between the custom 3x3 eigenvector implementation and those from MKL / Eigen / cuBLAS.

reyanshsolis:
> Looks good overall. It would be good to have some sense of performance comparison between the custom 3x3 eigenvector implementation and those from MKL / Eigen / cuBLAS.

The implementation is directly derived/referred from the legacy code 
https://github.com/isl-org/Open3D/blob/e5e8c2041ae15333905b40ec7c28a44346e63f61/cpp/open3d/geometry/EstimateNormals.cpp#L139

I tested with the legacy with the `RobustEigen3x3 Implementation (the default we use)`, and the `Eigen - EigenSolver`, the difference in time is under 3-4% as the bottleneck is `search` not the `computation` part itself.
The result also looks the same in both these cases:
image

I agree with the idea of using standard functions, instead of our `RobustEigen3x3` implementation, as the difference in performance given the current state of NNS is not much. 
With Eigen there were some issues like calling `host` function from `device`, I have been experimenting with it, we might need to enable some `device` mode in Eigen it seems. 

If you are fine with it, I would like to keep experimenting and bring the standard function in the future and merge this functionality for now, as we are getting good output, and it will be easier to update/improve the function than to keep adding changes in this PR.

reyanshsolis:
Updated performance breakdown and comparisons in the main comment.

reyanshsolis (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3691)* status: 10 of 16 files reviewed, 10 unresolved discussions (waiting on @ssheorey, @stotko, and @yxlao)

___
*[cpp/open3d/t/geometry/kernel/PointCloudImpl.h, line 185 at r8](https://reviewable.io/reviews/isl-org/open3d/3691#-Mgv2cHF8gKEJvy5rWy9:-MgzbN8r9rkDVVqdCSNJ:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/53d2f0de7bcdae7cdbd614d75a2fdc55550bdedc/cpp/open3d/t/geometry/kernel/PointCloudImpl.h#L185)):*

> ```
>     double num_indices = static_cast<double>(indices_count);
> ```
nit: Unnecessary. The compiler will take care of this since we only do `double / int32`.
</blockquote></details>

Done.
___
*[cpp/open3d/t/geometry/kernel/PointCloudImpl.h, line 209 at r8](https://reviewable.io/reviews/isl-org/open3d/3691#-Mgv3r6VEA9e-J_fj5ne:-MgzbPa4DNmbvbSZmYhU:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/53d2f0de7bcdae7cdbd614d75a2fdc55550bdedc/cpp/open3d/t/geometry/kernel/PointCloudImpl.h#L209)):*

nit: Use loop for clearer code. The compiler will figure out if unrolling is faster.
</blockquote></details>

Done.
___
*[cpp/open3d/t/geometry/kernel/PointCloudImpl.h, line 252 at r8](https://reviewable.io/reviews/isl-org/open3d/3691#-Mgv7w5m9PEh1C839vvm:-MgzbTHFANBl_vVAbTgJ:bs9jmdt) ([raw file](https://github.com/isl-org/open3d/blob/53d2f0de7bcdae7cdbd614d75a2fdc55550bdedc/cpp/open3d/t/geometry/kernel/PointCloudImpl.h#L252)):*

Is this an index building failure? Clarify error message e.g.: "Building FixedRadiusIndex failed." Is there any action that the user can take to fix this?
</blockquote></details>

Done. I am not sure about `any action that the user can take to fix this`, I will ask Junha.
___
*[cpp/open3d/t/geometry/kernel/PointCloudImpl.h, line 301 at r8](https://reviewable.io/reviews/isl-org/open3d/3691#-Mgv6SQOBjI6Q7zV2KsR:-Mgzbca8E4rHuv0jZnUN:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/53d2f0de7bcdae7cdbd614d75a2fdc55550bdedc/cpp/open3d/t/geometry/kernel/PointCloudImpl.h#L301)):*

Is this an index building failure? Clarify error message e.g.: "Building KnnIndex failed." Is there any action that the user can take to fix this?
</blockquote></details>

Done.
___
*[cpp/open3d/t/geometry/kernel/PointCloudImpl.h, line 313 at r8](https://reviewable.io/reviews/isl-org/open3d/3691#-Mgv7cSxCtUT4wvz8I2S:-MgzbgfP3jp1VNAIMxgK:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/53d2f0de7bcdae7cdbd614d75a2fdc55550bdedc/cpp/open3d/t/geometry/kernel/PointCloudImpl.h#L313)):*

Clarify error message. e.g. "Try increasing k."
</blockquote></details>

Done.

stotko (approved):
:lgtm:

> Reviewed 1 of 8 files at r5, 5 of 5 files at r9.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3691#-:-MhEMUL10tnWTWpDSo06:bnfp4nl)* status: all files reviewed, 8 unresolved discussions (waiting on @reyanshsolis, @ssheorey, @stotko, and @yxlao)

___
*[cpp/open3d/t/geometry/PointCloud.h, line 314 at r9](https://reviewable.io/reviews/isl-org/open3d/3691#-MhEK-gzBhpsQmken3l-:-MhEK-gzBhpsQmken3l0:b-bfvphl) ([raw file](https://github.com/isl-org/open3d/blob/95f5639682d8d27e78c1ce1910dce011f8686fa3/cpp/open3d/t/geometry/PointCloud.h#L314)):*
> ```objc
> 
> public:
>     /// \brief Function to estimate point normals. If the pointcloud normals
> ```

nit: `pointcloud` -> `point cloud`
___
*[cpp/open3d/t/geometry/PointCloud.h, line 315 at r9](https://reviewable.io/reviews/isl-org/open3d/3691#-MhEJsxe50298JXEeBXo:-MhEJsxg1D2gHtcxSeh8:bnkn94o) ([raw file](https://github.com/isl-org/open3d/blob/95f5639682d8d27e78c1ce1910dce011f8686fa3/cpp/open3d/t/geometry/PointCloud.h#L315)):*
> ```objc
> public:
>     /// \brief Function to estimate point normals. If the pointcloud normals
>     /// exists, the estimated normals are oriented with respect to the same.
> ```

nit: `exists`-> `exist`
___
*[cpp/open3d/t/geometry/kernel/PointCloudImpl.h, line 178 at r9](https://reviewable.io/reviews/isl-org/open3d/3691#-MhEL6bd1sP7qQrLnDqv:-MhEL6bd1sP7qQrLnDqw:b9czybp) ([raw file](https://github.com/isl-org/open3d/blob/95f5639682d8d27e78c1ce1910dce011f8686fa3/cpp/open3d/t/geometry/kernel/PointCloudImpl.h#L178)):*
> ```objc
> 
>     double centroid[3] = {0};
>     for (int32_t i = 0; i < indices_count; i++) {
> ```

nit: We can change `i++` to `++i` everywhere.
___
*[cpp/open3d/t/geometry/kernel/PointCloudImpl.h, line 209 at r9](https://reviewable.io/reviews/isl-org/open3d/3691#-MhEKfIJ8TRj7XiMIkWc:-MhEKfIJ8TRj7XiMIkWd:b-7nkmod) ([raw file](https://github.com/isl-org/open3d/blob/95f5639682d8d27e78c1ce1910dce011f8686fa3/cpp/open3d/t/geometry/kernel/PointCloudImpl.h#L209)):*
> ```objc
>     // Refer: https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance
>     const double normalization_factor =
>             static_cast<double>(indices_count - 1.0);
> ```

Use `indices_count - 1` within the parentheses instead. This avoids implicit casting of `indices_count` to `double` due to `1.0` being a `double` value.

reyanshsolis (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3691)* status: 14 of 16 files reviewed, 5 unresolved discussions (waiting on @ssheorey, @stotko, and @yxlao)

___
*[cpp/open3d/t/geometry/PointCloud.h, line 314 at r9](https://reviewable.io/reviews/isl-org/open3d/3691#-MhEK-gzBhpsQmken3l-:-MhEPGNG-lyO-kp9yvRO:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/95f5639682d8d27e78c1ce1910dce011f8686fa3/cpp/open3d/t/geometry/PointCloud.h#L314)):*

nit: `pointcloud` -> `point cloud`
</blockquote></details>

Done.
___
*[cpp/open3d/t/geometry/PointCloud.h, line 315 at r9](https://reviewable.io/reviews/isl-org/open3d/3691#-MhEJsxe50298JXEeBXo:-MhEPF1P0e-sQem2-Wv4:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/95f5639682d8d27e78c1ce1910dce011f8686fa3/cpp/open3d/t/geometry/PointCloud.h#L315)):*

nit: `exists`-> `exist`
</blockquote></details>

Done.
___
*[cpp/open3d/t/geometry/kernel/PointCloudImpl.h, line 178 at r9](https://reviewable.io/reviews/isl-org/open3d/3691#-MhEL6bd1sP7qQrLnDqv:-MhEPCnr4PzC0iRw8tMW:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/95f5639682d8d27e78c1ce1910dce011f8686fa3/cpp/open3d/t/geometry/kernel/PointCloudImpl.h#L178)):*

nit: We can change `i++` to `++i` everywhere.
</blockquote></details>

Done.
___
*[cpp/open3d/t/geometry/kernel/PointCloudImpl.h, line 209 at r9](https://reviewable.io/reviews/isl-org/open3d/3691#-MhEKfIJ8TRj7XiMIkWc:-MhEPEDQBK6FlmwcrZP8:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/95f5639682d8d27e78c1ce1910dce011f8686fa3/cpp/open3d/t/geometry/kernel/PointCloudImpl.h#L209)):*

Use `indices_count - 1` within the parentheses instead. This avoids implicit casting of `indices_count` to `double` due to `1.0` being a `double` value.
</blockquote></details>

Done.

stotko (commented):
> Reviewed 2 of 2 files at r10.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3691)* status: all files reviewed, 4 unresolved discussions (waiting on @reyanshsolis, @ssheorey, and @yxlao)

___
*[cpp/open3d/t/geometry/kernel/PointCloudImpl.h, line 209 at r9](https://reviewable.io/reviews/isl-org/open3d/3691#-MhEKfIJ8TRj7XiMIkWc:-MhERPNhAROfn2_ns_LE:b-urlyui) ([raw file](https://github.com/isl-org/open3d/blob/95f5639682d8d27e78c1ce1910dce011f8686fa3/cpp/open3d/t/geometry/kernel/PointCloudImpl.h#L209)):*

Done.
</blockquote></details>

I actually meant `const double normalization_factor = static_cast<double>(indices_count - 1);`. Sorry if that was not clear enough.

reyanshsolis (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3691)* status: 15 of 16 files reviewed, 4 unresolved discussions (waiting on @ssheorey, @stotko, and @yxlao)

___
*[cpp/open3d/t/geometry/kernel/PointCloudImpl.h, line 209 at r9](https://reviewable.io/reviews/isl-org/open3d/3691#-MhEKfIJ8TRj7XiMIkWc:-MhETtdiBJo3zp_1TLI6:bs7wumx) ([raw file](https://github.com/isl-org/open3d/blob/95f5639682d8d27e78c1ce1910dce011f8686fa3/cpp/open3d/t/geometry/kernel/PointCloudImpl.h#L209)):*

I actually meant `const double normalization_factor = static_cast<double>(indices_count - 1);`. Sorry if that was not clear enough.
</blockquote></details>

Done.

PR #3702, opened 2021-07-07
Add DEPENDS option to import and build 3rdparty functions

stotko:
Some readability cleanups:

- Prefix internal functions for 3rdparty dependency handling with `open3d_` for consistency.
- Add `DEPENDS` argument to import and build functions to make them more self-contained.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3702)

### DIFF
--- a/3rdparty/find_dependencies.cmake
+++ b/3rdparty/find_dependencies.cmake
@@ -29,7 +29,7 @@ set(Open3D_3RDPARTY_PRIVATE_TARGETS)
 
 find_package(PkgConfig QUIET)
 
-# build_3rdparty_library(name ...)
+# open3d_build_3rdparty_library(name ...)
 #
 # Builds a third-party library from source
 #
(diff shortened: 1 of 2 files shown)


stotko:
Marking as draft to postpone merging. #3703 should go first and potential upcoming merge conflicts (related to `zeromq`) that will affect this PR will be resolved later.

ssheorey (approved):
LGTM.

Related minor request:
I think we can get rid of these variables: JPEG_TARGET, ASSIMP_TARGET, WEBRTC_TARGET, GLFW_TARGET and so on. They are only used as place holders for `3rdparty_glfw`, etc.  We can use the `3rdparty_glfw`names directly. When we need to choose the system installed TARGET (e.g.: `glfw`) over Open3D built  3rdparty_TARGET (e.g. `3rdparty_glfw`), we can add an alias instead:
`add_library(3rdparty_glfw ALIAS glfw)`

> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3702#-:-Me1mb24DNp3z2JINGa2:b4ul3qw)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @yxlao)

stotko:
> LGTM.
> 
> Related minor request:
> I think we can get rid of these variables: JPEG_TARGET, ASSIMP_TARGET, WEBRTC_TARGET, GLFW_TARGET and so on. They are only used as place holders for `3rdparty_glfw`, etc. We can use the `3rdparty_glfw`names directly. When we need to choose the system installed TARGET (e.g.: `glfw`) over Open3D built 3rdparty_TARGET (e.g. `3rdparty_glfw`), we can add an alias instead:
> `add_library(3rdparty_glfw ALIAS glfw)`
> 
> > _[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3702#-:-Me1mb24DNp3z2JINGa2:b4ul3qw)_ status: 0 of 2 files reviewed, all discussions resolved (waiting on @yxlao)

Good point. I am also not happy with the current approach of using plain variables to wrap actual targets.

Thinking about this request, the non-trivial part here is making the new aliases work when installing Open3D and importing it afterwards via `find_package(Open3D)`. I can try to come up with a new function `open3d_find_package_3rdparty_library(...)` to handle this. The result would be a completely uniform 3rdparty target name convention, i.e. `Open3D::3rdparty_<name>` for every dependency. Such a larger change certainly requires more CI testing, so I suggest to defer that refactoring to a different PR.

yxlao (approved):

PR #3746, opened 2021-07-16
Detect ARM64 on linux, macos and windows

yxlao:
Previously we only detect ARM64 on Linux and does not care about ARM64 on macOS or Windows. 

Now, ARM64 is detected on all three platforms. `BUILD_WEBRTC` will be turned off by default if `OPEN3D_AARCH64` is set.

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/3746)


nitinkapgate:
Is there any plan to natively support Open3D on ARM M1 based MAC OS Machines? @yxlao

yxlao:
> Is there any plan to natively support Open3D on ARM M1 based MAC OS Machines? @yxlao

@nitinkapgate , we definitely want to support Apple silicon. It's in the backlog, we'll get to it at some point. 

As always, we welcome contributions from the community.

PR #3751, opened 2021-07-16
Update math for converting depth buffer values to view space Z

errissa:
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/3751)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,7 @@ _vimrc_local.vim
 .vs/
 .vscode/
 .clangd/
+.ccls-cache/
 
 # Build artifacts
 **/package-lock.json
--- a/cpp/open3d/visualization/gui/Application.cpp
+++ b/cpp/open3d/visualization/gui/Application.cpp
@@ -706,7 +706,8 @@ std::shared_ptr<geometry::Image> Application::RenderToDepthImage(
         rendering::View *view,
         rendering::Scene *scene,
         int width,
-        int height) {
+        int height,
+        bool z_in_view_space /* =false */) {
     std::shared_ptr<geometry::Image> img;
     auto callback = [&img](std::shared_ptr<geometry::Image> _img) {
         img = _img;
@@ -718,7 +719,7 @@ std::shared_ptr<geometry::Image> Application::RenderToDepthImage(
     // C++ callers do not need to know do this themselves.
     view->SetViewport(0, 0, width, height);
 
-    renderer.RenderToDepthImage(view, scene, callback);
+    renderer.RenderToDepthImage(view, scene, callback, z_in_view_space);
     renderer.BeginFrame();
     renderer.EndFrame();
 
--- a/cpp/open3d/visualization/gui/Application.h
+++ b/cpp/open3d/visualization/gui/Application.h
@@ -197,7 +197,8 @@ public:
             rendering::View *view,
             rendering::Scene *scene,
             int width,
-            int height);
+            int height,
+            bool z_in_view_space = false);
 
 private:
     Application();
--- a/cpp/open3d/visualization/rendering/Camera.h
+++ b/cpp/open3d/visualization/rendering/Camera.h
@@ -125,6 +125,9 @@ public:
     // is in the range [-1, 1] if the point is in view, or outside the range if
     // the point is out of view.
     virtual Eigen::Vector2f GetNDC(const Eigen::Vector3f& pt) const = 0;
+
+    /// Returns the view space depth (i.e., distance from camera) for the given
+    /// Z-buffer value
     virtual double GetViewZ(float z_buffer) const = 0;
 
     struct ProjectionInfo {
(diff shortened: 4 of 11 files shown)


benjaminum (changes requested):
> Reviewed 1 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3751)* status: 1 of 6 files reviewed, 2 unresolved discussions (waiting on @errissa)

___
*[cpp/open3d/visualization/rendering/Camera.h, line 128 at r1](https://reviewable.io/reviews/intel-isl/open3d/3751#-Men5dT71r34959YIcvd:-Men5dT8C8BvyHW_xsto:b-mermhe) ([raw file](https://github.com/intel-isl/open3d/blob/5080793fced07ff25935db2ae29c3bc2c015f892/cpp/open3d/visualization/rendering/Camera.h#L128)):*
> ```objc
>     // the point is out of view.
>     virtual Eigen::Vector2f GetNDC(const Eigen::Vector3f& pt) const = 0;
>     virtual double GetViewZ(float z_buffer) const = 0;
> ```

Please add function doc.
___
*[cpp/open3d/visualization/rendering/Renderer.cpp, line 144 at r1](https://reviewable.io/reviews/intel-isl/open3d/3751#-Men69E18pZbf81-oFvx:-Men69E18pZbf81-oFvy:b-gsqbhi) ([raw file](https://github.com/intel-isl/open3d/blob/5080793fced07ff25935db2ae29c3bc2c015f892/cpp/open3d/visualization/rendering/Renderer.cpp#L144)):*
> ```cpp
>                 if (z_in_view_space) {
>                     for (int i = 0; i < n_pixels; ++i) {
>                         pixels[i] = -z_near / pixels[i] + 1e-6f;
> ```

Why +1e-6f ?

errissa (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3751)* status: 1 of 6 files reviewed, 2 unresolved discussions (waiting on @benjaminum)

___
*[cpp/open3d/visualization/rendering/Camera.h, line 128 at r1](https://reviewable.io/reviews/isl-org/open3d/3751#-Men5dT71r34959YIcvd:-MguUguYBzz5Q_PJLqUi:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/5080793fced07ff25935db2ae29c3bc2c015f892/cpp/open3d/visualization/rendering/Camera.h#L128)):*

Please add function doc.
</blockquote></details>

Done.
___
*[cpp/open3d/visualization/rendering/Renderer.cpp, line 144 at r1](https://reviewable.io/reviews/isl-org/open3d/3751#-Men69E18pZbf81-oFvx:-MguUhsbAKPfHqEHfclY:b94hz5l) ([raw file](https://github.com/isl-org/open3d/blob/5080793fced07ff25935db2ae29c3bc2c015f892/cpp/open3d/visualization/rendering/Renderer.cpp#L144)):*

Why +1e-6f ?
</blockquote></details>

There was an error here. the +1e-6 should have been in the denominator to prevent divide-by-zero error. Another option here is to check for pixels[i] == 0.0 (which means far plane) and set pixels[i] to a constant large value (e.g., 1e9). That has the advantage of always returning the same value to indicate far plane. In this current code the value returned to indicate far place depends on `z_near`.

benjaminum (changes requested):
@errissa Please update the python bindings to make the new arg visible.

> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3751#-:-Mh0GDOk3C0HG9__OOmV:bnj7bdq)* status: 1 of 6 files reviewed, 1 unresolved discussion (waiting on @benjaminum)

___
*[cpp/open3d/visualization/rendering/Renderer.cpp, line 144 at r1](https://reviewable.io/reviews/isl-org/open3d/3751#-Men69E18pZbf81-oFvx:-Mh0FfRS9aRZRQrTrIV8:bkyanti) ([raw file](https://github.com/isl-org/open3d/blob/5080793fced07ff25935db2ae29c3bc2c015f892/cpp/open3d/visualization/rendering/Renderer.cpp#L144)):*

There was an error here. the +1e-6 should have been in the denominator to prevent divide-by-zero error. Another option here is to check for pixels[i] == 0.0 (which means far plane) and set pixels[i] to a constant large value (e.g., 1e9). That has the advantage of always returning the same value to indicate far plane. In this current code the value returned to indicate far place depends on `z_near`.
</blockquote></details>

OK, then let's remove the +1e-6 and return something constant. Do we return 1e9 already somewhere else? Would `inf` be a reasonable value?

errissa (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3751)* status: 1 of 11 files reviewed, 1 unresolved discussion (waiting on @benjaminum)

___
*[cpp/open3d/visualization/rendering/Renderer.cpp, line 144 at r1](https://reviewable.io/reviews/isl-org/open3d/3751#-Men69E18pZbf81-oFvx:-MhvM8Av4PsagRua-ykS:b-q08g1c) ([raw file](https://github.com/isl-org/open3d/blob/5080793fced07ff25935db2ae29c3bc2c015f892/cpp/open3d/visualization/rendering/Renderer.cpp#L144)):*

OK, then let's remove the +1e-6 and return something constant. Do we return 1e9 already somewhere else? Would `inf` be a reasonable value?
</blockquote></details>

Modified to return +inf which is the 'correct' value. No special constants are necessary this way. Updated `FilamentaCamera::GetViewZ` to also return +inf for consistency.

PR #3758, opened 2021-07-18
WebRTC datachannel HTML DOM element callbacks

ssheorey:
Register server callbacks triggered with messages through the WebRTC datachannel registered for specific HTML DOM elements  and events. An example of use is a radio button form in a browser that selects a tag. When the user clicks a button (input event), the browser sends a WebRTC datachannel message that triggers displaying geometry corresponding to that tag in the remote server.

This provides a general mechanism for JavaScript GUI elements without relying on a separate web server to relay messages.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3758)

### DIFF
--- a/cpp/open3d/visualization/webrtc_server/WebRTCWindowSystem.cpp
+++ b/cpp/open3d/visualization/webrtc_server/WebRTCWindowSystem.cpp
@@ -122,6 +122,9 @@ struct WebRTCWindowSystem::Impl {
 
     std::thread webrtc_thread_;
     bool sever_started_ = false;
+
+    std::unordered_map<std::string, std::function<void(std::string)>>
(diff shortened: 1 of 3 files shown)


yxlao:
The idea looks good to me in general. 

We can go one step further by providing a generic callback registration method. In #3759, I added a generic callback mechanism, and the `MouseEvent`/`ResizeEvent` handlers are converted to callbacks as well.

yxlao (commented):
The idea looks good to me in general.

We can go one step further by providing a generic callback registration method. In #3759, I added a generic callback mechanism, and the MouseEvent/ResizeEvent handlers are converted to callbacks as well.

PR #3732, opened 2021-07-12
Introduce CUDAScopedStream class for Multi-Stream Support

stotko:
Highlights:
- Rename `CUDADeviceSwitcher` to `CUDAScopedDevice` and remove `SwitchTo` function so simply its usage.
- Add `CUDAScopedStream` class for setting the global internal per-thread stream state.
- Add support for stream-based allocation (introduced in CUDA 11.2) and copy functions in `CUDAMemoryManager`.

Further changes:
- Add `cuda::{Get,Set}Stream` functions to access the global state. These functions are used to implement `CUDAScopedStream`.
- Add `cuda::{Get,Set}Device` to mirror the stream API. This makes `CUDAScopedDevice`'s implementation look similar to `CUDAScopedStream`. It further allows to hide the implementation detail that CUDA is already capable of managing the device state internally.
- Make all of Open3D's kernel calls aware of the current stream.
- Add unit tests to check thread locality.
- Add `CUDAScopedDeviceStream` in ML ops as a combination of both scoped wrappers. It infers the device from the stream and sets both states simultaneously.

Future work:
- Unify stream usage in ML ops to take advantage of `CUDAScopedStream`.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3732)

PR #3767, opened 2021-07-20
Cleanup CUDA utilities

stotko:
2nd followup of #3732:

- Merge `CUDAState.cuh` into `CUDAUtils.h` and make `cuda::set{Device,Stream}` and implementation detail.
- Reorder contents in merged `CUDAUtils.h` and move implementation details (which must be exposed though) to the end of the file.
- Rename `MemoryManagerCUDA` from `.cu` to `.cpp` file.
- Remove `OPEN3D_INLINE` macro which is always a wrapper around `inline` and thus not needed.

Other considerations;
- Python bindings for `CUDAScoped{Device,Stream}` are not provided as these classes are designed as low-level implementation details for high-level operations. Furthermore, `CUDAScopedStream` would also require bindings for `cudaStream_t`. This decision can be re-evaluated in the future when both classes gained more adoption throughout Open3D (and user code).

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3767)


yxlao (approved):
> Reviewed 16 of 16 files at r1.
> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3767)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ssheorey)

ssheorey (approved):
LGTM. Python bindings should be useful for users prototyping algorithms in Python. We can get around binding `cudaStream_t` by only exposing the version that manages streams internally. OK to look at this later, perhaps depending on user feedback.

> *[Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3767#-:-Mf9NHvV8hgC7_LLXCUZ:b-hi2qac)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @stotko)

PR #3790, opened 2021-07-23
RANSACConvergenceCriteria confidence python docstring

cosama:
I think the doc string for the `confidence` value in  `RANSACConvergenceCriteria` is wrong in the python documentation. At least to me the default value doesn't make sense when looking at the current description. Also the C++ string describes something different. I copied over the C++ doc string, however, I feel that it is still not optimal, as the formula mentioned there is hard to understand and uses some undefined values.

I suspect the `confidence`, is what it says, the closer to one the more confident RANSAC needs to be in it's solution, maybe a description along these lines would be better in the python and C++ reference. I am not an expert on this, so happy to change it if you provide me with some feedback.

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/3790)


ssheorey (changes requested):
Hi @cosama thanks for catching this!
See https://en.wikipedia.org/wiki/Random_sample_consensus#Parameters for an explanation of the parameters.
`ransac_n` is the number of correspondences used to fit the model at each step. Please go ahead and clarify this and this should be good to merge.
We usually try to keep the C++ and Python documentation consistent. Let us know / submit a PR if you find any other discrepancies.

cosama:
@ssheorey Thanks for getting back to me on this. I actually realized that the description mentions `desired probability of success` which is pretty clear with respect to the parameter being between 0 and 1. I can add the wikipedia link to the C++ and python documentation if you think that would be a good addition. Otherwise I think we can leave it as is.

ssheorey:
OK. We just need to explain `ransac_n` since that's defined in the C++ docs, but not in the Python docs and that `k` is `max_iterations`.

cosama:
I looked through the code and I actually feel the current documentation overall here doesn't make any sense. I feel like there is some documentation left over here from some api changes at some point. I updated it to better reflect this. @ssheorey let me know what you think.

ssheorey (approved):
Thanks @cosama LGTM!

PR #3796, opened 2021-07-26
Add tensor I/O support for .npz

yxlao:
### New
- `.npz` read: compress, uncompressed
- `.npz` write: uncompressed only
- Source code is mostly migrated from https://github.com/rogersce/cnpy

### Why not use `cnpy` as a library?
- `cnpy` is not well-maintained
- `cnpy` 's `NpyArray` struct does not provide `dtype`, you have to know the `dtype` before opening the file
  - https://github.com/rogersce/cnpy/blob/4e8810b1a8637695171ed346ce68f6984e585ef4/cnpy.h#L24
- Some bugs
  - https://github.com/rogersce/cnpy/blob/4e8810b1a8637695171ed346ce68f6984e585ef4/cnpy.cpp#L67
    ```cpp
    // It was
    std::string header(reinterpret_cast<char*>(buffer+9),header_len);
    // It should be
    std::string header(reinterpret_cast<char*>(buffer+10),header_len);
    ```
    Although, this does not affect the result, due to the way the rest of the header is parsed.
  - https://github.com/rogersce/cnpy/blob/4e8810b1a8637695171ed346ce68f6984e585ef4/cnpy.cpp#L106
    ```cpp
    // It was
    size_t res = fread(buffer,sizeof(char),11,fp);
    // It should be
    size_t res = fread(buffer,sizeof(char),10,fp); 
    ```
    Although, this does not affect the result as well, due to the way the rest of the header is parsed.    
  - https://github.com/rogersce/cnpy/blob/4e8810b1a8637695171ed346ce68f6984e585ef4/cnpy.cpp#L105
    This basically limits the header size to <= 256 chars. In fact, the maximum size of that NumPy `v1.0` allows is 2 ^ 16 = 25536. For `v2.0`, this number becomes 2 ^ 32 = 4GB. So, if the tensor has a very long shape, e.g. `np.zeros(shape=(1, 1, 1, 1, 1, ...., 1, 1))`, the original method might break.
- Does not support writing compressed `.npz`
  - It's easier for us to add support for this if it is in our code base.
- It's easier for us to add support for `v2.0` and `v3.0` of the file format.

### Future work
- For Open3D:
  - Geometry I/O to `.npz`
  - Pybind if needed. (Might not be needed if we already provide pybind for geometry I/O)
- For the npy/npz I/O:
  - Allow writing of compressed `.npz`.
    - A good reference for this: https://github.com/wolfv/xtensor-io/blob/master/include/xtensor-io/xnpz.hpp .
  - Avoid open-close file multiple times when writing multiple tensors to `.npz`. This is an artifact of `cnpy`.
  - Numpy `v2.0`, `v3.0`

### Appendix
Since it cannot write compressed `.npz` yet, we generate `tensors_compressed.npz` with Python for unit test. This file is very small (1.1KB) so it is checked in.

```python
import numpy as np
from pathlib import Path
import os

pwd = Path(os.path.dirname(os.path.realpath(__file__)))

if __name__ == '__main__':
    npz_file = pwd / "examples" / "test_data" / "tensors_compressed.npz"
    np.savez_compressed(
        npz_file,
        t0=np.array([[1, 2], [3, 4]], dtype=np.int32),
        t1=np.array([0, 2, 8, 10, 12, 14, 20, 22], dtype=np.float32).reshape(
            (2, 2, 2)),
        t2=np.array(3.14, dtype=np.float32),
        t3=np.ones((0,), dtype=np.float32),
        t4=np.ones((0, 0), dtype=np.float32),
        t5=np.ones((0, 1, 0), dtype=np.float32),
    )
```

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/3796)


stotko (changes requested):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3796)* status: all files reviewed, 5 unresolved discussions (waiting on @theNded and @yxlao)

___
*[cpp/open3d/t/io/NumpyIO.cpp, line 238 at r1](https://reviewable.io/reviews/isl-org/open3d/3796#-MfgA8WG3cH_qTWLj6z5:-MfgA8WG3cH_qTWLj6z6:bv0ao3) ([raw file](https://github.com/isl-org/open3d/blob/3bd55c68667c20127939bd94364ab234223deed7/cpp/open3d/t/io/NumpyIO.cpp#L238)):*
> ```cpp
> //   replaced from uint16_t to uint32_t.
> // - Version 3.0 uses utf8-encoded header string.
> static uint16_t ParseNpyPreamble(const char* preamble) {
> ```

`uint16_t` is just the internal size of the length variable. Should we return it as `size_t` (or `int64_t`) instead?
___
*[cpp/open3d/t/io/NumpyIO.cpp, line 318 at r1](https://reviewable.io/reviews/isl-org/open3d/3796#-Mfg6Xmo02qQtels8cXt:-Mfg6Xmp87RevGg6BpKl:b-c3ik4v) ([raw file](https://github.com/isl-org/open3d/blob/3bd55c68667c20127939bd94364ab234223deed7/cpp/open3d/t/io/NumpyIO.cpp#L318)):*
> ```cpp
> 
> template <typename T>
> static std::vector<char>& operator+=(std::vector<char>& lhs, const T rhs) {
> ```

Adding this operator to `std::vector`, even though just locally, should be avoided. It is confusing and (strictly speaking) formally not allowed by the C++ standard library. (There are only few exceptions to "extend"/specialize certain classes.)

Instead, we should define a small wrapper class (e.g. `NPZSerializer`, better name suggestions welcome) which has `operator+=`, `data()`, etc. and internally manages a `std::vector<char>`.
___
*[cpp/open3d/t/io/NumpyIO.cpp, line 355 at r1](https://reviewable.io/reviews/isl-org/open3d/3796#-Mfg95Sv1tsVEs9cCEJS:-Mfg95Sv1tsVEs9cCEJT:b-x9nw5j) ([raw file](https://github.com/isl-org/open3d/blob/3bd55c68667c20127939bd94364ab234223deed7/cpp/open3d/t/io/NumpyIO.cpp#L355)):*
> ```cpp
>     FILE* fp = nullptr;
>     if (append) {
>         fp = fopen(file_name.c_str(), "r+b");
> ```

We have our own `FOpen` in `utility/FileSystem.h`. Should we use this instead of the raw `fopen`?
___
*[cpp/open3d/t/io/NumpyIO.cpp, line 651 at r1](https://reviewable.io/reviews/isl-org/open3d/3796#-MfgBWkG2X7hBs1tfz4M:-MfgBWkG2X7hBs1tfz4N:bs3184x) ([raw file](https://github.com/isl-org/open3d/blob/3bd55c68667c20127939bd94364ab234223deed7/cpp/open3d/t/io/NumpyIO.cpp#L651)):*
> ```cpp
>     // It's possible to check tensor_name and only one selected numpy array,
>     // here we load all of them.
>     while (1) {
> ```

This only stops if the global header is found. What happens if there is no global header? Will it hang forever?
___
*[cpp/open3d/t/io/NumpyIO.cpp, line 717 at r1](https://reviewable.io/reviews/isl-org/open3d/3796#-MfgBEXr0zVUK4epIMCE:-MfgBEXr0zVUK4epIMCF:b-i8c2q8) ([raw file](https://github.com/isl-org/open3d/blob/3bd55c68667c20127939bd94364ab234223deed7/cpp/open3d/t/io/NumpyIO.cpp#L717)):*
> ```cpp
> 
>     // TODO: WriteNpzOneTensor is called multiple times inorder to write
>     // multiple tensors. This reqruies opening/closing files multiple times,
> ```

`reqruies` -> `requires`

theNded (changes requested):
> Reviewed 1 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3796)* status: all files reviewed, 8 unresolved discussions (waiting on @yxlao)

___
*[cpp/open3d/t/io/NumpyIO.cpp, line 554 at r1](https://reviewable.io/reviews/isl-org/open3d/3796#-Mfj44WA7NzwWFq_2qRP:-Mfj44WA7NzwWFq_2qRQ:bksa6zm) ([raw file](https://github.com/isl-org/open3d/blob/3bd55c68667c20127939bd94364ab234223deed7/cpp/open3d/t/io/NumpyIO.cpp#L554)):*
> ```cpp
> };
> 
> static NumpyArray CreateNumpyArrayFromFile(FILE* fp) {
> ```

Should these functions take a string as input, just like other factor functions for geometry IO?
___
*[cpp/open3d/t/io/NumpyIO.cpp, line 576 at r1](https://reviewable.io/reviews/isl-org/open3d/3796#-Mfj4Qn3CM7LjfBxi4Bd:-Mfj4Qn3CM7LjfBxi4Be:bb73xfv) ([raw file](https://github.com/isl-org/open3d/blob/3bd55c68667c20127939bd94364ab234223deed7/cpp/open3d/t/io/NumpyIO.cpp#L576)):*
> ```cpp
> 
> static NumpyArray CreateNumpyArrayFromCompressedFile(
>         FILE* fp,
> ```

Same
___
*[cpp/tests/t/io/NumpyIO.cpp, line 102 at r1](https://reviewable.io/reviews/isl-org/open3d/3796#-Mfj3bxN3HIou5Unc63O:-Mfj3bxN3HIou5Unc63P:bdolb81) ([raw file](https://github.com/isl-org/open3d/blob/3bd55c68667c20127939bd94364ab234223deed7/cpp/tests/t/io/NumpyIO.cpp#L102)):*
> ```cpp
> }
> 
> TEST_P(NumpyIOPermuteDevices, NpzWriteRead) {
> ```

Shall we test boundary cases, like I/O of empty dictionaries/files?

yxlao (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3796)* status: 2 of 4 files reviewed, 5 unresolved discussions (waiting on @stotko)

___
*[cpp/open3d/t/io/NumpyIO.cpp, line 238 at r1](https://reviewable.io/reviews/isl-org/open3d/3796#-MfgA8WG3cH_qTWLj6z5:-MfjhDUi62BUAD4ADzHD:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/ddb50d676f9cbf6c3904ae9db4b941ef3d4a812d/cpp/open3d/t/io/NumpyIO.cpp#L238)):*

`uint16_t` is just the internal size of the length variable. Should we return it as `size_t` (or `int64_t`) instead?
</blockquote></details>

Done.
___
*[cpp/open3d/t/io/NumpyIO.cpp, line 318 at r1](https://reviewable.io/reviews/isl-org/open3d/3796#-Mfg6Xmo02qQtels8cXt:-Mfjn9AY9kSp83buOp_z:bbf6a9g) ([raw file](https://github.com/isl-org/open3d/blob/ddb50d676f9cbf6c3904ae9db4b941ef3d4a812d/cpp/open3d/t/io/NumpyIO.cpp#L318)):*

Adding this operator to `std::vector`, even though just locally, should be avoided. It is confusing and (strictly speaking) formally not allowed by the C++ standard library. (There are only few exceptions to "extend"/specialize certain classes.)

Instead, we should define a small wrapper class (e.g. `NPZSerializer`, better name suggestions welcome) which has `operator+=`, `data()`, etc. and internally manages a `std::vector<char>`.
</blockquote></details>

Good idea. 

Added `class ByteSerializer : public std::vector<char>`.
___
*[cpp/open3d/t/io/NumpyIO.cpp, line 355 at r1](https://reviewable.io/reviews/isl-org/open3d/3796#-Mfg95Sv1tsVEs9cCEJS:-MfkmOoP1MBeSWZmW5eu:b-z4pt8c) ([raw file](https://github.com/isl-org/open3d/blob/ddb50d676f9cbf6c3904ae9db4b941ef3d4a812d/cpp/open3d/t/io/NumpyIO.cpp#L355)):*

We have our own `FOpen` in `utility/FileSystem.h`. Should we use this instead of the raw `fopen`?
</blockquote></details>

Done. Changed to the `CFile` class.
___
*[cpp/open3d/t/io/NumpyIO.cpp, line 651 at r1](https://reviewable.io/reviews/isl-org/open3d/3796#-MfgBWkG2X7hBs1tfz4M:-Mfjp30j1xAYtlpPoxRU:b-c8eqm3) ([raw file](https://github.com/isl-org/open3d/blob/ddb50d676f9cbf6c3904ae9db4b941ef3d4a812d/cpp/open3d/t/io/NumpyIO.cpp#L651)):*

This only stops if the global header is found. What happens if there is no global header? Will it hang forever?
</blockquote></details>

At every loop iteration, we advance the file pointer by 30 bytes (for local header) or more (for local data), so if the file is not a valid `.npz` file, it should eventually fail at an`fread` call.
___
*[cpp/open3d/t/io/NumpyIO.cpp, line 717 at r1](https://reviewable.io/reviews/isl-org/open3d/3796#-MfgBEXr0zVUK4epIMCE:-MfjoMzABo7eSAMA_PuO:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/ddb50d676f9cbf6c3904ae9db4b941ef3d4a812d/cpp/open3d/t/io/NumpyIO.cpp#L717)):*

`reqruies` -> `requires`
</blockquote></details>

Done.

yxlao:
@theNded Somehow I couldn't see your comments in Reviewable after a forced push. Here're the discussions.

## Comment 1
```cpp
cpp/open3d/t/io/NumpyIO.cpp, line 554 at r1 (raw file):

};

static NumpyArray CreateNumpyArrayFromFile(FILE* fp) {
```
> Should these functions take a string as input, just like other factor functions for geometry IO?

We use this design since we could be reading from the beginning of a file (`.npy`), or from the middle of a file (`.npz`), this function takes in a file pointer and advances the filer pointer while reading.

## Comment 2
```cpp
cpp/open3d/t/io/NumpyIO.cpp, line 576 at r1 (raw file):

static NumpyArray CreateNumpyArrayFromCompressedFile(
        FILE* fp,
```
> Same

Same as above. We want to read from the middle of the file.


## Comment 3
```cpp
cpp/tests/t/io/NumpyIO.cpp, line 102 at r1 (raw file):

}

TEST_P(NumpyIOPermuteDevices, NpzWriteRead) {
```
> Shall we test boundary cases, like I/O of empty dictionaries/files?

Yes, Added.

stotko (changes requested):
> Reviewed 2 of 2 files at r5.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3796)* status: all files reviewed, 6 unresolved discussions (waiting on @theNded and @yxlao)

___
*[cpp/open3d/t/io/NumpyIO.cpp, line 318 at r1](https://reviewable.io/reviews/isl-org/open3d/3796#-Mfg6Xmo02qQtels8cXt:-Mfm1LE12MujIDrOQdIj:bhjae2q) ([raw file](https://github.com/isl-org/open3d/blob/ddb50d676f9cbf6c3904ae9db4b941ef3d4a812d/cpp/open3d/t/io/NumpyIO.cpp#L318)):*

Good idea. 

Added `class ByteSerializer : public std::vector<char>`.
</blockquote></details>

Thanks, this is already a good improvement. I think it might be even better to make `std::vector<char>` a member variable of `ByteSerializer` rather than deriving from it. We only need a subset of `std::vector`'s functionality, mainly `data()`, `size()`, a range overload of `operator<<` to replace the `insert` call (see below), and perhaps some constructors for pre-allocation (see other comments).
___
*[cpp/open3d/t/io/NumpyIO.cpp, line 131 at r5](https://reviewable.io/reviews/isl-org/open3d/3796#-Mfm3FlY-SqeVhDouaZU:-Mfm3FlY-SqeVhDouaZV:b9yp5hs) ([raw file](https://github.com/isl-org/open3d/blob/2d73f6f7d8c11e5e6a79f59f8054aa0a486162b5/cpp/open3d/t/io/NumpyIO.cpp#L131)):*
> ```cpp
> }
> 
> static std::vector<char> CreateNumpyHeader(const core::SizeVector& shape,
> ```

We could return the `ByteSerializer` instance instead.
___
*[cpp/open3d/t/io/NumpyIO.cpp, line 276 at r5](https://reviewable.io/reviews/isl-org/open3d/3796#-Mfm3VYy9IP8nxoBdCih:-Mfm3VYy9IP8nxoBdCii:b-rup1nr) ([raw file](https://github.com/isl-org/open3d/blob/2d73f6f7d8c11e5e6a79f59f8054aa0a486162b5/cpp/open3d/t/io/NumpyIO.cpp#L276)):*
> ```cpp
>         FILE* fp) {
>     const size_t preamble_len = 10;  // Version 1.0 assumed.
>     std::vector<char> preamble(preamble_len);
> ```

Use `ByteSerializer` instead?
___
*[cpp/open3d/t/io/NumpyIO.cpp, line 283 at r5](https://reviewable.io/reviews/isl-org/open3d/3796#-Mfm3c9cAzv9nN5NkM0p:-Mfm3c9cAzv9nN5NkM0q:b-q64pkt) ([raw file](https://github.com/isl-org/open3d/blob/2d73f6f7d8c11e5e6a79f59f8054aa0a486162b5/cpp/open3d/t/io/NumpyIO.cpp#L283)):*
> ```cpp
>     const size_t header_len = ParseNpyPreamble(preamble.data());
> 
>     std::vector<char> header_chars(header_len, 0);
> ```

See above.
___
*[cpp/open3d/t/io/NumpyIO.cpp, line 348 at r5](https://reviewable.io/reviews/isl-org/open3d/3796#-Mfm44xGAJLmqG9mEIK9:-Mfm44xGAJLmqG9mEIKA:b-crwcxm) ([raw file](https://github.com/isl-org/open3d/blob/2d73f6f7d8c11e5e6a79f59f8054aa0a486162b5/cpp/open3d/t/io/NumpyIO.cpp#L348)):*
> ```cpp
>     FILE* fp = cfile.GetFILE();
> 
>     uint16_t nrecs = 0;
> ```

This could also be a `size_t` to avoid exposing the internal type.
___
*[cpp/open3d/t/io/NumpyIO.cpp, line 407 at r5](https://reviewable.io/reviews/isl-org/open3d/3796#-Mfm2Uty44dimavA0pqd:-Mfm2Uty44dimavA0pqe:blcpn4z) ([raw file](https://github.com/isl-org/open3d/blob/2d73f6f7d8c11e5e6a79f59f8054aa0a486162b5/cpp/open3d/t/io/NumpyIO.cpp#L407)):*
> ```cpp
>     global_header << static_cast<uint16_t>(0x0201);  // Second part of sig
>     global_header << static_cast<uint16_t>(20);      // Version made by
>     global_header.insert(global_header.end(), local_header.begin() + 4,
> ```

Add `operator<<` overload (e.g. with `std::pair` of iterators if only a single argument can be used for the operator) to make this consistent with the other usages here.

reyanshsolis (approved):
May we support optional `utility::CountingProgressReporter reporter` to update read/write progress for `NPZ` (example reference: `io/file_format/FilePLY`) ? 

> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3796#-:-Mfp2UZr-fW3GU120NDc:b585dyx)* status: all files reviewed, 6 unresolved discussions (waiting on @theNded and @yxlao)

reyanshsolis (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3796)* status: all files reviewed, 7 unresolved discussions (waiting on @theNded and @yxlao)

___
*[cpp/open3d/t/io/NumpyIO.cpp, line 751 at r5](https://reviewable.io/reviews/isl-org/open3d/3796#-Mfp5Hf-BTJ4YJCsDIps:-Mfp5Hf-BTJ4YJCsDIpt:bfnd2ab) ([raw file](https://github.com/isl-org/open3d/blob/2d73f6f7d8c11e5e6a79f59f8054aa0a486162b5/cpp/open3d/t/io/NumpyIO.cpp#L751)):*
> ```cpp
>     // TODO: Support writing in compressed mode: np.savez_compressed().
>     bool is_first_tensor = true;
>     for (auto it = tensor_map.begin(); it != tensor_map.end(); ++it) {
> ```

We may use `ProgressReporter` here, initialized with the count of `tensor_map length` and updated with each iteration of this loop.
For `Read`, as it is directly doing `memcpy`, we may update the reporter in two stages, `after header parsing` and `completion`.

yxlao (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3796)* status: 3 of 4 files reviewed, 7 unresolved discussions (waiting on @reyanshsolis, @stotko, @theNded, and @yxlao)

___
*[cpp/open3d/t/io/NumpyIO.cpp, line 318 at r1](https://reviewable.io/reviews/isl-org/open3d/3796#-Mfg6Xmo02qQtels8cXt:-MfqetgM8Hsc-yUrdfno:boiaa8a) ([raw file](https://github.com/isl-org/open3d/blob/ddb50d676f9cbf6c3904ae9db4b941ef3d4a812d/cpp/open3d/t/io/NumpyIO.cpp#L318)):*

Thanks, this is already a good improvement. I think it might be even better to make `std::vector<char>` a member variable of `ByteSerializer` rather than deriving from it. We only need a subset of `std::vector`'s functionality, mainly `data()`, `size()`, a range overload of `operator<<` to replace the `insert` call (see below), and perhaps some constructors for pre-allocation (see other comments).
</blockquote></details>

I did some experiments with the "ownership" version but I found it to be more verbose than the inheritance version, see: https://github.com/isl-org/Open3D/pull/3796/commits/11a6e90c717980a61383198c0c58a150bb907dec. 

The main reason for the verbosity is that we actually need quite a few functions from `std::vector`, such as `begin()`, `size()`, `resize()`, `data()`, so we need to forward them and change to capital letters. The changes have been reverted back to the inheritance version. To make it more explicit, I renamed the class to `CharVector`.  If you have strong feelings in favor of the ownership version, I can revert it back again.

For re-allocation, `reserve()` has been added to two cases of `ByteSerializer::Append()`. The other cases shall be fine as it is the vector's own manipulation functions.
___
*[cpp/open3d/t/io/NumpyIO.cpp, line 131 at r5](https://reviewable.io/reviews/isl-org/open3d/3796#-Mfm3FlY-SqeVhDouaZU:-Mfqh2whCKIwxvX6LO35:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/2d73f6f7d8c11e5e6a79f59f8054aa0a486162b5/cpp/open3d/t/io/NumpyIO.cpp#L131)):*

We could return the `ByteSerializer` instance instead.
</blockquote></details>

Done.
___
*[cpp/open3d/t/io/NumpyIO.cpp, line 348 at r5](https://reviewable.io/reviews/isl-org/open3d/3796#-Mfm44xGAJLmqG9mEIK9:-MfqaO2U0yas8kks-U4r:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/2d73f6f7d8c11e5e6a79f59f8054aa0a486162b5/cpp/open3d/t/io/NumpyIO.cpp#L348)):*

This could also be a `size_t` to avoid exposing the internal type.
</blockquote></details>

Done.
___
*[cpp/open3d/t/io/NumpyIO.cpp, line 407 at r5](https://reviewable.io/reviews/isl-org/open3d/3796#-Mfm2Uty44dimavA0pqd:-MfqaOl393sTV6fV6OaT:b-q17c7g) ([raw file](https://github.com/isl-org/open3d/blob/2d73f6f7d8c11e5e6a79f59f8054aa0a486162b5/cpp/open3d/t/io/NumpyIO.cpp#L407)):*

Add `operator<<` overload (e.g. with `std::pair` of iterators if only a single argument can be used for the operator) to make this consistent with the other usages here.
</blockquote></details>

Done. I renamed the `operator<<` to `Append()`, where it can take 1 or 2 arguments.
___
*[cpp/open3d/t/io/NumpyIO.cpp, line 751 at r5](https://reviewable.io/reviews/isl-org/open3d/3796#-Mfp5Hf-BTJ4YJCsDIps:-MfqeL1x7sWb4vv3_dNH:bc0btry) ([raw file](https://github.com/isl-org/open3d/blob/2d73f6f7d8c11e5e6a79f59f8054aa0a486162b5/cpp/open3d/t/io/NumpyIO.cpp#L751)):*

We may use `ProgressReporter` here, initialized with the count of `tensor_map length` and updated with each iteration of this loop.
For `Read`, as it is directly doing `memcpy`, we may update the reporter in two stages, `after header parsing` and `completion`.
</blockquote></details>

I would leave that as future work to limit the scope of this PR. 

In the short run, the `.npz` I/O should be sufficiently fast for most of our typical use cases, as it is only parsing very short headers and the tensor data is directly copied (possibly compressed with zlib).

stotko (commented):
> Reviewed 1 of 1 files at r7.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3796)* status: all files reviewed, 4 unresolved discussions (waiting on @reyanshsolis, @theNded, and @yxlao)

___
*[cpp/open3d/t/io/NumpyIO.cpp, line 318 at r1](https://reviewable.io/reviews/isl-org/open3d/3796#-Mfg6Xmo02qQtels8cXt:-MfrG1x28ksQU1dfiK7I:bxeg6kg) ([raw file](https://github.com/isl-org/open3d/blob/ddb50d676f9cbf6c3904ae9db4b941ef3d4a812d/cpp/open3d/t/io/NumpyIO.cpp#L318)):*

I did some experiments with the "ownership" version but I found it to be more verbose than the inheritance version, see: https://github.com/isl-org/Open3D/pull/3796/commits/11a6e90c717980a61383198c0c58a150bb907dec. 

The main reason for the verbosity is that we actually need quite a few functions from `std::vector`, such as `begin()`, `size()`, `resize()`, `data()`, so we need to forward them and change to capital letters. The changes have been reverted back to the inheritance version. To make it more explicit, I renamed the class to `CharVector`.  If you have strong feelings in favor of the ownership version, I can revert it back again.

For re-allocation, `reserve()` has been added to two cases of `ByteSerializer::Append()`. The other cases shall be fine as it is the vector's own manipulation functions.
</blockquote></details>

Yes, the ownership/composition version is, by design, more verbose than the inheritance one.

I still think that the former is better because I am not sure if the increase in verbosity is that bad. Regarding consistency with our overall API, having capitalized function names is actually better. The fact that we use `std::vector<char>` could also be considered an implementation detail. The probably strongest argument for the former version is that this class might be useful in other scenarios and could be moved outside this `.cpp` file. In that case, we would already have a clear and consistent API (that might of course change/be extended over time) where the function definitions eventually become hidden in a dedicated `CharVector.cpp`/`ByteSerializer.cpp` file.

So I am still in favor of the ownership/composition version, but also not totally against the inheritance version.

yxlao:
Merging as all review comments are addressed and discussed with @stotko

PR #3851, opened 2021-08-02
NPZ support for PointCloud IO

reyanshsolis:
Changes:
- Added `.npz` support for `tensor pointcloud io`. [No restriction on length shape and type of attribute].
- Added Unit Test.

Performance: [for redwood bedroom pointcloud with 16,164,488 points, colors, normals]
  | Legacy PLY | Tensor [after fix] PLY | NPZ
-- | -- | -- | --
File Size | 824 MB | 452 MB | 454 MB
Read Time | 2,800 - 2,900 ms | 5,000 - 6,764 ms | 160 - 175 ms
Write Time | 3,400 - 3,900 ms | 3,050 - 3,500 ms | 600 - 660 ms
Preserves Dtype. | No (converts dtype to F64) | Yes | Yes
Supports Custom Attributes | No | Partial (Only {N,1} shaped) | Yes

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/3851)

### DIFF
--- a/cpp/open3d/io/PointCloudIO.cpp
+++ b/cpp/open3d/io/PointCloudIO.cpp
@@ -160,5 +160,6 @@ bool WritePointCloud(const std::string &filename,
     p.update_progress = progress_updater;
     return WritePointCloud(filename, pointcloud, p);
 }
+
 }  // namespace io
 }  // namespace open3d
--- a/cpp/open3d/io/PointCloudIO.cpp
+++ b/cpp/open3d/io/PointCloudIO.cpp
@@ -29,6 +29,7 @@
 #include <iostream>
 #include <unordered_map>
 
+#include "open3d/t/io/PointCloudIO.h"
 #include "open3d/utility/FileSystem.h"
 #include "open3d/utility/Helper.h"
 #include "open3d/utility/Logging.h"
@@ -49,6 +50,7 @@ static const std::unordered_map<
                 {"ply", ReadPointCloudFromPLY},
                 {"pcd", ReadPointCloudFromPCD},
                 {"pts", ReadPointCloudFromPTS},
+                {"npz", ReadPointCloudFromNPZ},
         };
 
 static const std::unordered_map<
@@ -63,6 +65,7 @@ static const std::unordered_map<
                 {"ply", WritePointCloudToPLY},
                 {"pcd", WritePointCloudToPCD},
                 {"pts", WritePointCloudToPTS},
+                {"npz", WritePointCloudToNPZ},
         };
 
 std::shared_ptr<geometry::PointCloud> CreatePointCloudFromFile(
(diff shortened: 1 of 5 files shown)


yxlao (changes requested):
> Reviewed 5 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3851)* status: all files reviewed, 2 unresolved discussions (waiting on @reyanshsolis)

___
*[cpp/open3d/io/PointCloudIO.h, line 190 at r1](https://reviewable.io/reviews/isl-org/open3d/3851#-Mg8MFde9lVyqrkzL6FV:-Mg8MFde9lVyqrkzL6FW:b6pjfyf) ([raw file](https://github.com/isl-org/open3d/blob/5f1767752537251408e3b2aa829d4cbae1780e56/cpp/open3d/io/PointCloudIO.h#L190)):*
> ```objc
>                           const WritePointCloudOption &params);
> 
> bool ReadPointCloudFromNPZ(const std::string &filename,
> ```

Is this needed for the visualizer? If not, we can just drop it.
___
*[cpp/open3d/t/io/PointCloudIO.cpp, line 183 at r1](https://reviewable.io/reviews/isl-org/open3d/3851#-Mg8Mbz0DI1UFoWhn0cm:-Mg8Mbz0DI1UFoWhn0cn:b-2uro2r) ([raw file](https://github.com/isl-org/open3d/blob/5f1767752537251408e3b2aa829d4cbae1780e56/cpp/open3d/t/io/PointCloudIO.cpp#L183)):*
> ```cpp
>                            const ReadPointCloudOption &params) {
>     // Required checks are performed in the pointcloud constructor itself.
>     t::geometry::PointCloud pcd(ReadNpz(filename));
> ```

Does it work, if you do
```cpp
pointcloud = t::geometry::PointCloud(ReadNpz(filename));
```

reyanshsolis:
> Is this needed for the visualizer? If not, we can just drop it.

- Since our `Open3DViewer` uses `legacy geometry` for now if we want to support `.npz` in our current visualizer we may either support `.npz` in `legacy io` or check for extension locally in visualizer code to move this function there (I think, that will make the code more confusing). 

- We can also either drop this support for now and wait for the visualizer to move to `tensor based geometry` or keep it for now, and remove it then.

yxlao:
> Since our Open3DViewer uses legacy geometry for now if we want to support .npz in our current visualizer we may either support .npz in legacy io or check for extension locally in visualizer code to move this function there (I think, that will make the code more confusing).

I would drop it for now to keep things simple.

reyanshsolis (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3851)* status: 2 of 5 files reviewed, 2 unresolved discussions (waiting on @yxlao)

___
*[cpp/open3d/io/PointCloudIO.h, line 190 at r1](https://reviewable.io/reviews/isl-org/open3d/3851#-Mg8MFde9lVyqrkzL6FV:-MgKFnSyDzvy0nZbYf7n:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/5f1767752537251408e3b2aa829d4cbae1780e56/cpp/open3d/io/PointCloudIO.h#L190)):*

Is this needed for the visualizer? If not, we can just drop it.
</blockquote></details>

Done.
___
*[cpp/open3d/t/io/PointCloudIO.cpp, line 183 at r1](https://reviewable.io/reviews/isl-org/open3d/3851#-Mg8Mbz0DI1UFoWhn0cm:-MgKFvrr2tsdTVy26FTK:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/5f1767752537251408e3b2aa829d4cbae1780e56/cpp/open3d/t/io/PointCloudIO.cpp#L183)):*

Does it work, if you do
```cpp
pointcloud = t::geometry::PointCloud(ReadNpz(filename));
```
</blockquote></details>

Done.

yxlao (approved):

PR #3854, opened 2021-08-03
Removed CPU() and CUDA() wrappers for device transfer in cpp. 

reyanshsolis:
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/3854)


cpp/pybind/core/hashmap.cpp
```diff
@@ -105,8 +106,33 @@ void pybind_core_hashmap(py::module& m) {
...
+            [](const Hashmap& hashmap) {
+                return hashmap.To(
+                        core::Device(core::Device::DeviceType::CPU, 0));
+            },
+            "Transfer the hashmap to CPU. If the hashmap "
+            "is already on CPU, no copy will be performed.");
+    hashmap.def(
+            "cuda",
+            [](const Hashmap& hashmap, int device_id) {
+                if (!core::cuda::IsAvailable()) {
```
yxlao, line 120: Could these checks be made inside the `To()` function? Apply this to the rest of this PR for all pybinds.

cpp/pybind/core/hashmap.cpp
```diff
@@ -105,8 +106,33 @@ void pybind_core_hashmap(py::module& m) {
...
+                    utility::LogError(
+                            "CUDA is not available, cannot copy Hashmap.");
+                }
+                if (device_id < 0 || device_id >= core::cuda::DeviceCount()) {
+                    utility::LogError(
+                            "Invalid device_id {}, must satisfy 0 <= "
+                            "device_id < {}",
+                            device_id, core::cuda::DeviceCount());
+                }
+                return hashmap.To(core::Device(core::Device::DeviceType::CUDA,
```
yxlao, line 130: `core::Device("CUDA", device_id)`, the first argument can be a string. 
Apply this change to all `DeviceType::CUDA` and `DeviceType::CPU` in this PR,

cpp/pybind/core/tensor.cpp
```diff
@@ -472,6 +472,13 @@ void pybind_core_tensor(py::module& m) {
             "dtype"_a = py::none(), "device"_a = py::none());
 
     // Device transfer.
+    tensor.def(
+            "cpu",
+            [](const Tensor& tensor) {
+                return tensor.To(Device(Device::DeviceType::CPU, 0));
```
yxlao, line 478: `To(Device("CPU:0"))`, same for the rest of this PR

yxlao (changes requested):

reyanshsolis (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3854)* status: 0 of 25 files reviewed, 3 unresolved discussions (waiting on @yxlao)

___
*[cpp/pybind/core/hashmap.cpp, line 120 at r1](https://reviewable.io/reviews/isl-org/open3d/3854#-MgA30RFuVxFbBLXWXKQ-r1-120:-MgBeS_q3H7Syqr7LFfD:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/b5821d3c71b366abc3057a7168aac7aaebb47944/cpp/pybind/core/hashmap.cpp#L120)):*

Could these checks be made inside the `To()` function? Apply this to the rest of this PR for all pybinds.
</blockquote></details>

Done.
___
*[cpp/pybind/core/hashmap.cpp, line 130 at r1](https://reviewable.io/reviews/isl-org/open3d/3854#-MgA30RFuVxFbBLXWXKQ-r1-130:-MgBeT192DCMsE1IgX3a:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/b5821d3c71b366abc3057a7168aac7aaebb47944/cpp/pybind/core/hashmap.cpp#L130)):*

 `core::Device("CUDA", device_id)`, the first argument can be a string. 
Apply this change to all `DeviceType::CUDA` and `DeviceType::CPU` in this PR,
</blockquote></details>

Done.
___
*[cpp/pybind/core/tensor.cpp, line 478 at r1](https://reviewable.io/reviews/isl-org/open3d/3854#-MgA30RFuVxFbBLXWXKR-r1-478:-MgBeTy_3vgNVirN2Tr8:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/b5821d3c71b366abc3057a7168aac7aaebb47944/cpp/pybind/core/tensor.cpp#L478)):*

`To(Device("CPU:0"))`, same for the rest of this PR
</blockquote></details>

Done.

reyanshsolis:
@yxlao Do we want to support having CUDA Device object, even when Open3D is not built with CUDA.

eg. 
Even if Open3D is not built with CUDA, currently we allow this:
```python
device0 = o3c.Device("CUDA:-1")
print(device0) # CUDA:-1

device1 = o3c.Device("CUDA:1000")
print(device1) # CUDA:1000
```

Now, getting the `cuda::DeviceCount()` in `Device()` constructor will be difficult due to cyclic dependency, but we can block users from creating CUDA Device, when not built with CUDA enabled, or if the index is negative.

yxlao:
@reyanshsolis 

> Do we want to support having CUDA Device object, even when Open3D is not built with CUDA.

Yes. `o3c.Device("CUDA:1000")` should be allowed even when CUDA is not enabled. Even `o3c.Device("CUDA:1000")` should work. Similarly, if you only have one GPU, `o3c.Device("CUDA:1")` is allowed, but you can't allocate to this device.

> Now, getting the cuda::DeviceCount() in Device() constructor will be difficult due to cyclic dependency, 

This can be resolved by creating a `Device.cpp` and move all implementations from `Device.h` to `Device.cpp`. Remove all unnecessary includes from `Device.h` and add those includes in `Device.cpp`.

reyanshsolis:
Thanks for the clarification. 
If that is the case, then we might not need to make any changes to Device.h. Kindly let me know if any further changes are required.

yxlao (changes requested):
> Reviewed 16 of 23 files at r1, 8 of 9 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3854)* status: all files reviewed, 1 unresolved discussion (waiting on @reyanshsolis)

___
*[cpp/open3d/core/Tensor.cpp, line 557 at r3](https://reviewable.io/reviews/isl-org/open3d/3854#-MgOSJhwE3RARcGKbyEx:-MgOSJhwE3RARcGKbyEy:b8w50kz) ([raw file](https://github.com/isl-org/open3d/blob/a6d4a462bbf9e4e962c25cc149d69f1d003a4e1c/cpp/open3d/core/Tensor.cpp#L557)):*
> ```cpp
> // Checks if the device_id for CUDA is valid. For CPU Device ID == 0 is verified
> // in the Device constructor itself.
> static void AssertCUDADeviceID(const Device& device) {
> ```

We shouldn't need to check it here. It shall error gracefully in `kernel::Copy()`. 

If `kernel::Copy()` already errors gracefully, remove this check. Otherwise, update `kernel::Copy()`.

yxlao (approved):

PR #3857, opened 2021-08-03
moved data download dir

ShubhamAgarwal12:
Created a separate data directory for storing the test and downloaded data by default.


yxlao:
image
I'll expect this number to be relatively small, as we are not adding or removing any data files.

data/Crop/cropped.json
```diff
@@ -0,0 +1,31 @@
+{
```
yxlao, line 1: 1. We shouldn't see a data file being added. It should be shown as rename in the git history. The number of lines changed in git history should be relatively small.
2. Please follow the structure in https://github.com/isl-org/Open3D/tree/yixing/data-dir-mockup/data as we discussed. The local data should be placed in `data/common`.

yxlao (changes requested):

.gitignore
```diff
@@ -52,22 +52,22 @@ dist/
...
-examples/test_data/open3d_downloads
+data/common/downloads
 examples/python/reconstruction_system/dataset/
-examples/test_data/fountain_small
-examples/test_data/Armadillo.ply
-examples/test_data/Bunny.ply
-examples/test_data/eagle.ply
-examples/test_data/voxelized.ply
+data/common/fountain_small
+data/common/Armadillo.ply
```
yxlao, line 58: (Future work) These are downloaded datasets. Let's not change them in this PR to keep things simple. In the next PR, this shall be moved to the downloads folder and we no longer need to git ignore them.

data/Crop/cropped.json
ShubhamAgarwal12: 1. Moving it around in VSCode led to delete/add log in git. I've changed it now to rename.
2. Placed the local data in data/common

.gitignore
```diff
@@ -52,22 +52,22 @@ dist/
 compile_commands.json
 
 # Example resources
```
yxlao, line 54: Sort all these `Example resources`. If you're on VSCode, there is a command to sort lines.

.gitignore
```diff
@@ -52,22 +52,22 @@ dist/
...
-examples/test_data/open3d_downloads
+data/common/downloads
```
yxlao, line 55: `download` instead of `downloads`, see: https://github.com/isl-org/Open3D/tree/yixing/data-dir-mockup/data

cpp/benchmarks/t/pipelines/odometry/RGBDOdometry.cpp
```diff
@@ -64,9 +64,9 @@ static void ComputeOdometryResultPointToPlane(benchmark::State& state,
     const float depth_max = 3.0;
 
     t::geometry::Image src_depth = *t::io::CreateImageFromFile(
-            std::string(TEST_DATA_DIR) + "/RGBD/depth/00000.png");
+            std::string(TEST_DATA_DIR) + "/common/RGBD/depth/00000.png");
```
yxlao, line 67: Rename `TEST_DATA_DIR` to either:
- `OPEN3D_DATA_COMMON_DIR`: equivalent to `TEST_DATA_DIR/common`
- `OPEN3D_DATA_DOWNLOAD_DIR`: equivalent to `TEST_DATA_DIR/download`

Update CMake and the corresponding code.

CMakeLists.txt
```diff
@@ -335,9 +335,9 @@ if(WIN32)
 endif()
 
 # Downloads files from https://github.com/isl-org/open3d_downloads to
-# Open3D/examples/test_data/open3d_downloads.
-add_custom_target(open3d_downloads
-    COMMAND ${CMAKE_COMMAND} -P ${PROJECT_SOURCE_DIR}/examples/test_data/download_test_data.cmake
+# Open3D/data/downloads.
+add_custom_target(downloads
```
yxlao, line 339: `downloads` is too short for a cmake target name. Rename it to `download_data`.

yxlao (changes requested):
Great, looks much better now. 

If git got confused, one way is to squash all commits into one, then git will recognize those changes are renames but not delete/add files.

PR #3859, opened 2021-08-04
GitHub Action to lint Python code

cclauss:
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/3859)

### DIFF
--- a/.github/workflows/lint_python.yml
+++ b/.github/workflows/lint_python.yml
@@ -1,4 +1,4 @@
-name: lint_python
+name: Lint Python
 on: [pull_request, push]
 jobs:
   lint_python:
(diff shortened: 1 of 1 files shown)


.github/workflows/lint_python.yml
```diff
@@ -0,0 +1,24 @@
+name: lint_python
```
yxlao, line 1: Rename to `Lint Python` to be consistent with the rest of our Actions.

.github/workflows/lint_python.yml
```diff
@@ -0,0 +1,24 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+      - run: pip install --upgrade pip wheel
+      - run: pip install bandit black codespell flake8 isort mypy pytest pyupgrade safety
+      - run: bandit --recursive --skip B101 . || true  # B101 is assert statements
+      - run: black --check . || true
+      - run: codespell || true  # --ignore-words-list="" --skip="*.css,*.js,*.lock"
+      - run: flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+      - run: flake8 . --count --exit-zero --max-complexity=10 --max-line-length=88 --show-source --statistics
```
yxlao, line 15: I'd like to see an actual lint error generated by flake8`. 

Could you deliberately modify some Python code, such that it is dramatically correct but not up to the linting standard? After we get the results from CI confirming the linter is working, revert the commit.

.github/workflows/lint_python.yml
```diff
@@ -0,0 +1,24 @@
...
+      - run: isort --check-only --profile black . || true
```
yxlao, line 16: I would remove all the pytest related code here, as we are already doing an actual pytest in CI. See https://github.com/isl-org/Open3D/blob/44e9c4713bd86a83b6fa7141faaeb28558c41180/util/ci_utils.sh#L394 and https://github.com/isl-org/Open3D/runs/3237531611#step:8:1

yxlao (changes requested):

.github/workflows/lint_python.yml
cclauss: `--show-source` option is doing that.

.github/workflows/lint_python.yml
cclauss: pytest removed

.github/workflows/lint_python.yml
cclauss: pre-commit.ci can autofix errors but GitHub Actions makes that difficult to do.

yxlao:
@cclauss Thanks for your contribution. Sorry for the confusion for closing this PR. 

It would be great if we can show that this PR can automatically detect (not necessarily correct) Python warnings/errors. But currently, the linter still passes even if there are issues in the code.

cclauss:
You can make tests ___mandatory___ by removing `|| true` from the end of the line.  For flake8, `--exit-zero` plays a similar role.

PR #3885, opened 2021-08-09
Set openblas build ccache size to 1G

yxlao:
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/3885)


ssheorey (approved):
Can you also update the CI README with the cache size?

yxlao:
@ssheorey Yes, I'll do that. We'll also need similar mechanisms for CUDA docker build in GitHub Actions, so I'll do another round of cache summaries once everything is in place.

PR #3887, opened 2021-08-09
Make Normal Estimate optional for point clouds to speed up loading

errissa:
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/3887)

### DIFF
--- a/cpp/open3d/visualization/visualizer/GuiVisualizer.cpp
+++ b/cpp/open3d/visualization/visualizer/GuiVisualizer.cpp
@@ -349,7 +349,7 @@ struct GuiVisualizer::Impl {
     } settings_;
 
     rendering::TriangleMeshModel loaded_model_;
-
+    std::shared_ptr<geometry::PointCloud> loaded_pcd_;
     int app_menu_custom_items_index_ = -1;
     std::shared_ptr<gui::Menu> app_menu_;
 
@@ -369,6 +369,7 @@ struct GuiVisualizer::Impl {
         settings_.view_->ShowFileMaterialEntry(false);
 
         settings_.model_.SetMaterialsToDefault();
+        settings_.view_->EnableEstimateNormals(false);
         // model's OnChanged callback will get called (if set), which will
         // update everything.
     }
@@ -423,6 +424,11 @@ struct GuiVisualizer::Impl {
 
         UpdateLighting(renderer, settings_.model_.GetLighting());
 
+        // Does user want Point Cloud normals estimated?
+        if (settings_.model_.GetUserWantsEstimateNormals()) {
+            RunNormalEstimation();
+        }
+
         // Make sure scene redraws once changes have been applied
         scene_wgt_->ForceRedraw();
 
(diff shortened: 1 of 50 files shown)


reyanshsolis:
Hi @errissa,
Thanks a lot for this PR.

One issue I found (at least on my system), when we click on the `Estimate PCD Normals` button, the GUI becomes unusable until the process is completed. During this period, if the user tries to click anywhere in the viewer, it will hang, and crash most of the time. 

Also, the button for `Estimate PCD Normals` is disabled for `Mesh` but it is still there. I was wondering if there is a better way to hide it. I mean for example, if we can have a `tab` / `section` / `option` available when a `pointcloud` is loaded. 
In this `section` we may also have our other ops such as `Compute Normals`, `Remove Radius Outlier`, `Downsample PointCloud`, etc, which when selected can pop up adjustable parameters with a default value. 

Also, during processing, a `loading` / `processing` dialog may pop up to block users from playing with the GUI and crashing it. 

Many of these things can be part of future development, I just wanted to have a discussion about them, and maybe I can try a few of these things if the team finds it in the interest of the project.

reyanshsolis:
This PR, makes the loading of point-cloud with `normals` or `colors`, extremely fast as compared to the current master by skipping the normal computation step by default. For point cloud without `normals` and `colors` the `normals` are still computed by default (might be changed in the future if a convenient option is available to compute normals on demand). 

Here are some benchmarking for reference:
image

Here is a reference for the updated `Estimate PCD Normals` button added by this PR:
image

reyanshsolis:
Thanks! The dialog box really helps to prevent the user from crashing the viewer. 
Reference:
image

cpp/open3d/visualization/visualizer/GuiVisualizer.cpp
```diff
@@ -521,6 +507,68 @@ struct GuiVisualizer::Impl {
...
+                            scene3d->ClearGeometry();
+                            rendering::Material mat;
+                            scene3d->AddGeometry(MODEL_NAME, loaded_pcd_.get(),
+                                                 mat);
+                            UpdateSceneMaterial();
+                        });
+                gui::Application::GetInstance().PostToMainThread(
+                        visualizer_, [this]() { visualizer_->CloseDialog(); });
+            });
+        }
```
reyanshsolis, line 544: as for else condition, should we print some warnings on the console?

cpp/open3d/visualization/visualizer/GuiVisualizer.cpp
```diff
@@ -982,13 +1035,14 @@ void GuiVisualizer::LoadGeometry(const std::string &path) {
             if (success) {
                 utility::LogInfo("Successfully read {}", path.c_str());
                 UpdateProgress(ioProgressAmount);
-                if (!cloud->HasNormals()) {
+                if (!cloud->HasNormals() && !cloud->HasColors()) {
```
reyanshsolis, line 1038: [Discussing] May we make this optional too, with a console warning, for example: `The point cloud {file_name} file has neither normals nor colors, Please use 'Estimate PCD Normals' to add normals to the point cloud for better visualization`?

cpp/open3d/visualization/visualizer/GuiVisualizer.cpp
```diff
@@ -521,6 +507,68 @@ struct GuiVisualizer::Impl {
...
+                scene_wgt_->GetScene()->UpdateMaterial(
+                        settings_.normal_depth_material_);
+            } break;
+            case GuiSettingsModel::MaterialType::DEPTH: {
+                settings_.normal_depth_material_.shader = "depth";
+                scene_wgt_->GetScene()->UpdateMaterial(
+                        settings_.normal_depth_material_);
+            } break;
+
+            default:
```
reyanshsolis, line 567: do we want to print some warning for the `default` case?

reyanshsolis (commented):
LGTM

cpp/open3d/visualization/visualizer/GuiVisualizer.cpp
reyanshsolis: ignore this comment.

errissa: Discussed at meeting.

reyanshsolis (approved):
LGTM. In future PR we may add some `Warnings` to be printed on the console, for the `break` or `else` cases, when the user might not see the expected behaviour.

PR #3915, opened 2021-08-13
Testing : CUDA Device Available Assert

reyanshsolis:
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/3915)


reyanshsolis:
commit `0125dbf` : working fine.
Changes: Added `cuda::AssertCUDADeviceAvailable()`

reyanshsolis:
@yxlao 
The CI is getting stuck at
```
Running Open3D C++ unit tests...
Run ./bin/tests --gtest_shuffle --gtest_random_seed=SEED to repeat this test sequence.
[Open3D INFO] CPUInfo: 4 cores, 8 threads.
```
It keeps running after this step, but when we `Download log archive`, 
we see 
```
2021-08-13T17:07:49.2523067Z Running Open3D C++ unit tests...
2021-08-13T17:07:49.2523912Z Run ./bin/tests --gtest_shuffle --gtest_random_seed=SEED to repeat this test sequence.
2021-08-13T17:07:49.3209264Z [Open3D INFO] CPUInfo: 4 cores, 8 threads.
2021-08-13T17:37:21.0564165Z ##[error]The operation was canceled. 
```

cpp/open3d/core/CUDAUtils.cpp
```diff
@@ -96,6 +131,7 @@ int GetDevice() {
 }
 
 static void SetDevice(int device_id) {
+    AssertCUDADeviceAvailable(device_id);
```
reyanshsolis, line 134: this line is the minimum change, that triggers this behavior.

reyanshsolis:
Trying to move the `AssertCUDADeviceAvailable` to `MemoryManager` to check if it also triggers this behavior.

cpp/open3d/core/CUDAUtils.cpp
yxlao: Great, we can dig into `AssertCUDADeviceAvailable` to see which line causes the issue.

reyanshsolis:
using `cudaGetDeviceCount(&num_devices)` instead of `open3d::cude::DeviceCount()` might work. 

Possible reason: `cuda::SetDevice()` is being called from `CUDAScopedDevice` `constructor` and `destructor`. Since this is used everywhere, adding a function `AssetCUDADeviceAvailable` which uses `DeviceCount` which itself uses `cuda::CUDAState`, might be resulting in a larger binary size for tests, causing the CI to hang at just the starting of the test (executing `./bin/tests`).

Just my guess, not sure if it makes much sense.

reyanshsolis:
The guess solution seems to be working, creating another PR for this change to master.

reyanshsolis:
moved to PR #3922

PR #3919, opened 2021-08-13
Make docs archive available after wheels are ready.

ssheorey:
Upload docs archive with a prefix initially.
Rename to correct name after Ubuntu wheels are uploaded.
Ubuntu wheels take the longest to build (on master).
~~Upload enabled on all branches for testing.~~ Only master

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/3919)


yxlao (approved):
LGTM. The order is not guaranteed though. 
Works for now, we can refactor it later.

PR #3922, opened 2021-08-14
Assert CUDA Device-ID Available

reyanshsolis:
Changes: Added `AssertCUDADeviceAvailable` to check if the `device_id` is valid.

Context:
- Using `cuda::DeviceCount` in the function `AssertCUDADeviceAvailable` was resulting in GPU CI failure. The CI run was getting killed trying to run the cpp unit tests `./bin/tests`, however, the instance did not get terminated. 
- replacing `cuda::DeviceCount` with `cudaGetDeviceCount(&num_devices)` seems to solve the issue. 
- However, the previous version ran fine on the local system, so the issue might have to do with the CI environment. This is one of my guess : https://github.com/isl-org/Open3D/pull/3915#issuecomment-898823350 

TODO: 
- maybe we can directly use `cudaGetDeviceCount(&num_devices)` in `cuda::DeviceCount`.

This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/3922)

### DIFF
--- a/cpp/open3d/core/CUDAUtils.cpp
+++ b/cpp/open3d/core/CUDAUtils.cpp
@@ -90,9 +90,7 @@ void Synchronize(const Device& device) {
 
 void AssertCUDADeviceAvailable(int device_id) {
 #ifdef BUILD_CUDA_MODULE
-    int num_devices = 0;
-    OPEN3D_CUDA_CHECK(cudaGetDeviceCount(&num_devices));
-
+    int num_devices = cuda::DeviceCount();
     if (num_devices == 0) {
         utility::LogError(
                 "Invalid device 'CUDA:{}'. -DBUILD_CUDA_MODULE=ON, but no "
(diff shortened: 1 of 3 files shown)


yxlao:
I found the root cause of the problem. This bug only occurs in a multi-GPU setting. 

If there are more than one GPUs, we'll reach the code here:
https://github.com/isl-org/Open3D/blob/d92de2bffba5682ba12a73a269245c38a8e8baac/cpp/open3d/core/CUDAUtils.cpp#L224

This line introduces a circular dependency:
```cpp
DeviceCount()
  -> CUDAState::GetInstance() <-------------------------
    -> CUDAState::CUDAState()                          |
      -> CUDAScopedDevice::CUDAScopedDevice()          | CUDAState static instance
        -> cuda::SetDevice()                           | never finishes initialization.
          -> DeviceCount()                             |
            -> CUDAState::GetInstance() ----------------
```

With one GPU, `CUDAState::CUDAState()` does not call `CUDAScopedDevice` so there is no circular dependency.

yxlao (approved):

PR #3929, opened 2021-08-15
Wrap TEST_DATA_DIR in helper functions

yxlao:
- Data dir refactoring will be done in multiple PRs
- Add helper functions to get test data path
- Rename "dir" to "path" as it may return path to a file
- Currently:
  - `GetDataPathCommon()`: `Open3D/example/test_data`
  - `GetDataPathDownload()`: `Open3D/example/test_data/open3d_download`
- We are going to rename and move test data dir in the future, in the future:
  - `GetDataPathCommon()`: `Open3D/data/common`, also `OPEN3D_DATA_PATH_COMMON` in defs
  - `GetDataPathDownload()`: `Open3D/data/download`, also `OPEN3D_DATA_PATH_DOWNLOAD` in defs
  - With this getter API, later when we move the data dir, we don't need to change individual files.
- Future PR (#3930): Benchmarks folder should use `open3d::benchmarks` namespace by default for all files similar to what we use in the unit test.
- Future PR: Provide similar APIs for our Python examples, tests and tutorials.

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/3929)


reyanshsolis:
May we have the `GetDataPathCommon` and `GetDataPathDownload` in some common location (such as utility or a new directory of data handling, which may also be used for ongoing download functionality development), such that the same function becomes accessible to `benchmarks`, `tests` and `examples` also?

As the new dataset hander might have the default download location as that returned by `GetDataPathDownload`, it might be useful in cpp examples too.

benjaminum (changes requested):
Same question as @reyanshsolis , is there a reason for having both functions multiple times?

> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3929#-:-MhYNrLX3g7PcUUs9mwO:b-v7xo19)* status: 0 of 123 files reviewed, 1 unresolved discussion (waiting on @reyanshsolis and @yxlao)

___
*[cpp/benchmarks/Benchmarks.h, line 34 at r1](https://reviewable.io/reviews/isl-org/open3d/3929#-MhWt3p_CCVSzCgQdRBh:-MhWt3pa8vCQdNVyIdcA:b14yj1a) ([raw file](https://github.com/isl-org/open3d/blob/d59c1683308087ff57557ff3bfdca372ba1464ba/cpp/benchmarks/Benchmarks.h#L34)):*
> ```objc
> namespace benchmarks {
> 
> /// Returns Open3D common data directory. If a relative path will be appended
> ```

suggestion
```
/// Returns Open3D's common data directory. Appends \p relative_path if specified to the returned path.
```

yxlao (commented):
@benjaminum Yes, moved the utility to `cpp/open3d/utility/DataManager.h`.

The main note is that we should not bake in a `TEST_DATA_DIR` value in the Open3D main library so I make that part header-only. 

> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3929#-:-MiUaKNNENxplVskIjli:b10vuxf)* status: 0 of 127 files reviewed, 1 unresolved discussion (waiting on @reyanshsolis and @yxlao)

reyanshsolis:
nit: 
@yxlao may we use this header file as `ResourcePathManager`, (adding a `GetRenderingResourcesPath()` and using it in the visualiser code to fix the issue of constraints of launching a visualizer based example from inside `build/bin/examples`) ?

yxlao:
> may we use this header file as `ResourcePathManager`, (adding a `GetRenderingResourcesPath()` and using it in the visualiser code to fix the issue of constraints of launching a visualizer based example from inside `build/bin/examples`) ?
Or GetPathOpen3D might also do the job, as the absolute path might be used in the rendering code, as the relative structure of Materials and Resources are fixed.

I would limit the scope of change for each PR. `ResourcePath` can be handled in a future PR.

This PR focuses on one thing: avoid the direct usage of `TEST_DATA_DIR`.   `TEST_DATA_DIR` were previously scattered in different locations. By having a centralized location, it make future refacoring easier.

cpp/tests/t/pipelines/slac/SLAC.cpp
```diff
@@ -71,9 +71,8 @@ static bool IsPointCloudSimilar(t::geometry::PointCloud source,
 TEST_P(SLACPermuteDevices, DISABLED_RunSLACOptimizerForFragments) {
     core::Device device = GetParam();
 
-    std::string dataset_folder =
-            std::string(TEST_DATA_DIR) +
-            "/reconstruction_system/livingroom1_clean_micro";
+    std::string dataset_folder = utility::GetDataPathCommon(
+            "/reconstruction_system/livingroom1_clean_micro");
```
reyanshsolis, line 75: we may remove the first slash:
`/reconstruction_system/livingroom1_clean_micro` -> `reconstruction_system/livingroom1_clean_micro`.

Do we want to handle this automatically in the `GetDataPathCommon` (like join(path1, path2) in python?

cpp/tests/t/pipelines/slac/SLAC.cpp
```diff
@@ -86,8 +85,7 @@ TEST_P(SLACPermuteDevices, DISABLED_RunSLACOptimizerForFragments) {
         utility::LogError(
                 "No fragment found in {}, please make sure the test dataset "
                 "has been downloaded in "
-                "open3d_downloads/tests/reconstruction_system/ been "
-                "downloaded.",
+                "test_data/open3d_downloads/tests/reconstruction_system/.",
```
reyanshsolis, line 88: we may use `GetDataPathDownload` instead of `test_data/open3d_downloads`.

cpp/tests/t/pipelines/slac/SLAC.cpp
```diff
@@ -181,9 +179,8 @@ TEST_P(SLACPermuteDevices, DISABLED_RunSLACOptimizerForFragments) {
 TEST_P(SLACPermuteDevices, DISABLED_SLACIntegrate) {
     core::Device device = GetParam();
 
-    std::string dataset_folder =
-            std::string(TEST_DATA_DIR) +
-            "/reconstruction_system/livingroom1_clean_micro";
+    std::string dataset_folder = utility::GetDataPathCommon(
+            "/reconstruction_system/livingroom1_clean_micro");
```
reyanshsolis, line 183: same.

reyanshsolis (approved):

cpp/tests/t/pipelines/slac/SLAC.cpp
yxlao: Done.

We recommend removing the `/` at the beginning. But both will work since `//` or even `///` will be interpreted as `/` eventually by the system.

cpp/tests/t/pipelines/slac/SLAC.cpp
yxlao: Done.

cpp/tests/t/pipelines/slac/SLAC.cpp
yxlao: Done.

PR #3948, opened 2021-08-17
`get_data_path_common` and `get_data_path_download`.

reyanshsolis:
TODO:
- [x] python examples and tutorials. 
- [x] Check all the notebooks manually.
- [x] Check all the python examples.

This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/3948)


examples/test_data/download_utils.py
```diff
@@ -31,12 +31,27 @@
...
 
 # Typically "Open3D/examples/test_data", the test data dir.
 _test_data_dir = Path(__file__).parent.absolute().resolve()
 
 # Typically "Open3D/examples/test_data/open3d_downloads", the download dir.
-_download_dir = _test_data_dir / "open3d_downloads"
+_download_dir = os.path.join(_test_data_dir, "open3d_downloads")
+
+
+def get_data_path_common(relative_path=""):
```
ShubhamAgarwal12, line 43: Should we just return the common path here, and add relative path outside?

reyanshsolis: I think, this interface might be more convenient for users, also it is consistent with C++ version (DataManager.h).

ShubhamAgarwal12 (commented):
Couple points (May not be directly related to this PR):
1) Download can be specific to tutorial and we need not download everything at once.
2) I think download_utils.py is not needed inside the "test_data" folder. It is just a folder for storing test data.

Let's wait for @yxlao  also and see if he has more comments.

yxlao (changes requested):
The current solution i

> Reviewed 1 of 39 files at r4.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3948#-:-MkBT1iu9-HjBvLdwD10:b-iu1fr7)* status: 1 of 184 files reviewed, 2 unresolved discussions (waiting on @reyanshsolis, @ShubhamAgarwal12, and @yxlao)

___
*[examples/python/gui/multiple-windows.py, line 33 at r4](https://reviewable.io/reviews/isl-org/open3d/3948#-MkBSqTj6JuZHYCtL2WA:-MkBSqTj6JuZHYCtL2WB:b-k21on0) ([raw file](https://github.com/isl-org/open3d/blob/a3c50df535b80fbf4eaa59b86373f62f4f1ba47a/examples/python/gui/multiple-windows.py#L33)):*
> ```Python
> import time
> 
> from data_manager import get_data_path_common
> ```

Seems like we have to append the path to `data_manager` before we can import from it. Could you confirm that?

reyanshsolis:
> The current solution i
> 
> > Reviewed 1 of 39 files at r4.
> > _[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3948#-:-MkBT1iu9-HjBvLdwD10:b-iu1fr7)_ status: 1 of 184 files reviewed, 2 unresolved discussions (waiting on @reyanshsolis, @ShubhamAgarwal12, and @yxlao)
> 
> _[examples/python/gui/multiple-windows.py, line 33 at r4](https://reviewable.io/reviews/isl-org/open3d/3948#-MkBSqTj6JuZHYCtL2WA:-MkBSqTj6JuZHYCtL2WB:b-k21on0) ([raw file](https://github.com/isl-org/open3d/blob/a3c50df535b80fbf4eaa59b86373f62f4f1ba47a/examples/python/gui/multiple-windows.py#L33)):_
> 
> > ```
> > import time
> > 
> > from data_manager import get_data_path_common
> > ```
> 
> Seems like we have to append the path to `data_manager` before we can import from it. Could you confirm that?

Right, I fixed this in another branch, I will add the changes here as well.
Changing `test_data` dir. highlighted some other missing places as well, once they are fixed, I will bring the changes to this PR.

reyanshsolis:
@yxlao 
The documentation is not building properly in this PR. 
The cells are not being executed. 

On running it on my local system, it shows
`Execution of XXX.ipynb failed, this will cause Travis to fail.`

but when I test the tutorial, it works fine. 
If I make docs after running the tutorial, documentation is generated properly.

yxlao:
There are some issues with the current approach:
1. There are 4 locations where we can import `get_data_xxx`, which can cause confusions:
    ```python
    from data_manager import get_data_path_common
      - examples/python/data_manager.py
      - calls _download_all_files()
    
    from download_utils import get_data_path_common
      - examples/test_data/download_utils.py
      - the main get_data_path_common and get_data_path_download
    
    from open3d_test import get_data_path_common
      - python/test/open3d_test.py
      - imports download_utils
    
    from open3d_tutorial import get_data_path_common
      - examples/python/open3d_tutorial.py
      - imports data_manager
        - which, imports download_utils
    ```
2. We still cannot avoid `sys.path.append` in every file. For instance, all the example scripts will have to append the relative path to `data_manager.py`. Otherwise, `data_manager` cannot be imported.

My recommendation is to only keep **one** copy of `get_data_path_common()` and `get_data_path_download()` in `download_utils.py`. Every other file that uses the dataset folder should specify relative path w.r.t. `download_utils.py`. Also using `pathlib` can make the like shorter than `os.path`. It will look something like this:

```python
import sys
from pathlib import Path
sys.path.append(str(Path(__file__).parent / "../../test_data"))
from download_utils import get_data_path_common
```

yxlao:
> `Execution of XXX.ipynb failed, this will cause Travis to fail.`

For Jupyter notebooks, the path detection might need to work differently. Could you check https://stackoverflow.com/a/32711758/1255535 or some other solutions in the post?

reyanshsolis:
> There are some issues with the current approach:
> 
> 1. There are 4 locations where we can import `get_data_xxx`, which can cause confusions:
>    ```python
>    from data_manager import get_data_path_common
>      - examples/python/data_manager.py
>      - calls _download_all_files()
>    
>    from download_utils import get_data_path_common
>      - examples/test_data/download_utils.py
>      - the main get_data_path_common and get_data_path_download
>    
>    from open3d_test import get_data_path_common
>      - python/test/open3d_test.py
>      - imports download_utils
>    
>    from open3d_tutorial import get_data_path_common
>      - examples/python/open3d_tutorial.py
>      - imports data_manager
>        - which, imports download_utils
>    ```
> 2. We still cannot avoid `sys.path.append` in every file. For instance, all the example scripts will have to append the relative path to `data_manager.py`. Otherwise, `data_manager` cannot be imported.
> 
> My recommendation is to only keep **one** copy of `get_data_path_common()` and `get_data_path_download()` in `download_utils.py`. Every other file that uses the dataset folder should specify relative path w.r.t. `download_utils.py`. Also using `pathlib` can make the like shorter than `os.path`. It will look something like this:
> 
> ```python
> import sys
> from pathlib import Path
> sys.path.append(str(Path(__file__).parent / "../../test_data"))
> from download_utils import get_data_path_common
> ```

I agree, but linking all the files to `download_util.py` (which itself is inside `test_data`), seems equally inconvenient as linking to the `test_data` dir itself. 
One solution can be to include it in some `__init__` file, (but then we will be adding it to the core-open3d package).

yxlao:
> I agree, but linking all the files to `download_util.py` (which itself is inside `test_data`), seems equally inconvenient as linking to the `test_data` dir itself.
> One solution can be to include it in some `__init__` file (maybe open3d root init), (but then we will be adding it to the core-open3d package).

Yes, it is equally inconvenient, but at least we have a consistent location for the two helper functions.

PR #3954, opened 2021-08-18
AssertShape() accepts both shape and dynamic shape

yxlao:
... by implicit type conversion.

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/3954)

### DIFF
--- a/cpp/open3d/core/CMakeLists.txt
+++ b/cpp/open3d/core/CMakeLists.txt
@@ -11,6 +11,7 @@ target_sources(core PRIVATE
     MemoryManagerCPU.cpp
     MemoryManagerStatistic.cpp
     ShapeUtil.cpp
+    SizeVector.cpp
     Tensor.cpp
     TensorKey.cpp
     TensorList.cpp
--- a/cpp/open3d/core/EigenConverter.cpp
+++ b/cpp/open3d/core/EigenConverter.cpp
@@ -113,7 +113,7 @@ static std::vector<Eigen::Matrix<T, N, 1>> TensorToEigenVectorNxVector(
         utility::LogError("Internal error: dtype size mismatch {} != {}.",
                           dtype.ByteSize() * N, sizeof(Eigen::Matrix<T, N, 1>));
     }
-    tensor.AssertShapeCompatible({utility::nullopt, N});
+    tensor.AssertShape({utility::nullopt, N});
 
     // Eigen::VectorNx is not a "fixed-size vectorizable Eigen type" thus it is
     // safe to write directly into std vector memory, see:
--- a/cpp/open3d/core/Indexer.cpp
+++ b/cpp/open3d/core/Indexer.cpp
@@ -26,6 +26,8 @@
 
 #include "open3d/core/Indexer.h"
 
+#include <numeric>
+
 namespace open3d {
 namespace core {
 
--- a/cpp/open3d/core/SizeVector.cpp
+++ b/cpp/open3d/core/SizeVector.cpp
@@ -38,8 +38,136 @@
 namespace open3d {
 namespace core {
 
+DynamicSizeVector::DynamicSizeVector(
+        const std::initializer_list<utility::optional<int64_t>>& dim_sizes)
+    : std::vector<utility::optional<int64_t>>(dim_sizes) {}
+
+DynamicSizeVector::DynamicSizeVector(
+        const std::vector<utility::optional<int64_t>>& dim_sizes)
+    : std::vector<utility::optional<int64_t>>(dim_sizes) {}
+
+DynamicSizeVector::DynamicSizeVector(const DynamicSizeVector& other)
+    : std::vector<utility::optional<int64_t>>(other) {}
+
+DynamicSizeVector::DynamicSizeVector(int64_t n, int64_t initial_value)
+    : std::vector<utility::optional<int64_t>>(n, initial_value) {}
+
 DynamicSizeVector::DynamicSizeVector(const SizeVector& dim_sizes)
     : DynamicSizeVector(dim_sizes.begin(), dim_sizes.end()) {}
+
+DynamicSizeVector& DynamicSizeVector::operator=(const DynamicSizeVector& v) {
+    static_cast<std::vector<utility::optional<int64_t>>*>(this)->operator=(v);
+    return *this;
+}
+
+DynamicSizeVector& DynamicSizeVector::operator=(DynamicSizeVector&& v) {
+    static_cast<std::vector<utility::optional<int64_t>>*>(this)->operator=(v);
+    return *this;
+}
+
+std::string DynamicSizeVector::ToString() const {
+    std::stringstream ss;
+    ss << "{";
+    bool first = true;
+    for (const utility::optional<int64_t>& element : *this) {
+        if (first) {
+            first = false;
+        } else {
+            ss << ", ";
+        }
+        if (element.has_value()) {
+            ss << fmt::format("{}", element.value());
+        } else {
+            ss << "None";
+        }
+    }
+    ss << "}";
+    return ss.str();
+}
+
+bool DynamicSizeVector::IsDynamic() const {
+    return std::any_of(
+            this->begin(), this->end(),
+            [](const utility::optional<int64_t>& v) { return !v.has_value(); });
+}
+
+SizeVector::SizeVector(const std::initializer_list<int64_t>& dim_sizes)
+    : std::vector<int64_t>(dim_sizes) {}
+
+SizeVector::SizeVector(const std::vector<int64_t>& dim_sizes)
+    : std::vector<int64_t>(dim_sizes) {}
+
+SizeVector::SizeVector(const SizeVector& other) : std::vector<int64_t>(other) {}
+
+SizeVector::SizeVector(int64_t n, int64_t initial_value)
+    : std::vector<int64_t>(n, initial_value) {}
+
+SizeVector::SizeVector(const DynamicSizeVector& dim_sizes) {
+    this->resize(dim_sizes.size());
+    std::transform(dim_sizes.begin(), dim_sizes.end(), this->begin(),
+                   [](const auto& v) { return v.value(); });
+}
+
+SizeVector& SizeVector::operator=(const SizeVector& v) {
+    static_cast<std::vector<int64_t>*>(this)->operator=(v);
+    return *this;
+}
+
+SizeVector& SizeVector::operator=(SizeVector&& v) {
+    static_cast<std::vector<int64_t>*>(this)->operator=(v);
+    return *this;
+}
+
+int64_t SizeVector::NumElements() const {
+    if (this->size() == 0) {
+        return 1;
+    }
+    return std::accumulate(
+            this->begin(), this->end(), 1LL,
+            [this](const int64_t& lhs, const int64_t& rhs) -> int64_t {
+                if (lhs < 0 || rhs < 0) {
+                    utility::LogError(
+                            "Shape {} cannot contain negative dimensions.",
+                            this->ToString());
+                }
+                return std::multiplies<int64_t>()(lhs, rhs);
+            });
+}
+
+int64_t SizeVector::GetLength() const {
+    if (size() == 0) {
+        utility::LogError("Cannot get length of a 0-dimensional shape.");
+    } else {
+        return operator[](0);
+    }
+}
+
+std::string SizeVector::ToString() const { return fmt::format("{}", *this); }
+
+void SizeVector::AssertCompatible(const DynamicSizeVector& dsv,
+                                  const std::string msg) const {
+    if (!IsCompatible(dsv)) {
+        if (msg.empty()) {
+            utility::LogError("Shape {} is not compatible with {}.", ToString(),
+                              dsv.ToString());
+        } else {
+            utility::LogError("Shape {} is not compatible with {}: {}",
+                              ToString(), dsv.ToString(), msg);
+        }
+    }
+}
+
+bool SizeVector::IsCompatible(const DynamicSizeVector& dsv) const {
+    if (size() != dsv.size()) {
+        return false;
+    }
+    for (size_t i = 0; i < size(); ++i) {
+        if (dsv[i].has_value() && dsv[i].value() != at(i)) {
+            return false;
+        }
+    }
+    return true;
+}
 
 }  // namespace core
 }  // namespace open3d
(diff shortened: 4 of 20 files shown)


stotko (changes requested):
> Reviewed 18 of 18 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3954)* status: all files reviewed, 6 unresolved discussions (waiting on @yxlao)

___
*[cpp/open3d/core/SizeVector.h, line 51 at r1](https://reviewable.io/reviews/isl-org/open3d/3954#-MhO3DoKBfz6KRqElbYx:-MhO3DoLAYhZAJpX0hU8:b7du805) ([raw file](https://github.com/isl-org/open3d/blob/4b393ff6bceb213310798ee2b573358aa77dd69c/cpp/open3d/core/SizeVector.h#L51)):*
> ```objc
> /// ```
> class DynamicSizeVector : public std::vector<utility::optional<int64_t>> {
>     using optint64_t = utility::optional<int64_t>;
> ```

Although this PR does not touch these lines: I think `optint64_t` is a bit confusing. My first thought was that it is something like the typedefs here: https://en.cppreference.com/w/cpp/header/cstdint

Removing the typedef should make it clear that this is not some integer but actually an `optional` class instance.
___
*[cpp/open3d/core/SizeVector.h, line 70 at r1](https://reviewable.io/reviews/isl-org/open3d/3954#-MhO3tmaDUtWddfnDK00:-MhO3tmaDUtWddfnDK01:b-1ekc3p) ([raw file](https://github.com/isl-org/open3d/blob/4b393ff6bceb213310798ee2b573358aa77dd69c/cpp/open3d/core/SizeVector.h#L70)):*
> ```objc
>         : std::vector<optint64_t>(first, last) {}
> 
>     DynamicSizeVector(const SizeVector& dim_sizes);
> ```

Since this cannot be inlined as `SizeVector` will be defined after this class is parsed, would it make sense to move all member definitions into `SizeVector.cpp`? In other words, is it worth to fully inline everything?
___
*[cpp/open3d/core/SizeVector.h, line 70 at r1](https://reviewable.io/reviews/isl-org/open3d/3954#-MhONIkR8xkzKtL9uACH:-MhONIkR8xkzKtL9uACI:b193udz) ([raw file](https://github.com/isl-org/open3d/blob/4b393ff6bceb213310798ee2b573358aa77dd69c/cpp/open3d/core/SizeVector.h#L70)):*
> ```objc
>         : std::vector<optint64_t>(first, last) {}
> 
>     DynamicSizeVector(const SizeVector& dim_sizes);
> ```

Make this `explicit` to avoid accidental implicit conversions.
___
*[cpp/open3d/core/SizeVector.cpp, line 29 at r1](https://reviewable.io/reviews/isl-org/open3d/3954#-MhO4GXWB7Ly0esoT4HX:-MhO4GXWB7Ly0esoT4HY:b-bjnwyr) ([raw file](https://github.com/isl-org/open3d/blob/4b393ff6bceb213310798ee2b573358aa77dd69c/cpp/open3d/core/SizeVector.cpp#L29)):*
> ```cpp
> #include "open3d/core/SizeVector.h"
> 
> #include <cstddef>
> ```

Why do we need all of these includes?
___
*[cpp/open3d/core/SizeVector.cpp, line 35 at r1](https://reviewable.io/reviews/isl-org/open3d/3954#-MhONq3W7Q16EEDZj2cu:-MhONq3W7Q16EEDZj2cv:brhumzt) ([raw file](https://github.com/isl-org/open3d/blob/4b393ff6bceb213310798ee2b573358aa77dd69c/cpp/open3d/core/SizeVector.cpp#L35)):*
> ```cpp
> #include <vector>
> 
> #include "open3d/utility/Logging.h"
> ```

Same here.
___
*[cpp/open3d/core/Tensor.cpp, line 1584 at r1](https://reviewable.io/reviews/isl-org/open3d/3954#-MhOMzB0BRkf9-JToyZv:-MhOMzB1A_9-PQu6jKCA:b-ud6bu4) ([raw file](https://github.com/isl-org/open3d/blob/4b393ff6bceb213310798ee2b573358aa77dd69c/cpp/open3d/core/Tensor.cpp#L1584)):*
> ```cpp
>     } else {
>         SizeVector static_shape(expected_shape.size());
>         std::transform(expected_shape.begin(), expected_shape.end(),
> ```

This could be moved to a new constructor `explicit SizeVector(const DynamicSizeVector&)`. That way we would have bidirectional conversion functions between `SizeVector` and `DynamicSizeVector`.

yxlao (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3954)* status: all files reviewed, 6 unresolved discussions (waiting on @stotko and @yxlao)

___
*[cpp/open3d/core/SizeVector.h, line 70 at r1](https://reviewable.io/reviews/isl-org/open3d/3954#-MhONIkR8xkzKtL9uACH:-MhOzHgD8T_KnFO-MGR5:b-y9j5cq) ([raw file](https://github.com/isl-org/open3d/blob/4b393ff6bceb213310798ee2b573358aa77dd69c/cpp/open3d/core/SizeVector.h#L70)):*

Make this `explicit` to avoid accidental implicit conversions.
</blockquote></details>

`SizeVector` is a special case of `DynamicSizeVector`.  The motivation for this PR is to avoid  the need for `AssertShapeCompatible` and to support all of the following usages with the same function name `AssertShape`:
```cpp
t.AssertShape({100, 3});
t.AssertShape({utility::nullopt, 3});

t.AssertShape(SizeVector{100, 3});
t.AssertShape(DynamicSizeVector{utility::nullopt, 3});
```

My first approach was function overloading:
```cpp
void Tensor::AssertShape(const SizeVector& shape);
void Tensor::AssertShape(const DynamicSizeVector& shape);
```
But the compiler complains about ambiguity since an initializer list like `{100, 3}` can be interpreted in both ways. That's why we came to this alternative solution of allowing `SizeVector` to implicitly convert to `DynamicSizeVector`. Do you see other possible solutions?

stotko (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3954)* status: all files reviewed, 6 unresolved discussions (waiting on @yxlao)

___
*[cpp/open3d/core/SizeVector.h, line 70 at r1](https://reviewable.io/reviews/isl-org/open3d/3954#-MhONIkR8xkzKtL9uACH:-MhP8_ip0RU423M8k47q:bewvkcx) ([raw file](https://github.com/isl-org/open3d/blob/4b393ff6bceb213310798ee2b573358aa77dd69c/cpp/open3d/core/SizeVector.h#L70)):*

`SizeVector` is a special case of `DynamicSizeVector`.  The motivation for this PR is to avoid  the need for `AssertShapeCompatible` and to support all of the following usages with the same function name `AssertShape`:
```cpp
t.AssertShape({100, 3});
t.AssertShape({utility::nullopt, 3});

t.AssertShape(SizeVector{100, 3});
t.AssertShape(DynamicSizeVector{utility::nullopt, 3});
```

My first approach was function overloading:
```cpp
void Tensor::AssertShape(const SizeVector& shape);
void Tensor::AssertShape(const DynamicSizeVector& shape);
```
But the compiler complains about ambiguity since an initializer list like `{100, 3}` can be interpreted in both ways. That's why we came to this alternative solution of allowing `SizeVector` to implicitly convert to `DynamicSizeVector`. Do you see other possible solutions?
</blockquote></details>

I see. In that case I am wondering: What is the use case of `DynamicSizeVector`? Is it designed to *partially* check the size, i.e. only check a subset of the dimensions? `SizeVector` seems to be used throughout the complete code base whereas `DynamicSizeVector` does not seem to be adopted (yet).

In case the above assumption is correct: Could we exclusively provide `void Tensor::AssertShape(const DynamicSizeVector& shape);` and only provide (implicit) conversion functions from `SizeVector` to `DynamicSizeVector`?

yxlao (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3954)* status: 15 of 18 files reviewed, 6 unresolved discussions (waiting on @stotko)

___
*[cpp/open3d/core/SizeVector.h, line 51 at r1](https://reviewable.io/reviews/isl-org/open3d/3954#-MhO3DoKBfz6KRqElbYx:-MhPBAcS9c3QLH_F-F9o:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/4b393ff6bceb213310798ee2b573358aa77dd69c/cpp/open3d/core/SizeVector.h#L51)):*

Although this PR does not touch these lines: I think `optint64_t` is a bit confusing. My first thought was that it is something like the typedefs here: https://en.cppreference.com/w/cpp/header/cstdint

Removing the typedef should make it clear that this is not some integer but actually an `optional` class instance.
</blockquote></details>

Done.
___
*[cpp/open3d/core/SizeVector.h, line 70 at r1](https://reviewable.io/reviews/isl-org/open3d/3954#-MhO3tmaDUtWddfnDK00:-MhPDNbW398mEExzR_Mq:bmpdnt8) ([raw file](https://github.com/isl-org/open3d/blob/4b393ff6bceb213310798ee2b573358aa77dd69c/cpp/open3d/core/SizeVector.h#L70)):*

Since this cannot be inlined as `SizeVector` will be defined after this class is parsed, would it make sense to move all member definitions into `SizeVector.cpp`? In other words, is it worth to fully inline everything?
</blockquote></details>

Moved to SizeVector.cpp
___
*[cpp/open3d/core/SizeVector.h, line 70 at r1](https://reviewable.io/reviews/isl-org/open3d/3954#-MhONIkR8xkzKtL9uACH:-MhPA9jeDL_4MrpGX85c:b-4sdugv) ([raw file](https://github.com/isl-org/open3d/blob/4b393ff6bceb213310798ee2b573358aa77dd69c/cpp/open3d/core/SizeVector.h#L70)):*
>  Is it designed to partially check the size, i.e. only check a subset of the dimensions?

`DynamicSizeVector` is currently used to check the shape of a tensor. For example, we can check the shape of a tensor matches `{N, 3}` where `N` is can be any value. 

> In case the above assumption is correct: Could we exclusively provide void Tensor::AssertShape(const DynamicSizeVector& shape); and only provide (implicit) conversion functions from SizeVector to DynamicSizeVector?

Yes. That is the original implementation of this PR. There's only one `Tensor::AssertShape`.
___
*[cpp/open3d/core/SizeVector.cpp, line 29 at r1](https://reviewable.io/reviews/isl-org/open3d/3954#-MhO4GXWB7Ly0esoT4HX:-MhPDC_XEAyt1RkRr18i:b-51pj00) ([raw file](https://github.com/isl-org/open3d/blob/4b393ff6bceb213310798ee2b573358aa77dd69c/cpp/open3d/core/SizeVector.cpp#L29)):*

Why do we need all of these includes?
</blockquote></details>

Simplified. Also moved most implementations to this file.
___
*[cpp/open3d/core/SizeVector.cpp, line 35 at r1](https://reviewable.io/reviews/isl-org/open3d/3954#-MhONq3W7Q16EEDZj2cu:-MhPDJbuB-gJ79JNxMvb:b-51pj00) ([raw file](https://github.com/isl-org/open3d/blob/4b393ff6bceb213310798ee2b573358aa77dd69c/cpp/open3d/core/SizeVector.cpp#L35)):*

Same here.
</blockquote></details>

Simplified. Also moved most implementations to this file.
___
*[cpp/open3d/core/Tensor.cpp, line 1584 at r1](https://reviewable.io/reviews/isl-org/open3d/3954#-MhOMzB0BRkf9-JToyZv:-MhPEU5O6RmB8wtstSqE:b226196) ([raw file](https://github.com/isl-org/open3d/blob/4b393ff6bceb213310798ee2b573358aa77dd69c/cpp/open3d/core/Tensor.cpp#L1584)):*

This could be moved to a new constructor `explicit SizeVector(const DynamicSizeVector&)`. That way we would have bidirectional conversion functions between `SizeVector` and `DynamicSizeVector`.
</blockquote></details>

Done. Note, we now have
```cpp
// This is not explicit
DynamicSizeVector::DynamicSizeVector(const SizeVector& dim_sizes);

// This is explicit
explicit SizeVector::SizeVector(const DynamicSizeVector& dim_sizes);
```

stotko (approved):
> Reviewed 2 of 3 files at r2, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3954)* status: all files reviewed, 2 unresolved discussions (waiting on @stotko and @yxlao)

___
*[cpp/open3d/core/SizeVector.h, line 93 at r3](https://reviewable.io/reviews/isl-org/open3d/3954#-MhSUi0L9scVfZ6Wj39B:-MhSUi0M4hapfjARRoG4:bu4prd0) ([raw file](https://github.com/isl-org/open3d/blob/028c993eebdee796dcc3e6107c77c486a72e9ee6/cpp/open3d/core/SizeVector.h#L93)):*
> ```objc
>         : std::vector<int64_t>(first, last) {}
> 
>     explicit SizeVector(const DynamicSizeVector& dim_sizes);
> ```

Just another idea: We could move this to `DynamicSizeVector` and define a function `explicit operator DynamicSizeVector() const` there. This would keep all functionality related to the `DynamicSizeVector` at a single place without any direct coupling to `SizeVector`. This might require changing `SizeVector static_shape(expected_shape);` to `SizeVector static_shape = expected_shape;`.

Changing to above proposal or keeping the current solution is both fine, however.

yxlao (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3954)* status: 17 of 20 files reviewed, 2 unresolved discussions (waiting on @stotko)

___
*[cpp/open3d/core/SizeVector.h, line 93 at r3](https://reviewable.io/reviews/isl-org/open3d/3954#-MhSUi0L9scVfZ6Wj39B:-MhW_Pbg5AxJunuF1gjI:b-em5v5a) ([raw file](https://github.com/isl-org/open3d/blob/028c993eebdee796dcc3e6107c77c486a72e9ee6/cpp/open3d/core/SizeVector.h#L93)):*

Just another idea: We could move this to `DynamicSizeVector` and define a function `explicit operator DynamicSizeVector() const` there. This would keep all functionality related to the `DynamicSizeVector` at a single place without any direct coupling to `SizeVector`. This might require changing `SizeVector static_shape(expected_shape);` to `SizeVector static_shape = expected_shape;`.

Changing to above proposal or keeping the current solution is both fine, however.
</blockquote></details>

Done. Added `DynamicSizeVector::ToSizeVector()`. The `ToXXX` syntax is also used in other locations of Open3D.

PR #3969, opened 2021-08-23
Fix typo of online-processing.py

muskie82:
TensorMap doesn't have "points" key. It has "positions".
https://github.com/isl-org/Open3D/blob/4f70ff19da8f65380a3835c10cbadb1a01c1b80e/cpp/open3d/t/geometry/TensorMap.h#L43

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/3969)

### DIFF
--- a/examples/python/gui/online-processing.py
+++ b/examples/python/gui/online-processing.py
@@ -174,7 +174,7 @@ class PipelineModel:
                 log.warning(f"No valid depth data in frame {frame_id})")
                 continue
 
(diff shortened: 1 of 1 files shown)


reyanshsolis (approved):

ssheorey (approved):
Thanks @muskie82 for catching this. There was a recent API change where 'points' was replaced with 'positions' for better consistency with other attribute names.

PR #3981, opened 2021-08-26
Legacy PTS Read IO : Fixed to support generic types supported in Tensor PTS Write IO

reyanshsolis:
Changes:
- Tensor PTS IO supported writing the data in formats such as `X Y Z I R G B`, `X Y Z R G B`, `X Y Z I`, `X Y Z`, however, legacy assumed the data to be either in `X Y Z I R G B` and `X Y Z` form. So, the file is written by tensor pts io (for ex. in `X Y Z R G B`), was not supported / wrongly read, by legacy io and therefore current visualizer.

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/3981)


errissa (changes requested):
> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3981)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @reyanshsolis)

*[a discussion](https://reviewable.io/reviews/isl-org/open3d/3981#-MiWxG01E8vUwv8yT5zq:-MiWxG01E8vUwv8yT5zr:b-4wdbc3) (no related file):*
This change seems to work but I have a question and a comment:

1) Why only support XYZIRGB and XYZRGB? The T Geometry version supports several other formats not supported here. If we're already going through the trouble of modifying the legacy loader we might as well support the other formats.

2) I would restructure the `while` loop to match as much as possible the T Geometry version for the sake of consistency. See comment in code below.

---
___
*[cpp/open3d/io/file_format/FilePTS.cpp, line 69 at r1](https://reviewable.io/reviews/isl-org/open3d/3981#-MiWx-qy60i_oDjOox-F:-MiWx-qzCbmL7YI4zkam:b-6tgnxs) ([raw file](https://github.com/isl-org/open3d/blob/d38d2a1a698248a5f8bcc6c664f49290c0c6d71d/cpp/open3d/io/file_format/FilePTS.cpp#L69)):*
> ```cpp
>         size_t idx = 0;
>         while (idx < num_of_pts && (line_buffer = file.ReadLine())) {
>             if (num_of_fields == 0) {
> ```

I realize this code did not change, however, there is no need to have this conditional check inside the loop. See the t geometry implementation of PTS loading.

reyanshsolis (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3981)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @errissa)

*[a discussion](https://reviewable.io/reviews/isl-org/open3d/3981#-MiWxG01E8vUwv8yT5zq:-Mj8aIjF7vltatg_eW9F:b-v7eu8c) (no related file):*

This change seems to work but I have a question and a comment:

1) Why only support XYZIRGB and XYZRGB? The T Geometry version supports several other formats not supported here. If we're already going through the trouble of modifying the legacy loader we might as well support the other formats.

2) I would restructure the `while` loop to match as much as possible the T Geometry version for the sake of consistency. See comment in code below.
</blockquote></details>

I tried to review the possibility in the code. Tensor only supports `intensity` attribute additional to `legacy`, however in `legacy` we can not add custom attribute/intensity attribute, therefore other format support can not be supported in legacy.
Only for legacy read, the strategy is to read and ignore intensity attribute, just to support the file formats of PTS.

Restructured the while loop by moving the `num_fields` section outside, like `tensor`.

---
___
*[cpp/open3d/io/file_format/FilePTS.cpp, line 69 at r1](https://reviewable.io/reviews/isl-org/open3d/3981#-MiWx-qy60i_oDjOox-F:-Mj8aAz1-8X9_7AwF8eb:bd0hv8l) ([raw file](https://github.com/isl-org/open3d/blob/d38d2a1a698248a5f8bcc6c664f49290c0c6d71d/cpp/open3d/io/file_format/FilePTS.cpp#L69)):*

I realize this code did not change, however, there is no need to have this conditional check inside the loop. See the t geometry implementation of PTS loading.
</blockquote></details>

Done. Moved it outside while loop like tensor implementation.

errissa (approved):
> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3981)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @reyanshsolis)

PR #3986, opened 2021-08-27
Refactor HashMap and backends

theNded:
- Change Block and BlockHash to MiniVec and MiniVecHash
- Remove duplicate functions in DeviceHashBackends and collect shared code snippets in the HashMap caller
- Dispatch value pointers to vectorize copy of value chunks

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/3986)

### DIFF
--- a/cpp/benchmarks/core/HashMap.cpp
+++ b/cpp/benchmarks/core/HashMap.cpp
@@ -220,11 +220,11 @@ void HashClearInt(benchmark::State& state,
     }
 }
 
-void HashRehashInt(benchmark::State& state,
-                   int capacity,
-                   int duplicate_factor,
-                   const Device& device,
-                   const HashBackendType& backend) {
+void HashReserveInt(benchmark::State& state,
+                    int capacity,
+                    int duplicate_factor,
+                    const Device& device,
+                    const HashBackendType& backend) {
     int slots = std::max(1, capacity / duplicate_factor);
     HashData<int, int> data(capacity, slots);
 
@@ -254,7 +254,7 @@ void HashRehashInt(benchmark::State& state,
         cuda::Synchronize(device);
         state.ResumeTiming();
 
-        hashmap.Rehash(2 * capacity);
+        hashmap.Reserve(2 * capacity);
 
         cuda::Synchronize(device);
         state.PauseTiming();
@@ -445,11 +445,11 @@ void HashClearInt3(benchmark::State& state,
     }
 }
 
-void HashRehashInt3(benchmark::State& state,
-                    int capacity,
-                    int duplicate_factor,
-                    const Device& device,
-                    const HashBackendType& backend) {
+void HashReserveInt3(benchmark::State& state,
+                     int capacity,
+                     int duplicate_factor,
+                     const Device& device,
+                     const HashBackendType& backend) {
     int slots = std::max(1, capacity / duplicate_factor);
     HashData<Int3, int> data(capacity, slots);
 
@@ -482,7 +482,7 @@ void HashRehashInt3(benchmark::State& state,
         cuda::Synchronize(device);
         state.ResumeTiming();
 
-        hashmap.Rehash(2 * capacity);
+        hashmap.Reserve(2 * capacity);
 
         cuda::Synchronize(device);
         state.PauseTiming();
@@ -542,8 +542,8 @@ ENUM_BM_BACKEND(HashFindInt)
 ENUM_BM_BACKEND(HashFindInt3)
 ENUM_BM_BACKEND(HashClearInt)
 ENUM_BM_BACKEND(HashClearInt3)
-ENUM_BM_BACKEND(HashRehashInt)
-ENUM_BM_BACKEND(HashRehashInt3)
+ENUM_BM_BACKEND(HashReserveInt)
+ENUM_BM_BACKEND(HashReserveInt3)
 
 }  // namespace core
 }  // namespace open3d
--- a/cpp/open3d/core/hashmap/CPU/CreateCPUHashBackend.cpp
+++ b/cpp/open3d/core/hashmap/CPU/CreateCPUHashBackend.cpp
@@ -59,8 +59,9 @@ std::shared_ptr<DeviceHashBackend> CreateCPUHashBackend(
 
     std::shared_ptr<DeviceHashBackend> device_hashmap_ptr;
     DISPATCH_DTYPE_AND_DIM_TO_TEMPLATE(key_dtype, dim, [&] {
-        device_hashmap_ptr = std::make_shared<TBBHashBackend<key_t, hash_t>>(
-                init_capacity, key_dsize, value_dsizes, device);
+        device_hashmap_ptr =
+                std::make_shared<TBBHashBackend<key_t, hash_t, eq_t>>(
+                        init_capacity, key_dsize, value_dsizes, device);
     });
     return device_hashmap_ptr;
 }
--- a/cpp/open3d/core/hashmap/CPU/TBBHashBackend.h
+++ b/cpp/open3d/core/hashmap/CPU/TBBHashBackend.h
@@ -37,7 +37,7 @@
 
 namespace open3d {
 namespace core {
-template <typename Key, typename Hash>
+template <typename Key, typename Hash, typename Eq>
 class TBBHashBackend : public DeviceHashBackend {
 public:
     TBBHashBackend(int64_t init_capacity,
@@ -77,13 +77,13 @@ public:
     std::vector<int64_t> BucketSizes() const override;
     float LoadFactor() const override;
 
-    std::shared_ptr<tbb::concurrent_unordered_map<Key, buf_index_t, Hash>>
+    std::shared_ptr<tbb::concurrent_unordered_map<Key, buf_index_t, Hash, Eq>>
     GetImpl() const {
         return impl_;
     }
 
 protected:
-    std::shared_ptr<tbb::concurrent_unordered_map<Key, buf_index_t, Hash>>
+    std::shared_ptr<tbb::concurrent_unordered_map<Key, buf_index_t, Hash, Eq>>
             impl_;
 
     std::shared_ptr<CPUHashBackendBufferAccessor> buffer_accessor_;
@@ -97,8 +97,8 @@ protected:
     void Allocate(int64_t capacity);
 };
 
-template <typename Key, typename Hash>
-TBBHashBackend<Key, Hash>::TBBHashBackend(
+template <typename Key, typename Hash, typename Eq>
+TBBHashBackend<Key, Hash, Eq>::TBBHashBackend(
         int64_t init_capacity,
         int64_t key_dsize,
         const std::vector<int64_t>& value_dsizes,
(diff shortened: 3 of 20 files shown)


theNded (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3986)* status: 10 of 16 files reviewed, 4 unresolved discussions (waiting on @stotko and @theNded)

___
*[cpp/open3d/core/hashmap/DeviceHashBackend.h, line 110 at r2](https://reviewable.io/reviews/isl-org/open3d/3986#-Mi5tfrm3ax1PelIxvG1:-Mi9YcqZ0eKaUEVIhkep:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/3664084418fb9213f5a7151dab50f1f6e7e2f987/cpp/open3d/core/hashmap/DeviceHashBackend.h#L110)):*

I think only providing `capacity` as the argument and infer `bucket_count` as some implementation-specific detail would be better. TBB and stdgpu infer the value internally, and Slab uses `bucket_count = 2 * capacity`. Also see the discussion about `Rehash()` below.
</blockquote></details>

Done.
___
*[cpp/open3d/core/hashmap/Dispatch.h, line 103 at r2](https://reviewable.io/reviews/isl-org/open3d/3986#-Mi5ngUtEG9AnK0Wsbmg:-Mi9X7k12fDHzFL6oUPw:b-v6s1o3) ([raw file](https://github.com/isl-org/open3d/blob/3664084418fb9213f5a7151dab50f1f6e7e2f987/cpp/open3d/core/hashmap/Dispatch.h#L103)):*

These typedefs only seem to be used in the dispatch macro above. Could we remove them?
</blockquote></details>

Removed. 
___
*[cpp/open3d/core/hashmap/HashMap.cpp, line 66 at r2](https://reviewable.io/reviews/isl-org/open3d/3986#-Mi5sBQD58GF_owZjOLs:-Mi9YFU8AedyJLFXE7od:b-yzcct8) ([raw file](https://github.com/isl-org/open3d/blob/3664084418fb9213f5a7151dab50f1f6e7e2f987/cpp/open3d/core/hashmap/HashMap.cpp#L66)):*

I think `Rehash()` makes more sense with specifying the new expected *capacity* rather than *buckets*. The latter could be some implementation-specific detail that is inferred from the requested capacity.

Since that is likely a larger change (with some implications), we could discuss this first. Deferring to a future PR would also be okay.
</blockquote></details>

I agree, but the std unordered_map takes the bucket count as input: https://en.cppreference.com/w/cpp/container/unordered_map/rehash.
Maybe we can rename it as resize, and revisit the functionalities for rehash?

stotko (changes requested):
> Reviewed 6 of 6 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3986)* status: all files reviewed, 3 unresolved discussions (waiting on @theNded)

___
*[cpp/open3d/core/hashmap/Dispatch.h, line 79 at r3](https://reviewable.io/reviews/isl-org/open3d/3986#-MiLG6XU-fd-2agQS7k4:-MiLG6XV4v8-97Vs58AZ:b47ynj1) ([raw file](https://github.com/isl-org/open3d/blob/59ae9c94311d3ce4a2c07948c214f71326ddf023/cpp/open3d/core/hashmap/Dispatch.h#L79)):*
> ```objc
>     [&] {                                              \
>         if (DIVISOR == 16) {                           \
>             using block_t = int4;                      \
> ```

Here, the removed typedefs are still used. Note that theses are the CUDA vector types and may also work here. But we should definitely use `MiniVec<T, N>` here instead.
___
*[cpp/open3d/core/hashmap/HashMap.cpp, line 66 at r2](https://reviewable.io/reviews/isl-org/open3d/3986#-Mi5sBQD58GF_owZjOLs:-MiLGmbp0YylcP6hLGiH:b-38hpqa) ([raw file](https://github.com/isl-org/open3d/blob/3664084418fb9213f5a7151dab50f1f6e7e2f987/cpp/open3d/core/hashmap/HashMap.cpp#L66)):*

I agree, but the std unordered_map takes the bucket count as input: https://en.cppreference.com/w/cpp/container/unordered_map/rehash.
Maybe we can rename it as resize, and revisit the functionalities for rehash?
</blockquote></details>

I think `reserve` is a better name as it is also exposed in `std::unordered_map` for exactly this purpose of specifying the new capacity: https://en.cppreference.com/w/cpp/container/unordered_map/reserve. I am leaning towards dropping `Rehash` and replace it by a `Reserve` function.

theNded (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3986)* status: 8 of 20 files reviewed, 3 unresolved discussions (waiting on @stotko and @theNded)

___
*[cpp/open3d/core/hashmap/Dispatch.h, line 79 at r3](https://reviewable.io/reviews/isl-org/open3d/3986#-MiLG6XU-fd-2agQS7k4:-MiMM2V31rXdu2LN_kSL:b-97g9bn) ([raw file](https://github.com/isl-org/open3d/blob/59ae9c94311d3ce4a2c07948c214f71326ddf023/cpp/open3d/core/hashmap/Dispatch.h#L79)):*

Here, the removed typedefs are still used. Note that theses are the CUDA vector types and may also work here. But we should definitely use `MiniVec<T, N>` here instead.
</blockquote></details>

They are used here https://github.com/isl-org/Open3D/blob/wei/hashmap-upgrade/cpp/open3d/core/hashmap/CUDA/StdGPUHashBackend.h#L391, https://github.com/isl-org/Open3D/blob/wei/hashmap-upgrade/cpp/open3d/core/hashmap/CUDA/StdGPUHashBackend.h#L360
in CUDA to increase the performance of copy -- this is because memcpy or byte-by-byte copying is slow in CUDA kernels, and int4 is a built-in primitive that allows fast copying. 
We can use MiniVec instead, but the underlying copy operation might be reduced to smaller primitives (int).
This dispatch pattern is not used in C++ since memcpy in C++ is relatively fast. 
https://github.com/isl-org/Open3D/blob/wei/hashmap-upgrade/cpp/open3d/core/hashmap/CPU/TBBHashBackend.h#L225

Do you think it is reasonable to wrap this dispatch in the CUDACC macro and keep it as it is?
___
*[cpp/open3d/core/hashmap/HashMap.cpp, line 66 at r2](https://reviewable.io/reviews/isl-org/open3d/3986#-Mi5sBQD58GF_owZjOLs:-MiMNl03-sIuW2jbE5FS:b-4t9upe) ([raw file](https://github.com/isl-org/open3d/blob/3664084418fb9213f5a7151dab50f1f6e7e2f987/cpp/open3d/core/hashmap/HashMap.cpp#L66)):*

I think `reserve` is a better name as it is also exposed in `std::unordered_map` for exactly this purpose of specifying the new capacity: https://en.cppreference.com/w/cpp/container/unordered_map/reserve. I am leaning towards dropping `Rehash` and replace it by a `Reserve` function.
</blockquote></details>

Sounds reasonable, upgraded with interface and doc changes.

stotko (commented):
> Reviewed 12 of 12 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3986)* status: all files reviewed, 3 unresolved discussions (waiting on @theNded)

___
*[cpp/open3d/core/hashmap/Dispatch.h, line 79 at r3](https://reviewable.io/reviews/isl-org/open3d/3986#-MiLG6XU-fd-2agQS7k4:-MiMcCe18LH2kzHbgkRe:bfinpmb) ([raw file](https://github.com/isl-org/open3d/blob/59ae9c94311d3ce4a2c07948c214f71326ddf023/cpp/open3d/core/hashmap/Dispatch.h#L79)):*

They are used here https://github.com/isl-org/Open3D/blob/wei/hashmap-upgrade/cpp/open3d/core/hashmap/CUDA/StdGPUHashBackend.h#L391, https://github.com/isl-org/Open3D/blob/wei/hashmap-upgrade/cpp/open3d/core/hashmap/CUDA/StdGPUHashBackend.h#L360
in CUDA to increase the performance of copy -- this is because memcpy or byte-by-byte copying is slow in CUDA kernels, and int4 is a built-in primitive that allows fast copying. 
We can use MiniVec instead, but the underlying copy operation might be reduced to smaller primitives (int).
This dispatch pattern is not used in C++ since memcpy in C++ is relatively fast. 
https://github.com/isl-org/Open3D/blob/wei/hashmap-upgrade/cpp/open3d/core/hashmap/CPU/TBBHashBackend.h#L225

Do you think it is reasonable to wrap this dispatch in the CUDACC macro and keep it as it is?
</blockquote></details>

I see. Wrapping them by `__CUDACC__` and adding a comment somewhere (probably above `DISPATCH_DIVISOR_SIZE_TO_BLOCK_T`) should be sufficient.
___
*[cpp/open3d/core/hashmap/HashMap.cpp, line 69 at r4](https://reviewable.io/reviews/isl-org/open3d/3986#-MiMcnYv2g12RD54SOWF:-MiMcnYv2g12RD54SOWG:by3cuzn) ([raw file](https://github.com/isl-org/open3d/blob/8a66cfe91e7fd4e490f75e3e9857336248117e9a/cpp/open3d/core/hashmap/HashMap.cpp#L69)):*
> ```cpp
>     int64_t count = Size();
>     if (capacity <= count) {
>         utility::LogDebug("Target capacity smaller then current size, abort.");
> ```

There is no `return` statement here, but the message indicates that we should stop at this point. Where is this handled?

theNded (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3986)* status: 18 of 20 files reviewed, 3 unresolved discussions (waiting on @stotko and @theNded)

___
*[cpp/open3d/core/hashmap/Dispatch.h, line 79 at r3](https://reviewable.io/reviews/isl-org/open3d/3986#-MiLG6XU-fd-2agQS7k4:-MiMhAVa-QinkxjTMzE6:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/59ae9c94311d3ce4a2c07948c214f71326ddf023/cpp/open3d/core/hashmap/Dispatch.h#L79)):*

I see. Wrapping them by `__CUDACC__` and adding a comment somewhere (probably above `DISPATCH_DIVISOR_SIZE_TO_BLOCK_T`) should be sufficient.
</blockquote></details>

Done.
___
*[cpp/open3d/core/hashmap/HashMap.cpp, line 69 at r4](https://reviewable.io/reviews/isl-org/open3d/3986#-MiMcnYv2g12RD54SOWF:-MiMhBF_1850eOX_rH1C:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/8a66cfe91e7fd4e490f75e3e9857336248117e9a/cpp/open3d/core/hashmap/HashMap.cpp#L69)):*

There is no `return` statement here, but the message indicates that we should stop at this point. Where is this handled?
</blockquote></details>

Done.

stotko (approved):
> Reviewed 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3986)* status: all files reviewed, 1 unresolved discussion (waiting on @theNded)

PR #3987, opened 2021-08-27
Tensor Registration - `GetInformationMatrixFromPointClouds`

reyanshsolis:
Changes:
- Renamed `ComputeTransform*` kernel codes, to `Registration*` kernel codes, for consistency with other kernel code naming patterns.
- Added `GetInformationMatrixFromPointClouds` function, kernels, unit-tests, pybinds.

The contribution has been derived from https://github.com/theNded/Open3D/blob/cuda/src/Cuda/Registration/RegistrationCuda.h

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/3987)

### DIFF
--- a/cpp/open3d/t/pipelines/CMakeLists.txt
+++ b/cpp/open3d/t/pipelines/CMakeLists.txt
@@ -3,8 +3,8 @@ add_subdirectory(kernel)
 add_library(tpipelines OBJECT $<TARGET_OBJECTS:tpipelines_kernel>)
 
 target_sources(tpipelines PRIVATE
-    kernel/ComputeTransform.cpp
-    kernel/ComputeTransformCPU.cpp
+    kernel/Registration.cpp
+    kernel/RegistrationCPU.cpp
     kernel/FillInLinearSystem.cpp
     kernel/FillInLinearSystemCPU.cpp
     kernel/RGBDOdometry.cpp
@@ -14,7 +14,7 @@ target_sources(tpipelines PRIVATE
 
 if (BUILD_CUDA_MODULE)
     target_sources(tpipelines PRIVATE
-        kernel/ComputeTransformCUDA.cu
+        kernel/RegistrationCUDA.cu
         kernel/FillInLinearSystemCUDA.cu
         kernel/RGBDOdometryCUDA.cu
         kernel/TransformationConverter.cu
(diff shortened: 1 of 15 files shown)


theNded (changes requested):
> Reviewed 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3987)* status: all files reviewed, 1 unresolved discussion (waiting on @reyanshsolis)

___
*[cpp/open3d/t/pipelines/registration/Registration.cpp, line 363 at r2](https://reviewable.io/reviews/isl-org/open3d/3987#-Mi6sJElCiqJhIga70BQ:-Mi9ZLhd6M_hk9Zym0GO:b-i88yqn) ([raw file](https://github.com/isl-org/open3d/blob/d07734c23b7c83d26382864be43ae7c853893467/cpp/open3d/t/pipelines/registration/Registration.cpp#L363)):*

counts is of type `Int32`, counts.Sum({0}) gives us tensor of `{1, 1} Int32`, 
so either we can do static_cast<double>(counts.Sum({0}).Item<int32_t>()) or ` counts.Sum({0}).To(core::Float64).Item<double>()`
the later ensures that the tensor will be of type `Float64` before using `Item`, avoiding throw from `Item` (in case NNS switch from Int32 to Int64).
</blockquote></details>

But the number of correspondences is an integer, why should we convert to a float? If it was a legacy issue from my code then maybe we can fix it now?

reyanshsolis (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3987)* status: 14 of 15 files reviewed, 1 unresolved discussion (waiting on @theNded)

___
*[cpp/open3d/t/pipelines/registration/Registration.cpp, line 363 at r2](https://reviewable.io/reviews/isl-org/open3d/3987#-Mi6sJElCiqJhIga70BQ:-Mi9cwcg34Ll7CxQCwAQ:b-l2167r) ([raw file](https://github.com/isl-org/open3d/blob/d07734c23b7c83d26382864be43ae7c853893467/cpp/open3d/t/pipelines/registration/Registration.cpp#L363)):*

But the number of correspondences is an integer, why should we convert to a float? If it was a legacy issue from my code then maybe we can fix it now?
</blockquote></details>

oh my bad, sorry. I copied that part from the ICP code, where it was double because I used it to compute fitness (double). Fixed it now.

theNded (approved):
> Reviewed 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3987)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @reyanshsolis)

PR #3991, opened 2021-08-28
Fixed wrong maximum count character in strncpy.

### LINKED ISSUE #3990: [-Werror=stringop-truncation] error occurs due to strncpy (Fedora34 GCC11)
**Describe the bug**

I get a compilation error with [-Werror=stringop-truncation] in GCC11 on Fedora 34.
```
Consolidate compiler generated dependencies of target core
[ 35%] Building CXX object cpp/open3d/core/CMakeFiles/core.dir/Dtype.cpp.o
In file included from /usr/include/string.h:519,
                 from /usr/include/c++/11/cstring:42,
                 from /home/xxx/WorkSpace/open3d/Open3D/cpp/open3d/core/Dtype.h:29,
                 from /home/xxx/WorkSpace/open3d/Open3D/cpp/open3d/core/Dtype.cpp:27:
In function 'char* strncpy(char*, const char*, size_t)',
    inlined from 'open3d::core::Dtype::Dtype(open3d::core::Dtype::DtypeCode, int64_t, const string&)' at /home/xxx/WorkSpace/open3d/Open3D/cpp/open3d/core/Dtype.cpp:80:21:
/usr/include/bits/string_fortified.h:95:34: error: 'char* __builtin_strncpy(char*, const char*, long unsigned int)' specified bound 16 equals destination size [-Werror=stringop-truncation]
   95 |   return __builtin___strncpy_chk (__dest, __src, __len,
      |          ~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~
   96 |                                   __glibc_objsize (__dest));
      |                                   ~~~~~~~~~~~~~~~~~~~~~~~~~
cc1plus: all warnings being treated as errors
make[2]: *** [cpp/open3d/core/CMakeFiles/core.dir/build.make:104: cpp/open3d/core/CMakeFiles/core.dir/Dtype.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:2191: cpp/open3d/core/CMakeFiles/core.dir/all] Error 2
make: *** [Makefile:136: all] Error 2

```

**To Reproduce**
```
$ git clone https://github.com/isl-org/Open3D.git
$ cd Open3D/
$ mkdir build 
$ cd build
$ cmake ..
$ make
```

- Operating system: Fedora 34
- Python version: 3.9.6
- Open3D version: commit id 2a7c912b6e458ad571bee6e2942ca4a4940f2bd9
- Compiler version (if built from source): gcc 11.2.1

**Additional context**

- There are no compilation errors on Ubuntu 18.04.
- The problem is caused by specifying the wrong size for count in `strncpy`.
  https://github.com/isl-org/Open3D/blob/2a7c912b6e458ad571bee6e2942ca4a4940f2bd9/cpp/open3d/core/Dtype.cpp#L77-L81
- Since `max_name_len_` is specified, the compiler determines that truncation will occur, resulting in an error of [-Werror = stringop-truncation].
- It's a warning from GCC 8, but at least it's an error in GCC 11.  
  https://stackoverflow.com/questions/50198319/gcc-8-wstringop-truncation-what-is-the-good-practice
- The size excluding `\n` is compared immediately before. Therefore, `max_name_len_-1` should be specified for strncpy.

NobuoTsukamoto:
- Fixed #3990 .
- Fixed strncpy compilation error with GCC 11 on Fedora 34.

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/3991)

### DIFF
--- a/cpp/open3d/core/Dtype.cpp
+++ b/cpp/open3d/core/Dtype.cpp
@@ -72,8 +72,6 @@ const Dtype Bool = Dtype::Bool;
 
 Dtype::Dtype(DtypeCode dtype_code, int64_t byte_size, const std::string &name)
     : dtype_code_(dtype_code), byte_size_(byte_size) {
-    (void)dtype_code_;
-    (void)byte_size_;
     if (name.size() > max_name_len_ - 1) {
         utility::LogError("Name {} must be shorter.", name);
     } else {
--- a/cpp/open3d/core/Dtype.cpp
+++ b/cpp/open3d/core/Dtype.cpp
@@ -77,7 +77,7 @@ Dtype::Dtype(DtypeCode dtype_code, int64_t byte_size, const std::string &name)
     if (name.size() > max_name_len_ - 1) {
         utility::LogError("Name {} must be shorter.", name);
     } else {
-        std::strncpy(name_, name.c_str(), max_name_len_);
+        std::strncpy(name_, name.c_str(), max_name_len_ - 1);
     }
 }
 
--- a/cpp/open3d/core/Dtype.cpp
+++ b/cpp/open3d/core/Dtype.cpp
@@ -75,7 +75,8 @@ Dtype::Dtype(DtypeCode dtype_code, int64_t byte_size, const std::string &name)
     if (name.size() > max_name_len_ - 1) {
         utility::LogError("Name {} must be shorter.", name);
     } else {
-        std::strncpy(name_, name.c_str(), max_name_len_ - 1);
+        std::strncpy(name_, name.c_str(), max_name_len_);
+        name_[max_name_len_ - 1] = '\0';
     }
 }
 


reyanshsolis (approved):
Thanks for the fix!

NobuoTsukamoto:
- In some cases, `strncpy` does not provide null termination. For this reason, the modification of the `strncpy` argument count was not appropriate.
- I think `strcpy` should be used instead of `strncpy` because it does size checking.
- It was previously corrected from `strncpy` to `strcpy` by a static analysis tool pointing it out.
- For this reason, avoid the warning by inserting \n.

I referred to the following Red Hat Developer blog.
https://developers.redhat.com/blog/2018/05/24/detecting-string-truncation-with-gcc-8#

PR #3996, opened 2021-08-30
Introduce initial ISPC language support

stotko:
This PR introduces initial ISPC language support. Depending on the platform, either CMake's native first-class language support for ISPC (Makefile and Ninja generators) or a legacy emulation module is used.

#### Highlights:

- Full support for x86 target architectures. ARM support is also present, but disabled and untested due to missing prebuilt binaries of the ISPC compiler.
- Automatic cross-platform downloading of the prebuild ISPC compiler via the `Open3DFetchISPCCompiler` module. The environment variable `$ENV{ISPC}` is set to the path of the downloaded compiler. Respects user-provided paths to the ISPC compiler and skips downloading in that case.
- CMake language emulation module `Open3DISPC` with drop-in replacements for `enable_language()`, `add_library()`, `add_executable()` and `target_sources()` which are prefixed with `open3d_ispc_`. See the limitations section below for additional notes on how to use it.
- Setup of instruction sets similar to CUDA architectures via the `Open3DMakeISPCInstructionSets` module.
- Functionality to detect the chosen ISA by the internal ISPC runtime dispatcher in C++ code including a unit test.
- Support for `.isph` and `.ispc` files in the style checking scripts (both CMake and Python based).
- Support for `.isph` and `.ispc` files for generating `Doxygen`-based documentation.

#### Language Emulation Features:

- Support for `$ENV{ISPC}` and `CMAKE_ISPC_COMPILER` variables when enabling the ISPC language. This also sets the variables `CMAKE_ISPC_COMPILER`, `CMAKE_ISPC_COMPILER_ID` and `CMAKE_ISPC_COMPILER_VERSION`.
- Support for the following properties to build ISPC sources files:
    - `(CMAKE_)ISPC_OUTPUT_EXTENSION`
    - `(CMAKE_)ISPC_HEADER_DIRECTORY`
    - `(CMAKE_)ISPC_HEADER_SUFFIX`
    - `(CMAKE_)ISPC_FLAGS`
    - `INCLUDE_DIRECTORIES`
    - `COMPILE_DEFINITIONS`
    - `COMPILE_OPTIONS`
- Support for almost all functions that update the properties mentioned above. This includes `target_include_directories()`, `target_compile_definitions()`, `target_compile_options()`.
- Support for the generator expressions `BUILD_INTERFACE`, `$<COMPILE_LANGUAGE:ISPC>` and `$<COMPILE_LANG_AND_ID:ISPC,...>`.
- Support for compiling with multiple instruction sets. 
- Building in the correct subdirectories and printing the correct build information string (at least for Makefile generators).
- Support for automatic dependency scanning and handling (on Makefile generators only).
- Option `ISPC_FORCE_LEGACY` to force language emulation on all platforms.
- Option `ISPC_PRINT_LEGACY_COMPILE_COMMANDS` to print the generated build commands for debugging purposes.

#### Language Emulation Limitations:

- No support for `PUBLIC` or `INTERFACE` sources.
- No support for adding ISPC source files in `open3d_ispc_add_library()` and `open3d_ispc_add_executable()`.
- No support for adding ISPC source files in directories outside the one where the target has been created.
- ISPC source files must be specified as relative paths. Absolute paths are not supported.
- Properties must be set before before processing the directory where the target has been created is finished. All properties that are set after adding a ISPC source file will be ignored.
- Only first-order `INTERFACE` properties from linked libraries are populated. Higher orders of indirection are ignored.
- Generator expressions not listed above, i.e. "unsupported" ones, are always ignored.
- Dependency scanning and handling for ISPC source files does not work on non-Makefile generators.
- Object libraries do not contain compiled ISPC code in the `$<TARGET_OBJECTS:...>` generator expression. The workaround function `open3d_ispc_target_sources_TARGET_OBJECTS` must be used in this case to propagate all object files correctly.

#### Further Notes:

- Tested legacy language emulation support for `BUILD_SHARED` set to both `ON` and `OFF` on all CI platforms via `ISPC_FORCE_LEGACY=ON`. However, this is turned off by default to use the superior first-class language support by CMake.
- `ISPC_PRINT_LEGACY_COMPILE_COMMANDS` is set to `ON` by default to ease debugging of the emulation module. With increasing adoption of vectorized code written in ISPC, this increases the verbosity of the configuration output.

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/3996)


yxlao (changes requested):
> Reviewed 30 of 30 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3996)* status: all files reviewed, 4 unresolved discussions (waiting on @ssheorey and @stotko)

___
*[cmake/Open3DISPC.cmake, line 20 at r1](https://reviewable.io/reviews/isl-org/open3d/3996#-MihTvZ8AC_Gr8Ta7PC2:-MihTvZ8AC_Gr8Ta7PC3:b-vh14au) ([raw file](https://github.com/isl-org/open3d/blob/9e828719d360ca158c6fc52d12272a2a9b138594/cmake/Open3DISPC.cmake#L20)):*
> ```CMake
> 
> # Internal helper function.
> function(open3d_get_target_relative_object_dir target output_dir)
> ```

nit: for internal helper functions, can we add prefix or suffix such as `_` or `__` to indicate it is internal?
___
*[cpp/open3d/CMakeLists.txt, line 82 at r1](https://reviewable.io/reviews/isl-org/open3d/3996#-Mih_fuQCFSFMZd_6N9d:-Mih_fuQCFSFMZd_6N9e:b-i6gaog) ([raw file](https://github.com/isl-org/open3d/blob/9e828719d360ca158c6fc52d12272a2a9b138594/cpp/open3d/CMakeLists.txt#L82)):*
> ```CMake
> endif()
> 
> # Apply the workaround linking on all defined object libraries.
> ```

Would it be possible to do this step automatically if `open3d_ispc_add_library(xxx OBJECT)` is called?

E.g. run `open3d_ispc_target_sources_ISPC_TARGET_OBJECTS` inside `open3d_ispc_add_library()`; or maintain a list in `open3d_ispc_add_library()` and apply `open3d_ispc_target_sources_ISPC_TARGET_OBJECTS` to the list at the end.
___
*[cpp/open3d/utility/ISA.h, line 32 at r1](https://reviewable.io/reviews/isl-org/open3d/3996#-MihaGC0Bijz2M5d79E5:-MihaGC0Bijz2M5d79E6:bi26hme) ([raw file](https://github.com/isl-org/open3d/blob/9e828719d360ca158c6fc52d12272a2a9b138594/cpp/open3d/utility/ISA.h#L32)):*
> ```objc
> 
> /// Set of known ISA targets.
> enum class ISATarget {
> ```

Maybe print this message in `CPUInfo::Print()` and say the code is compiled/executed with this certain ISA.
Is this a compile-time decision or a run-time decision?
___
*[cpp/open3d/utility/ISA.cpp, line 29 at r1](https://reviewable.io/reviews/isl-org/open3d/3996#-MihT3Pb0kAIKHE28Rc3:-MihT3Pb0kAIKHE28Rc4:bshkxkd) ([raw file](https://github.com/isl-org/open3d/blob/9e828719d360ca158c6fc52d12272a2a9b138594/cpp/open3d/utility/ISA.cpp#L29)):*
> ```cpp
> #include "open3d/utility/ISA.h"
> 
> #include "ISA_ispc.h"
> ```

Is this file generated? Maybe provide some comments in the code of how the code is generated.

stotko (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3996)* status: 21 of 33 files reviewed, 4 unresolved discussions (waiting on @ssheorey and @yxlao)

*[a discussion](https://reviewable.io/reviews/isl-org/open3d/3996#-MivKSKt7rxkQGBFDzOv:-MivKSKt7rxkQGBFDzOw:bgebj9g) (no related file):*
I have further relaxed the ordering constraints of `open3d_ispc_target_sources`. It can now be used anywhere within the same `CMakeLists.txt` file or any subdirectory. This allows to keep the current pattern of first specifying the sources and then other properties like include directories, compile definitions, etc. where the latter will still be correctly detected now.

---
___
*[cmake/Open3DISPC.cmake, line 20 at r1](https://reviewable.io/reviews/isl-org/open3d/3996#-MihTvZ8AC_Gr8Ta7PC2:-MivFn8u9lAlLcGLSYVk:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/9e828719d360ca158c6fc52d12272a2a9b138594/cmake/Open3DISPC.cmake#L20)):*

nit: for internal helper functions, can we add prefix or suffix such as `_` or `__` to indicate it is internal?
</blockquote></details>

Done.
___
*[cpp/open3d/CMakeLists.txt, line 82 at r1](https://reviewable.io/reviews/isl-org/open3d/3996#-Mih_fuQCFSFMZd_6N9d:-MivG-sB3UGZeeZPHFXF:b-8i5wul) ([raw file](https://github.com/isl-org/open3d/blob/9e828719d360ca158c6fc52d12272a2a9b138594/cpp/open3d/CMakeLists.txt#L82)):*

Would it be possible to do this step automatically if `open3d_ispc_add_library(xxx OBJECT)` is called?

E.g. run `open3d_ispc_target_sources_ISPC_TARGET_OBJECTS` inside `open3d_ispc_add_library()`; or maintain a list in `open3d_ispc_add_library()` and apply `open3d_ispc_target_sources_ISPC_TARGET_OBJECTS` to the list at the end.

</blockquote></details>

This is not directly possible as we need a call of the form `my_func(Open3D PRIVATE xxx)` such that a dependency to `xxx` can be set to the main `Open3D` library target. There are only two candidates that come to my mind:

1. `target_link_libraries`: Since linking object libraries is natively supported by CMake and also propagates object files, this would be the best solution. However, the linked targets are not directly "replaced" by their interface properties, so CMake still needs to know what is behind that linked target. In particular, this would require to also install all object libraries.
2. `open3d_ispc_target_sources_TARGET_OBJECTS`: This would basically simulate the call `target_sources(Open3D PRIVATE $<TARGET_OBJECTS:xxx>) ` without requiring to install the object libaries.

I have picked option 2. The slightly less clean syntax sounds like the better trade-off here.
___
*[cpp/open3d/utility/ISA.cpp, line 29 at r1](https://reviewable.io/reviews/isl-org/open3d/3996#-MihT3Pb0kAIKHE28Rc3:-MivJKPu8zN2cXL9LNLG:bnsoiq4) ([raw file](https://github.com/isl-org/open3d/blob/9e828719d360ca158c6fc52d12272a2a9b138594/cpp/open3d/utility/ISA.cpp#L29)):*

Is this file generated? Maybe provide some comments in the code of how the code is generated.
</blockquote></details>

Yes, this file is auto-generated by the ISPC compiler. Each `xxx.ispc` file implicitly comes with a `xxx_ispc.h` file containing all `export`ed functionality. IMHO, I think it is not necessary to add further comments here because the file itself has `ispc` in its name and is also guarded by `BUILD_ISPC_MODULE`.
___
*[cpp/open3d/utility/ISA.h, line 32 at r1](https://reviewable.io/reviews/isl-org/open3d/3996#-MihaGC0Bijz2M5d79E5:-MivImxi-3wNHBroKjFZ:bq3wqyz) ([raw file](https://github.com/isl-org/open3d/blob/9e828719d360ca158c6fc52d12272a2a9b138594/cpp/open3d/utility/ISA.h#L32)):*

Maybe print this message in `CPUInfo::Print()` and say the code is compiled/executed with this certain ISA.
Is this a compile-time decision or a run-time decision?
</blockquote></details>

Done. Since the ISA is decided at run-time and there is also GPU ISAs, i.e. `GENX`, I have added a `ISAInfo` class instead rather than extending `CPUInfo`.

stotko (commented):
Good catch. Added `.isph` and `.ispc` files to the list of accepted file patterns.

> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3996#-:-Mj8dJWqEJeAEELDSZwD:b-yfs30f)* status: 20 of 34 files reviewed, 6 unresolved discussions (waiting on @ssheorey and @yxlao)

___
*[CMakeLists.txt, line 151 at r2](https://reviewable.io/reviews/isl-org/open3d/3996#-Mj24Io9A41p5wyK4Kl9:-Mj8daDv2xbMEt2qs7iH:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/1f8393fb0e7e790175b1e1c0b1163d0e9edc0ef7/CMakeLists.txt#L151)):*

[nit] I suggest we make this a FATAL_ERROR, so that the user doesn't accidentally miss the warning message and wonder if ISPC worked. Also default value of  BUILD_ISPC_MODULE should be OFF for ARM, so this warning doesn't pop-up all the time on ARM.
</blockquote></details>

Done.
___
*[cmake/Open3DSetGlobalProperties.cmake, line 105 at r2](https://reviewable.io/reviews/isl-org/open3d/3996#-Mj5VlvkBs-HFJBJMdv_:-Mj8dczC8LkJO9F_sIMd:beqz7r1) ([raw file](https://github.com/isl-org/open3d/blob/1f8393fb0e7e790175b1e1c0b1163d0e9edc0ef7/cmake/Open3DSetGlobalProperties.cmake#L105)):*

[nit] Does auto-detection work for arch?
</blockquote></details>

Yes, I tested auto-detection with CI in an earlier version of this PR and have not seen any issues so far. However, it is recommended to explicitly set the architecture for better guiding the ISPC compiler.
___
*[cpp/open3d/utility/ISA.h, line 32 at r1](https://reviewable.io/reviews/isl-org/open3d/3996#-MihaGC0Bijz2M5d79E5:-Mj8eM5Z7ZL2zW4BnMcV:b-95dzq) ([raw file](https://github.com/isl-org/open3d/blob/9e828719d360ca158c6fc52d12272a2a9b138594/cpp/open3d/utility/ISA.h#L32)):*

Is this ISPC specific? [Add a comment in that case].
</blockquote></details>

Yes, this is specific to ISPC kernel code. Added a comment to the class and slightly tweaked the printed message.

yxlao (approved):

yxlao:
(clicked by accident, @ssheorey has already approved the PR)

PR #3998, opened 2021-08-30
Use new tensor check api

yxlao:
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/3998)


reyanshsolis:
Since, at most of the places, we try to assert more than one of `dtype, device, shape` property. Do you think having a wrapper such as `AssertTensorProperties(const Device& device, const Dtype& dtype, utility::optional<Shape> shape = nullptr)` might be helpful ?

stotko (approved):
Such a combined check looks simpler at first glance, but I think it is the opposite and makes things more complicated:

- We are now checking more properties at the same time, so they are passed in some arbitrary but defined order. It is harder to remember that order than using the individual checks which only have 2 arguments where the order is clear.
- The other signature has the same problem. For example, what is the second `true` in `AssertTensorHasSameProperties(t, s, false, true, true);`? This could be improved by having some flag-like `enum class` with `operator|`. But applying these flags to the first signature is not that easy as it is not clear what the type of the reference object should be.

IMHO, the proposed new API should be sufficient for now. Other thoughts?

> Reviewed 30 of 30 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/3998#-:-MiQUJkj2fzctZ4Nk7ao:b-8m3lt8)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @yxlao)

yxlao:
> We are now checking more properties at the same time, so they are passed in some arbitrary but defined order. It is harder to remember that order than using the individual checks which only have 2 arguments where the order is clear.

In theory, we could have:

1. overloaded checker functions with different arguments or different orders of arguments:
    ```cpp
    void AssertProperties_(const Shape&);
    void AssertProperties_(const Device&, const Shape&);
    void AssertProperties_(const Device&, const Dtype&, const Shape&);
    void AssertProperties_(const Shape&, const Device&, const Dtype&);
    ```
2.  And then, have some macro magic to determine whether there are 1, 2 or 3 arguments and call the corresponding overloaded functions. `__VA_ARGS__` is one possible solution but we have the technical challenge as described below. 

The main technical challenge is that the use of initializer lists makes it hard for the macro magic to determine the number of arguments. This can be resolved if we always require the user to put `()` or the full class constructor. when using the function. 

E.g.
```cpp
// allowed
AssertProperties_(({100, 3}), Dtype::Int32); // Allowed, use ()
AssertProperties_(SizeVector({100, 3}), Dtype::Int32); // Allowed, use full class constructor
// not allowed
AssertProperties_({100, 3}, Dtype::Int32); // "," is interpreted as separators for arguments
```

If we can accept this limitation (always require the user to put `()` or the full class constructor. when using the function), there is a possibility that we can achieve arbitrary order and arbitrary number of arguments in the tensor checkers.

PR #4009, opened 2021-09-01
Build Docker with scripts

yxlao:
`docker_build.sh` is used to build Open3D docker images for all supported scenarios. This can be used in CI and on local machines. The objective is to allow developers to emulate CI environments for debugging or build release artifacts such as Python wheels locally.

### Example usage
```bash
./docker_build.sh openblas_arm64
./docker_build.sh cuda_wheel_py36_dev
```

### Design decisions
- Use a flat list of options.
  We don't want to have a cartesian product of different combinations of options. E.g., to support Ubuntu {18.04, 20.04} with Python {3.7, 3.8}, we don't specify the OS and Python version separately, instead, we have a flat list of combinations: [u1804_py37, u1804_py38, u2004_py37, u2004_py38].
- No external environment variables.
  This script should not make assumptions on external environment variables. This make the Docker image reproducible across different machines.

### Other changes
`BUILD_COMMON_CUDA_ARCHS=ON` for CI, always. With ccache, the time gap between the two cases is not significant. Setting it to `ON` reduces the combination of cases that we need to take care of.

E.g. in #4008
- `BUILD_COMMON_CUDA_ARCHS=OFF` (in PR): 58m 25s (https://github.com/isl-org/Open3D/runs/3488679056)
- `BUILD_COMMON_CUDA_ARCHS=ON`(in master): 1h 1m 5s (https://github.com/isl-org/Open3D/runs/3490723783)

### Next
`docker_build.sh` can be used to build Dcoker images on Github Actions for GPU CI (merging with `gce-ubuntu-docker.yml`). Hopefully this will give us a more stable GPU CI.

This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4009)

### DIFF
--- a/.github/workflows/ubuntu-cuda.yml
+++ b/.github/workflows/ubuntu-cuda.yml
@@ -79,6 +79,7 @@ jobs:
             ${{ env.CONDA_PKG_NAME }}
           if-no-files-found: error
       - name: GCloud CLI setup (GCE_SA_KEY_DOCS_CI)
+        if: ${{ github.ref == 'refs/heads/master' }}
         uses: google-github-actions/setup-gcloud@master
         with:
           version: ${{ env.GCE_CLI_GHA_VERSION }}
@@ -86,9 +87,11 @@ jobs:
           project_id: ${{ secrets.GCE_DOCS_PROJECT }}
           export_default_credentials: true
       - name: Upload ccache to GCS
+        if: ${{ github.ref == 'refs/heads/master' }}
         run: |
           gsutil cp ${GITHUB_WORKSPACE}/${{ env.CCACHE_TAR_NAME }}.tar.gz gs://open3d-ci-cache/
       - name: GCloud CLI setup (GCE_SA_KEY_GPU_CI)
+        if: ${{ github.ref == 'refs/heads/master' }}
         uses: google-github-actions/setup-gcloud@master
         with:
           version: ${{ env.GCE_CLI_GHA_VERSION }}
@@ -96,6 +99,7 @@ jobs:
           project_id: ${{ secrets.GCE_DOCS_PROJECT }}
           export_default_credentials: true
       - name: Upload wheel to GCS
+        if: ${{ github.ref == 'refs/heads/master' }}
         run: |
           gsutil cp ${GITHUB_WORKSPACE}/${{ env.PIP_PKG_NAME }} gs://open3d-releases-master/python-wheels/
           echo "Download pip package at: https://storage.googleapis.com/open3d-releases-master/python-wheels/${{ env.PIP_PKG_NAME }}"
--- a/.github/workflows/ubuntu-cuda.yml
+++ b/.github/workflows/ubuntu-cuda.yml
@@ -31,16 +31,9 @@ jobs:
           - is_master: false
             python_version: 3.9
     env:
-      # Environment variables are used as Docker build args.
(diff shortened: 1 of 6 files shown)


.github/workflows/ubuntu-cuda.yml
```diff
@@ -56,65 +49,57 @@ jobs:
...
+            .github/workflows/docker_build.sh cuda_wheel_py39_dev
+          elif [ "${{ env.PYTHON_VERSION }}" = "3.6" ] && [ "${{ env.DEVELOPER_BUILD }}" = "OFF" ]; then
+            .github/workflows/docker_build.sh cuda_wheel_py36
+          elif [ "${{ env.PYTHON_VERSION }}" = "3.7" ] && [ "${{ env.DEVELOPER_BUILD }}" = "OFF" ]; then
+            .github/workflows/docker_build.sh cuda_wheel_py37
+          elif [ "${{ env.PYTHON_VERSION }}" = "3.8" ] && [ "${{ env.DEVELOPER_BUILD }}" = "OFF" ]; then
+            .github/workflows/docker_build.sh cuda_wheel_py38
+          elif [ "${{ env.PYTHON_VERSION }}" = "3.9" ] && [ "${{ env.DEVELOPER_BUILD }}" = "OFF" ]; then
+            .github/workflows/docker_build.sh cuda_wheel_py39
+          fi
```
ssheorey, line 68: [nit] Slightly shorter if we separate the PYTHON_VERSION and DEVELOPER_BUILD checking:
```
if [ "${{ env.DEVELOPER_BUILD }}" = "ON" ]; then
    DEV="_dev"
else
    DEV=""
# Check PYTHON_VERSION and run
.github/workflows/docker_build.sh cuda_wheel_py37${DEV}
```

ssheorey (approved):

.github/workflows/ubuntu-cuda.yml
yxlao: Yes, I considered that initially as well. We could also go further and remove the `.` in python version number, i.e. `3.6` -> `36`, the code will look like:

```bash
PY_FLAG="_${${{ env.PYTHON_VERSION }}//.}"
if [ "${{ env.DEVELOPER_BUILD }}" = "ON" ]; then
  DEV_FLAG="_dev"
else
  DEV_FLAG=""
fi
DOCKER_OPTION="cuda_wheel${PY_FLAG}${DEV_FLAG}"
.github/workflows/docker_build.sh ${DOCKER_OPTION}
```

However, with the more verbose approach:
```bash
if   [ "${{ env.PYTHON_VERSION }}" = "3.6" ] && [ "${{ env.DEVELOPER_BUILD }}" = "ON"  ]; then
  .github/workflows/docker_build.sh cuda_wheel_py36_dev
elif [ "${{ env.PYTHON_VERSION }}" = "3.7" ] && [ "${{ env.DEVELOPER_BUILD }}" = "ON"  ]; then
  .github/workflows/docker_build.sh cuda_wheel_py37_dev
elif [ "${{ env.PYTHON_VERSION }}" = "3.8" ] && [ "${{ env.DEVELOPER_BUILD }}" = "ON"  ]; then
  .github/workflows/docker_build.sh cuda_wheel_py38_dev
elif ...
  ...
fi
```

it is more explicit and a developer can directly copy the command:
```bash
.github/workflows/docker_build.sh cuda_wheel_py36_dev
```
to run on the local machine.

PR #4017, opened 2021-09-06
TIO PLY Read Performance Improvement

reyanshsolis:
Changes:
- Faster Read.
- More memory efficient (avoiding memory copy).

#### Tensor Reads Custom Attributes with various Dtype support
 ```
 curvature [shape: {196133, 1}, stride: {1, 1}, Float32]
 colors [shape: {196133, 3}, stride: {3, 1}, UInt8]
 normals [shape: {196133, 3}, stride: {3, 1}, Float32]
 positions [shape: {196133, 3}, stride: {3, 1}, Float32]
```
#### Legacy only reads points/normals/colors as double.

PLY Read | Legacy | Master | This PR
-- | -- | -- | --
ASCII | 115 | 139 | 121
BIN | 22.2 | 41.0 | 22.8
BIN_COMP. | 22.2 | 41.3 | 22.8

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4017)

### DIFF
--- a/cpp/open3d/t/io/file_format/FilePLY.cpp
+++ b/cpp/open3d/t/io/file_format/FilePLY.cpp
@@ -50,10 +50,6 @@ struct PLYReaderState {
         int64_t size_;
         int64_t current_size_;
     };
-
-    // Allow fast access of attr_state by name.
-    std::unordered_map<std::string, std::shared_ptr<AttrState>>
-            name_to_attr_state_;
     // Allow fast access of attr_state by index.
     std::vector<std::shared_ptr<AttrState>> id_to_attr_state_;
     utility::CountingProgressReporter *progress_bar_;
@@ -78,7 +74,7 @@ static int ReadAttributeCallback(p_ply_argument argument) {
 
     ++attr_state->current_size_;
 
-    if (attr_state->stride_ == 0 && attr_state->current_size_ % 1000 == 0) {
+    if (attr_state->offset_ == 0 && attr_state->current_size_ % 1000 == 0) {
         state_ptr->progress_bar_->Update(attr_state->current_size_);
     }
     return 1;
@@ -154,6 +150,27 @@ static core::Dtype GetDtype(e_ply_type type) {
     }
 }
 
+static std::tuple<std::string, int, int> GetNameStrideOffsetForAttribute(
+        const std::string &name) {
+    // Positions attribute.
+    if (name == "x") return std::make_tuple("positions", 3, 0);
+    if (name == "y") return std::make_tuple("positions", 3, 1);
+    if (name == "z") return std::make_tuple("positions", 3, 2);
+
+    // Normals attribute.
+    if (name == "nx") return std::make_tuple("normals", 3, 0);
+    if (name == "ny") return std::make_tuple("normals", 3, 1);
+    if (name == "nz") return std::make_tuple("normals", 3, 2);
+
+    // Colors attribute.
+    if (name == "red") return std::make_tuple("colors", 3, 0);
+    if (name == "green") return std::make_tuple("colors", 3, 1);
+    if (name == "blue") return std::make_tuple("colors", 3, 2);
+
+    // Other attribute.
+    return std::make_tuple(name, 1, 0);
+}
+
 bool ReadPointCloudFromPLY(const std::string &filename,
                            geometry::PointCloud &pointcloud,
                            const open3d::io::ReadPointCloudOption &params) {
@@ -190,9 +207,8 @@ bool ReadPointCloudFromPLY(const std::string &filename,
         return false;
     }
 
-    bool positions_init = false;
-    bool normals_init = false;
-    bool colors_init = false;
+    std::unordered_map<std::string, bool> primary_attr_init = {
+            {"positions", false}, {"normals", false}, {"colors", false}};
 
     p_ply_property attribute = ply_get_next_property(element, nullptr);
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by stotko — CHANGES_REQUESTED** [review]
> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4017)* status: all files reviewed, 2 unresolved discussions (waiting on @reyanshsolis)

___
*[cpp/open3d/t/io/file_format/FilePLY.cpp, line 81 at r1](https://reviewable.io/reviews/isl-org/open3d/4017#-MizM5UCAv3M8KArNqzm:-MizM5UCAv3M8KArNqzn:b-8ctwt) ([raw file](https://github.com/isl-org/open3d/blob/8bd655555255656c19478da7d46410a424116b22/cpp/open3d/t/io/file_format/FilePLY.cpp#L81)):*
> ```cpp
>     ++attr_state->current_size_;
> 
>     if (attr_state->stride_ == 0 && attr_state->current_size_ % 1000 == 0) {
> ```

Why is the progress bar only updated for attributes with stride `0`? This would exclude any updates for the common attributes `positions`, `normals`, and `colors`.
___
*[cpp/open3d/t/io/file_format/FilePLY.cpp, line 226 at r1](https://reviewable.io/reviews/isl-org/open3d/4017#-MizN0MGDJ7VUl29FsmI:-MizN0MGDJ7VUl29FsmJ:bp2c6jh) ([raw file](https://github.com/isl-org/open3d/blob/8bd655555255656c19478da7d46410a424116b22/cpp/open3d/t/io/file_format/FilePLY.cpp#L226)):*
> ```cpp
>             const std::string attr_name = std::string(name);
> 
>             if (attr_name == "x" || attr_name == "y" || attr_name == "z") {
> ```

This block has the same structure for `positions`, `normals`, and `colors`. Could we factor this into a lambda function that gets called here for all 3 common attributes? I am thinking of something like this (untested draft):
```
auto InitAttributeState = [&](const std::unordered_map<std::string>& attr_offsets, const auto& pointcloud_set_func) {
    // Use attr_offsets for:
    // 1. The first if condition to init the point cloud. Also use the bound member function to set the tensor.
    // 2. Setting the offsets.
};
```
Feel free to come up with a better signature for the lambda, e.g. use `std::function` for the member function, a different container, and some return values to indicate whether the attribute has been set (to cope with the `else` cases).

**Review by reyanshsolis — COMMENTED** [review]
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4017)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @stotko)

___
*[cpp/open3d/t/io/file_format/FilePLY.cpp, line 81 at r1](https://reviewable.io/reviews/isl-org/open3d/4017#-MizM5UCAv3M8KArNqzm:-Mj-n0O56EfrVl6AlE3r:bcps070) ([raw file](https://github.com/isl-org/open3d/blob/8bd655555255656c19478da7d46410a424116b22/cpp/open3d/t/io/file_format/FilePLY.cpp#L81)):*

Why is the progress bar only updated for attributes with stride `0`? This would exclude any updates for the common attributes `positions`, `normals`, and `colors`.
</blockquote></details>

Done.
My bad, changed it to `offset_`. The idea is to update it only once for a row (like legacy, and unlike prev. version of this file). 
___
*[cpp/open3d/t/io/file_format/FilePLY.cpp, line 226 at r1](https://reviewable.io/reviews/isl-org/open3d/4017#-MizN0MGDJ7VUl29FsmI:-Mj-nJ0p7-ZqIYC7Ug7S:b-uzcj2m) ([raw file](https://github.com/isl-org/open3d/blob/8bd655555255656c19478da7d46410a424116b22/cpp/open3d/t/io/file_format/FilePLY.cpp#L226)):*

This block has the same structure for `positions`, `normals`, and `colors`. Could we factor this into a lambda function that gets called here for all 3 common attributes? I am thinking of something like this (untested draft):
```
auto InitAttributeState = [&](const std::unordered_map<std::string>& attr_offsets, const auto& pointcloud_set_func) {
    // Use attr_offsets for:
    // 1. The first if condition to init the point cloud. Also use the bound member function to set the tensor.
    // 2. Setting the offsets.
};
```
Feel free to come up with a better signature for the lambda, e.g. use `std::function` for the member function, a different container, and some return values to indicate whether the attribute has been set (to cope with the `else` cases).
</blockquote></details>

Using the function `GetNameStrideOffsetForAttribute` function, to resolve the code duplication, as I thought lambda function might make this slightly complicated. If you think lambda will be better than the updated version in this revision, I will try that.

stotko (approved):
> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4017)* status: all files reviewed, 1 unresolved discussion (waiting on @stotko)

___
*[cpp/open3d/t/io/file_format/FilePLY.cpp, line 226 at r1](https://reviewable.io/reviews/isl-org/open3d/4017#-MizN0MGDJ7VUl29FsmI:-Mj040Ql5TLXCPH59QfH:b-85i63v) ([raw file](https://github.com/isl-org/open3d/blob/8bd655555255656c19478da7d46410a424116b22/cpp/open3d/t/io/file_format/FilePLY.cpp#L226)):*

Using the function `GetNameStrideOffsetForAttribute` function, to resolve the code duplication, as I thought lambda function might make this slightly complicated. If you think lambda will be better than the updated version in this revision, I will try that. 
</blockquote></details>

LGTM.

PR #1683, opened 2020-04-03
Add seed argument to mesh sampling functions (#1682)

rowoflo:
This MR adds an additional argument to the `TriangleMesh::SamplePoints...` methods that is used to set the seed for the random number generator. The argument defaults to `-1`, which is a magic value that keeps the current functionality (the random number generator seed being set to a random number; thus, producing a new sequence which each trial). Any value besides `-1` will be used as the seed and will produce the same random number sequence for each trail. 

This MR addresses issue #1682 .

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/1683)

PR #4028, opened 2021-09-09
Add seed argument to PointCloud segment plane functions

ktsujister:
This MR adds an additional argument to the PointCloud::SegmentPlane method that is used to set the seed for the random number generator. The argument defaults to -1, which is a magic value that keeps the current functionality (the random number generator seed being set to a random number; thus, producing a new sequence which each trial). Any value besides -1 will be used as the seed and will produce the same random number sequence for each trail.

This is based on #1683.

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4028)


ktsujister:
I found style error after CI ran `style-check` and pushed f449139 to fix it.
@yxlao  Please let me know if there's anything else that needs to be fixed.

yxlao (changes requested):
> Reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4028)* status: all files reviewed, 4 unresolved discussions (waiting on @ktsujister)

___
*[cpp/open3d/geometry/PointCloud.h, line 335 at r1](https://reviewable.io/reviews/isl-org/open3d/4028#-MohKhzn7OhYR8SeWWGz:-MohKhzn7OhYR8SeWWH-:b-2k6cyb) ([raw file](https://github.com/isl-org/open3d/blob/f4491391154f87da205173a10baed30f59a7bc96/cpp/open3d/geometry/PointCloud.h#L335)):*
> ```objc
>     /// \param num_iterations Number of iterations.
>     /// \param seed Sets the seed value used in the random
>     /// generator, set to -1 to use a random seed value with each function call.
> ```

Update docs
___
*[cpp/open3d/geometry/PointCloud.h, line 342 at r1](https://reviewable.io/reviews/isl-org/open3d/4028#-MohJzRC9GSIo75eAwAe:-MohJzRC9GSIo75eAwAf:b6ae2e7) ([raw file](https://github.com/isl-org/open3d/blob/f4491391154f87da205173a10baed30f59a7bc96/cpp/open3d/geometry/PointCloud.h#L342)):*
> ```objc
>             const int ransac_n = 3,
>             const int num_iterations = 100,
>             int seed = -1) const;
> ```

Use `utility::optional<int> seed = utility::nullopt`. See https://stackoverflow.com/a/63657046/1255535.

(This is similar to `std::optional`. Once we move to C++17 (0.15 release), we'll be fully switching to `std::optional`.)
___
*[cpp/pybind/geometry/pointcloud.cpp, line 194 at r1](https://reviewable.io/reviews/isl-org/open3d/4028#-MohKcy34Ima53zX--xc:-MohKcy34Ima53zX--xd:bo542xe) ([raw file](https://github.com/isl-org/open3d/blob/f4491391154f87da205173a10baed30f59a7bc96/cpp/pybind/geometry/pointcloud.cpp#L194)):*
> ```cpp
>                  "algorithm.",
>                  "distance_threshold"_a, "ransac_n"_a, "num_iterations"_a,
>                  "seed"_a = -1)
> ```

default value set to `py::none()`
___
*[cpp/pybind/geometry/pointcloud.cpp, line 346 at r1](https://reviewable.io/reviews/isl-org/open3d/4028#-MohKgGU40vO3d4wNIf7:-MohKgGU40vO3d4wNIf8:b-2k6cyb) ([raw file](https://github.com/isl-org/open3d/blob/f4491391154f87da205173a10baed30f59a7bc96/cpp/pybind/geometry/pointcloud.cpp#L346)):*
> ```cpp
>              {"num_iterations", "Number of iterations."},
>              {"seed",
>               "Seed value used in the random generator, set to -1 to use a "
> ```

Update docs

yxlao (approved):
> Reviewed 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4028)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ktsujister)

PR #4034, opened 2021-09-10
Default AABB and OBB color to white instead of black

errissa:
Legacy OBB and AABB volumes default to black for color. When converting OBB and AABB to LineSets to use with new visualizer this results in black per-vertex colors which makes it impossible to specify a color since (per-vertex black * any color) = black. This PR changes the default to white which will work with the new visualizer more sensibly.

### DIFF
--- a/cpp/open3d/geometry/BoundingVolume.cpp
+++ b/cpp/open3d/geometry/BoundingVolume.cpp
@@ -41,8 +41,9 @@ OrientedBoundingBox& OrientedBoundingBox::Clear() {
     center_.setZero();
     extent_.setZero();
     R_ = Eigen::Matrix3d::Identity();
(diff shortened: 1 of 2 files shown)


saurabheights:
@errissa Will it be OK, if I add a PR to change the color from white to [rgb(119, 82, 254)](https://www.google.com/search?q=rgb(119%2C+82%2C+254)) (appealing color with uncommon value). The issue with white color is that by default, bounding box is invisible in white background of `o3d.visualization.Visualizer` which is quite commonly used. It ends up causing unnecessary complexity for someone starting with open3d and probably sometimes even for experienced users.

The choice of picking uncommon value is to prevent any conflict with different background colors used in any other visualization.

P.S. Any other color will do as well.

PR #4039, opened 2021-09-10
Add vectorized ParallelFor function and helper macros

stotko:
#### Highlights:

- `ParallelFor` with additional argument to call the vectorized code path. Falls back to other `ParallelFor` if `BUILD_ISPC_MODULE` is `OFF`.
- `OPEN3D_VECTORIZED` and `OPEN3D_TEMPLATE_VECTORIZED` macros to automatically create a lambda function calling the specified ISPC kernel. Only work for up to 9 additional parameters.
- `OPEN3D_EXPORT_VECTORIZED` and `OPEN3D_EXPORT_TEMPLATE_VECTORIZED` macros to define the ISPC kernels to be called by `ParallelFor`. Only work for up to 9 additional parameters.
- `OPEN3D_INSTANTIATE_TEMPLATE` and `OPEN3D_INSTANTIATE_TEMPLATE_WITH_BOOL` macros to emulate template functions and their instantiation in ISPC code.
- Unit tests to cover the introduced functions and macros.

#### Benchmarks (Intel i7-10750H):

Command: `./bin/benchmarks --benchmark_filter=ParallelFor`

```
[Open3D INFO] CPUInfo: 6 cores, 12 threads.
[Open3D INFO] ISAInfo: AVX2 instruction set used in ISPC code.
2021-09-14T11:44:43+02:00
Running ./bin/benchmarks
Run on (12 X 4727.5 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x6)
  L1 Instruction 32 KiB (x6)
  L2 Unified 256 KiB (x6)
  L3 Unified 12288 KiB (x1)
Load Average: 1.76, 1.27, 1.13
-----------------------------------------------------------------------------
Benchmark                                   Time             CPU   Iterations
-----------------------------------------------------------------------------
ParallelForScalar/CPU100                 1.27 us         1.27 us       542632
ParallelForScalar/CPU1000                1.26 us         1.26 us       533485
ParallelForScalar/CPU10000               2.14 us         2.14 us       329166
ParallelForScalar/CPU100000              11.4 us         11.4 us        55214
ParallelForScalar/CPU1000000              105 us          105 us         6769
ParallelForScalar/CPU10000000            3714 us         3714 us          188
ParallelForScalar/CPU100000000          39998 us        39995 us           18
ParallelForVectorized/CPU100             1.38 us         1.38 us       510147
ParallelForVectorized/CPU1000            1.40 us         1.40 us       496748
ParallelForVectorized/CPU10000           1.51 us         1.51 us       462849
ParallelForVectorized/CPU100000          3.59 us         3.59 us       197242
ParallelForVectorized/CPU1000000         42.2 us         42.2 us        16401
ParallelForVectorized/CPU10000000        3659 us         3658 us          194
ParallelForVectorized/CPU100000000      39473 us        39473 us           18
```

Command: `OPENMP_NUM_THREADS=1 ./bin/benchmarks --benchmark_filter=ParallelFor`

```
[Open3D INFO] CPUInfo: 6 cores, 12 threads.
[Open3D INFO] ISAInfo: AVX2 instruction set used in ISPC code.
2021-09-14T12:08:34+02:00
Running ./bin/benchmarks
Run on (12 X 4795.83 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x6)
  L1 Instruction 32 KiB (x6)
  L2 Unified 256 KiB (x6)
  L3 Unified 12288 KiB (x1)
Load Average: 0.61, 0.46, 0.57
-----------------------------------------------------------------------------
Benchmark                                   Time             CPU   Iterations
-----------------------------------------------------------------------------
ParallelForScalar/CPU100                0.554 us        0.554 us      1252981
ParallelForScalar/CPU1000               0.940 us        0.940 us       733749
ParallelForScalar/CPU10000               4.97 us         4.97 us       139518
ParallelForScalar/CPU100000              44.6 us         44.6 us        15436
ParallelForScalar/CPU1000000              444 us          444 us         1570
ParallelForScalar/CPU10000000            5207 us         5207 us          130
ParallelForScalar/CPU100000000          52967 us        52968 us           13
ParallelForVectorized/CPU100            0.597 us        0.597 us      1170738
ParallelForVectorized/CPU1000           0.662 us        0.662 us      1039586
ParallelForVectorized/CPU10000           1.40 us         1.40 us       502559
ParallelForVectorized/CPU100000          14.0 us         14.0 us        50152
ParallelForVectorized/CPU1000000          145 us          145 us         4678
ParallelForVectorized/CPU10000000        3924 us         3924 us          178
ParallelForVectorized/CPU100000000      42717 us        42718 us           16
```
#### Observations:
- Performance improvement becomes more significant with less cores used.
- Array size in relation to CPU Cache seems to have significant positive effect:
   - **Small**: Compute-bound, overhead exceeds actual work.
   - **Mid-sized**: Compute-bound, 2-4x speed improvements with `AVX2`.
   - **Large**: Memory-bound, acceleration less significant.
- More compute-intense kernels could benefit more from vectorization.

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4039)

### DIFF
--- a/cmake/Open3DISPC.cmake
+++ b/cmake/Open3DISPC.cmake
@@ -321,12 +321,7 @@ function(open3d_ispc_make_build_rules_ target)
 
         # Add files to ISPC_OBJECTS property.
         # This will later be used to resolve library.
-        get_target_property(TARGET_ISPC_OBJECTS ${target} ISPC_OBJECTS)
-        if (NOT TARGET_ISPC_OBJECTS)
-            set(TARGET_ISPC_OBJECTS "")
-        endif()
-        list(APPEND TARGET_ISPC_OBJECTS ${TARGET_OBJECT_FILES})
-        set_target_properties(${target} PROPERTIES ISPC_OBJECTS "${TARGET_ISPC_OBJECTS}")
+        set_property(TARGET ${target} APPEND PROPERTY ISPC_OBJECTS "${TARGET_OBJECT_FILES}")
     endif()
 endfunction()
 
@@ -522,7 +517,7 @@ function(open3d_ispc_target_sources target)
             endif()
         endforeach()
 
-        set_target_properties(${target} PROPERTIES ISPC_SOURCES "${TARGET_ISPC_SOURCES}")
+        set_property(TARGET ${target} APPEND PROPERTY ISPC_SOURCES "${TARGET_ISPC_SOURCES}")
     endif()
 endfunction()
 
--- a/cpp/benchmarks/CMakeLists.txt
+++ b/cpp/benchmarks/CMakeLists.txt
@@ -1,4 +1,4 @@
-add_executable(benchmarks)
+open3d_ispc_add_executable(benchmarks)
 
 add_subdirectory(core)
 add_subdirectory(geometry)
--- a/cpp/benchmarks/core/CMakeLists.txt
+++ b/cpp/benchmarks/core/CMakeLists.txt
@@ -1,6 +1,13 @@
 target_sources(benchmarks PRIVATE
     HashMap.cpp
     MemoryManager.cpp
+    ParallelFor.cpp
     Reduction.cpp
     Zeros.cpp
 )
+
+if (BUILD_ISPC_MODULE)
+    open3d_ispc_target_sources(benchmarks PRIVATE
+        ParallelFor.ispc
+    )
+endif()
--- a/cpp/benchmarks/core/ParallelFor.cpp
+++ b/cpp/benchmarks/core/ParallelFor.cpp
@@ -1,0 +1,108 @@
+// ----------------------------------------------------------------------------
+// -                        Open3D: www.open3d.org                            -
+// ----------------------------------------------------------------------------
+// The MIT License (MIT)
+//
+// Copyright (c) 2018-2021 www.open3d.org
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+// IN THE SOFTWARE.
+// ----------------------------------------------------------------------------
+
+#include "open3d/core/ParallelFor.h"
+
+#include <benchmark/benchmark.h>
+
+#include <vector>
+
+#ifdef BUILD_ISPC_MODULE
+#include "ParallelFor_ispc.h"
+#endif
+
+namespace open3d {
+namespace core {
+
+void ParallelForScalar(benchmark::State& state, int size) {
+    std::vector<float> input(size);
+    std::vector<float> output(size);
+    std::iota(input.begin(), input.end(), 0.0f);
+
+    // Warmup.
+    {
+        core::ParallelFor(core::Device("CPU:0"), size, [&](int64_t idx) {
+            float x = input[idx];
+            float x2 = x * x;
+            output[idx] = x2;
+        });
+    }
+
+    for (auto _ : state) {
+        core::ParallelFor(core::Device("CPU:0"), size, [&](int64_t idx) {
+            float x = input[idx];
+            float x2 = x * x;
+            output[idx] = x2;
+        });
+    }
+}
+
+void ParallelForVectorized(benchmark::State& state, int size) {
+    std::vector<float> input(size);
+    std::vector<float> output(size);
+    std::iota(input.begin(), input.end(), 0.0f);
+
+    // Warmup.
+    {
+        core::ParallelFor(
+                core::Device("CPU:0"), size,
+                [&](int64_t idx) {
+                    float x = input[idx];
+                    float x2 = x * x;
+                    output[idx] = x2;
+                },
+                OPEN3D_VECTORIZED(SquareKernel, input.data(), output.data()));
+    }
+
+    for (auto _ : state) {
+        core::ParallelFor(
+                core::Device("CPU:0"), size,
+                [&](int64_t idx) {
+                    float x = input[idx];
+                    float x2 = x * x;
+                    output[idx] = x2;
+                },
+                OPEN3D_VECTORIZED(SquareKernel, input.data(), output.data()));
+    }
+}
+
+#define ENUM_BM_SIZE(FN)                                                       \
+    BENCHMARK_CAPTURE(FN, CPU##100, 100)->Unit(benchmark::kMicrosecond);       \
+    BENCHMARK_CAPTURE(FN, CPU##1000, 1000)->Unit(benchmark::kMicrosecond);     \
+    BENCHMARK_CAPTURE(FN, CPU##10000, 10000)->Unit(benchmark::kMicrosecond);   \
+    BENCHMARK_CAPTURE(FN, CPU##100000, 100000)->Unit(benchmark::kMicrosecond); \
+    BENCHMARK_CAPTURE(FN, CPU##1000000, 1000000)                               \
+            ->Unit(benchmark::kMicrosecond);                                   \
+    BENCHMARK_CAPTURE(FN, CPU##10000000, 10000000)                             \
+            ->Unit(benchmark::kMicrosecond);                                   \
+    BENCHMARK_CAPTURE(FN, CPU##100000000, 100000000)                           \
+            ->Unit(benchmark::kMicrosecond);
+
+ENUM_BM_SIZE(ParallelForScalar)
+ENUM_BM_SIZE(ParallelForVectorized)
+
+}  // namespace core
+}  // namespace open3d
--- a/cpp/benchmarks/core/ParallelFor.ispc
+++ b/cpp/benchmarks/core/ParallelFor.ispc
@@ -23,30 +23,15 @@
 // FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
 // IN THE SOFTWARE.
 // ----------------------------------------------------------------------------
-// Copyright 2018 Google Inc. All rights reserved.
-//
-// Licensed under the Apache License, Version 2.0 (the "License");
-// you may not use this file except in compliance with the License.
-// You may obtain a copy of the License at
-//
-//     http://www.apache.org/licenses/LICENSE-2.0
-//
-// Unless required by applicable law or agreed to in writing, software
-// distributed under the License is distributed on an "AS IS" BASIS,
-// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-// See the License for the specific language governing permissions and
-// limitations under the License.
 
-#include <benchmark/benchmark.h>
+#include "open3d/core/ParallelFor.isph"
 
-#include "open3d/Open3D.h"
-
-int main(int argc, char** argv) {
-    open3d::utility::CPUInfo::GetInstance().Print();
-    open3d::utility::ISAInfo::GetInstance().Print();
-    benchmark::Initialize(&argc, argv);
-    if (benchmark::ReportUnrecognizedArguments(argc, argv)) {
-        return 1;
-    }
-    benchmark::RunSpecifiedBenchmarks();
+static inline void Square(int64_t idx,
+                          float* uniform input,
+                          float* uniform output) {
+    float x = input[idx];
+    float x2 = x * x;
+    output[idx] = x2;
 }
+
+OPEN3D_EXPORT_VECTORIZED(SquareKernel, Square, float* uniform, float* uniform)
--- a/cpp/open3d/Macro.h
+++ b/cpp/open3d/Macro.h
@@ -28,9 +28,6 @@
 
 #include <cassert>
 
-#define OPEN3D_CONCAT_IMPL(s1, s2) s1##s2
-#define OPEN3D_CONCAT(s1, s2) OPEN3D_CONCAT_IMPL(s1, s2)
-
 // https://gcc.gnu.org/wiki/Visibility updated to use C++11 attribute syntax
 #if defined(_WIN32) || defined(__CYGWIN__)
 #define OPEN3D_DLL_IMPORT __declspec(dllimport)
--- a/cpp/open3d/core/ParallelFor.h
+++ b/cpp/open3d/core/ParallelFor.h
@@ -27,11 +27,13 @@
 #pragma once
 
 #include <cstdint>
-#include <vector>
+#include <type_traits>
 
 #include "open3d/core/Device.h"
 #include "open3d/utility/Logging.h"
+#include "open3d/utility/Overload.h"
 #include "open3d/utility/Parallel.h"
+#include "open3d/utility/Preprocessor.h"
 
 #ifdef __CUDACC__
 #include <cuda.h>
@@ -50,7 +52,7 @@ static constexpr int64_t OPEN3D_PARFOR_THREAD = 4;
 
 /// Calls f(n) with the "grid-stride loops" pattern.
 template <int64_t block_size, int64_t thread_size, typename func_t>
-__global__ void __ElementWiseKernel(int64_t n, func_t f) {
+__global__ void ElementWiseKernel_(int64_t n, func_t f) {
     int64_t items_per_block = block_size * thread_size;
     int64_t idx = blockIdx.x * items_per_block + threadIdx.x;
 #pragma unroll
@@ -62,20 +64,9 @@ __global__ void __ElementWiseKernel(int64_t n, func_t f) {
     }
 }
 
-/// Run a function in parallel with CUDA.
-///
-/// \param device The device for the ParallelFor to run on.
-/// \param n The number of workloads.
-/// \param func The function to be executed in parallel. The function should
-/// take an int64_t workload index and returns void, i.e., `void func(int64_t)`.
-///
-/// \note This is optimized for uniform work items, i.e. where each call to \p
-/// func takes the same time.
-/// \note If you use a lambda function, capture only the required variables
-/// instead of all to prevent accidental race conditions. If you want the
-/// kernel to be used on both CPU and CUDA, capture the variables by value.
+/// Run a function in parallel on CUDA.
 template <typename func_t>
-void ParallelFor(const Device& device, int64_t n, const func_t& func) {
+void ParallelForCUDA_(const Device& device, int64_t n, const func_t& func) {
     if (device.GetType() != Device::DeviceType::CUDA) {
         utility::LogError("ParallelFor for CUDA cannot run on device {}.",
                           device.ToString());
@@ -88,7 +79,7 @@ void ParallelFor(const Device& device, int64_t n, const func_t& func) {
     int64_t items_per_block = OPEN3D_PARFOR_BLOCK * OPEN3D_PARFOR_THREAD;
     int64_t grid_size = (n + items_per_block - 1) / items_per_block;
 
-    __ElementWiseKernel<OPEN3D_PARFOR_BLOCK, OPEN3D_PARFOR_THREAD>
+    ElementWiseKernel_<OPEN3D_PARFOR_BLOCK, OPEN3D_PARFOR_THREAD>
             <<<grid_size, OPEN3D_PARFOR_BLOCK, 0, core::cuda::GetStream()>>>(
                     n, func);
     OPEN3D_GET_LAST_CUDA_ERROR("ParallelFor failed.");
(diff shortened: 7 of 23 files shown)


ssheorey (commented):
Suggestions about renaming for clarity / brevity:
```
ParallelForVectorized -> overloaded ParallelFor

OPEN3D_VECTORIZED_LAMBDA -> OPEN3D_VECTORIZED
OPEN3D_TEMPLATE_VECTORIZED_LAMBDA -> OPEN3D_TEMPLATE_VECTORIZED
OPEN3D_EXPORT_VECTORIZED_LAMBDA -> OPEN3D_EXPORT_VECTORIZED
OPEN3D_EXPORT_TEMPLATE_VECTORIZED_LAMBDA -> OPEN3D_EXPORT_TEMPLATE_VECTORIZED

LambdaKernel -> ISPCKernel
```

In general, only use `Lambda` in names if only lambda functions would work there, otherwise use Kernel (if C++/CUDA/ISPC) or ISPCKernel (ISPC specific).

Could you add a benchmark for any simple vectorized code? Vector assignment (load/store) should work.

The `/Zc:preprocessor` option may be causing errors in 3rd party code (tinyfiledialogs,..?). Restricting this option to only Open3D code should fix this.

Visual Studio 16.5.0 was released pretty recently Mar 16, 2020. Can users build without ISPC if they have an older version? Does `/experimental:preprocessor` work on an older version? Visual Studio is tricky because you need a subscription to receive updates in the commercial version, in additional to the initial purchase. So some commercial users may have older versions.
[If you want to test older versions: https://docs.microsoft.com/en-us/visualstudio/releases/2019/history]

> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4039#-:-MjGVTB81ouw1HaSZzWA:b-al88fo)* status: 0 of 17 files reviewed, 9 unresolved discussions (waiting on @stotko and @yxlao)

___
*[cpp/open3d/core/ParallelForVectorized.h, line 62 at r1](https://reviewable.io/reviews/isl-org/open3d/4039#-MjQmIGk7N-VMNjYZERb:-MjQmIGk7N-VMNjYZERc:b-u5tasb) ([raw file](https://github.com/isl-org/open3d/blob/fb5ef4d4850a51edb1be851ef4b4fcfbbac17897/cpp/open3d/core/ParallelForVectorized.h#L62)):*
> ```objc
>                            int64_t n,
>                            const vec_func_t& vec_func,
>                            const func_t& func) {
> ```

For simplicity, can we overload `ParallelFor`? If `vec_func` is the last argument and is provided, we dispatch to ISPC (if enabled).
Use would be:
```
ParallelFor(device, n, func);   // ISPC vec_func not implemented
ParallelFor(device, n, func, vec_func);   // If ISPC vec_func is implemented
```
___
*[cpp/open3d/core/ParallelForVectorized.h, line 108 at r1](https://reviewable.io/reviews/isl-org/open3d/4039#-MjQwCXd3twb8KYnAk_1:-MjQwCXd3twb8KYnAk_2:bgdo3jp) ([raw file](https://github.com/isl-org/open3d/blob/fb5ef4d4850a51edb1be851ef4b4fcfbbac17897/cpp/open3d/core/ParallelForVectorized.h#L108)):*
> ```objc
>     }
> 
> /// OPEN3D_VECTORIZED_LAMBDA(LambdaKernel, ...)
> ```

`OPEN3D_VECTORIZED(ISPCKernel, ...)` should be sufficient and is shorter. The `LAMBDA` is unnecessary. Also renaming `LambdaKernel` to `ISPCKernel` is clearer, since the arguments must be ISPC functions, not ordinary C++ lambda functions.
___
*[cpp/open3d/core/ParallelForVectorized.isph, line 47 at r1](https://reviewable.io/reviews/isl-org/open3d/4039#-MjQz4G_54nbMkW6_pP8:-MjQz4G_54nbMkW6_pP9:b-kjfdvu) ([raw file](https://github.com/isl-org/open3d/blob/fb5ef4d4850a51edb1be851ef4b4fcfbbac17897/cpp/open3d/core/ParallelForVectorized.isph#L47)):*
> ```
> ///
> /// Example:
> ///
> ```

\code...\endcode
___
*[cpp/open3d/utility/Overload.h, line 35 at r1](https://reviewable.io/reviews/isl-org/open3d/4039#-MjQoiUN35TGXcdbbyGM:-MjQoiUN35TGXcdbbyGN:b-7vzz4n) ([raw file](https://github.com/isl-org/open3d/blob/fb5ef4d4850a51edb1be851ef4b4fcfbbac17897/cpp/open3d/utility/Overload.h#L35)):*
> ```objc
> /// See Overload(...) function on how to use it.
> ///
> /// Note: In C++17, this could be simplified to:
> ```

\note
___
*[cpp/open3d/utility/Overload.h, line 37 at r1](https://reviewable.io/reviews/isl-org/open3d/4039#-MjQoktO-4T-qDKEhXws:-MjQoktO-4T-qDKEhXwt:b-shxyvm) ([raw file](https://github.com/isl-org/open3d/blob/fb5ef4d4850a51edb1be851ef4b4fcfbbac17897/cpp/open3d/utility/Overload.h#L37)):*
> ```objc
> /// Note: In C++17, this could be simplified to:
> ///
> /// template <typename... Ts>
> ```

  \code
...
  \endcode
___
*[cpp/open3d/utility/Overload.h, line 62 at r1](https://reviewable.io/reviews/isl-org/open3d/4039#-MjQp61NCIMOcA4wjRYn:-MjQp61NCIMOcA4wjRYo:b-kjfdvu) ([raw file](https://github.com/isl-org/open3d/blob/fb5ef4d4850a51edb1be851ef4b4fcfbbac17897/cpp/open3d/utility/Overload.h#L62)):*
> ```objc
> /// Overloads an arbitrary set of (lambda) functions.
> ///
> /// Example:
> ```

\code...\endcode
___
*[cpp/open3d/utility/Preprocessor.h, line 61 at r1](https://reviewable.io/reviews/isl-org/open3d/4039#-MjQqGNL-ALKZHKLq_w-:-MjQqGNL-ALKZHKLq_w0:b-kjfdvu) ([raw file](https://github.com/isl-org/open3d/blob/fb5ef4d4850a51edb1be851ef4b4fcfbbac17897/cpp/open3d/utility/Preprocessor.h#L61)):*
> ```objc
> /// Example:
> ///
> /// #define FOO_1(x1) foo(x1)
> ```

\code...\endcode
___
*[cpp/tests/core/ParallelFor.cpp, line 40 at r1](https://reviewable.io/reviews/isl-org/open3d/4039#-MjQr7BR762H-2oHGYf8:-MjQr7BR762H-2oHGYf9:b-u00bvr) ([raw file](https://github.com/isl-org/open3d/blob/fb5ef4d4850a51edb1be851ef4b4fcfbbac17897/cpp/tests/core/ParallelFor.cpp#L40)):*
> ```cpp
>     std::vector<int64_t> v(N);
> 
>     core::ParallelFor(core::Device("CPU:0"), v.size(),
> ```

Also for "CUDA:0"?
___
*[cpp/tests/utility/Preprocessor.cpp, line 58 at r1](https://reviewable.io/reviews/isl-org/open3d/4039#-MjR-4Z3AWCtuHy8P-KK:-MjR-4Z3AWCtuHy8P-KL:b9qfbfl) ([raw file](https://github.com/isl-org/open3d/blob/fb5ef4d4850a51edb1be851ef4b4fcfbbac17897/cpp/tests/utility/Preprocessor.cpp#L58)):*
> ```cpp
> TEST(Preprocessor, NUM_ARGS) {
>     // Detecting 0 Arguments is not supported and instead returns 1.
>     // EXPECT_EQ(OPEN3D_NUM_ARGS(), 0);
> ```

In this case, go ahead and have this as an actual test case that checks the 1 return value. This at least tests that the code runs with a known return value.

stotko (commented):
Done.

- I have reverted setting the `/Zc:preprocessor` flag and instead introduced a `OPEN3D_FIX_MSVC_()` macro. The 3rdparty issues could be fixed either by explicitly forcing a newer Windows SDK version (but further limits backwards compatibility) or by limiting it to the main Open3D library as suggested (but does not allow to use it outside the library without the flag). Both options exclude a subset of use cases and users and could potentially lead to more complaints. In Visual Studio 16.4 or older, `/experimental:preprocessor` enables the incomplete new preprocessor which at least has one blocker bug with boost and a few other smaller limitations. I agree that the support is still quite new, so we should probably wait for a wider adoption of the new preprocessor.

- The numbers from the test are not very representative as the `EXPECT_EQ` check dominate the runtime cost. In general I observed a performance improvement of ~2x between `AVX2` and `DISABLED` (i.e. the scalar lambda function) ISAs. I will add detailed benchmark results for each concrete application of the new vectorized `ParallelFor` (in a separate PR).

> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4039#-:-MjUMJ1H5od-TrT_zoFW:bavsu2f)* status: 0 of 18 files reviewed, 9 unresolved discussions (waiting on @ssheorey and @yxlao)

___
*[cpp/open3d/utility/Overload.h, line 35 at r1](https://reviewable.io/reviews/isl-org/open3d/4039#-MjQoiUN35TGXcdbbyGM:-MjULfST7DazZ_zNrVSg:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/fb5ef4d4850a51edb1be851ef4b4fcfbbac17897/cpp/open3d/utility/Overload.h#L35)):*

\note
</blockquote></details>

Done.
___
*[cpp/open3d/utility/Overload.h, line 37 at r1](https://reviewable.io/reviews/isl-org/open3d/4039#-MjQoktO-4T-qDKEhXws:-MjULgDaC3bU_Vv0cVXA:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/fb5ef4d4850a51edb1be851ef4b4fcfbbac17897/cpp/open3d/utility/Overload.h#L37)):*

  \code
...
  \endcode
</blockquote></details>

Done.
___
*[cpp/open3d/utility/Overload.h, line 62 at r1](https://reviewable.io/reviews/isl-org/open3d/4039#-MjQp61NCIMOcA4wjRYn:-MjULgsaC6ePX_gINmMf:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/fb5ef4d4850a51edb1be851ef4b4fcfbbac17897/cpp/open3d/utility/Overload.h#L62)):*

\code...\endcode
</blockquote></details>

Done.
___
*[cpp/open3d/utility/Preprocessor.h, line 61 at r1](https://reviewable.io/reviews/isl-org/open3d/4039#-MjQqGNL-ALKZHKLq_w-:-MjULhpj-eSreft9YiY8:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/fb5ef4d4850a51edb1be851ef4b4fcfbbac17897/cpp/open3d/utility/Preprocessor.h#L61)):*

\code...\endcode
</blockquote></details>

Done.
___
*[cpp/tests/core/ParallelFor.cpp, line 40 at r1](https://reviewable.io/reviews/isl-org/open3d/4039#-MjQr7BR762H-2oHGYf8:-MjULjKt0h2rz5n7pzcp:bx990e4) ([raw file](https://github.com/isl-org/open3d/blob/fb5ef4d4850a51edb1be851ef4b4fcfbbac17897/cpp/tests/core/ParallelFor.cpp#L40)):*

Also for "CUDA:0"?
</blockquote></details>

This would require moving the `"CUDA:0"` device to a separate `ParallelFor.cu` file and doing some manually `memcpy`s. Should I still add that test case?
___
*[cpp/tests/utility/Preprocessor.cpp, line 58 at r1](https://reviewable.io/reviews/isl-org/open3d/4039#-MjR-4Z3AWCtuHy8P-KK:-MjUMDLYDhoE5_RZCNtk:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/fb5ef4d4850a51edb1be851ef4b4fcfbbac17897/cpp/tests/utility/Preprocessor.cpp#L58)):*

In this case, go ahead and have this as an actual test case that checks the 1 return value. This at least tests that the code runs with a known return value.
</blockquote></details>

Done.
___
*[cpp/open3d/core/ParallelForVectorized.h, line 62 at r1](https://reviewable.io/reviews/isl-org/open3d/4039#-MjQmIGk7N-VMNjYZERb:-MjUMFN984WuMoLcCMyi:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/fb5ef4d4850a51edb1be851ef4b4fcfbbac17897/cpp/open3d/core/ParallelForVectorized.h#L62)):*

For simplicity, can we overload `ParallelFor`? If `vec_func` is the last argument and is provided, we dispatch to ISPC (if enabled).
Use would be:
```
ParallelFor(device, n, func);   // ISPC vec_func not implemented
ParallelFor(device, n, func, vec_func);   // If ISPC vec_func is implemented
```
</blockquote></details>

Done.
___
*[cpp/open3d/core/ParallelForVectorized.h, line 108 at r1](https://reviewable.io/reviews/isl-org/open3d/4039#-MjQwCXd3twb8KYnAk_1:-MjUMG8WDm-Q_fqU0bKC:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/fb5ef4d4850a51edb1be851ef4b4fcfbbac17897/cpp/open3d/core/ParallelForVectorized.h#L108)):*

`OPEN3D_VECTORIZED(ISPCKernel, ...)` should be sufficient and is shorter. The `LAMBDA` is unnecessary. Also renaming `LambdaKernel` to `ISPCKernel` is clearer, since the arguments must be ISPC functions, not ordinary C++ lambda functions.
</blockquote></details>

Done.
___
*[cpp/open3d/core/ParallelForVectorized.isph, line 47 at r1](https://reviewable.io/reviews/isl-org/open3d/4039#-MjQz4G_54nbMkW6_pP8:-MjUMH285POIRMtKUNVm:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/fb5ef4d4850a51edb1be851ef4b4fcfbbac17897/cpp/open3d/core/ParallelForVectorized.isph#L47)):*

\code...\endcode
</blockquote></details>

Done.

ssheorey (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4039)* status: 0 of 18 files reviewed, 1 unresolved discussion (waiting on @stotko and @yxlao)

___
*[cpp/tests/core/ParallelFor.cpp, line 40 at r1](https://reviewable.io/reviews/isl-org/open3d/4039#-MjQr7BR762H-2oHGYf8:-MjV684y2ep-WdVlOlHA:bem1n0d) ([raw file](https://github.com/isl-org/open3d/blob/fb5ef4d4850a51edb1be851ef4b4fcfbbac17897/cpp/tests/core/ParallelFor.cpp#L40)):*

This would require moving the `"CUDA:0"` device to a separate `ParallelFor.cu` file and doing some manually `memcpy`s. Should I still add that test case?
</blockquote></details>

We can create a `Tensor` (instead of `std::vector`) on CPU / CUDA and fill it in the lambda function to test both devices, similar to the tests for `Tensor`. I think we don't have a CUDA test for `ParallelFor` right now, so this would be useful.

yxlao (approved):
> Reviewed 5 of 17 files at r1, 11 of 13 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4039)* status: 16 of 18 files reviewed, 3 unresolved discussions (waiting on @stotko)

___
*[cpp/open3d/core/ParallelFor.h, line 138 at r3](https://reviewable.io/reviews/isl-org/open3d/4039#-MjY0vPn3G5G3Ic17C_N:-MjY0vPn3G5G3Ic17C_O:bth99v8) ([raw file](https://github.com/isl-org/open3d/blob/b4a5d03439787c0295868651ef7247a880cd4c43/cpp/open3d/core/ParallelFor.h#L138)):*
> ```objc
> /// \param vec_func The vectorized function to be executed in parallel. The
> /// function should be provided using the OPEN3D_VECTORIZED macro, e.g.,
> /// `OPEN3D_VECTORIZED(MyISPCKernel, some_used_variable)`.
> ```

It would be best if we can provide a minimal code example here to demonstrate the usage of 
`OPEN3D_VECTORIZED` and `OPEN3D_EXPORT_VECTORIZED` and how it interacts with `ParallelFor`
___
*[cpp/open3d/core/ParallelFor.h, line 165 at r3](https://reviewable.io/reviews/isl-org/open3d/4039#-MjXt1tp3dwmFwenl7Uh:-MjXt1tp3dwmFwenl7Ui:b-jnjedq) ([raw file](https://github.com/isl-org/open3d/blob/b4a5d03439787c0295868651ef7247a880cd4c43/cpp/open3d/core/ParallelFor.h#L165)):*
> ```objc
> #else
> 
>     ParallelFor(device, n, func);
> ```

nit: it would be more symmetrical and explicit to dispatch it as
```cpp
#ifdef __CUDACC__
    ParallelForCUDA_(device, n, func);
#else
    ParallelForCPU_(device, n, func);
#endif
```

stotko (commented):
I actually decided to go ahead and put some minimal benchmark in this PR. See the results in the description of the PR above.

> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4039#-:-MjYh8vjCSG6pY5fmgQ-:b-lhk1g1)* status: 13 of 24 files reviewed, 2 unresolved discussions (waiting on @ssheorey and @yxlao)

___
*[cpp/open3d/core/ParallelFor.h, line 138 at r3](https://reviewable.io/reviews/isl-org/open3d/4039#-MjY0vPn3G5G3Ic17C_N:-MjYgv5S0hdLru_dRKX7:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/b4a5d03439787c0295868651ef7247a880cd4c43/cpp/open3d/core/ParallelFor.h#L138)):*

It would be best if we can provide a minimal code example here to demonstrate the usage of 
`OPEN3D_VECTORIZED` and `OPEN3D_EXPORT_VECTORIZED` and how it interacts with `ParallelFor`
</blockquote></details>

Done.
___
*[cpp/open3d/core/ParallelFor.h, line 165 at r3](https://reviewable.io/reviews/isl-org/open3d/4039#-MjXt1tp3dwmFwenl7Uh:-MjYgvuG5gkyFdVk99Sj:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/b4a5d03439787c0295868651ef7247a880cd4c43/cpp/open3d/core/ParallelFor.h#L165)):*

nit: it would be more symmetrical and explicit to dispatch it as
```cpp
#ifdef __CUDACC__
    ParallelForCUDA_(device, n, func);
#else
    ParallelForCPU_(device, n, func);
#endif
</blockquote></details>

Done.
___
*[cpp/tests/core/ParallelFor.cpp, line 40 at r1](https://reviewable.io/reviews/isl-org/open3d/4039#-MjQr7BR762H-2oHGYf8:-MjYgyay7TSaNlhaxZVU:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/fb5ef4d4850a51edb1be851ef4b4fcfbbac17897/cpp/tests/core/ParallelFor.cpp#L40)):*

We can create a `Tensor` (instead of `std::vector`) on CPU / CUDA and fill it in the lambda function to test both devices, similar to the tests for `Tensor`. I think we don't have a CUDA test for `ParallelFor` right now, so this would be useful.
</blockquote></details>

Done.


<!-- Sent from Reviewable.io -->

stotko:
I have added two new helper functions on the ISPC side:

- `OPEN3D_ENSURE_EXPORTED()`: Used to make a `struct` or `enum` visible in in the corresponding `xxx_ispc.h` header independent of whether it is used in another `export`ed function.
- Typedefs for the fixed size integer types, e.g. `int64` -> `int64_t`: This makes the syntax look much more like C/C++.

Since common, standard-like functionality is currently handled in `Helper.h`, putting above functionality into `Helper.isph` seemed like a natural choice.

@yxlao @ssheorey Please have a look again at this PR. Functionality-wise it should be complete now and CI should pass as well.

yxlao:
@stotko I am observing a strange behavior in my system.
1. Clean the `build` directory, and build from scratch. For example, configure and run `make test`.
2. Modify any CMake file, e.g. change comment formatting in the root `CMakeLists.txt`. 
3. Rerun `make test`.
   - The compilation eventually fails at:
     ```
     [ 43%] Building ISPC object cpp/open3d/utility/CMakeFiles/utility.dir/ISAInfo.ispc.o
     Consolidate compiler generated dependencies of target core
     Error: Unknown option "-pthread"
     ```
   - In my case, `ccache` is also no longer working properly, as it treats everything as recompilation. Perhaps this is related to the issue above (wrong compilation flag).

This is true both in `master` and in this PR.

stotko:
> @stotko I am observing a strange behavior:
> 
>     1. Clean the `build` directory, and build from scratch. For example, configure and run `make test`.
> 
>     2. Modify any CMake file, e.g. change comment formatting in the root `CMakeLists.txt`.
> 
>     3. Rerun `make test`.
>        
>        * The compilation eventually fails at:
>          ```
>          [ 43%] Building ISPC object cpp/open3d/utility/CMakeFiles/utility.dir/ISAInfo.ispc.o
>          Consolidate compiler generated dependencies of target core
>          Error: Unknown option "-pthread"
>          ```
>        * In my case, `ccache` is also no longer working properly, as it treats everything as recompilation. Perhaps this is related to the compilation flag issue above.
> 
> 
> This is true both in `master` and in this PR. It means that every time we update cmake, we have to clean the `build` directory and rebuild.

Interesting. It seems like the "state" of the project, i.e. the values of some CMake variables, always changes in such cases. The build error you mentioned is likely just a part of that state change where the `Threads::Threads` target changed its properties for some reason and ignores our setting of `THREADS_PREFER_PTHREAD_FLAG`.

Since this also happens on `master` (for a longer time I think), the root cause is unrelated to this PR and needs further investigation. In the meantime, we could consider forcing the legacy ISPC support which does not suffer from the `-pthread` flag issue due to its several limitations. Other options could be to report this issue upstream to ISPC and/or CMake to "fix" first-class ISPC language support.

yxlao:
> Interesting. It seems like the "state" of the project, i.e. the values of some CMake variables, always changes in such cases. The build error you mentioned is likely just a part of that state change where the `Threads::Threads` target changed its properties for some reason and ignores our setting of `THREADS_PREFER_PTHREAD_FLAG`.
> 
> Since this also happens on `master` (for a longer time I think), the root cause is unrelated to this PR and needs further investigation. In the meantime, we could consider forcing the legacy ISPC support which does not suffer from the `-pthread` flag issue due to its several limitations. Other options could be to report this issue upstream to ISPC and/or CMake to "fix" first-class ISPC language support.

The issue happens to me after the first ISPC PR #3996. Did you observe similar issues?

stotko:
> > Interesting. It seems like the "state" of the project, i.e. the values of some CMake variables, always changes in such cases. The build error you mentioned is likely just a part of that state change where the `Threads::Threads` target changed its properties for some reason and ignores our setting of `THREADS_PREFER_PTHREAD_FLAG`.
> > Since this also happens on `master` (for a longer time I think), the root cause is unrelated to this PR and needs further investigation. In the meantime, we could consider forcing the legacy ISPC support which does not suffer from the `-pthread` flag issue due to its several limitations. Other options could be to report this issue upstream to ISPC and/or CMake to "fix" first-class ISPC language support.
> 
> The issue happens to me after the first ISPC PR #3996. Did you observe similar issues?

I observed the general issue of triggering recompilation of every source file after a change in a `CMakeLists.txt` file. The `-pthread` issue is likely just another thing that "gets broken" for some reason. The ISPC support itself should be fine. Instead, I suspect that it could related to `ccache` itself and/or `CMAKE_<LANG>_COMPILER_LAUNCHER`, but I do not have any concrete proofs for that.

PR #4043, opened 2021-09-13
throw exception when slicing a 0-dim tensor

yxlao:
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4043)

### DIFF
--- a/python/test/core/test_core.py
+++ b/python/test/core/test_core.py
@@ -441,9 +441,11 @@ def test_getitem(device):
                             np_t[0:2, 1:3, 0:4][0:1, 0:2, 2:3])
 
     # Slice a zero-dim tensor
-    with pytest.raises(Exception):
+    with pytest.raises(RuntimeError,
+                       match=r"Cannot slice a scalar \(0-dim\) tensor."):
         o3c.Tensor.ones((), device=device)[:]
-    with pytest.raises(Exception):
+    with pytest.raises(RuntimeError,
+                       match=r"Cannot slice a scalar \(0-dim\) tensor."):
         o3c.Tensor.ones((), device=device)[0:1]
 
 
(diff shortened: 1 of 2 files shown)


python/test/core/test_core.py
```diff
@@ -440,6 +440,12 @@ def test_getitem(device):
     np.testing.assert_equal(o3_t[0:2, 1:3, 0:4][0:1, 0:2, 2:3].cpu().numpy(),
                             np_t[0:2, 1:3, 0:4][0:1, 0:2, 2:3])
 
+    # Slice a zero-dim tensor
+    with pytest.raises(Exception):
```
ssheorey, line 444: `with pytest.raises(RuntimeError, match="Cannot slice a scalar (0-dim) tensor.")`
(more specific exception class with error message)
also below...

python/test/core/test_core.py
```diff
@@ -548,6 +554,12 @@ def test_setitem(device):
     o3_t[2, 2] = False
     np.testing.assert_equal(o3_t.cpu().numpy(), np_t)
 
+    # Slice a zero-dim tensor
+    with pytest.raises(Exception):
+        o3c.Tensor.ones((), device=device)[:] = 0
+    with pytest.raises(Exception):
+        o3c.Tensor.ones((), device=device)[0:1] = 0
+
```
ssheorey, line 562: More specific test here as well.

ssheorey (approved):

python/test/core/test_core.py
yxlao: done

python/test/core/test_core.py
yxlao: done

PR #4062, opened 2021-09-18
Fix unresolved symbols with BUILD_SHARED_LIBS=ON and BUILD_GUI=OFF

benjaminum:
This PR fixes unresolved symbols when linking to the shared libs without gui module by moving MessageProcessor.cpp to the gui source files.

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4062)


ssheorey (approved):
Tested on Ubuntu 20.04 with BUILD_SHARED_LIBS=ON BUILD_GUI=OFF
No error when building tests, pybind. No error when running tests.

PR #4069, opened 2021-09-20
install ml ops shared libraries

benjaminum:
This PR adds install commands to install the op libraries to allow using them in C++ projects

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4069)


ssheorey (commented):
@benjaminum Can you add an example for how to use the ops libraries with a C++ app?
We have such an example for the main Open3D library here:
https://github.com/isl-org/open3d-cmake-external-project

yxlao (approved):

ssheorey (approved):

PR #4076, opened 2021-09-22
Avoid -pthread flag with ISPC compiler

yxlao:
### Fix
This overwrites the default flags for `Threads::Threads`:
https://github.com/Kitware/CMake/blob/efed2f8706b46d2f5c7711a9426b7e23a08b9709/Modules/FindThreads.cmake#L243-L247

### Testing

Copy and run this script as to the root Open3D directory (as `Open3D/build_and_rebuild.sh`).

```bash
#!/usr/bin/env bash
set -eu

# Clean up
rm -rf build
mkdir build
rm -f build.log
rm -f rebuild.log

# First build
pushd build
cmake -DBUILD_BENCHMARKS=ON -DBUILD_UNIT_TESTS=ON .. &> >(tee -a "../build.log")
make VERBOSE=1 Open3D -j$(nproc) &> >(tee -a "../build.log")
popd

# Critical step: touch
touch CMakeLists.txt

# Second build
pushd build
make VERBOSE=1 Open3D -j$(nproc) &> >(tee -a "../rebuild.log")
popd
```

- On `master` :
    - First build: no `-pthread` flag at all.
    - Second build:  `-pthread` added for every file, causing recompilations. `-pthread` is also used in ISPC, causing the compilation to fail with
        ```bash
        Error: Unknown option "-pthread".
        ```
- With this PR:
    - First build:  `-pthread` added for every file, except for ISPC. 
    - Second build:  `-pthread` added for every file, except for ISPC. No recompilations, since it is the same as the first build.

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4076)

### DIFF
--- a/3rdparty/find_dependencies.cmake
+++ b/3rdparty/find_dependencies.cmake
@@ -462,12 +462,6 @@ open3d_find_package_3rdparty_library(3rdparty_threads
     PACKAGE Threads
     TARGETS Threads::Threads
 )
-if(THREADS_HAVE_PTHREAD_ARG)
-    set_property(TARGET Threads::Threads
-                 PROPERTY INTERFACE_COMPILE_OPTIONS
-                 "$<$<COMPILE_LANG_AND_ID:CUDA,NVIDIA>:SHELL:-Xcompiler -pthread>"
-                 "$<$<AND:$<NOT:$<COMPILE_LANGUAGE:ISPC>>,$<NOT:$<COMPILE_LANG_AND_ID:CUDA,NVIDIA>>>:-pthread>")
-endif()
 
 # Assimp
 include(${Open3D_3RDPARTY_DIR}/assimp/assimp.cmake)
--- a/3rdparty/find_dependencies.cmake
+++ b/3rdparty/find_dependencies.cmake
@@ -462,6 +462,12 @@ open3d_find_package_3rdparty_library(3rdparty_threads
     PACKAGE Threads
     TARGETS Threads::Threads
 )
+if(THREADS_HAVE_PTHREAD_ARG)
+    set_property(TARGET Threads::Threads
+                 PROPERTY INTERFACE_COMPILE_OPTIONS
+                 "$<$<COMPILE_LANG_AND_ID:CUDA,NVIDIA>:SHELL:-Xcompiler -pthread>"
+                 "$<$<AND:$<NOT:$<COMPILE_LANGUAGE:ISPC>>,$<NOT:$<COMPILE_LANG_AND_ID:CUDA,NVIDIA>>>:-pthread>")
+endif()
 
 # Assimp
 include(${Open3D_3RDPARTY_DIR}/assimp/assimp.cmake)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -309,9 +309,20 @@ add_compile_definitions("${HARDENING_DEFINITIONS}")
 
 # Explicitly specifiy the preference of using -pthread over -lpthread.
 # This must be defined here since CUDA calls find_package(Threads) internally.
-# -pthread is disabled since some compilers, e.g. the ISPC compiler, have
-# problems with this compile flag and issue errors.
 set(THREADS_PREFER_PTHREAD_FLAG TRUE)
+
+# Overwrites property for Thread::Threads in find_package(Threads)
+# For CUDA, "-pthread" is replaced with "-Xcompiler -pthread" (CMake's default)
+# For ISPC, "-pthread" is disabled
+macro(open3d_patch_findthreads_module_)
+    if(TARGET Threads::Threads AND THREADS_HAVE_PTHREAD_ARG)
+        set_property(TARGET Threads::Threads
+                     PROPERTY INTERFACE_COMPILE_OPTIONS
+                     "$<$<COMPILE_LANG_AND_ID:CUDA,NVIDIA>:SHELL:-Xcompiler -pthread>"
+                     "$<$<AND:$<NOT:$<COMPILE_LANG_AND_ID:CUDA,NVIDIA>>,$<NOT:$<COMPILE_LANGUAGE:ISPC>>>:-pthread>")
+    endif()
+endmacro()
+cmake_language(EVAL CODE "cmake_language(DEFER CALL open3d_patch_findthreads_module_)")
 
 # Build CUDA module by default if CUDA is available
 if(BUILD_CUDA_MODULE)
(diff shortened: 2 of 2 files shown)


stotko (commented):
Thanks for working on this! I already took a look into this and it seems like the `FindThreads` module is buggy as it sets `FindThreads_FOUND` to `FALSE` before processing anything. This is likely the reason why setting the preference via `THREADS_PREFER_PTHREAD_FLAG` does not reliably work. We should probably report this to CMake.

Regarding the workaround we have to come up with, I have a few suggestions for brainstorming as there is multiple ways to approach this problem.

> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4076#-:-MkGOucLE7enXqpappsQ:b-6ominh)* status: all files reviewed, 2 unresolved discussions (waiting on @yxlao)

___
*[3rdparty/find_dependencies.cmake, line 465 at r1](https://reviewable.io/reviews/isl-org/open3d/4076#-MkGPvE-AFe4Unn_R_ZP:-MkGPvE-AFe4Unn_R_ZQ:br662wp) ([raw file](https://github.com/isl-org/open3d/blob/1fa38412184db67fe952fde7b8c8a33a3d4d4617/3rdparty/find_dependencies.cmake#L465)):*
> ```CMake
>     TARGETS Threads::Threads
> )
> if(THREADS_HAVE_PTHREAD_ARG)
> ```

This internal **cache** variable is responsible for turning on `-pthread`. When I set this here
```
set(THREADS_HAVE_PTHREAD_ARG FALSE CACHE INTERNAL "Forced to FALSE since THREADS_PREFER_PTHREAD_FLAG is set to FALSE" FORCE)
```
then only the `benchmark` target links whereas all other targets are not recompiled at all. 

**Pros:** Does not change the `Threads::Threads` target.
**Cons:** Changes internal cache variable + `benchmark` still links, but does not recompile.
___
*[3rdparty/find_dependencies.cmake, line 466 at r1](https://reviewable.io/reviews/isl-org/open3d/4076#-MkGQo8FDgfWVrwpZM7q:-MkGQo8G9_1SwW5Qef7u:b-bzkk9i) ([raw file](https://github.com/isl-org/open3d/blob/1fa38412184db67fe952fde7b8c8a33a3d4d4617/3rdparty/find_dependencies.cmake#L466)):*
> ```CMake
> )
> if(THREADS_HAVE_PTHREAD_ARG)
>     set_property(TARGET Threads::Threads
> ```

This is exactly what we could report to CMake as a solution for the underlying problem with ISPC.

We should move this code closer to where we set `THREADS_PREFER_PTHREAD_FLAG`.  There are two possible similar, but lightly different ideas here:
```
set(THREADS_PREFER_PTHREAD_FLAG TRUE)

macro(open3d_patch_findthreads_module_)
    if(TARGET Threads::Threads AND THREADS_HAVE_PTHREAD_ARG)
        set_property(TARGET Threads::Threads
                     PROPERTY INTERFACE_COMPILE_OPTIONS
                     "$<$<COMPILE_LANG_AND_ID:CUDA,NVIDIA>:SHELL:-Xcompiler -pthread>"
                     "$<$<AND:$<NOT:$<COMPILE_LANGUAGE:ISPC>>,$<NOT:$<COMPILE_LANG_AND_ID:CUDA,NVIDIA>>>:-pthread>")
    endif()
endmacro()

# Option 1:
find_package(Threads REQUIRED)
open3d_patch_findthreads_module_()

# Option 2:
cmake_language(DEFER "open3d_patch_findthreads_module_()")
```

I am leaning towards Option 2 as it does not force to find the package at this stage. What is your opinion @yxlao?

yxlao (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4076)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @stotko)

___
*[3rdparty/find_dependencies.cmake, line 465 at r1](https://reviewable.io/reviews/isl-org/open3d/4076#-MkGPvE-AFe4Unn_R_ZP:-MkGnb_EEJk6KEHktACT:b-qxafb0) ([raw file](https://github.com/isl-org/open3d/blob/1fa38412184db67fe952fde7b8c8a33a3d4d4617/3rdparty/find_dependencies.cmake#L465)):*

This internal **cache** variable is responsible for turning on `-pthread`. When I set this here
```
set(THREADS_HAVE_PTHREAD_ARG FALSE CACHE INTERNAL "Forced to FALSE since THREADS_PREFER_PTHREAD_FLAG is set to FALSE" FORCE)
```
then only the `benchmark` target links whereas all other targets are not recompiled at all. 

**Pros:** Does not change the `Threads::Threads` target.
**Cons:** Changes internal cache variable + `benchmark` still links, but does not recompile.
</blockquote></details>

Implemented Option 2 so we'll be changing `Threads::Threads` for now.
___
*[3rdparty/find_dependencies.cmake, line 466 at r1](https://reviewable.io/reviews/isl-org/open3d/4076#-MkGQo8FDgfWVrwpZM7q:-MkGnXlZ5OfN236y9-tB:bygbw89) ([raw file](https://github.com/isl-org/open3d/blob/1fa38412184db67fe952fde7b8c8a33a3d4d4617/3rdparty/find_dependencies.cmake#L466)):*

This is exactly what we could report to CMake as a solution for the underlying problem with ISPC.

We should move this code closer to where we set `THREADS_PREFER_PTHREAD_FLAG`.  There are two possible similar, but lightly different ideas here:
```
set(THREADS_PREFER_PTHREAD_FLAG TRUE)

macro(open3d_patch_findthreads_module_)
    if(TARGET Threads::Threads AND THREADS_HAVE_PTHREAD_ARG)
        set_property(TARGET Threads::Threads
                     PROPERTY INTERFACE_COMPILE_OPTIONS
                     "$<$<COMPILE_LANG_AND_ID:CUDA,NVIDIA>:SHELL:-Xcompiler -pthread>"
                     "$<$<AND:$<NOT:$<COMPILE_LANGUAGE:ISPC>>,$<NOT:$<COMPILE_LANG_AND_ID:CUDA,NVIDIA>>>:-pthread>")
    endif()
endmacro()

# Option 1:
find_package(Threads REQUIRED)
open3d_patch_findthreads_module_()

# Option 2:
cmake_language(DEFER "open3d_patch_findthreads_module_()")
```

I am leaning towards Option 2 as it does not force to find the package at this stage. What is your opinion @yxlao?
</blockquote></details>

Done. Implemented Option 2. 
Yes, we shall report this to the upstream cmake.

stotko (approved):
> Reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4076)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @yxlao)

PR #3956, opened 2021-08-18
Tensorboard plugin

ssheorey:
Initial Open3D for TensorBoard plugin. Save intermediate 3D data (point clouds, meshes, line sets) during an iterative optimization procedure (e.g.: 3D semantic segmentation, registration,...) and visualize data through TensorBoard.

### Features:
- Summary writing: Save 3D data to msgpack files on disk and write an index in TensorBoard event files. Async non-blocking writer. Point cloud, mesh and line sets supported.
- De-duplication of written 3D summaries: e.g. If point cloud labels or colors keep changing with iteration, but normals are the same, you can ask Open3D to use normals from step=0 for all iterations. ~~However, primary keys (`*_positions`, `*_indices` cannot be de-duplicated as of now due to validation checks in serialization.)~~ Supported through PR #3963 
- Cached reading of 3D data.
- Display 3D data. Multiple geometry elements to be displayed can be selected with `tags`. The user can scroll through time with a slider for `step`. Geometry element can be written in batches and a specific `batch_idx` can be chosen for display. Finally, multiple experiment `run`s can be displayed in different windows for comparison.
- Displays in multiple windows can show the same iteration step or batch index.
- Tested with large meshes: >30M points and >60M triangles.
- [x] Deduplication for primary keys (`*_positions`, `*_indices`).
- [x] Test with PyTorch and remove TensorFlow dependency.
- [x] Switch http handling from civetweb to TensorBoard. (civetweb needs extra port and browser CORS permissions may prevent secure remote access). Currently this causes a deadlock. This blocks use on Chrome. -> **Fixed**
- [x] Add unit tests for self functions and TensorBoard / TF / PyTorch APIs. TF 2.6.0 has a breaking change for a private function from 2.5.0 -> **Fixed** Ask for explicit `logdir` in TF, since there is no public API and the private API was removed in 2.6.0.
- [x] Validation of msgpack data with CRC-32C
- [x] Jupyter support -> Works without any customization.

### Notes for reviewers: 
- TensorBoard plugin discovery doesn't play nice with our method of including the 3DML repo, so keeping the code in the main repo. for now.
- ~~LineSet needs RPC/mspgpack support from PR #3821.~~ Merged.
- ~~Ready for review now, but set as draft till PR  #3821 is merged.~~ Merged.

 ### TODO (future PRs):
- [ ] Support for materials [Depends on PR #3858 and PR #3985]
- [ ] Allow syncing camera viewpoints for displayed runs.
[...]

PR #3858, opened 2021-08-03
Add suppport for Material Properties to Tensor Geometries

errissa:

PR #4078, opened 2021-09-22
Tensorboard plugin materials support

ssheorey:
For sample use see: `python/test/visualization/tensorboard_plugin.py`:`test_tensorflow_summary()`

Updated usage over previous Geometry add_3d():
```
add_3d(name, data, step, logdir=None, max_outputs=1, description=None)
    Write 3D geometry data as summary.
    
    Args:
      name (str): A name or tag for this summary. The summary tag used for TensorBoard
        will be this name prefixed by any active name scopes.
      data (dict): A dictionary of tensors representing 3D data. Tensorflow,
        PyTorch, Numpy and Open3D tensors are supported. The following keys
        are supported:
          - ``vertex_positions``: shape `(B, N, 3)` where B is the number of
                point clouds and must be same for each key. N is the number of
                3D points. Will be cast to ``float32``.
          - ``vertex_colors``: shape `(B, N, 3)` Will be converted to ``uint8``.
          - ``vertex_normals``: shape `(B, N, 3)` Will be cast to ``float32``.
          -  ``texture_uvs``: shape `(B, N, 2)` UV coordinates for applying
                material texture maps. Will be cast to ``float32``.
          - ``triangle_indices``: shape `(B, Nf, 3)`. Will be cast to ``uint32``.
          - ``line_indices``: shape `(B, Nl, 2)`. Will be cast to ``uint32``.
          - ``material_name``: shape `(B,)` and dtype ``str``. PBR material name is
            required to specify any material properties.  Open3D built-in
            materials: ``defaultLit``, ``defaultUnlit``, ``unlitLine``.
          - ``material_scalar_*PROPERTY*``: Any material scalar property with
            float values of shape `(B,)`. e.g. To specify the property
            `metallic`, use the key `material_scalar_metallic`.
          -  ``material_vector_*PROPERTY*``: Any material 4-vector property with
             float values of shape `(B, 4)` e.g. To specify the property
            `baseColor`, use the key `material_vector_baseColor`.
          -  ``material_texture_map_*TEXTURE_MAP*``: PBR Material texture maps.
             e.g. ``material_texture_map_metallic`` represents a texture map
             describing the metallic property for rendering.
             Values are Tensors with shape `(B, Nr, Nc, C)`, corresponding to a
             batch of texture maps with `C` channels and shape `(Nr, Nc)`. The
             geometry must have ``texture_uvs`` coordinates to use any texture
             map.
    
        For batch_size B=1, the tensors may drop a rank (e.g. (N,3)
        vertex_positions, (4,) material vector properties or float scalar ()
        material scalar properties.)
    
        Floating point color and texture map data will be clipped to the range
        [0,1] and converted to ``uint8`` range [0,255]. ``uint16`` data will be
        compressed to the range [0,255].
    
        Any data tensor (with ndim>=2, including batch_size, i.e. excluding
        ``material_name`` and ``material_scalar_*PROPERTY*``), may be replaced by
        an ``int`` scalar referring to a previous step. This allows reusing a
        previously written property in case that it does not change at different
        steps.
    
        Please see the `Filament Materials Guide
        <https://google.github.io/filament/Materials.html#materialmodels>`__ for
        a complete explanation of material properties.
...
```
Depends on these PRs:
- #3858 
- #3985 
- #3956


cpp/open3d/t/geometry/TriangleMesh.cpp
```diff
@@ -197,6 +197,31 @@ geometry::TriangleMesh TriangleMesh::FromLegacy(
                 core::eigen_converter::EigenVector3dVectorToTensor(
                         mesh_legacy.triangle_normals_, float_dtype, device));
     }
+    if (mesh_legacy.HasTriangleUvs()) {
+        // Legacy uses UVs as triangle attributes
+        // Check length and issue warning on mismatch
+        if (mesh_legacy.triangle_uvs_.size() != mesh_legacy.vertices_.size()) {
```
reyanshsolis, line 203: UVs map vertices of each triangle to a 2D coordinate [Image of (h, w) : (1, 1)]. 
So, each vertex in a triangle has a 2D UV coordinate, therefore 3 such coordinates per triangle.

A vertex can have multiple UV coordinates (for example, if we un-wrap a cylinder on a 2D plane, the cut vertices will be having 2 UV coordinates).

My suggestion would be, to use `texture_uvs`, as a `triangle tensor map attribute`, of shape `[N, 3, 2]`, where `N` is the Length of `Triangle Indices`. 
In Legacy, we have a vector of `Eigen::Vector2d` of length `3xN`,
so, we need to reshape [N x 3, 2] (output of `EigenVector2dVectorToTensor`) -> [N, 3, 2] in tensor.

so the following might work: (not tested)
```cpp
// we have a function in legacy to check UV shape.
if(!mesh.HasTriangleUvs()) {
            utility::LogWarning(
                    "Ignoring UV coordinates since length(UV) ({}) != "
                    "3 x n_triangles ({})",
                    mesh_legacy.triangle_uvs_.size(),
                    mesh_legacy.triangles_.size());
} else {
            mesh.SetVertexAttr(
                    "texture_uvs",
                    core::eigen_converter::EigenVector2dVectorToTensor(
                            mesh_legacy.triangle_uvs_, float_dtype, device)).Reshape(mesh_legacy.triangles_.size(), 3, 2);
}
```

For `Tensor` -> `Legacy`
We will need to `Reshape` `[N, 3, 2]` back to `[N x 3, 2]`, and then use `TensorToEigenVector2dVector`

PR #4080, opened 2021-09-23
Simplify ISPC language emulation module

stotko:
This PR simplifies parts of the logic of the ISPC language emulation module and relaxes some limitations:

- Replace `ISPC_FORCE_LEGACY` with `ISPC_USE_LEGACY_EMULATION` to expose precise control to enable and disable (not support by old option) emulation.
- Move the default condition, i.e. checking for Make and Ninja generators, to the top-level CMake file to a single and consistent place. Furthermore, check for `CMAKE_ISPC_COMPILER_LOADED` to not use emulation if first-class support is already enabled.
- Recursively process the target dependency tree rather than only considering first-level dependenciesto capture (almost) all properties. `PROCESSED_TARGET_LIBRARIES` is used to break potential cyclic dependencies.
- Remove `open3d_ispc_target_sources` macro and process the `SOURCES` property instead of the custom `ISPC_SOURCES` property to setup build rules.
- Support sources files with absolute file paths.
- Infer ISPC output extension from either C or C++ language and issue an error if none is available.

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4080)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -341,8 +341,12 @@ if (BUILD_ISPC_MODULE)
 
     message(STATUS "Using ISPC instruction sets: ${CMAKE_ISPC_INSTRUCTION_SETS}")
 
-    option(ISPC_FORCE_LEGACY "Force legacy ISPC language emulation over first-class CMake support" OFF)
-    mark_as_advanced(ISPC_FORCE_LEGACY)
+    if (CMAKE_ISPC_COMPILER_LOADED OR (CMAKE_GENERATOR MATCHES "Make" OR CMAKE_GENERATOR MATCHES "Ninja"))
+        option(ISPC_USE_LEGACY_EMULATION "Use legacy ISPC language emulation over first-class CMake support" OFF)
+    else()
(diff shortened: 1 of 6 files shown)


yxlao:
Could you remind me of the use case for the emulation mode? Shall we assume the user is using the latest CMake with ISPC support already?

stotko:
> Could you remind me of the use case for the emulation mode? Shall we assume the user is using the latest CMake with ISPC support already?

The ISPC language emulation module is required to allow building ISPC on Visual Studio generators. CMake only supports Make and Ninja generators so far. Therefore, we need that module to bring ISPC support to Windows. By default, we use CMake's first-class support if it is available (Linux + MacOS) and the emulation otherwise (Windows).

yxlao (approved):
> Reviewed 6 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4080)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @stotko)

PR #4089, opened 2021-09-24
uninstall before install-pip-package

yxlao:
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4089)


yxlao:
Minor change in PR, merge directly.

PR #4092, opened 2021-09-24
silent wget in docker build

yxlao:
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4092)


yxlao:
Simple PR, merge directly

PR #4097, opened 2021-09-25
Support for multiple depth format in Python reconstruction example

yutashx:
The purpose of this PR is to support multiple formats of depth images in `examples/python/reconstruction_system`.
The current reconstruction_system does not reflect the `depth_map_type` parameter in `reconstruction_system/initialize_config.py`.

So I improved it to load depth images with the corresponding function `o3d.geometry,RGBDImage.create_rgbd_from_<depth_map_type>_format` when the `depth_map_type` parameter is `redwood`, `sun`, `nyu`, and `tum`. 
I have also improved it to read depth images with `o3d.geometry.RGBDImage.create_from_color_and_depth` if parameters other than `redwood`, `sun`, `nyu`, and `tum` are set.
Also, since the color and depth images in the NYU dataset have ppm and pgm extensions, I modified `example/python/utility/file.py` to read them.
I referred to the [tutorial RGBD images](http://www.open3d.org/docs/release/tutorial/geometry/rgbd_image.html) when implementing `read_nyu_pgm`.

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4097)

### DIFF
--- a/examples/python/reconstruction_system/make_fragments.py
+++ b/examples/python/reconstruction_system/make_fragments.py
@@ -29,8 +29,8 @@
 import math
 import re
 import sys
-import numpy as np
 import matplotlib.image as mpimg
+import numpy as np
 import open3d as o3d
 sys.path.append("../utility")
 from file import join, make_clean_folder, get_rgbd_file_lists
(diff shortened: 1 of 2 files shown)


ssheorey:
Thank you for this PR. Please update and re-submit with the new data API.

PR #4099, opened 2021-09-25
Tensor iterator

yxlao:
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4099)

### DIFF
--- a/cpp/open3d/core/Tensor.cpp
+++ b/cpp/open3d/core/Tensor.cpp
@@ -205,7 +205,7 @@ struct Tensor::Iterator::Impl {
                            // of the tensor for Iterator::operator->.
 };
 
-Tensor::Iterator::Iterator(Tensor* tensor, int64_t index)
+Tensor::Iterator::Iterator(pointer tensor, int64_t index)
     : impl_(std::make_unique<Impl>()) {
     impl_->tensor_ = tensor;
     impl_->index_ = index;
@@ -272,7 +272,7 @@ struct Tensor::ConstIterator::Impl {
                            // of the tensor for ConstIterator::operator->.
 };
 
-Tensor::ConstIterator::ConstIterator(const Tensor* tensor, int64_t index)
+Tensor::ConstIterator::ConstIterator(pointer tensor, int64_t index)
     : impl_(std::make_unique<Impl>()) {
     impl_->tensor_ = tensor;
     impl_->index_ = index;
--- a/cpp/open3d/core/Tensor.cpp
+++ b/cpp/open3d/core/Tensor.cpp
@@ -197,6 +197,73 @@ public:
     }
 };
 
+struct Tensor::Iterator::Impl {
+    Tensor* tensor_;
+    int64_t index_;
+    Tensor tensor_slice_;  // Stores temporary tensor slice with shared memory
+                           // as the original tensor. This allows taking the &
+                           // of the tensor for Iterator::operator->.
+};
+
+Tensor::Iterator::Iterator(Tensor* tensor, int64_t index)
+    : impl_(std::make_unique<Impl>()) {
+    impl_->tensor_ = tensor;
+    impl_->index_ = index;
+}
+
+Tensor::Iterator::Iterator(const Tensor::Iterator& other) {
+    impl_ = std::make_unique<Impl>();
+    impl_->tensor_ = other.impl_->tensor_;
+    impl_->index_ = other.impl_->index_;
+}
+
+// Empty destructor since Impl is incomplete type in Tensor.h.
+// https://stackoverflow.com/a/34073093/1255535
+Tensor::Iterator::~Iterator() {}
+
+Tensor::Iterator::reference Tensor::Iterator::operator*() const {
+    return impl_->tensor_->operator[](impl_->index_);
+}
+
+Tensor::Iterator::pointer Tensor::Iterator::operator->() {
+    impl_->tensor_slice_ = impl_->tensor_->operator[](impl_->index_);
+    return &impl_->tensor_slice_;
+}
+
+Tensor::Iterator& Tensor::Iterator::operator++() {
+    impl_->index_++;
+    return *this;
+}
+
+Tensor::Iterator Tensor::Iterator::operator++(int) {
+    Iterator tmp(impl_->tensor_, impl_->index_);
+    impl_->index_++;
+    return tmp;
+}
+
+bool operator==(const Tensor::Iterator& a, const Tensor::Iterator& b) {
+    return a.impl_->tensor_ == b.impl_->tensor_ &&
+           a.impl_->index_ == b.impl_->index_;
+}
+
+bool operator!=(const Tensor::Iterator& a, const Tensor::Iterator& b) {
+    return !(a == b);
+}
+
+Tensor::Iterator Tensor::begin() {
+    if (NumDims() == 0) {
+        utility::LogError("Cannot iterate a scalar (0-dim) tensor.");
+    }
+    return Iterator(this, 0);
+}
+
+Tensor::Iterator Tensor::end() {
+    if (NumDims() == 0) {
+        utility::LogError("Cannot iterate a scalar (0-dim) tensor.");
+    }
+    return Iterator(this, shape_[0]);
+}
+
 // Equivalent to `Tensor& operator=(const Tensor& other) & = default;`.
 // Manual implentaiton is need to avoid MSVC bug (error C2580:  multiple
 // versions of a defaulted special member functions are not allowed.)
--- a/cpp/open3d/core/Tensor.cpp
+++ b/cpp/open3d/core/Tensor.cpp
@@ -264,6 +264,75 @@ Tensor::Iterator Tensor::end() {
     return Iterator(this, shape_[0]);
 }
 
+struct Tensor::ConstIterator::Impl {
+    const Tensor* tensor_;
+    int64_t index_;
+    Tensor tensor_slice_;  // Stores temporary tensor slice with shared memory
+                           // as the original tensor. This allows taking the &
+                           // of the tensor for ConstIterator::operator->.
+};
+
+Tensor::ConstIterator::ConstIterator(const Tensor* tensor, int64_t index)
+    : impl_(std::make_unique<Impl>()) {
+    impl_->tensor_ = tensor;
+    impl_->index_ = index;
+}
+
+Tensor::ConstIterator::ConstIterator(const Tensor::ConstIterator& other)
+    : impl_(std::make_unique<Impl>()) {
+    impl_->tensor_ = other.impl_->tensor_;
+    impl_->index_ = other.impl_->index_;
+}
+
+// Empty destructor since Impl is incomplete type in Tensor.h.
+// https://stackoverflow.com/a/34073093/1255535
+Tensor::ConstIterator::~ConstIterator() {}
+
+Tensor::ConstIterator::reference Tensor::ConstIterator::operator*() const {
+    return impl_->tensor_->operator[](impl_->index_);
+}
+
+Tensor::ConstIterator::pointer Tensor::ConstIterator::operator->() const {
+    impl_->tensor_slice_ = impl_->tensor_->operator[](impl_->index_);
+    return &impl_->tensor_slice_;
+}
+
+Tensor::ConstIterator& Tensor::ConstIterator::operator++() {
+    impl_->index_++;
+    return *this;
+}
+
+Tensor::ConstIterator Tensor::ConstIterator::operator++(int) {
+    ConstIterator tmp(impl_->tensor_, impl_->index_);
+    impl_->index_++;
+    return tmp;
+}
+
+bool Tensor::ConstIterator::operator==(
+        const Tensor::ConstIterator& other) const {
+    return impl_->tensor_ == other.impl_->tensor_ &&
+           impl_->index_ == other.impl_->index_;
+}
+
+bool Tensor::ConstIterator::operator!=(
+        const Tensor::ConstIterator& other) const {
+    return !(*this == other);
+}
+
+Tensor::ConstIterator Tensor::cbegin() const {
+    if (NumDims() == 0) {
+        utility::LogError("Cannot iterate a scalar (0-dim) tensor.");
+    }
+    return ConstIterator(this, 0);
+}
+
+Tensor::ConstIterator Tensor::cend() const {
+    if (NumDims() == 0) {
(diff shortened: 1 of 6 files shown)


ssheorey (approved):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4099)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @stotko and @yxlao)

___
*[cpp/tests/core/Tensor.cpp, line 3043 at r1](https://reviewable.io/reviews/isl-org/open3d/4099#-MkZ95FE6C8fbTEHB1zY:-MkZ95FFD2pHbC5nffAc:bk9vjwe) ([raw file](https://github.com/isl-org/open3d/blob/4e3d0b4fdb94f746c2e34d1b210340ca5a36f2d8/cpp/tests/core/Tensor.cpp#L3043)):*
> ```cpp
>         t_slice.AsRvalue() = 10;
>     }
>     EXPECT_TRUE(t.AllClose(core::Tensor::Init<int>({10, 10, 10}, device)));
> ```

[nit] Should be exact equal, since there is no arithmetic operation to introduce numeric errors. Also in the checks below.

yxlao (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4099)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @ssheorey and @stotko)

___
*[cpp/tests/core/Tensor.cpp, line 3043 at r1](https://reviewable.io/reviews/isl-org/open3d/4099#-MkZ95FE6C8fbTEHB1zY:-Mk_Qxcq2xxMfbqO7Bcr:b-gm26ew) ([raw file](https://github.com/isl-org/open3d/blob/4e3d0b4fdb94f746c2e34d1b210340ca5a36f2d8/cpp/tests/core/Tensor.cpp#L3043)):*

[nit] Should be exact equal, since there is no arithmetic operation to introduce numeric errors. Also in the checks below.
</blockquote></details>

Done. Previously we don't have this function. Added `Tensor::AllEqual()`.

stotko (changes requested):
> Reviewed 2 of 5 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4099)* status: all files reviewed, 4 unresolved discussions (waiting on @ssheorey and @yxlao)

___
*[cpp/open3d/core/Tensor.h, line 1074 at r2](https://reviewable.io/reviews/isl-org/open3d/4099#-MkaEAnf1f7LUeCCEsCJ:-MkaEAnf1f7LUeCCEsCK:b4maquz) ([raw file](https://github.com/isl-org/open3d/blob/74d301588cf0da64529b20a4b9498337845f1d60/cpp/open3d/core/Tensor.h#L1074)):*
> ```objc
> 
>     /// Iterator for Tensor.
>     struct Iterator {
> ```

We also need a `ConstIterator` class with `using value_type = const Tensor`, etc. to be used in contexts where we want to iterate over a `const Tensor& tensor` instance. This includes `ConstIterator begin() const` (and `end` respectively) overloads. For completeness, we should also define `cbegin()` and `cend()` as further "overloads" to the above `const` versions.
___
*[cpp/open3d/core/Tensor.h, line 1088 at r2](https://reviewable.io/reviews/isl-org/open3d/4099#-MkaFlRu7IEjvNZeJ_wS:-MkaFlRu7IEjvNZeJ_wT:bl42d6n) ([raw file](https://github.com/isl-org/open3d/blob/74d301588cf0da64529b20a4b9498337845f1d60/cpp/open3d/core/Tensor.h#L1088)):*
> ```objc
>         ~Iterator();
>         reference operator*() const;
>         pointer operator->();
> ```

nit: Could also be made `const`.
___
*[cpp/open3d/core/Tensor.h, line 1092 at r2](https://reviewable.io/reviews/isl-org/open3d/4099#-MkaDJsAC6a_8eoR6t3J:-MkaDJsBCwXXUaDS8Jtj:b-bf4zi9) ([raw file](https://github.com/isl-org/open3d/blob/74d301588cf0da64529b20a4b9498337845f1d60/cpp/open3d/core/Tensor.h#L1092)):*
> ```objc
>         Iterator operator++(int);
>         friend bool operator==(const Iterator& a, const Iterator& b);
>         friend bool operator!=(const Iterator& a, const Iterator& b);
> ```

nit: Only `operator==` needs to be a friend.

Alternatively, we could remove both friend declarations and define a `bool operator(const Iterator& other) const` member function. I think just having `operator==` in its current form as a friend is clear enough.

yxlao:
While it is possible to make `Iterator` inherit `ConstIterator`, it will make the code less symmetrical. The current approach has some duplicated code but it should be easier to read than the inheritance case.

stotko (changes requested):
> Reviewed 4 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4099)* status: all files reviewed, 6 unresolved discussions (waiting on @ssheorey and @yxlao)

___
*[cpp/open3d/core/Tensor.h, line 1078 at r4](https://reviewable.io/reviews/isl-org/open3d/4099#-MkkLcKC8lBfBMygR9hd:-MkkLcKC8lBfBMygR9he:b-ylmkqa) ([raw file](https://github.com/isl-org/open3d/blob/9f64e53d9b7104dd4d2791e3a5f9a87dbd96ed09/cpp/open3d/core/Tensor.h#L1078)):*
> ```objc
>         using difference_type = std::ptrdiff_t;
>         using value_type = Tensor;
>         using pointer = Tensor*;
> ```

See comment for `ConstIterator`. The same consistency enforcement can be applied here as well.
___
*[cpp/open3d/core/Tensor.h, line 1084 at r4](https://reviewable.io/reviews/isl-org/open3d/4099#-MkkM9S6-apDWLesBfLo:-MkkM9S6-apDWLesBfLp:b-9zjvuc) ([raw file](https://github.com/isl-org/open3d/blob/9f64e53d9b7104dd4d2791e3a5f9a87dbd96ed09/cpp/open3d/core/Tensor.h#L1084)):*
> ```objc
>         // Iterator must be constructible, copy-constructible, copy-assignable,
>         // destructible and swappable.
>         Iterator(Tensor* tensor, int64_t index);
> ```

nit: The first parameter could be changed to `pointer` type.
___
*[cpp/open3d/core/Tensor.h, line 1104 at r4](https://reviewable.io/reviews/isl-org/open3d/4099#-MkkKyvt4LoRJfVKDF9t:-MkkKyvt4LoRJfVKDF9u:bgjxosn) ([raw file](https://github.com/isl-org/open3d/blob/9f64e53d9b7104dd4d2791e3a5f9a87dbd96ed09/cpp/open3d/core/Tensor.h#L1104)):*
> ```objc
>         using difference_type = std::ptrdiff_t;
>         using value_type = const Tensor;
>         using pointer = Tensor*;
> ```

This should be `const Tensor*`. I would suggest to define it as `value_type*` which would automatically enforce consistency. Same for `reference` which should be `value_type` in this case.
___
*[cpp/open3d/core/Tensor.h, line 1110 at r4](https://reviewable.io/reviews/isl-org/open3d/4099#-MkkMIy866bJviIyCcFx:-MkkMIy866bJviIyCcFy:b-9zjvuc) ([raw file](https://github.com/isl-org/open3d/blob/9f64e53d9b7104dd4d2791e3a5f9a87dbd96ed09/cpp/open3d/core/Tensor.h#L1110)):*
> ```objc
>         // ConstIterator must be constructible, copy-constructible,
>         // copy-assignable, destructible and swappable.
>         ConstIterator(const Tensor* tensor, int64_t index);
> ```

nit: The first parameter could be changed to `pointer` type.
___
*[cpp/tests/core/Tensor.cpp, line 3207 at r4](https://reviewable.io/reviews/isl-org/open3d/4099#-MkkNNKn9ZGqy1bM0_op:-MkkNNKn9ZGqy1bM0_oq:b283g5o) ([raw file](https://github.com/isl-org/open3d/blob/9f64e53d9b7104dd4d2791e3a5f9a87dbd96ed09/cpp/tests/core/Tensor.cpp#L3207)):*
> ```cpp
>     }
>     for (core::Tensor &&t_slice : AsConst(t)) {
>         t_slice.AsRvalue() = 10;
> ```

This should not be possible for the object `t` which is marked as `const`. Fixing `reference` and `pointer` as suggested in the comment above, might be already sufficient to take care of this. Please double check this.

stotko:
> While it is possible to make `Iterator` inherit `ConstIterator`, it will make the code less symmetrical. The current approach has some duplicated code but it should be easier to read than the inheritance case.

I agree. Inheritance would also allow some weird `dynamic_cast` stuff which users really should not try. I also added further comments. LGTM, however the semantics for the `ConstIterator` should be double-checked such that attempting to change the state of the underlying `Tensor` should be a compiler error.

yxlao (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4099)* status: all files reviewed, 4 unresolved discussions (waiting on @ssheorey and @stotko)

___
*[cpp/open3d/core/Tensor.h, line 1078 at r4](https://reviewable.io/reviews/isl-org/open3d/4099#-MkkLcKC8lBfBMygR9hd:-MklFT2K6mtDEihKa4ud:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/9f64e53d9b7104dd4d2791e3a5f9a87dbd96ed09/cpp/open3d/core/Tensor.h#L1078)):*

See comment for `ConstIterator`. The same consistency enforcement can be applied here as well.
</blockquote></details>

Done.
___
*[cpp/open3d/core/Tensor.h, line 1104 at r4](https://reviewable.io/reviews/isl-org/open3d/4099#-MkkKyvt4LoRJfVKDF9t:-MklEY_3EHnTLlOqY5S3:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/9f64e53d9b7104dd4d2791e3a5f9a87dbd96ed09/cpp/open3d/core/Tensor.h#L1104)):*

This should be `const Tensor*`. I would suggest to define it as `value_type*` which would automatically enforce consistency. Same for `reference` which should be `value_type` in this case.
</blockquote></details>

Done.
___
*[cpp/tests/core/Tensor.cpp, line 3207 at r4](https://reviewable.io/reviews/isl-org/open3d/4099#-MkkNNKn9ZGqy1bM0_op:-MklFGkJ9vVpuQ81VZaE:b-236swe) ([raw file](https://github.com/isl-org/open3d/blob/9f64e53d9b7104dd4d2791e3a5f9a87dbd96ed09/cpp/tests/core/Tensor.cpp#L3207)):*

This should not be possible for the object `t` which is marked as `const`. Fixing `reference` and `pointer` as suggested in the comment above, might be already sufficient to take care of this. Please double check this.
</blockquote></details>

That's right. This test is removed.

stotko (approved):
> Reviewed 3 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4099)* status: all files reviewed, 3 unresolved discussions (waiting on @ssheorey and @yxlao)

*[a discussion](https://reviewable.io/reviews/isl-org/open3d/4099#-MklIRiS4nv3_gBGqO-z:-MklIRiS4nv3_gBGqO0-:bdm89vp) (no related file):*
LGTM. There is just a minor consistency issue and the `AsRvalue()` problem. The latter might have side effects to the remaining code base, so it is okay to address this in a separate PR.

---
___
*[cpp/open3d/core/Tensor.h, line 1084 at r4](https://reviewable.io/reviews/isl-org/open3d/4099#-MkkM9S6-apDWLesBfLo:-MklHWlQ8TYHA1PTW5Z0:bepdg0w) ([raw file](https://github.com/isl-org/open3d/blob/9f64e53d9b7104dd4d2791e3a5f9a87dbd96ed09/cpp/open3d/core/Tensor.h#L1084)):*

nit: The first parameter could be changed to `pointer` type.
</blockquote></details>

The implementation has been updated, but the declarations seems not.
___
*[cpp/open3d/core/Tensor.h, line 1110 at r4](https://reviewable.io/reviews/isl-org/open3d/4099#-MkkMIy866bJviIyCcFx:-MklHIOs9hGJcIBecMV6:bepdg0w) ([raw file](https://github.com/isl-org/open3d/blob/9f64e53d9b7104dd4d2791e3a5f9a87dbd96ed09/cpp/open3d/core/Tensor.h#L1110)):*

nit: The first parameter could be changed to `pointer` type.
</blockquote></details>

The implementation has been updated, but the declarations seems not.
___
*[cpp/tests/core/Tensor.cpp, line 3222 at r4](https://reviewable.io/reviews/isl-org/open3d/4099#-MklHaIUCVa1rHkYpRHB:-MklHaIUCVa1rHkYpRHC:b-jy8w5e) ([raw file](https://github.com/isl-org/open3d/blob/9f64e53d9b7104dd4d2791e3a5f9a87dbd96ed09/cpp/tests/core/Tensor.cpp#L3222)):*
> ```cpp
>     for (core::Tensor::ConstIterator iter = t.cbegin(); iter != t.cend();
>          ++iter) {
>         iter->AsRvalue() = 10;
> ```

This rvalue casting still seems to be possible to force writing to the `const` tensor. It looks like we currently have
```
Tensor AsRvalue() const;
```
in `Tensor.h` which does not seem to fully respect `const` semantics. I think this should actually be
```
Tensor AsRvalue();
const Tensor AsRvalue() const;
```
instead.

yxlao (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4099)* status: 4 of 6 files reviewed, 3 unresolved discussions (waiting on @ssheorey, @stotko, and @yxlao)

___
*[cpp/tests/core/Tensor.cpp, line 3222 at r4](https://reviewable.io/reviews/isl-org/open3d/4099#-MklHaIUCVa1rHkYpRHB:-MklXKDTEhefVG4el-sw:bg803ts) ([raw file](https://github.com/isl-org/open3d/blob/9f64e53d9b7104dd4d2791e3a5f9a87dbd96ed09/cpp/tests/core/Tensor.cpp#L3222)):*

This rvalue casting still seems to be possible to force writing to the `const` tensor. It looks like we currently have
```
Tensor AsRvalue() const;
```
in `Tensor.h` which does not seem to fully respect `const` semantics. I think this should actually be
```
Tensor AsRvalue();
const Tensor AsRvalue() const;
```
instead.
</blockquote></details>

That's right. Good catch. 

`Tensor.h` has been updated. Tests are updated.

stotko:
The `Tensor/TensorPermuteDevicePairs.AllEqual` always fails in `Ubuntu GPU CI`, see e.g. https://github.com/isl-org/Open3D/pull/4099/checks?check_run_id=3744048736

PR #4101, opened 2021-09-26
explicitly set max ccache size in docker build

yxlao:
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4101)


yxlao:
Minor change, merge directly.

PR #4102, opened 2021-09-26
chown inside docker, avoids sudo

yxlao:
This avoids `sudo` from the host.

This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4102)


yxlao:
Minor change, directly merge.

PR #4105, opened 2021-09-26
Asserts in `t`

reyanshsolis:
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4105)


yxlao (changes requested):
> Reviewed 6 of 13 files at r1, 2 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4105)* status: 8 of 14 files reviewed, 7 unresolved discussions (waiting on @reyanshsolis and @yxlao)

___
*[cpp/open3d/t/geometry/Image.cpp, line 114 at r2](https://reviewable.io/reviews/isl-org/open3d/4105#-MkgPo6u1APGVSx0wAHb:-MkgPo6v3gzpIEHC5Yne:b-nfk2ga) ([raw file](https://github.com/isl-org/open3d/blob/4c20a5e115cd43a68f4ba8275a3e88be5d128c78/cpp/open3d/t/geometry/Image.cpp#L114)):*
> ```cpp
>             scale = 1.0 /
>                     static_cast<double>(std::numeric_limits<uint16_t>::max());
>         } else if (GetDtype() == core::UInt32) {
> ```

We don't need to scale unit32 and uint64. This is specified in the `Image.h` header. 

Intuitively,  uint32 and uint64 have pretty large ranges already, so it makes sense to convert it to float directly without scaling. 

In fact, we can revisit the semantics in the future. But for now, let's keep it unchanged. 
___
*[cpp/open3d/t/geometry/Image.cpp, line 243 at r2](https://reviewable.io/reviews/isl-org/open3d/4105#-MkgQm7M7gJFtbUa14EB:-MkgQm7M7gJFtbUa14EC:b-4fe0yu) ([raw file](https://github.com/isl-org/open3d/blob/4c20a5e115cd43a68f4ba8275a3e88be5d128c78/cpp/open3d/t/geometry/Image.cpp#L243)):*
> ```cpp
> 
>     core::AssertTensorDtypes(
>             AsTensor(), {core::Bool, core::UInt8, core::UInt16, core::Float32});
> ```

`ipp` and `npp` are not the same here. So we cannot check their dtypes directly. So this refactoring is not equivalent. 

I suggest keeping our original way of checking `{dtype, num_channels}` in a bundle. for all functions in `Image.cpp`. We can revisit to refactor `Image.cpp` and support all unsupported cases with direct tensor kernels.
___
*[cpp/open3d/t/geometry/PointCloud.cpp, line 121 at r2](https://reviewable.io/reviews/isl-org/open3d/4105#-MkgU_o99PcYtMCl4YS9:-MkgU_o99PcYtMCl4YSA:bwccabv) ([raw file](https://github.com/isl-org/open3d/blob/4c20a5e115cd43a68f4ba8275a3e88be5d128c78/cpp/open3d/t/geometry/PointCloud.cpp#L121)):*
> ```cpp
> PointCloud PointCloud::Clone() const { return To(GetDevice(), /*copy=*/true); }
> 
> PointCloud PointCloud::Append(const PointCloud &other) const {
> ```

e.g. here we shall already check `other's device. 
___
*[cpp/open3d/t/geometry/PointCloud.cpp, line 122 at r2](https://reviewable.io/reviews/isl-org/open3d/4105#-MkgRkKZ60h2OX5fS1W7:-MkgRkKZ60h2OX5fS1W8:b40kfn) ([raw file](https://github.com/isl-org/open3d/blob/4c20a5e115cd43a68f4ba8275a3e88be5d128c78/cpp/open3d/t/geometry/PointCloud.cpp#L122)):*
> ```cpp
> 
> PointCloud PointCloud::Append(const PointCloud &other) const {
>     // TODO (@rishabh): Add Append to Tensor -> TensorMap -> Use that in
> ```

Remove this TODO as we are already doing it.
___
*[cpp/open3d/t/geometry/PointCloud.cpp, line 614 at r2](https://reviewable.io/reviews/isl-org/open3d/4105#-MkgSE3eB1azpFvURi0o:-MkgSE3eB1azpFvURi0p:b-5ca20g) ([raw file](https://github.com/isl-org/open3d/blob/4c20a5e115cd43a68f4ba8275a3e88be5d128c78/cpp/open3d/t/geometry/PointCloud.cpp#L614)):*
> ```cpp
>             normalization_factor =
>                     1.0 /
>                     static_cast<double>(std::numeric_limits<uint32_t>::max());
> ```

Same, don't change the behavior in a refactoring PR. 
___
*[cpp/open3d/t/geometry/TSDFVoxelGrid.cpp, line 124 at r2](https://reviewable.io/reviews/isl-org/open3d/4105#-MkgSwjR0Zpd8gbXmmzx:-MkgSO992GrPvWnwNJUH:b-u10u07) ([raw file](https://github.com/isl-org/open3d/blob/4c20a5e115cd43a68f4ba8275a3e88be5d128c78/cpp/open3d/t/geometry/TSDFVoxelGrid.cpp#L124)):*
> ```cpp
>                               const core::Tensor &extrinsics,
>                               float depth_scale,
>                               float depth_max) {
> ```

Per "multi-device" support (which is the original name for this PR), the main purpose is to ensure all inputs are in a compatible device. Here, we need to check that:
- `depth`, `color`: are they on the same device? (and, how do we assert a `Geometry` classes' device)? E.g. shall we design some new helper functions (asserting geometry device, or simply assert device is the same)?
- `intrinsics`, `extrinsics`: do they need to be on the same device? do they need to be on the same device as `depth`, `color`? 

In addition to refactoring current checks, we shall go through every public-facing function and ask the same question above and **add** new checks for dtype, device, shape. 
___
*[cpp/open3d/t/pipelines/kernel/Registration.cpp, line 39 at r2](https://reviewable.io/reviews/isl-org/open3d/4105#-MkgTaaO4t5dzuqVfAA-:-MkgTaaO4t5dzuqVfAA0:b8otcti) ([raw file](https://github.com/isl-org/open3d/blob/4c20a5e115cd43a68f4ba8275a3e88be5d128c78/cpp/open3d/t/pipelines/kernel/Registration.cpp#L39)):*
> ```cpp
> static void AssertEmptyTensor(const core::Tensor &tensor,
>                               const std::string &name) {
>     if (!tensor.GetLength()) {
> ```

GetLength() returns an `int` so let's not take the `!` of the int. E.g. we can do `tensor.GetLength() == 0`.

`tensor.GetLength()` throws excpetion for 0-dim tensor. But a 0-dim tensor is not empty. What do we exactly mean by empty tensor? 

Can we replace this by checking the shape to be `{0, N}` or something like that?

yxlao:
We'll need to go through ll public APIs again and see if we need to add checks for device/dtype/shape. For example, in `PointCloud.cpp`, we are missing device checks for `PointCloud::Append()`, `PointCloud::Transform()`, `PointCloud::Rotate()`, ...

yxlao (approved):
> Reviewed 2 of 13 files at r1, 2 of 4 files at r2, 9 of 9 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4105)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @reyanshsolis)

yxlao:
LGTM, after the discussion of the issue on the device type of transformation matrix and etc.

Issue #3799, opened 2021-07-26
read_triangle_mesh function is returning a wrong number of vertices 


dhorka:
**Describe the bug**
 o3d.io.read_triangle_mesh returns a wrong number of vertices. I have double-check it with meshlab, and I observed that the read_triangle_mesh function of the current version of open3d is returning a different number of vertices than meshlab. In version 0.11.2 that was working fine. 

**To Reproduce**

mesh = o3d.io.read_triangle_mesh(filename)
print(mesh)

**Expected behavior**
Obtain the same number of vertices obtained in version 0.11.2 and in meshlab.

**Screenshots**
0.11.2
8334.obj : TriangleMesh with 2323 points and 4647 triangles.
0.12
8334.obj : TriangleMesh with 13941 points and 4647 triangles.
0.13
8334.obj : TriangleMesh with 13941 points and 4647 triangles

![Screenshot_20210726_105429](https://user-images.githubusercontent.com/39189267/126964638-c663edf4-4756-4d63-9207-52335ca2418c.png)

**Environment (please complete the following information):**

- Operating system: Ubuntu 20
- Python version: Python 3.8
- Open3D version: 0.13
- Is this remote workstation?: no
- How did you install Open3D?: pip

**Additional context**
I am attaching the [file](https://drive.google.com/file/d/1l5Y3O7DRpHbCJGJ79TeXfnLAn-SpHvvF/view?usp=sharing) used for the testing.

reyanshsolis:
Hi @dhorka, Thanks for reporting this. For loading `.obj` models, we use `ASSIMP` which might be performing some additional post-processing on the model. I will update you soon.
 
@errissa do you have any idea about the same?

Thanks

TrepangCat:
I have the same problem. And when I write the mesh as .obj file, I find that some vertices repeat several times.

reyanshsolis:
This PR https://github.com/isl-org/Open3D/pull/4119 solves this issue.

Thanks for your patience, 
Rishabh Singh

PR #4119, opened 2021-10-01
Join Identical Vertices [Fixing #3799]

reyanshsolis:
Issue: https://github.com/isl-org/Open3D/issues/3799

`ASSIMP`: Face normals are shared between all points of a single face, so a single point can have multiple normals, which forces the library to duplicate vertices in some cases.

`aiProcess_JoinIdenticalVertices` Identifies and joins identical vertex data sets within all imported meshes.

- `aiProcessPreset_TargetRealtime_Fast` includes the removed flags.

#### Please note: We may still have duplicated vertices sometimes added by the `ASSIMP loader`. 

Refactored unit test: In test `TriangleMeshLegecyCompatibility`, using `cube` instead of `monkey`, because of the following issue:
- Reading `monkey` is fine, Writing `monkey` to a file and then reading in `legacy` and `tensor` is giving a slightly different number of vertices. Suspected reason: While `tensor -> eigen` there might be some precision loss, which leads to some vertices getting very close, which are then removed by the `JoinIdenticalVertices` step.

  

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4119)

### DIFF
--- a/cpp/open3d/io/file_format/FileASSIMP.cpp
+++ b/cpp/open3d/io/file_format/FileASSIMP.cpp
@@ -55,10 +55,13 @@ FileGeometry ReadFileGeometryTypeFBX(const std::string& path) {
     return FileGeometry(CONTAINS_TRIANGLES | CONTAINS_POINTS);
 }
 
-const unsigned int kPostProcessFlags =
-        aiProcess_GenNormals | aiProcess_RemoveRedundantMaterials |
-        aiProcess_Triangulate | aiProcess_GenUVCoords | aiProcess_SortByPType |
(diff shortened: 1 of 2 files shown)


errissa (approved):
> Reviewed 1 of 1 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4119)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @yxlao)

yxlao (approved):
LGTM after test passing.

PR #4125, opened 2021-10-03
[WIP] Tutorial update

theNded:
Tutorials for
- [x] Hash map
- [x] Voxel Block Grid 
- [x] TSDF integration for Voxel Block Grid
- [ ] Ray Casting for Voxel Block Grid
- [x] Customized integration  
- [x] Dense SLAM system
- [ ] SLAM GUI in python


theNded:
You have been emphasizing the shape `()`, but I would think they are not very intuitive. These are nuances that bother users.

Unlike professional tensor developers, conventional users (including me) will regard `()` as an empty shape corresponding to an empty Tensor with zero elements rather than a scalar shape. The fact that `()` corresponds to a scalar does not present in our Tensor tutorial, or [numpy construction](https://numpy.org/doc/stable/reference/generated/numpy.empty.html), or [numpy shape](https://numpy.org/doc/stable/reference/generated/numpy.shape.html) (there is even an example implying it is 0 shape or empty), or [Torch tensor](https://pytorch.org/docs/stable/tensors.html), or [Torch size](https://pytorch.org/docs/stable/generated/torch.Tensor.size.html). 
I would prefer to change the input shape to be consistent using `reshape((N, 1))`, rather than emphasizing the `()` shape.

The trailing comma is similarly less intuitive, but it has been well-documented in [numpy glossary](https://numpy.org/devdocs/glossary.html#term-n) and used in constructors, so I accept the suggestions with it.

reyanshsolis:
Added review comments (mostly grammatical corrections) on `ReviewNB` discussions. 
Still working on reviewing the rest of the PR.

reyanshsolis:
For future reference:
- Changing the line start number to one less than the first line of the code snippet solved the issue.
- Also `:linenos:` was not needed. 
- In order to de-indent the code by 4 spaces `:dedent: 4` must be used. However, using that is clipping the `code path` comment. So, skipping the same. (block looks fine to me).

theNded:
@reyanshsolis thanks for the great catch!

PR #4131, opened 2021-10-04
`Concatenate` op for Tensor

reyanshsolis:
- Adds `o3d.core.Tensor.concatenate((a1, a2, ...), axis = 0)` op for tensor, similar to [numpy.concatenate](https://numpy.org/doc/stable/reference/generated/numpy.concatenate.html)

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4131)


yxlao:
The python tests seem to be failing. We can simplify the regex match to only match a substring.

PR #4133, opened 2021-10-04
temporary fix for eigen download

yxlao:
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4133)


yxlao:
Eigen's back. No need to update.

PR #4139, opened 2021-10-05
Asserts in `NNS`

reyanshsolis:
- Added asserts to `nns`.
- Yet to fix the policy on `point / query splits device`.

### DIFF
--- a/cpp/open3d/core/nns/FaissIndex.cpp
+++ b/cpp/open3d/core/nns/FaissIndex.cpp
@@ -49,18 +49,18 @@ namespace nns {
 FaissIndex::FaissIndex() {}
 
 FaissIndex::FaissIndex(const Tensor &dataset_points) {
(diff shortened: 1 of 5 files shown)


reyanshsolis:
@junha-l for `Faiss`, in unit-test the the `query` was on `CPU` device and `dataset` was on `device [may be CPU / CUDA]`.
Is there an auto device transfer somewhere in `Faiss` to `CUDA`?

junha-l:
> @junha-l for `Faiss`, in unit-test the the `query` was on `CPU` device and `dataset` was on `device [may be CPU / CUDA]`. Is there an auto device transfer somewhere in `Faiss` to `CUDA`?
> 
> Kindly check the added checks in this PR.

Yes, `Faiss` automatically transfers CPU query to CUDA if the index is GPU tensor. 
I think it's better to restrict the query to be on CUDA, as you did in this PR.

PR #4141, opened 2021-10-05
Asserts in `Tensor.h`

reyanshsolis:


reyanshsolis:
@yxlao may you please take a look at `*Set*`, and `operator=` functions, if `asserts` are required?

yxlao:
> may you please take a look at *Set*, and operator= functions, if asserts are required?
In general, let's be strict and add them unless the unit tests are not passing. Record the cases where the unit tests are failing.

In this PR, can you also take care of `Tensor.cpp`?

PR #4221, opened 2021-10-31
:book: recommends release build type for user

### LINKED ISSUE #4207: Document Bug, The steps in 'Build from source' does not optimize correctly.
**Describe the bug**
Following 'Build from source' will perform a debug build using `-O2` . This results in serious performance degradation for the end user. This is a problem because self-building is common in ARM environments.

**To Reproduce**

when cmake

```shell
mkdir build
cd build
cmake ..
find . -name '*.make' | xargs grep O2
```

when make

```shell
make VERBOSE=1 | grep O2
```

or 

```
cmake --build . --config Release --target install VERBOSE=1 | grep O2
```

**Expected behavior**

At `3. Config` step, the correct command is:

```shell
cmake -DCMAKE_BUILD_TYPE=Release
```

check result

```shell
find . -name '*.make' | xargs grep O3
cmake --build . --config Release --target install VERBOSE=1 | grep O3
```

Ah ... do you need this?

- Operating system: Ubuntu 20.04
- Python version: Python 3.8
- Open3D version: 0.13.0
- Is this remote workstation?: yes
- Compiler version (if built from source): gcc 9.3.0

yxlao:
Fixes #4207

@ssheorey , shall we recommend the `Release` build in the documentation. Or, shall make `Release` build the default build type in CMake (currently `RelWithDebInfo` is the default for single-config generators)? 

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4221)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -100,8 +100,8 @@ endif()
 # https://blog.kitware.com/cmake-and-the-default-build-type/
 if(NOT CMAKE_CONFIGURATION_TYPES)
     if(NOT CMAKE_BUILD_TYPE)
-        message(STATUS "Setting build type to RelWithDebInfo as none was specified.")
-        set(CMAKE_BUILD_TYPE RelWithDebInfo CACHE STRING "Choose the type of build." FORCE)
+        message(STATUS "Setting build type to Release as none was specified.")
+        set(CMAKE_BUILD_TYPE Release CACHE STRING "Choose the type of build." FORCE)
         # Set the possible values of build type for cmake-gui.
         set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS
                     "Debug" "Release" "MinSizeRel" "RelWithDebInfo")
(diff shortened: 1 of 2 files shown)


ssheorey (approved):
Yes, switching to `Release` sounds good. It should also reduce disk space requirements for building Open3D for most users who do not want to debug it.

We can put a comment for Windows users that they need to match the Open3D `CMAKE_BUILD_TYPE` to that of their own code.

Can we have the `-DCMAKE_INSTALL_PREFIX=../../open3d-install` as default? This is a good safeguard for new users who may unintentionally use `sudo` for this. Should work in Windows too.

yxlao:
> Yes, switching to `Release` sounds good. It should also reduce disk space requirements for building Open3D for most users who do not want to debug it.
> 
> We can put a comment for Windows users that they need to match the Open3D `CMAKE_BUILD_TYPE` to that of their own code.
> 
> Can we have the `-DCMAKE_INSTALL_PREFIX=../../open3d-install` as default? This is a good safeguard for new users who may unintentionally use `sudo` for this. Should work in Windows too.

I've changed to `Release` by default. However, I don't think we shall set `CMAKE_INSTALL_PREFIX` by default. If we do that, users will have to specify the system install path when do want to install globally, but the system install path is more tricky to set.

PR #4219, opened 2021-10-30
Standarize Ubuntu CUDA CI in Docker

yxlao:
### Changes

- Added standardized Dockerfile and build scripts for 4 Ubuntu CUDA CI test cases.
- Docker image: now we use Nvidia's image, instead of the plain Ubuntu image.
- Faster CUDA CI: with sufficient ccache hit, the total build time can be < 50 mins for each container. Currently, it takes about 2.5 hours for each config.
- By design, there is one and only one way to build and test the containers. No flexibility is allowed.
  - Similar to our previous work, `docker_build.sh` only takes the "docker type" parameter. 
  - Environment variables or other parameters are not allowed, on purpose.
  - This makes it easy to build a local container image that is consistent with the one built on CI.
  - You're not supposed to use the Dockerfiles directly. Use `docker_build.sh`.
- GCE VM image: `ubuntu-os-docker-gpu-2004-lts` in unchanged (thanks @ssheorey)
  - This was the key to getting this PR working. The CUDA drivers are sufficiently new so no modification is required.
  - Previously I used the `deeplearning-platform-release` image, which creates a lot of problems, e.g. ssh connection drops unpredictably while running on GCE.
- Multi-gpu tests still require 2 local GPUs.

### Usage

[Install Nvidia Docker](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/install-guide.html#setting-up-nvidia-container-toolkit) first.

```bash
cd .github/workflows

# To build Docker
./docker_build.sh 2-bionic
./docker_build.sh 3-ml-shared-bionic
./docker_build.sh 4-ml-bionic
./docker_build.sh 5-ml-focal

# To test Docker (runs all unit tests)
# Yes, you'll not need to run your own `docker run`
./docker_test.sh 2-bionic
./docker_test.sh 3-ml-shared-bionic
./docker_test.sh 4-ml-bionic
./docker_test.sh 5-ml-focal
```

Now we can build and run CI-equivalent CUDA tests on a local machine. 

### Incoming PRs
- Merge `Dockerfile.wheel` and `Dockerfile.cuda`.
- We can build with a CPU-only instance or build with Github actions, and run with a GPU instance to save cost.
- Refactor `util/docker`, `util/ci_utils.sh` and etc.
  - This PR does not do the clean-up to keep things simple.
  - We shall separate Ubuntu and macOS tests scripts. This makes the test scripts less conditional and easier to follow. Our Windows test scripts are also separated. 
- Documentation and readme updates.
[...]

PR #4222, opened 2021-10-31
rename ubuntu wheel ci

yxlao:
This will make the changesets smaller in #4219 since #4219 overwrites `.github/workflows/ubuntu-cuda.yml`.

This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4222)


yxlao:
Minor change, merge directly.

Issue #3838, opened 2021-07-31
Multiple calls to compute_vertex_normals() may change some of the vertex_normals


guanming001:
**Describe the bug**
Multiple calls to `compute_vertex_normals()` may change some of the `vertex_normals`

**To Reproduce**
```python
import numpy as np
import open3d as o3d

mesh = o3d.geometry.TriangleMesh.create_cylinder()
# mesh = o3d.geometry.TriangleMesh.create_arrow()
# mesh = o3d.geometry.TriangleMesh.create_box()
# mesh = mesh.subdivide_loop(1)

# 1st call of compute_vertex_normals()
mesh.compute_vertex_normals()
n1 = np.asarray(mesh.vertex_normals).copy()

# 2nd call of compute_vertex_normals()
mesh.compute_vertex_normals()
n2 = np.asarray(mesh.vertex_normals).copy()

print(np.allclose(n1, n2))

# Visualization
pcd1 = o3d.geometry.PointCloud()
pcd2 = o3d.geometry.PointCloud()
pcd1.points = mesh.vertices
pcd2.points = mesh.vertices
pcd1.normals = o3d.utility.Vector3dVector(n1)
pcd2.normals = o3d.utility.Vector3dVector(n2)
o3d.visualization.draw_geometries([mesh, pcd1, pcd2])
```

**Expected behavior**
Shouldn't the vertex normals remain the same as there is no change to the vertices and faces?
i.e. the output of `print(np.allclose(n1, n2))` should be True

**Screenshots**
![image](https://user-images.githubusercontent.com/38412160/127731082-cd048039-f48e-4ad4-9262-85456699c12f.png)

**Environment (please complete the following information):**

- Operating system: Ubuntu 18.04
- Python version: Python 3.7
- Open3D version: 0.12.0
- Is this remote workstation?: No
- How did you install Open3D?: pip

PR #4224, opened 2021-11-02
fix issue#3838, called multiple times may change vertex normal

chunibyo-wly:
#3838 
add an unit test in this condition

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4224)


yxlao (approved):

yxlao:
@chunibyo-wly I changed it to always recompute. The vertex coordinates may change, so it is required to recompute always.

chunibyo-wly:
I did not notice that, thanks for your reminder

PR #4245, opened 2021-11-05
No need for recursive clone and submodules

yxlao:
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4245)


yxlao:
minor refactoring, merge directly

PR #4252, opened 2021-11-06
GitHub Action to lint Python (again)

cclauss:
#3859

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4252)


cclauss:
Some tests are quite fast (~1 minute) and other tests are slow (>45 minutes).  It would help developer productivity (and the planet!) if each slow [`job needs`](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions?query=needs#jobsjob_idneeds) that the faster jobs have completed successfully.

yxlao:
> Some tests are quite fast (~1 minute) and other tests are slow (> 45 minutes). It would help developer productivity (and the planet!) if each slow [`job needs`](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idneeds) that the faster jobs have completed successfully.

Yes, that's a good point. From [here](https://github.community/t/depend-on-another-workflow/16311), there isn't a simple way to depend on another workflow (i.e. another `.yml` file). We may implement implement some workaround in the future to depend on a job, rather than the workflow.

.github/workflows/lint_python.yml
```diff
@@ -0,0 +1,23 @@
...
+jobs:
+  lint_python:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+      - run: pip install --upgrade pip wheel
+      - run: pip install bandit black codespell flake8 isort mypy pyupgrade safety
+      - run: bandit --recursive --skip B101,B108,B110,B310,B311,B403,B404,B603 .
+      - run: black --check . || true
```
yxlao, line 12: Remove `black` as we're using `yapf` and it is already converted in the style check.

.github/workflows/lint_python.yml
```diff
@@ -0,0 +1,23 @@
...
+      - run: codespell || true  # --ignore-words-list="ans,kno,lof,te,thisy" --skip="*.css,*.js,*.lock"
```
yxlao, line 13: - Limit the scope to `.py`
- Only check `python/`
- Future work: we can extend it to check the C++/CUDA files and documentations.

.github/workflows/lint_python.yml
```diff
@@ -0,0 +1,23 @@
...
+on: [pull_request, push]
+jobs:
+  lint_python:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+      - run: pip install --upgrade pip wheel
+      - run: pip install bandit black codespell flake8 isort mypy pyupgrade safety
+      - run: bandit --recursive --skip B101,B108,B110,B310,B311,B403,B404,B603 .
```
yxlao, line 11: - I would add a comment on each of the linter used here, explain what they are testing.
- Open3D's Python package installation requires compiling C++/CUDA code. Could you point out which of these tests needs to be done after installation of the Open3D Python package?

.github/workflows/lint_python.yml
```diff
@@ -0,0 +1,23 @@
...
+      - run: black --check . || true
+      - run: codespell || true  # --ignore-words-list="ans,kno,lof,te,thisy" --skip="*.css,*.js,*.lock"
+      - run: flake8 . --count --select=E501,E9,F63,F7,F82 --max-complexity=18
+                      --max-line-length=215 --show-source --statistics
+      - run: flake8 . --count --exit-zero --max-complexity=10 --max-line-length=88 --show-source --statistics
+      - run: isort --check-only --profile black . || true
```
yxlao, line 17: `--profile google`, also limit the scope to `python/`

.github/workflows/lint_python.yml
```diff
@@ -0,0 +1,23 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+      - run: pip install --upgrade pip wheel
+      - run: pip install bandit black codespell flake8 isort mypy pyupgrade safety
+      - run: bandit --recursive --skip B101,B108,B110,B310,B311,B403,B404,B603 .
+      - run: black --check . || true
+      - run: codespell || true  # --ignore-words-list="ans,kno,lof,te,thisy" --skip="*.css,*.js,*.lock"
+      - run: flake8 . --count --select=E501,E9,F63,F7,F82 --max-complexity=18
```
yxlao, line 14: - Limit the scope to `python/`, `max-line-length=80` (since we're using the google style)
- Also, why do we have two `flake8`s?

.github/workflows/lint_python.yml
```diff
@@ -0,0 +1,23 @@
+name: Lint Python
+on: [pull_request, push]
```
yxlao, line 2: Let's be consistent to the dispatch pattern of `style.yml`
```yml
on:
  workflow_dispatch:
  push:
    branches:
      - master
  pull_request:
    types: [opened, reopened, synchronize] # Rebuild on new pushes to PR
```

.github/workflows/lint_python.yml
```diff
@@ -0,0 +1,23 @@
...
+jobs:
+  lint_python:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
```
yxlao, line 8: what is the default python version?

.github/workflows/lint_python.yml
```diff
@@ -0,0 +1,23 @@
...
+      - run: black --check . || true
+      - run: codespell || true  # --ignore-words-list="ans,kno,lof,te,thisy" --skip="*.css,*.js,*.lock"
+      - run: flake8 . --count --select=E501,E9,F63,F7,F82 --max-complexity=18
+                      --max-line-length=215 --show-source --statistics
+      - run: flake8 . --count --exit-zero --max-complexity=10 --max-line-length=88 --show-source --statistics
+      - run: isort --check-only --profile black . || true
+      - run: pip install -r requirements.txt || true
+      # - run: pip install -e .
+      - run: mkdir -p .mypy_cache
+      - run: mypy --ignore-missing-imports --install-types --non-interactive . || true
```
yxlao, line 21: same, for all the remaining tests, limit the scope to `python/`

.github/workflows/lint_python.yml
```diff
@@ -0,0 +1,23 @@
+name: Lint Python
```
yxlao, line 1: Rename to `lint-python.yml` to be consistent with the other workflows.

yxlao (changes requested):

.github/workflows/lint_python.yml
cclauss:
```
./python/open3d/visualization/tensorboard_plugin/plugin_data_pb2.py:45:216: E501 line too long 2444 characters)
```

The `--select=E501,E9,F63,F7,F82` run focuses on _show-stopper_ issues that effect program correctness and runtime safety.

The `--exit-zero` run is designed to show us ___all___ issues (some of which we may not choose to fix).

.github/workflows/lint_python.yml
cclauss: Black is now gone.

.github/workflows/lint_python.yml
cclauss: Codespell is gone.  It is extremely fast, processing the entire repo in 1 second including error message output but we can add it in a separate PR.

.github/workflows/lint_python.yml
cclauss: `isort` has better algorithms for finding files that contain Python code and can process the entire repo in a few seconds. For instance `./examples` and `./util` also have Python files and given that this is open source we should make those files readable as well as those in `./python`.

.github/workflows/lint_python.yml
cclauss: Python 3.10.0 and allowing it to float means that our tests will find incompatibilities in new Python releases before our users do.

.github/workflows/lint_python.yml
cclauss: All of these tests should be completed in a minute or two max.  All these tools have better algorithms for finding files that contain Python code and can process the entire repo in seconds.  We these tools to find bugs in our examples and utilities before our users do.

.github/workflows/lint_python.yml
cclauss: Done.

.github/workflows/lint_python.yml
yxlao: It's mainly about noise. In practice, `build/` and `3rdparty_downloads/` will be created during C++ compilation. LInters will exam 3rd-party Python code unless we strict the directory. 

You're right that the examples shall be covered. Then, could you restrict the directory to `examples/python` and `python`?

yxlao (changes requested):

ssheorey:
@yxlao corresponding Open3D-ML PR: https://github.com/isl-org/Open3D-ML/pull/333

PR #4254, opened 2021-11-07
move ProgressBar to a standalone file

yxlao:
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4254)


cclauss:
This is a perfect example of https://github.com/isl-org/Open3D/pull/4252#issuecomment-962578232 because the 2 minute `Style check` failed, none of these other jobs should be run.

yxlao:
minor refactoring, merge directly

PR #4262, opened 2021-11-08
Add Holes filling algorithm for CPU TSDF Pipelines.

nachovizzo:
# Description

This is as described in Section 4 of the original paper. Note: This
change has been made intentionally backward compatible, so no
`extract_triangle_mesh` calls change the behavior.

This simple fix is usually missing on most TSDF implementations, and
depending on the application it might be the desired behavior(mostly when
reconstructing "objects" and not scenes).

## Reference

Section 4 of the original [paper](https://graphics.stanford.edu/papers/volrange/volrange.pdf)
image

### Before (and now without specifying the `fill_holes = True` flag):
image

### After (when explicitly saying to fill_holes):

image

Closeup:

ScreenCapture_2021-11-08-14-45-18

## Note

The "fill_holes" might be a bit misleading, this is **NOT** a mesh-repair pipeline. Just the original hole-filling idea described in the Curless and Levoy paper.

## ToDo

- [ ] I quickly checked the new Tensor API but I'm still unfamiliar with it, so not sure if to add this change there are not, or how to test it.
- [ ]  Look for an example application (Bunny-reconstruction for example) where this hole filling shine :star:

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4262)

### DIFF
--- a/cpp/open3d/pipelines/integration/ScalableTSDFVolume.cpp
+++ b/cpp/open3d/pipelines/integration/ScalableTSDFVolume.cpp
@@ -221,8 +221,8 @@ std::shared_ptr<geometry::PointCloud> ScalableTSDFVolume::ExtractPointCloud() {
     return pointcloud;
 }
 
-std::shared_ptr<geometry::TriangleMesh>
-ScalableTSDFVolume::ExtractTriangleMesh() {
+std::shared_ptr<geometry::TriangleMesh> ScalableTSDFVolume::ExtractTriangleMesh(
+        bool fill_holes /* = false*/) {
     // implementation of marching cubes, based on
     // http://paulbourke.net/geometry/polygonise/
     auto mesh = std::make_shared<geometry::TriangleMesh>();
@@ -298,7 +298,7 @@ ScalableTSDFVolume::ExtractTriangleMesh() {
                                                        .color_.cast<double>();
                                 }
                             }
-                            if (w[i] == 0.0f) {
+                            if (w[i] == 0.0f && !fill_holes) {
                                 cube_index = 0;
                                 break;
                             } else {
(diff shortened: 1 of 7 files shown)


theNded (commented):
The application on the demo scene looks good, but introduces black regions caused by undefined normals computed from undefined voxel TSDFs. Need to fix this before merging, either by interpolation/extrapolation or analytical TSDF gradient.

> Reviewed 6 of 7 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4262#-:-MpXblKo8rMwTrl9ovez:b-5dmr4a)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nachovizzo)

theNded:
> The application on the demo scene looks good, but introduces black regions caused by undefined normals computed from undefined voxel TSDFs. Need to fix this before merging, either by interpolation/extrapolation or analytical TSDF gradient.
> 
> > Reviewed 6 of 7 files at r1, 1 of 1 files at r2, all commit messages.
> > _[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4262#-:-MpXblKo8rMwTrl9ovez:b-5dmr4a)_ status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nachovizzo)

It also introduces a thick layer of boundary in addition to the originally thin isosurface. I would suggest to check the statistics and make sure that it only compensates for the **hole** regions instead of growing the **whole** mesh.

nachovizzo:
@theNded Thanks for reviewing and apologize for the delay.

I'm currently out of time to address reviews on this PR. I will close it and we can re-open it in the future in case someone is interested.

Apologize for making you review it :(

Thanks again

PR #4274, opened 2021-11-09
Explicitly set line set/bounding volume color

errissa:
* Internal change to default bounding volume color caused unexpected colors in example.

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4274)


errissa:
Before this PR:
2021-11-10-143814_1024x768_scrot

After this PR:
2021-11-10-144405_1024x768_scrot

ssheorey (approved):

PR #4289, opened 2021-11-12
Demo Scene Updates

errissa:
- Remove image size references from file names so demo scene assets can be updated without having to update demo scene script.

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4289)


yxlao (approved):

ssheorey (changes requested):
Could you replace the monkey with the helmet, as discussed?

ssheorey (approved):

PR #4291, opened 2021-11-12
Hide refracted materials when rendering to depth image

errissa:
This PR solves crash when trying to double click in `O3DVisualizer` window when there are objects with "defaultLitSSR" materials.

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4291)

### DIFF
--- a/cpp/open3d/visualization/rendering/filament/FilamentRenderToBuffer.cpp
+++ b/cpp/open3d/visualization/rendering/filament/FilamentRenderToBuffer.cpp
@@ -119,7 +119,7 @@ void FilamentRenderToBuffer::Configure(const View* view,
 
     // Create a proper copy of the View with scen attached
     CopySettings(view);
-    auto* downcast_scene = dynamic_cast<FilamentScene*>(scene);
+    auto* downcast_scene = static_cast<FilamentScene*>(scene);
     if (downcast_scene) {
         view_->SetScene(*downcast_scene);
         scene_ = downcast_scene;
@@ -154,7 +154,7 @@ void FilamentRenderToBuffer::SetDimensions(const std::uint32_t width,
 
 void FilamentRenderToBuffer::CopySettings(const View* view) {
     view_ = new FilamentView(engine_, EngineInstance::GetResourceManager());
-    auto* downcast = dynamic_cast<const FilamentView*>(view);
+    auto* downcast = static_cast<const FilamentView*>(view);
     if (downcast) {
         view_->CopySettingsFrom(*downcast);
     }
--- a/cpp/open3d/visualization/rendering/filament/FilamentRenderToBuffer.cpp
+++ b/cpp/open3d/visualization/rendering/filament/FilamentRenderToBuffer.cpp
@@ -122,6 +122,7 @@ void FilamentRenderToBuffer::Configure(const View* view,
     auto* downcast_scene = dynamic_cast<FilamentScene*>(scene);
     if (downcast_scene) {
         view_->SetScene(*downcast_scene);
+        scene_ = downcast_scene;
     }
     SetDimensions(width, height);
 }
@@ -189,6 +190,7 @@ void FilamentRenderToBuffer::ReadPixelsCallback(void*, size_t, void* user) {
 
 void FilamentRenderToBuffer::Render() {
     frame_done_ = false;
+    scene_->HideRefractedMaterials();
     if (renderer_->beginFrame(swapchain_)) {
         renderer_->render(view_->GetNativeView());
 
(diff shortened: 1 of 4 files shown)


benjaminum (changes requested):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4291)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @benjaminum and @errissa)

___
*[cpp/open3d/visualization/rendering/filament/FilamentRenderToBuffer.cpp, line 122 at r1](https://reviewable.io/reviews/isl-org/open3d/4291#-MoXt_ib2UzT1qpL96OW:-MoXt_ic8qjdOM5rYdka:buc7kmu) ([raw file](https://github.com/isl-org/open3d/blob/b8c8b6d4aceec09cd05a6cf33b33a7d85b1fe8ce/cpp/open3d/visualization/rendering/filament/FilamentRenderToBuffer.cpp#L122)):*
> ```cpp
>     // Create a proper copy of the View with scen attached
>     CopySettings(view);
>     auto* downcast_scene = dynamic_cast<FilamentScene*>(scene);
> ```

Will this ever fail? Do we need to check the pointer throughout the code?
___
*[cpp/open3d/visualization/rendering/filament/FilamentRenderToBuffer.cpp, line 217 at r1](https://reviewable.io/reviews/isl-org/open3d/4291#-MoXuGzSB48dPUBVuS2g:-MoXuGzSB48dPUBVuS2h:b-s3kre8) ([raw file](https://github.com/isl-org/open3d/blob/b8c8b6d4aceec09cd05a6cf33b33a7d85b1fe8ce/cpp/open3d/visualization/rendering/filament/FilamentRenderToBuffer.cpp#L217)):*
> ```cpp
>         renderer_->endFrame();
>     }
>     scene_->HideRefractedMaterials(false);
> ```

If there are some geometries that were already hidden this will unhide them, right?
I think this is OK for now but will need to be changed when we give more control about visibility to the user.

errissa (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4291)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @benjaminum)

___
*[cpp/open3d/visualization/rendering/filament/FilamentRenderToBuffer.cpp, line 122 at r1](https://reviewable.io/reviews/isl-org/open3d/4291#-MoXt_ib2UzT1qpL96OW:-MoZ6Dbc9OOi_ioNSWP5:b6orege) ([raw file](https://github.com/isl-org/open3d/blob/b8c8b6d4aceec09cd05a6cf33b33a7d85b1fe8ce/cpp/open3d/visualization/rendering/filament/FilamentRenderToBuffer.cpp#L122)):*

Will this ever fail? Do we need to check the pointer throughout the code?
</blockquote></details>

This can never fail because Filament is the only back end we have. In a future where we have another rendering back end this still shouldn't fail because this would only be active when the Filament back end is active.
___
*[cpp/open3d/visualization/rendering/filament/FilamentRenderToBuffer.cpp, line 217 at r1](https://reviewable.io/reviews/isl-org/open3d/4291#-MoXuGzSB48dPUBVuS2g:-MoZ6mXPC2-K16s3U-re:bta1dvg) ([raw file](https://github.com/isl-org/open3d/blob/b8c8b6d4aceec09cd05a6cf33b33a7d85b1fe8ce/cpp/open3d/visualization/rendering/filament/FilamentRenderToBuffer.cpp#L217)):*

If there are some geometries that were already hidden this will unhide them, right?
I think this is OK for now but will need to be changed when we give more control about visibility to the user.
</blockquote></details>

That is true. Fixed.

benjaminum (approved):
Please see the comment about `dynamic_cast` before mergeing. Apart from that it :lgtm: 

> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4291#-:-MocIBwD5x3PRmnmbnYo:b-jq7z1v)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @benjaminum and @errissa)

___
*[cpp/open3d/visualization/rendering/filament/FilamentRenderToBuffer.cpp, line 122 at r1](https://reviewable.io/reviews/isl-org/open3d/4291#-MoXt_ib2UzT1qpL96OW:-MocHtStBrKoJISH5faE:bzgneyd) ([raw file](https://github.com/isl-org/open3d/blob/b8c8b6d4aceec09cd05a6cf33b33a7d85b1fe8ce/cpp/open3d/visualization/rendering/filament/FilamentRenderToBuffer.cpp#L122)):*

This can never fail because Filament is the only back end we have. In a future where we have another rendering back end this still shouldn't fail because this would only be active when the Filament back end is active.
</blockquote></details>

Consider using `static_cast` or throw something if `downcast_scene` is `null`

PR #4304, opened 2021-11-17
Fix color map optimization numerical issues

### LINKED ISSUE #4266: Degraded color map optimization quality in most recent version
**Describe the bug**
There may be an issue with color map optimization in Open3D versions 0.13.0 and up. There's a noticeable difference in the quality of the results when using 0.12.0 versus 0.13.0 (or the latest development version). In the recent versions, the residual falls for a while (first 100 iterations or so, depending on dataset of course) before climbing indefinitely. This is not seen on version 0.12.0.

**To Reproduce**
Install Open3D version 0.13.0 (I used pip/venv).

Download the fountain sample dataset:
```
wget https://github.com/intel-isl/open3d_downloads/releases/download/open3d_tutorial/fountain.zip
unzip fountain.zip
```

Create a config file for the fountain data:
```
{
    "name": "Fountain",
    "path_dataset": "<path/to/fountain/dataset>",
    "template_global_traj": "scene/key.log"
}
```

Run the script `color_map_optimization_for_reconstruction_system.py`, which is included with the Open3D repository:
```
python3 color_map_optimization_for_reconstruction_system.py --config <path/to/fountain/config> --sample_rate 1
```

Observe the results. Repeat the color map optimization with the same data and Open3D version 0.12.0 and notice the difference in outcome.

Let me know if I can provide additional info, logs etc. Hopefully I'm not missing something silly.

**Screenshots**
Here are the (expected) results from Open3D 0.12.0 (first is rigid, second non-rigid):

0 12 0-rigid
0 12 0-nonrigid

And here are the results from Open3D 0.13.0 and latest development (rigid, non-rigid):

0 13 0-rigid
0 13 0-nonrigid

`max_iteration` was set to 300 for all of these. The issue is exacerbated when using higher iteration count.

**Environment**
- Operating system: Ubuntu 20.04.3
- Python version: 3.8.10
- Open3D version: 0.12.0, 0.13.0 and latest development
- How did you install Open3D?: pip

yxlao:
## Fixed
Fixes #4266.

1. Change flooring to rounding.
2. Change "double -> float -> int" to "double -> int" to match 0.12.
3. The optimized camera trajectory is returned and can be used in the next round of optimization.
4. Added more strict unit tests, from `rtol=1/256` to `rtol=10e-6`. Also, camera parameters are tested.

## Testing
1. Save [test_012.py.txt](https://github.com/isl-org/Open3D/files/7552723/test_012.py.txt) as `Open3D/test_012.py`.
2. Save [test_013.py.txt](https://github.com/isl-org/Open3D/files/7552724/test_013.py.txt) as `Open3D/test_013.py`.
3. Run tests
    ```bash
    git checkout yixing/issue-4266-fix # Or some branch close to this commit point

    # Test 0.12
    pip uninstall open3d --yes
    pip install open3d==0.12.0
    OMP_NUM_THREADS=1 pytest test_012.py -s > test_012_out.txt

    # Test current branch
    mkdir build && cd build
    cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DBUILD_SHARED_LIBS=OFF ..
    cd ..
    OMP_NUM_THREADS=1 pytest test_013.py -s > test_fixed_branch.txt
    ```
The results are reported in [test_012_out.txt](https://github.com/isl-org/Open3D/files/7552740/test_012_out.txt) and [test_fixed_branch.txt](https://github.com/isl-org/Open3D/files/7552741/test_fixed_branch.txt). The results match exactly numerically.

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4304)


theNded (approved):
> Reviewed 10 of 10 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4304)* status: all files reviewed, 1 unresolved discussion (waiting on @yxlao)

___
*[cpp/open3d/pipelines/color_map/ColorMapUtils.cpp, line 77 at r1](https://reviewable.io/reviews/isl-org/open3d/4304#-MoiJaGZBkD7jBEezHpw:-MoiJaG_E4i9dw7RShoe:bosvitl) ([raw file](https://github.com/isl-org/open3d/blob/11a5e7d2781ff924259ded127aebf1a8db8e69f6/cpp/open3d/pipelines/color_map/ColorMapUtils.cpp#L77)):*
> ```cpp
>         }
>         if (img.TestImageBoundary(u, v, image_boundary_margin)) {
>             int u_round = int(round(u));
> ```

I think the major cause was simply the missing `round` operation at this line, since `int` cast is truncation instead of round. Double->Float shouldn't make much difference.

yxlao (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4304)* status: all files reviewed, 1 unresolved discussion (waiting on @theNded)

___
*[cpp/open3d/pipelines/color_map/ColorMapUtils.cpp, line 77 at r1](https://reviewable.io/reviews/isl-org/open3d/4304#-MoiJaGZBkD7jBEezHpw:-MoizPC4E87Pciw-c1Ga:b43k7on) ([raw file](https://github.com/isl-org/open3d/blob/11a5e7d2781ff924259ded127aebf1a8db8e69f6/cpp/open3d/pipelines/color_map/ColorMapUtils.cpp#L77)):*

I think the major cause was simply the missing `round` operation at this line, since `int` cast is truncation instead of round. Double->Float shouldn't make much difference.
</blockquote></details>

Yes. Double->Float only has a very minor effect.

PR #4309, opened 2021-11-18
fix tensormap return reference

yxlao:
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4309)

### DIFF
--- a/cpp/pybind/t/geometry/tensormap.cpp
+++ b/cpp/pybind/t/geometry/tensormap.cpp
@@ -90,7 +90,7 @@ static py::class_<Map, holder_type> bind_tensor_map(py::handle scope,
                 if (it == m.end()) throw py::key_error();
                 return it->second;
             },
-            // py::return_value_policy::copy is used as the safetest option.
+            // py::return_value_policy::copy is used as the safest option.
             // The goal is to make TensorMap works similarly as putting Tensors
             // into a python dict, i.e., {"a": Tensor(xx), "b": Tensor(XX)}.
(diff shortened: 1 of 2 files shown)


python/test/t/geometry/test_tensormap.py
```diff
@@ -69,3 +69,158 @@ def test_tensormap(device):
...
+    a_alias = tm["a"]
+    b_alias = tm["b"]
+    tm["a"], tm["b"] = tm["b"], tm["a"]
+    np.testing.assert_equal(a_alias.numpy(), [100])
+    np.testing.assert_equal(b_alias.numpy(), [200])
+    np.testing.assert_equal(tm["a"].numpy(), [200])
+    np.testing.assert_equal(tm["b"].numpy(), [100])
+
+
+def test_numpy_dict_modify():
```
ssheorey, line 178: We can skip this test (`@skip`) since this doesn't test Open3D code.

cpp/pybind/t/geometry/tensormap.cpp
```diff
@@ -90,8 +90,22 @@ static py::class_<Map, holder_type> bind_tensor_map(py::handle scope,
                 if (it == m.end()) throw py::key_error();
                 return it->second;
             },
-            py::return_value_policy::reference_internal  // ref + keepalive
-    );
+            // py::return_value_policy::copy is used as the safetest option.
```
ssheorey, line 93: safest

ssheorey (approved):
Looks good. Can merge when CI passes.

PR #4317, opened 2021-11-19
Add caching to prevent redundant creation of textures on GPU

errissa:
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4317)

### DIFF
--- a/cpp/open3d/visualization/rendering/filament/FilamentResourceManager.cpp
+++ b/cpp/open3d/visualization/rendering/filament/FilamentResourceManager.cpp
@@ -71,6 +71,21 @@ namespace open3d {
 namespace visualization {
 namespace rendering {
 
+// Implementation of FNV-1a hash function. See
+// https://en.wikipedia.org/wiki/Fowler-Noll-Vo_hash_function for details
+std::uint64_t fnv1a_block_hash(const void* const block, std::size_t len) {
+    const uint64_t kFNVPrime = 0x00000100000001B3;
+    std::uint64_t hash = 0xcbf29ce484222325;
+
+    const auto* ptr = static_cast<const unsigned char*>(block);
+    for (unsigned int i = 0; i < len; ++i) {
+        hash ^= static_cast<uint64_t>(*ptr++);
+        hash *= kFNVPrime;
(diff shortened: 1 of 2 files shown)


cpp/open3d/visualization/rendering/filament/FilamentResourceManager.cpp
```diff
@@ -441,10 +456,16 @@ TextureHandle FilamentResourceManager::CreateTexture(const char* path,
 TextureHandle FilamentResourceManager::CreateTexture(
         const std::shared_ptr<geometry::Image>& img, bool srgb) {
     TextureHandle handle;
-    if (img->HasData()) {
+    auto hash = fnv1a_block_hash(img->data_.data(), img->data_.size());
```
ssheorey, line 459: Are uint8 / float / double images supported? The hash function assumes int data here.

ssheorey (commented):
Curious if just `img->data_.data()` or `hash(data(), width, height)` works as well. They should be much faster to calculate. For (data(), width, height), `std::unordered_map<tuple, image>` should work directly.

cpp/open3d/visualization/rendering/filament/FilamentResourceManager.cpp
errissa: The hash function takes a void* and treats the memory block as a block of octets so it should work with any image format.

errissa:
> Curious if just `img->data_.data()` or `hash(data(), width, height)` works as well. They should be much faster to calculate. For (data(), width, height), `std::unordered_map<tuple, image>` should work directly.

Perhaps in the future but with the current necessary translations between legacy and new materials new memory sometimes gets allocated so `img->data_.data()` is not always the same for the same image.

ssheorey (approved):
This greatly reduces memory consumption in the TB plugin with the demo scene. GPU memory is around 720MB on the first load, increases by about 30MB on every reload. CPU memory usage is about 1.7GB (all of tensorboard) on first load, and increases by about 50MB per reload.

Solves the Out of Memory error after 2 reloads on my laptop.

PR #4321, opened 2021-11-19
Tensor Registration Tutorial Fix

reyanshsolis:
image

The tutorial is not supposed to be executed while building doc.

Tutorial Source Link: 
[t_icp_registration.zip](https://github.com/isl-org/Open3D/files/7592880/t_icp_registration.zip)

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4321)


yxlao:
LGTM. 
One concern is that the size of the Jupyter notebook is now 13.3 MB. If we modify the file in the future, there will be another 13.3 MB binary added to the git history.

docs/make_docs.py
```diff
@@ -458,6 +468,13 @@ def run(self):
                 with open(nb_path, "w", encoding="utf-8") as f:
                     nbformat.write(nb, f)
 
+        url = "https://github.com/isl-org/Open3D/files/7592880/t_icp_registration.zip"
+        output_file = "t_icp_registration.ipynb"
+        output_file_path = os.path.join(
```
yxlao, line 473: use pathlib

yxlao: there are also other `path.join()` use cases, we can do the refactoring later.

yxlao (approved):

PR #4330, opened 2021-11-22
add threads synchronization to RemoveStatisticalOutlier

### LINKED ISSUE #1381: RemoveStatisticalOutliers bug : valid_distances erratic behavior due to OpenMP looping
Variable `valid_distances` not consistent over multiple runs due to `OpenMP` parallelism.  It's probably best to calculate valid_distances after the kdTree loop, using `avg_distances[I] >= 0 logic`;

marcov868:
Closes #1381

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4330)

### DIFF
--- a/cpp/open3d/geometry/PointCloud.cpp
+++ b/cpp/open3d/geometry/PointCloud.cpp
@@ -551,22 +551,19 @@ PointCloud::RemoveStatisticalOutliers(size_t nb_neighbors,
     utility::OMPProgressBar progress_bar(points_.size(),
                                          "Remove statistical outliers: ", true);
 
-#pragma omp parallel for schedule(static) \
-        num_threads(utility::EstimateMaxThreads())
+#pragma omp parallel for reduction(+ : valid_distances) schedule(static) num_threads(utility::EstimateMaxThreads())
     for (int i = 0; i < int(points_.size()); i++) {
         std::vector<int> tmp_indices;
         std::vector<double> dist;
         kdtree.SearchKNN(points_[i], int(nb_neighbors), tmp_indices, dist);
         double mean = -1.0;
         if (dist.size() > 0u) {
-#pragma omp atomic
             valid_distances++;
             std::for_each(dist.begin(), dist.end(),
                           [](double &d) { d = std::sqrt(d); });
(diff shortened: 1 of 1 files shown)


yxlao (changes requested):
> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4330)* status: all files reviewed, 2 unresolved discussions (waiting on @marcov868)

___
*[cpp/open3d/geometry/PointCloud.cpp, line 562 at r1](https://reviewable.io/reviews/isl-org/open3d/4330#-Mp66Vw66GYBKxAldYwH:-Mp66Vw77HMxYg6ZYEdz:bodssfz) ([raw file](https://github.com/isl-org/open3d/blob/2ed63c4c09c30d4d59b70ee252dc9738fce36cbe/cpp/open3d/geometry/PointCloud.cpp#L562)):*
> ```cpp
>         double mean = -1.0;
>         if (dist.size() > 0u) {
> #pragma omp atomic
> ```

Could we try the OpenMP reduction instead? 

Something like:
```cpp
#pragma omp parallel for reduction(+ : valid_distances)
```

See: https://docs.microsoft.com/en-us/cpp/parallel/concrt/convert-an-openmp-loop-that-uses-a-reduction-variable?view=msvc-170
___
*[cpp/open3d/geometry/PointCloud.cpp, line 569 at r1](https://reviewable.io/reviews/isl-org/open3d/4330#-Mp66ffI4IQLyT7er9jw:-Mp66ffJ35cyuxgZSKlh:b-wnee30) ([raw file](https://github.com/isl-org/open3d/blob/2ed63c4c09c30d4d59b70ee252dc9738fce36cbe/cpp/open3d/geometry/PointCloud.cpp#L569)):*
> ```cpp
>         }
>         avg_distances[i] = mean;
> #pragma omp critical
> ```

This is not necessary. 

The progress bar is already `utility::OMPProgressBar` (merged a few days agao). It takes care of OMP internally.

yxlao (approved):
Thanks! LGTM.

PR #4343, opened 2021-11-24
Copy tensorboard.md from Open3D_ML_ROOT for docs

ssheorey:
Path to Open3D_ML_ROOT is different in CI (inside Open3D) compared to common developer machine (sibling to Open3D).

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4343)

### DIFF
--- a/docs/make_docs.py
+++ b/docs/make_docs.py
@@ -465,8 +465,8 @@ class JupyterDocsBuilder:
                 except nbconvert.preprocessors.execute.CellExecutionError:
                     print("Execution of {} failed, this will cause CI to fail.".
                           format(nb_path.name))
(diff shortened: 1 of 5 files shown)


docs/make_docs.py
```diff
@@ -301,6 +303,12 @@ def run(self):
         """
         Call Sphinx command with hard-coded "html" target
         """
+        # Copy docs files from Open3D-ML repo
+        OPEN3D_ML_ROOT = os.environ.get("OPEN3D_ML_ROOT", "../../Open3D-ML")
```
yxlao, line 307: Since we have `self.current_file_dir`, shall we make `../../Open3D-ML` relative to current file_dir?

yxlao (approved):

PR #4396, opened 2021-12-09
Data downloader, libcurl, and boringssl

reyanshsolis:
This PR:
- [x] Adds `boringssl` and `curl` to Open3D build system.
- [x] Adds `DownloadFromURL`, `GetSHA256`

TODO:
- [ ] pybind and unit-tests
- [ ] progress bar for downloader 

This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4396)


yxlao:
- [x] When WebRTC=ON, do not compile BoringSSL. By this, we avoid the need to install `golang`.
- [x] Pin BoringSSL version to be the same as the one used by WebRTC.
- [ ] Unify BoringSSL and OpenSSL to the same name.
- [ ] When WebRTC=OFF, use OpenSSL. On Windows, OpenSSL cannot be built easily with CMake. We'll need to enforce `WebRTC=ON` for Windows.
- [ ] Fix all remaining CI issues.
- [ ] On Ubuntu/macOS, don't use CMake to build `lbcurl` since the CMake configuration takes too long.

yxlao (approved):

PR #4413, opened 2021-12-12
Improve consistency in use of pixel size parameter

errissa:
The following behaviors change with this PR:

- If `point_size` is set in `draw` call then all point clouds will have their point sizes set to `point_size` regardless of the `point_size` specified in the material properties.
- If `point_size` is _not_ set in `draw` call then point clouds will be rendered with the `point_size` set in their respective material properties. Note: prior to this PR _all_ point clouds would have their point size set to the point size of the last point cloud geometry listed in the `draw` call.
- The point size slider in the User Interface will not be updated to match the point size of a point cloud's material settings. Instead, it works as a global override.

The following script demonstrates the new behavior:
```
import open3d as o3d
from open3d.core import Tensor, concatenate
from open3d.visualization import rendering, draw
from open3d.ml.vis import Colormap

# Create a helix point cloud. We will use the Z values to assign colors.
values = Tensor.arange(start=0.0, stop=1.0, step=0.001,
                       dtype=o3d.core.float32).reshape((-1, 1))
values2 = Tensor.arange(start=1.0, stop=2.0, step=0.001,
                       dtype=o3d.core.float32).reshape((-1, 1))
period = 0.2
xyz = concatenate(
    ((6.28 / period * values).sin(), (6.28 / period * values).cos(), values), 1)
xyz2 = concatenate(
    ((6.28 / period * values).sin(), (6.28 / period * values).cos(), values2), 1)

pcd = o3d.t.geometry.PointCloud(xyz)
pcd2 = o3d.t.geometry.PointCloud(xyz2)

# Use a special point property to specify colormap lookup values for the point
# cloud.
pcd.point['__visualization_scalar'] = values
pcd2.point['__visualization_scalar'] = values2

# Use a default rainbow colormap
colormap = Colormap.make_rainbow()

# Add alpha channel and convert to Gradient Points
colormap = list(
    rendering.Gradient.Point(pt.value, pt.color + [1.0])
    for pt in colormap.points)

# Now create the materials - each point cloud will have a different point size
material = rendering.MaterialRecord()
material.shader = "unlitGradient"
material.gradient = rendering.Gradient(colormap)
material.gradient.mode = rendering.Gradient.GRADIENT
material.scalar_min = 0.0
material.scalar_max = 1.0
material.point_size = 9.0
material2 = rendering.MaterialRecord()
material2.shader = "unlitGradient"
material2.gradient = rendering.Gradient(colormap)
material2.gradient.mode = rendering.Gradient.GRADIENT
material2.scalar_min = 1.0
material2.scalar_max = 2.0
material2.point_size = 3.0

# Draw two stacked helixes with different point sizes
draw([{
    'name': 'helix',
    'geometry': pcd,
    'material': material},
    { 'name': 'helix2',
      'geometry' : pcd2,
      'material': material2}],
show_skybox=False)

# Draw two stacked helixes overriding material point sizes.
draw([{
    'name': 'helix',
    'geometry': pcd,
    'material': material},
    { 'name': 'helix2',
      'geometry' : pcd2,
      'material': material2}],
point_size=9,
show_skybox=False)
```
2021-12-12-120035_1024x768_scrot

2021-12-12-120046_1024x768_scrot

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4413)

### DIFF
--- a/cpp/open3d/visualization/visualizer/O3DVisualizer.cpp
+++ b/cpp/open3d/visualization/visualizer/O3DVisualizer.cpp
@@ -1259,7 +1259,6 @@ struct O3DVisualizer::Impl {
         UpdateSelectionPointSize(px);
 
         px = int(ConvertToScaledPixels(px));
-        utility::LogWarning("Point size in set: {}", px);
         for (auto &o : objects_) {
             o.material.point_size = float(px);
             OverrideMaterial(o.name, o.material, ui_state_.scene_shader);
--- a/cpp/open3d/visualization/visualizer/O3DVisualizer.cpp
+++ b/cpp/open3d/visualization/visualizer/O3DVisualizer.cpp
@@ -311,6 +311,7 @@ struct O3DVisualizer::Impl {
 
     UIState ui_state_;
     bool can_auto_show_settings_ = true;
+    bool point_size_override = false;
 
     double min_time_ = 0.0;
     double max_time_ = 0.0;
@@ -837,9 +838,8 @@ struct O3DVisualizer::Impl {
         if (material) {
             mat = *material;
             is_default_color = false;
-            auto t_cloud =
-                    std::dynamic_pointer_cast<t::geometry::PointCloud>(tgeom);
-            ui_state_.point_size = static_cast<int>(mat.point_size);
+            mat.point_size = ConvertToScaledPixels(int(mat.point_size));
+            mat.line_width = mat.line_width * window_->GetScaling();
         } else if (model) {
             // Adding a triangle mesh model. Shader needs to be set to
             // defaultLit for O3D shader handling logic to work.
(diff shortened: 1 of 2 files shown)


ssheorey (changes requested):
Add docstring for Python `draw` detailing each parameter. Include example calls.
[Future] We should have a similar function call in C++ as well.

errissa (commented):
I'm working separately on documenting `draw` along with tutorials. I think that should be a separate PR.

> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4413#-:-MqoiR572ceEwn1BVPBc:bo32nxp)* status: 0 of 2 files reviewed, all discussions resolved

ssheorey (commented):
Thanks @errissa this clarifies and improves `point_size` behavior. 

Can we treat `line_width` in the same way as `point_size`? 

[Future PR] Are we able to set the point size for meshes and line sets? For example, can we get the graphic that Benjamin produced for the 0.14 release video, with very large end points in a line segment.

errissa:
A feature planned for near future will more properly and thoroughly address the issues here. Also, the feature will change the behavior of the global point size setting compared to what the behavior would be with this PR and we don't want to introduce confusion when the new feature is integrated.

PR #4418, opened 2021-12-13
fix issue 4409

### LINKED ISSUE #4409: SegmentPlane function ransac_n looks like useless
**Describe the bug**
In `SegmentPlane` function, when ransac_n bigger than 3, and inliers array size will bigger than 3.
But as you can see, the element which indices bigger than 3 had not been used.
And also, plane only use the first 3 element to construct plane, I speculate`GetPlaneFromPoints` function is a better choice than `ComputeTrianglePlane` function.
When user change the parameter ransac_n they may be confused.

https://github.com/isl-org/Open3D/blob/f4459dfcf1536999641fe0234a67dee865f4de3d/cpp/open3d/geometry/PointCloudSegmentation.cpp#L162-L190

chunibyo-wly:
Fixes #4409.
As the doc say, ransac_n is the inliers to initial plane.
But in current implement, only 3 point used.
So I make this change, and also passed unit test.

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4418)

### DIFF
--- a/cpp/open3d/geometry/PointCloudSegmentation.cpp
+++ b/cpp/open3d/geometry/PointCloudSegmentation.cpp
@@ -182,7 +182,8 @@ std::tuple<Eigen::Vector4d, std::vector<size_t>> PointCloud::SegmentPlane(
         // inliers.
         if (ransac_n == 3) {
             plane_model = TriangleMesh::ComputeTrianglePlane(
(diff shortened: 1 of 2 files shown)


theNded (approved):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4418)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @yxlao)

Issue #4382, opened 2021-12-07
SceneWidget obtains wrong coordinates of point when _on_mouse event


AkiraGiShinichi:
**Describe the bug**
Hi!
I try this sample: [mouse-and-point-coord](https://github.com/isl-org/Open3D/blob/master/examples/python/gui/mouse-and-point-coord.py)
It can show the coordinates of point that I click
But It can't be sure that the coordinates are correct. Hence I added the spheres to visualize the clicked points.
It seems that the spheres are displayed at wrong positions, not at where my mouse clicks(I create the spheres at the coordinates that are obtained and displayed)

**To Reproduce**

1. use sample  [mouse-and-point-coord](https://github.com/isl-org/Open3D/blob/master/examples/python/gui/mouse-and-point-coord.py)

2. modify a bit:

```python
import numpy as np
import open3d as o3d
import open3d.visualization.gui as gui
import open3d.visualization.rendering as rendering
import os

SCRIPT_DIR = os.path.dirname(os.path.realpath(__file__))

class ExampleApp:

    def __init__(self, cloud):
        # We will create a SceneWidget that fills the entire window, and then
        # a label in the lower left on top of the SceneWidget to display the
        # coordinate.
        app = gui.Application.instance
        self.window = app.create_window("Open3D - GetCoord Example", 1024, 768)
        # Since we want the label on top of the scene, we cannot use a layout,
        # so we need to manually layout the window's children.
        self.window.set_on_layout(self._on_layout)
        self.widget3d = gui.SceneWidget()
        self.window.add_child(self.widget3d)
        self.info = gui.Label("")
        self.info.visible = False
        self.window.add_child(self.info)

        self.widget3d.scene = rendering.Open3DScene(self.window.renderer)

        self.mat = rendering.Material()
        self.mat.shader = "defaultUnlit"
        # Point size is in native pixels, but "pixel" means different things to
        # different platforms (macOS, in particular), so multiply by Window scale
        # factor.
        self.mat.point_size = 3 * self.window.scaling
        self.widget3d.scene.add_geometry("Point Cloud", cloud, self.mat)

        bounds = self.widget3d.scene.bounding_box
        center = bounds.get_center()
        self.widget3d.setup_camera(60, bounds, center)
        self.widget3d.look_at(center, center - [0, 0, 3], [0, -1, 0])

        self.widget3d.set_on_mouse(self._on_mouse_widget3d)

        self.picked_points = []

    def _on_layout(self, layout_context):
[...]

Issue #4244, opened 2021-11-05
mouse-and-point-coord.py example giving different coordinates from different camera angle


JiayouQin:
**IMPORTANT: Please use the following template to report the bug.**

Before submitting:

1. Search for [similar issues](https://github.com/isl-org/Open3D/issues).
2. For Python issues, please test with the [latest development wheel](http://www.open3d.org/docs/latest/getting_started.html#development-version-pip). It may be already fixed!
_____

**Describe the bug**

Hi, I was trying one of the examples  and  have encontered some issue:
the world coordnates the example gives me seems to be very random, each time I change camera view angle the same point gives me different coordinates. 
I rendered a sphere to the coordnates it gave me and they were all over the place.
they do deviate alot as I move around, they are not approximations.

**To Reproduce**
file path: examples/python/gui/mouse-and-point-coord.py

**cd examples/python/gui
python mouse-and-point-coord.py**

to make things easier visually I made the following changes:

mat renamed to self.mat for referencing
named self.click=0
added the following code to render a sphere after

> **text = "({:.3f}, {:.3f}, {:.3f})".format(world[0], world[1], world[2])**

```
mesh = o3d.geometry.TriangleMesh.create_sphere(radius=15/1000)
mesh.translate((world[0],world[1],world[2]), relative=False)
self.widget3d.scene.add_geometry(f"Point Cloud{self.click}", mesh, self.mat)
self.click += 1
```

**Expected behavior**

Expecting the point to be exact same value and spheres are in the same spot

**Screenshots**
![scr](https://user-images.githubusercontent.com/46877249/140458046-c4e18778-1373-4d9b-9ba2-27bbb771521a.png)
The red mark shows where I click, and the grey spheres poistioned from the coordinates I get.

**Environment (please complete the following information):**

- Operating system: (Windows 10 64-bit)
- Python version: (Python 3.8)
- Open3D version:  (open3d-0.13.0-cp38)
- Is this remote workstation?: (No)
- How did you install Open3D?: (pip)

**Additional context**
I monkeyed around the code for a bit and found that if I don't set any camera angle and do not rotate the camera I the coordinates are correct except that the y axis is reversed. 
(I cannot see anything using the original pcd this way so I loaded another ply file)
If I turn the camera the coordinates are incoorect again.

JiayouQin:
Hello @prewettg, sorry to bother you here but I've been trying to working this out for  a  few days...any help would be  appreciated!

[...]

PR #4431, opened 2021-12-15
Correct unproject behavior to match documentation

errissa:
Addresses issues #4244 and #4382. The documentation indicate that inputs to `Camera::Unproject` and `visualization.rendering.Camera.unproject` expect `x,y` in screen coordinates relative to upper-left of screen but the implementation expected lower-right origin.

The following modified version of the `mouse-and-point-coord.py` demo provided by the authors of both referenced issues can be used to test:

```
import numpy as np
import open3d as o3d
import open3d.visualization.gui as gui
import open3d.visualization.rendering as rendering
import os

SCRIPT_DIR = os.path.dirname(os.path.realpath(__file__))


# This example displays a point cloud and if you Ctrl-click on a point
# (Cmd-click on macOS) it will show the coordinates of the point.
# This example illustrates:
# - custom mouse handling on SceneWidget
# - getting a the depth value of a point (OpenGL depth)
# - converting from a window point + OpenGL depth to world coordinate
class ExampleApp:

    def __init__(self, cloud):
        # We will create a SceneWidget that fills the entire window, and then
        # a label in the lower left on top of the SceneWidget to display the
        # coordinate.
        app = gui.Application.instance
        self.window = app.create_window("Open3D - GetCoord Example", 1024, 768)
        # Since we want the label on top of the scene, we cannot use a layout,
        # so we need to manually layout the window's children.
        self.window.set_on_layout(self._on_layout)
        self.widget3d = gui.SceneWidget()
        self.window.add_child(self.widget3d)
        self.info = gui.Label("")
        self.info.visible = False
        self.window.add_child(self.info)

        self.widget3d.scene = rendering.Open3DScene(self.window.renderer)

        mat = rendering.MaterialRecord()
        mat.shader = "defaultLit"
        # Point size is in native pixels, but "pixel" means different things to
        # different platforms (macOS, in particular), so multiply by Window scale
        # factor.
        mat.point_size = 3 * self.window.scaling
        self.widget3d.scene.add_geometry("Point Cloud", cloud, mat)

        bounds = self.widget3d.scene.bounding_box
        center = bounds.get_center()
        self.widget3d.setup_camera(60, bounds, center)
        self.widget3d.look_at(center, center - [0, 0, 12], [0, -1, 0])
        self.widget3d.set_on_mouse(self._on_mouse_widget3d)
        self.picked_points = []

    def _on_layout(self, layout_context):
        r = self.window.content_rect
        self.widget3d.frame = r
        pref = self.info.calc_preferred_size(layout_context,
                                             gui.Widget.Constraints())
        self.info.frame = gui.Rect(r.x,
                                   r.get_bottom() - pref.height, pref.width,
                                   pref.height)

    def _on_mouse_widget3d(self, event):
        # We could override BUTTON_DOWN without a modifier, but that would
        # interfere with manipulating the scene.
        if event.type == gui.MouseEvent.Type.BUTTON_DOWN and event.is_modifier_down(
                gui.KeyModifier.CTRL):

            def depth_callback(depth_image):
                # Coordinates are expressed in absolute coordinates of the
                # window, but to dereference the image correctly we need them
                # relative to the origin of the widget. Note that even if the
                # scene widget is the only thing in the window, if a menubar
                # exists it also takes up space in the window (except on macOS).
                x = event.x - self.widget3d.frame.x
                y = event.y - self.widget3d.frame.y
                # Note that np.asarray() reverses the axes.
                depth = np.asarray(depth_image)[y, x]

                if depth == 1.0:  # clicked on nothing (i.e. the far plane)
                    world = [0, 0, 0]
                    text = ""
                else:
                    world = self.widget3d.scene.camera.unproject(
                        event.x, event.y, depth, self.widget3d.frame.width,
                        self.widget3d.frame.height)
                    text = "({:.3f}, {:.3f}, {:.3f})".format(
                        world[0], world[1], world[2])
                self.picked_points.append(world)

                # This is not called on the main thread, so we need to
                # post to the main thread to safely access UI items.
                def update_label():
                    self.info.text = text
                    self.info.visible = (text != "")
                    # We are sizing the info label to be exactly the right size,
                    # so since the text likely changed width, we need to
                    # re-layout to set the new frame.
                    self.window.set_needs_layout()

                    mat = o3d.visualization.rendering.MaterialRecord()
                    mat.shader = "defaultLit"
                    # NOTE: USE 0.01 IF TESTING WITH POINT CLOUD
                    sphere = o3d.geometry.TriangleMesh.create_sphere(radius=0.1)
                    sphere.paint_uniform_color([0.1, 0.1, 0.7])
                    sphere.compute_vertex_normals()
                    sphere.translate(world)
                    self.widget3d.scene.add_geometry(f"point {len(self.picked_points)}", sphere, mat)

                gui.Application.instance.post_to_main_thread(
                    self.window, update_label)

            self.widget3d.scene.scene.render_to_depth_image(depth_callback)
            return gui.Widget.EventCallbackResult.HANDLED
        return gui.Widget.EventCallbackResult.IGNORED

def main():
    app = gui.Application.instance
    app.initialize()

    # This example will also work with a triangle mesh, or any 3D object.
    # If you use a triangle mesh you will probably want to set the material
    # shader to "defaultLit" instead of "defaultUnlit".

    # TEST WITH SPHERE...
    cloud = o3d.geometry.TriangleMesh.create_sphere(5.0)
    cloud.compute_vertex_normals()

    # OR POINTCLOUD...
    # cloud = o3d.io.read_point_cloud("examples/test_data/ICP/cloud_bin_0.pcd")

    ex = ExampleApp(cloud)

    app.run()


if __name__ == "__main__":
    main()
```

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4431)

### DIFF
--- a/cpp/open3d/visualization/gui/SceneWidget.cpp
+++ b/cpp/open3d/visualization/gui/SceneWidget.cpp
@@ -488,8 +488,7 @@ private:
         if (win_z < far_z) {
             auto vp = scene_->GetView()->GetViewport();
             auto point = scene_->GetCamera()->Unproject(
-                    float(x), float(vp[3] - y), win_z, float(vp[2]),
-                    float(vp[3]));
+                    float(x), float(y), win_z, float(vp[2]), float(vp[3]));
             SetCenterOfRotation(point);
             interactor_->Rotate(0, 0);  // update now
         }
--- a/cpp/open3d/visualization/rendering/Camera.h
+++ b/cpp/open3d/visualization/rendering/Camera.h
@@ -114,6 +114,9 @@ public:
     virtual ProjectionMatrix GetProjectionMatrix() const = 0;
     virtual Transform GetCullingProjectionMatrix() const = 0;
 
+    /// Returns world space coordinates given an x,y position in screen
+    /// coordinates relative to upper left, the screen dimensions, and z is the
+    /// depth value (0.0 - 1.0)
     virtual Eigen::Vector3f Unproject(float x,
                                       float y,
                                       float z,
--- a/cpp/open3d/visualization/rendering/filament/FilamentCamera.cpp
+++ b/cpp/open3d/visualization/rendering/filament/FilamentCamera.cpp
@@ -347,7 +347,8 @@ void FilamentCamera::SetModelMatrix(const Transform& view) {
 Eigen::Vector3f FilamentCamera::Unproject(
         float x, float y, float z, float view_width, float view_height) const {
     Eigen::Vector4f gl_pt(2.0f * x / view_width - 1.0f,
-                          2.0f * y / view_height - 1.0f, 2.0f * z - 1.0f, 1.0f);
+                          2.0f * (view_height - y) / view_height - 1.0f,
+                          2.0f * z - 1.0f, 1.0f);
 
     auto proj = GetProjectionMatrix();
     Eigen::Vector4f obj_pt = (proj * GetViewMatrix()).inverse() * gl_pt;


benjaminum (approved):
On Linux it works as expected -> :lgtm: 

On Mac a single point appears somewhere on the first click and no coordinates are shown. I get the same behavior on Mac without the patch. Probably a different problem or a problem with my machine.

> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4431#-:-Mra3F3b7D2jOS_9gs8w:b-ehgzeh)* status: 0 of 3 files reviewed, all discussions resolved

errissa:
> On Mac a single point appears somewhere on the first click and no coordinates are shown. I get the same behavior on Mac without the patch. Probably a different problem or a problem with my machine.

I was able to reproduce this with and without the patch. When the point does appear it appears correctly with this PR but incorrectly without it so the issue doesn't appear to be related to this PR.

PR #4438, opened 2021-12-16
fix nan in normal image

Fuhrmann-sep:
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4438)

### DIFF
--- a/cpp/open3d/t/geometry/kernel/ImageImpl.h
+++ b/cpp/open3d/t/geometry/kernel/ImageImpl.h
@@ -250,7 +250,7 @@ void CreateNormalMapCPU
                     float normal_norm =
                             sqrt(normal[0] * normal[0] + normal[1] * normal[1] +
                                  normal[2] * normal[2]);
-                    normal_norm = max(normal_norm, EPSILON);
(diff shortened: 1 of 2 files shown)


ssheorey (changes requested):
Hi @Fuhrmann-sep thanks for submitting this PR. To pass the CI checks, can you:
- Run `make apply-style` to pass the style checks.
- The normal CI tests are failing since this will change the values of the estimated normals. An alternate way to avoid NaN is:
`normal_norm = std::max(normal_norm, EPSILON);` instead of adding 1e-5 (a magic number). Define `constexpr float EPSILON = 1e-5f`;  so that readers understand the code. This will ensure that the normals are unchanged when they are not NaN.

Fuhrmann-sep:
Hello, @ssheorey thank you for the reminder. Modifications are done and another commit is added.

ssheorey:
> Hello, @ssheorey thank you for the reminder. Modifications are done and another commit is added.

Thanks @Fuhrmann-sep You need `std::max`.

ssheorey (approved):

PR #4462, opened 2021-12-18
rm -rf folder, std::experimental::filesystem, macOS 10.15+

yxlao:
- Allow `DeleteDirectory()` to remove directory with its contents.
- Uses `std::experimental::filesystem`, can be changed to `std::filesystem` after C++17 is enabled.
- Requires macOS 10.15+ (Initial release date: October 7, 2019).

This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4462)

### DIFF
--- a/cmake/Open3DSetGlobalProperties.cmake
+++ b/cmake/Open3DSetGlobalProperties.cmake
@@ -6,6 +6,23 @@ function(open3d_set_global_properties target)
     target_compile_features(${target} PUBLIC cxx_std_14)
 
     # std::filesystem (C++17) or std::experimental::filesystem (C++14)
+    #
+    # Ref: https://en.cppreference.com/w/cpp/filesystem:
+    #      Using this library may require additional compiler/linker options.
+    #      GNU implementation prior to 9.1 requires linking with -lstdc++fs and
+    #      LLVM implementation prior to LLVM 9.0 requires linking with -lc++fs.
+    # Ref: https://gitlab.kitware.com/cmake/cmake/-/issues/17834
+    #      It's non-trivial to determine the link flags for CMake.
+    #
+    # The linkage can be "-lstdc++fs" or "-lc++fs" or ""(empty). In our
+    # experiments, the behaviour doesn't quite match the specifications.
+    #
+    # - On Ubuntu 20.04:
+    #   - "-lstdc++fs" works with with GCC 7/10 and Clang 7/12
+    #   - "" does not work with GCC 7/10 and Clang 7/12
+    #
+    # - On latest macOS/Windows with the default compiler:
+    #   - "" works.
     if(UNIX AND NOT APPLE)
         target_link_libraries(${target} PRIVATE stdc++fs)
     endif()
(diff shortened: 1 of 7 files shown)


cpp/open3d/utility/FileSystem.cpp
```diff
@@ -47,6 +47,15 @@
 #include <unistd.h>
 #endif
```
ssheorey, line 48: `using std::experimental::filesystem;` will shorten the filesystem calls in this file. We just need to change that to `using std:filesystem;` later.

cpp/open3d/utility/FileSystem.cpp
```diff
@@ -47,6 +47,15 @@
...
 
+#ifdef WIN32
+#define _SILENCE_EXPERIMENTAL_FILESYSTEM_DEPRECATION_WARNING
+#endif
+#ifdef __APPLE__
+// CMAKE_OSX_DEPLOYMENT_TARGET "10.15" or newer
+#define _LIBCPP_NO_EXPERIMENTAL_DEPRECATION_WARNING_FILESYSTEM
```
ssheorey, line 55: What happens if this is done on 10.14? Would it just ignore it and give a warning any way? Perhaps we can suppress the warning to allow building on 10.14?

cmake/Open3DSetGlobalProperties.cmake
```diff
@@ -5,6 +5,11 @@ function(open3d_set_global_properties target)
     # Tell CMake we want a compiler that supports C++14 features
     target_compile_features(${target} PUBLIC cxx_std_14)
 
+    # std::filesystem (C++17) or std::experimental::filesystem (C++14)
+    if(UNIX AND NOT APPLE)
+        target_link_libraries(${target} PRIVATE stdc++fs)
```
ssheorey, line 10: Is this a compiler specific decision instead of OS specific? I see on cppreference:
`Using this library may require additional compiler/linker options. GNU implementation requires linking with -lstdc++fs and LLVM implementation requires linking with -lc++fs.`

ssheorey (approved):

cpp/open3d/utility/FileSystem.cpp
yxlao: Done. Changed to `namespace fs = std::experimental::filesystem;`.

cpp/open3d/utility/FileSystem.cpp
yxlao: It will not build for 10.14 at all. That's why we also change the `CMAKE_OSX_DEPLOYMENT_TARGET` to `10.15`.

cpp/tests/utility/FileSystem.cpp
```diff
@@ -353,11 +353,18 @@ TEST(FileSystem, DeleteDirectory) {
...
 
+    status = utility::filesystem::MakeDirectory(path + "/sub_dir");
+    EXPECT_TRUE(status);
+    FILE *file = utility::filesystem::FOpen(path + "/file_name.txt", "w");
+    fclose(file);
+
     status = utility::filesystem::DeleteDirectory(path);
     EXPECT_TRUE(status);
 
+    // std::filesystem::remove_all() won't throw error if the directory doesn't
```
reyanshsolis, line 364: Do we want to use `utility::filesystem::DirectoryExists()` to check, and throw a warning if the directory does not exist?

yxlao: I'd like it to behave like `rm -rf`, which will be silent if the dir does not exist.

PR #4463, opened 2021-12-18
Enable Python 3.10 (>= Python 3.6)

### LINKED ISSUE #4461: build python module for python 3.10.1
Is there a specific reason why we can't build the python module for python 3.10.1 ?
When running cmake it fails because it expects version 3.6 - 3.9:
-- Open3D 0.14.1+96bb48c5a
-- Could NOT find Python3: Found unsuitable version "3.10.1", required range is "3.6...<3.10"

Best, A.

yxlao:
For now, we don't release 3.10 wheel since PyTorch/TensorFlow does not support 3.10. 

However, we can still enable the 3.10 build-from-source option for users. 

Fixes #4461

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4463)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -246,7 +246,7 @@ if(NOT DEFINED Python3_FIND_REGISTRY)
     # Only consider PATH variable on Windows by default
     set(Python3_FIND_REGISTRY NEVER)
 endif()
-find_package(Python3 3.6...<3.10
+find_package(Python3 3.6...3.10
              COMPONENTS Interpreter Development
 )
 if (Python3_FOUND)
@@ -256,7 +256,7 @@ if (Python3_FOUND)
         "Deprecated path to the Python executable (for 3rdparty only)" FORCE)
 else()
     if (BUILD_PYTHON_MODULE)
-        message(FATAL_ERROR "BUILD_PYTHON_MODULE=ON requires Python 3.6-3.9. Please ensure it is in PATH.")
+        message(FATAL_ERROR "BUILD_PYTHON_MODULE=ON requires Python 3.6-3.10. Please ensure it is in PATH.")
     endif()
 endif()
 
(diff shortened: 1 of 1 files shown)


ssheorey (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4463)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @ssheorey and @yxlao)

___
*[CMakeLists.txt, line 249 at r1](https://reviewable.io/reviews/isl-org/open3d/4463#-MrNSK8Q3NbhO37AMhq_:-MrNSK8Q3NbhO37AMhqa:b-cjtj0f) ([raw file](https://github.com/isl-org/open3d/blob/c2a344e85c932630d34e341095d6ed383d593723/CMakeLists.txt#L249)):*
> ```CMake
>     set(Python3_FIND_REGISTRY NEVER)
> endif()
> find_package(Python3 3.6...3.10
> ```

`find_package(Python3 3.6`
I think Python API is stable enough that we can just say >=3.6 instead of >=3.6 and <=3.10. Everything should very likely work for 3.11 as well when it comes out.

odysseusPolytropos (changes requested):
If the intent was to accept all minor versions Python 3.10, this change is not correct. It will reject e.g. 3.10.1. To accept all Python 3.10 (but excluding 3.11 and following) line 249 should be:
 find_package(Python3 3.6...<3.11)

yxlao:
Thanks @ssheorey @odysseusPolytropos . Updated to support Python 3.6+.

odysseusPolytropos (approved):
:+1:

odysseusPolytropos (approved):

MrBenGriffin:
both TensorFlow and PyTorch are now supported in 3.10

PR #4188, opened 2021-10-23
expose API for Fast Global Registration given correspondences

plusk01:
- make FGR API more explicit and parallel the RANSAC registration API by:
	- renaming in c++ `FastGlobalRegistration` --> `FastGlobalRegistrationBasedOnFeatureMatching`
	- adding c++ `FastGlobalRegistrationBasedOnCorrespondence`, python `registration_fgr_based_on_correspondence`
- rename existing python API to be more explicit: `registration_fast_based_on_feature_matching` --> `registration_fgr_based_on_feature_matching`
- add function documentation in c++
- updates in `global_registration.ipynb` and `examples/python/reconstruction_system/register_fragments.py` for renamed `registration_fgr_based_on_feature_matching`

I felt `registration_fgr_*` was more descriptive than `registration_fast_*`, but happy to revert that change if we don't want to change the API

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4188)

PR #4479, opened 2021-12-22
fix call to AdvancedMatching in FGR

plusk01:
the `FastGlobalRegisrationBasedOnCorrespondence` function from https://github.com/isl-org/Open3D/pull/4188 has a bug in it (does not affect the original `FastGlobalRegistrationBasedOnFeatureMatching`)

the check on if `source` or `target` has more points (which was copied from `FastGlobalRegistrationBasedOnFeatureMatching`)  is useful so that the smaller set of points is used as the query in `InitialMatching`. This check was copied into the new `FastGlobalRegisrationBasedOnCorrespondence` function, but does not use `InitialMatching` -- i.e., the initial matches come from  the user. In this case, it does not make sense to swap the source/target based on number of points and in fact by swapping them but not swapping the "columns" of `corresvec` results in incorrect initial matches.

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4479)

### DIFF
--- a/cpp/open3d/pipelines/registration/FastGlobalRegistration.cpp
+++ b/cpp/open3d/pipelines/registration/FastGlobalRegistration.cpp
@@ -316,13 +316,7 @@ RegistrationResult FastGlobalRegistrationBasedOnCorrespondence(
     }
 
     if (option.tuple_test_) {
-        // for AdvancedMatching ensure the first point cloud is the larger one
-        if (source.points_.size() > target.points_.size()) {
-            corresvec = AdvancedMatching(source, target, corresvec, option);
-        } else {
-            corresvec = AdvancedMatching(target, source, corresvec, option);
-            for (auto& p : corresvec) std::swap(p.first, p.second);
-        }
+        corresvec = AdvancedMatching(source, target, corresvec, option);
     }
 
     Eigen::Matrix4d transformation;
@@ -359,8 +353,8 @@ RegistrationResult FastGlobalRegistrationBasedOnFeatureMatching(
 
     std::vector<std::pair<int, int>> corres;
     if (option.tuple_test_) {
-        // for AdvancedMatching ensure the first point cloud is the larger one
-        if (source.points_.size() > target.points_.size()) {
+        // use the smaller point cloud as the query during knn search
+        if (source.points_.size() >= target.points_.size()) {
             corres = AdvancedMatching(
                     source, target,
                     InitialMatching(source_feature, target_feature), option);


theNded (changes requested):
> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4479)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @plusk01 and @yxlao)

___
*[cpp/open3d/pipelines/registration/FastGlobalRegistration.cpp, line 323 at r1](https://reviewable.io/reviews/isl-org/open3d/4479#-Mrbx2sn9q1hqmgph25M:-Mrbx2sn9q1hqmgph25N:b-kvg8a8) ([raw file](https://github.com/isl-org/open3d/blob/9e384236edfb2914854fe60f93f866c8c4e5f085/cpp/open3d/pipelines/registration/FastGlobalRegistration.cpp#L323)):*
> ```cpp
>             corresvec = AdvancedMatching(source, target, corresvec, option);
>         } else {
>             corresvec = AdvancedMatching(target, source, corresvec, option);
> ```

How about changing the order of these two lines?
```
  for (auto& p : corresvec) std::swap(p.first, p.second);
  corresvec = AdvancedMatching(target, source, corresvec, option);
```

plusk01:
thanks @theNded - that is essentially what i am doing by removing the `if` statement altogether. In this case (when the correspondences are specified by the user and so the `InitialMatching` function is not called), there is no point checking if the `source` or `target` has more points.

theNded:
#4188 has introduced bugs in #4481 and #4522, and this PR does not address them. I will have to revisit #4188 and consider whether to revert it before moving forward with this PR.

theNded:
Sorry for coming back late, I forgot about this PR.
Actually, after the changes from this PR, the problem from #4481 reappears. Would you like to investigate?

plusk01:
now it's my turn to apologize for coming back late @theNded :)
i just rebased this PR on master and tested on the example of #4481 - I did not see the issue again. I am wondering if you had rebased to include the fixes from #4547 ?

nkapgate:
I am seeing some crashes due to memory leak on Linux and Intel macs, is this issue fixable?

plusk01:
Hi @nkapgate, do you have an small example that this PR is failing on?

ssheorey:
Check for crash, else merge.

theNded:
> Check for crash, else merge.

We can merge now. FGR has a lot of issues and is very slow now, no longer a "fast" method. Needs a complete rewrite. I will collect and fix existing issues in a separate PR.

theNded (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4479#-:-NMZtlsv2l2z6oGn8VFj:bnfp4nl)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @plusk01 and @ssheorey)

ssheorey (approved):
Tested no crash with example code from #4481. LGTM.

PR #4494, opened 2021-12-24
Data: Extract From ZIP

reyanshsolis:
- Patch Zlib to expose contrib/minizip/unzip.h and ioapi.h, and enable extraction of encrypted files.
- Add `ExtractZIP` (extract from zip functionality using unzip.h)
- General `Extract` API, which calls `ExtractZIP` for `.zip` extension.
- Unit Test.

TODO: (in next PR)
- [ ] Print Progress
- [ ] Avoid Overwrite
- [ ] Try Win32 API

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4494)


cpp/tests/data/Extract.cpp
```diff
@@ -0,0 +1,61 @@
...
+#include "open3d/utility/FileSystem.h"
+#include "open3d/utility/Helper.h"
+#include "open3d/utility/Logging.h"
+#include "tests/Tests.h"
+
+namespace open3d {
+namespace tests {
+
+TEST(Extract, ExtractFromZIP) {
+    std::string file_path = utility::GetDataPathCommon("test_data_00.zip");
```
yxlao, line 38: use the downloader to get the test data

cpp/open3d/data/Extract.h
```diff
@@ -0,0 +1,40 @@
...
+// ----------------------------------------------------------------------------
+
+#pragma once
+
+#include <string>
+
+namespace open3d {
+namespace data {
+
+bool Extract(const std::string& filename,
```
yxlao, line 34: Let's refactor the interface. After reading more about "returning boolean" vs. "throw exception", I believe throwing exception is more appropriate here. 

- When an extraction fails, it is an "exceptional" case. 
- We'll return `true/false` when we expect the result to be either "true" or "false". In our case, we only expect the result to be "true". Otherwise, something unexpected happened. 
- In practice, we want to stop the execution and notify the user when a file fails to extract. 
- If we return true/false, we cannot make sure the developer always remember to do the true/false check. 

For the same reason, I am also changing the downloader APIs to throw exceptions rather than returning boolean flags. We can apply the same principle to other I/O functions.

yxlao (changes requested):

cpp/tests/data/Extract.cpp
reyanshsolis: Done.

cpp/open3d/data/ExtractZIP.h
```diff
@@ -0,0 +1,42 @@
...
+
+#include <string>
+
+namespace open3d {
+namespace data {
+
+bool ExtractFromZIP(const std::string& filename,
+                    const std::string& extract_dir,
+                    const std::string& password = "",
+                    const bool print_progress = true);
```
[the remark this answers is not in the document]
reyanshsolis: Done.

cpp/open3d/data/Extract.h
```diff
@@ -0,0 +1,40 @@
...
+             const std::string& extract_dir,
+             const std::string& password = "",
```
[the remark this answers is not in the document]
reyanshsolis: Done. [Removed from API. However, the method is still implemented, such that it may be exposed in the future if required].

cpp/open3d/data/Extract.cpp
```diff
@@ -0,0 +1,79 @@
...
+    if (map_itr == file_extension_to_extract_function.end()) {
+        utility::LogWarning(
+                "Extraction Failed: unknown file extension for "
+                "{} (format: {}).",
+                filename, format);
+        return false;
+    }
+
+    bool success =
+            map_itr->second(filename, extract_dir, password, print_progress);
```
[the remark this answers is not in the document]
reyanshsolis: Done.

cpp/open3d/data/Extract.cpp
```diff
@@ -0,0 +1,79 @@
...
+             const std::string& extract_dir,
+             const std::string& password,
+             const bool print_progress) {
+    std::string format =
+            utility::filesystem::GetFileExtensionInLowerCase(filename);
+
+    utility::LogDebug("Format {} File {}", format, filename);
+
+    auto map_itr = file_extension_to_extract_function.find(format);
+    if (map_itr == file_extension_to_extract_function.end()) {
```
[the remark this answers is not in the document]
reyanshsolis: Done.

cpp/open3d/data/Extract.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+#include "open3d/data/Extract.h"
+
+#include <unordered_map>
+
+#include "open3d/data/ExtractZIP.h"
+#include "open3d/utility/FileSystem.h"
+#include "open3d/utility/Helper.h"
+#include "open3d/utility/Logging.h"
+#include "open3d/utility/ProgressReporters.h"
```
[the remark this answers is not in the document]
reyanshsolis: Done. [waiting to check with CI once].

cpp/open3d/data/ExtractZIP.cpp
```diff
@@ -0,0 +1,252 @@
...
+
+namespace open3d {
+namespace data {
+
+static int ExtractCurrentFile(unzFile uf, std::string password) {
+    char filename_inzip[256];
+    char *filename_withoutpath;
+    char *p;
+    int err = UNZ_OK;
+    FILE *fout = NULL;
```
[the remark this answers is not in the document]
reyanshsolis: Done.

cpp/open3d/data/ExtractZIP.cpp
```diff
@@ -0,0 +1,252 @@
...
+
+#define CASESENSITIVITY (0)
+// If required in future, the `WRITEBUFFERSIZE` size can be increased to 16384.
+#define WRITEBUFFERSIZE (8192)
+#define MAXFILENAME (256)
+
+namespace open3d {
+namespace data {
+
+static int ExtractCurrentFile(unzFile uf, std::string password) {
```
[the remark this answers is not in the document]
reyanshsolis: Done.

cpp/open3d/data/Extract.h
reyanshsolis: Done. [changed to `void` function, throwing `exception`].

cpp/open3d/data/Extract.cpp
```diff
@@ -0,0 +1,65 @@
...
+        std::function<void(const std::string&, const std::string&)>>
+        file_extension_to_extract_function{
+                {"zip", ExtractFromZIP},
+        };
+
+void Extract(const std::string& filename, const std::string& extract_dir) {
+    const std::string format =
+            utility::filesystem::GetFileExtensionInLowerCase(filename);
+
+    utility::LogDebug("Format {} File {}", format, filename);
```
yxlao, line 51: If we want the user to see it, use `LogInfo`.
We rarely use `LogDebug`. 
Here, I'll skip logging, since this information is not useful.

cpp/open3d/data/Extract.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+    if (file_extension_to_extract_function.count(format) == 0) {
+        utility::LogError(
+                "Extraction Failed: unknown file extension for "
+                "{} (format: {}).",
+                filename, format);
+    }
+
+    file_extension_to_extract_function.at(format)(filename, extract_dir);
+    utility::LogDebug("Successfully extracted {}.", filename);
```
yxlao, line 61: Now, this information is useful. The user needs to be aware of where is the file. 
I will put it before the extraction, such that when error occurs during the extraction, we have already printed the message.

The message shall show src and dst of the extraction.

cpp/open3d/data/Extract.h
```diff
@@ -0,0 +1,41 @@
...
+#pragma once
+
+#include <string>
+
+namespace open3d {
+namespace data {
+
+/// \brief Function to extract compressed files.
+/// \param filename Filename including path-prefix.
+/// Example: "/path/to/file/file.zip"
```
yxlao, line 36: Is this a file name or file path?
Does the `extract_dir` contain the prefix?
Shall we use the absolute path directly, since this is not used by the user?

cpp/open3d/data/Extract.cpp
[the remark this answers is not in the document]
yxlao: Looks like `ProgressReporters.h` is not used. Could you double-check the headers? (Including other files.)

cpp/open3d/data/ExtractZIP.cpp
```diff
@@ -0,0 +1,237 @@
...
+
+#define CASESENSITIVITY (0)
+// If required in future, the `WRITEBUFFERSIZE` size can be increased to 16384.
+#define WRITEBUFFERSIZE (8192)
+#define MAXFILENAME (256)
+
+namespace open3d {
+namespace data {
+
+static int ExtractCurrentFile(unzFile uf, const std::string &password) {
```
yxlao, line 67: This is a copied C funtion, it is returning an int. We shall not throw an exception here. Let it return an integer and throw exception after that. 

Let's revert to the previous state.

cpp/open3d/data/ExtractZIP.cpp
```diff
@@ -0,0 +1,237 @@
...
+    void *buf;
+    uInt size_buf;
+
+    unz_file_info64 file_info;
+    err = unzGetCurrentFileInfo64(uf, &file_info, filename_inzip,
+                                  sizeof(filename_inzip), nullptr, 0, nullptr,
+                                  0);
+
+    if (err != UNZ_OK) {
+        utility::LogError("Error {} with zipfile in unzGetCurrentFileInfo.",
```
yxlao, line 82: This information isn't very helpful. We shall call it error code (since it is an integer), or convert it to string. 

Here's what we did for curl:
```cpp
            utility::LogWarning("Download failed with error code: {}.",
                                curl_easy_strerror(res));
```

(We shouldn't `LogError` in this function anyway, this error shall be printing after `ExtractCurrentFile` returns.)

cpp/tests/data/Extract.cpp
```diff
@@ -0,0 +1,79 @@
...
+    // Extract the test zip file.
+    EXPECT_NO_THROW(data::Extract(file_path, download_dir));
+    std::string extracted_folder = download_dir + "/test_data";
+    std::string output_file = extracted_folder + "/lena_color.jpg";
+
+    // Check if the extracted file exists.
+    EXPECT_TRUE(utility::filesystem::FileExists(output_file));
+
+    // Delete test file.
+    std::remove(output_file.c_str());
```
yxlao, line 60: can we use the `utility::` function to delete the file as well?

cpp/tests/data/Extract.cpp
```diff
@@ -0,0 +1,79 @@
...
+    utility::filesystem::DeleteDirectory(extracted_folder);
```
yxlao, line 61: `EXPECT_TRUE()`, when we delete the directory. 
This is the case where the developer forgot to check the return code, and that was the reason why we moved to throwing exception for the extract function.

yxlao (changes requested):

cpp/open3d/data/Extract.cpp
reyanshsolis: Done.

cpp/open3d/data/Extract.cpp
reyanshsolis: Done.

cpp/open3d/data/Extract.h
reyanshsolis: Yes, `extract_dir` (file path) is the absolute path, it contains `data_root/prefix`. 
`filename` contains `path/to/file/filename.zip` (so, its `filename_with_prefix`). 
Do you have some suggestions to make it more clear?

cpp/open3d/data/ExtractZIP.cpp
reyanshsolis: Done.

cpp/open3d/data/ExtractZIP.cpp
reyanshsolis: Done.

cpp/tests/data/Extract.cpp
reyanshsolis: Done.

cpp/tests/data/Extract.cpp
reyanshsolis: Done.

cpp/open3d/data/ExtractZIP.cpp
reyanshsolis: Reverted the function to the previous state, with LogWarning, since it might be useful for debugging the error source in the future. The error is thrown in `ExtractAll` with the error code.

cpp/open3d/data/Extract.h
yxlao: ### Comments

- `name` can be confusing, since `name` typically means something different from `path`.
- Whether it is a relative path or an absolute path, we shall keep both arguments consistent.

### Suggestions

- Option 1: 
  - `const std::string& file_path`, `const std::string& extract_dir`
  - Both use absolute path. That is do not add any prefix in the `Extract()` function.
- Option 2:
  - `const std::string& file_path_from_data_root`, `const std::string& extract_dir_from_data_root`.
  - It is fine that the name is long since the user doesn't see it.

I prefer Option 1. Since it makes the `Extract()` function decouples from the dataset class design.

yxlao:
One additional request: 
- Migrate to use the new filesystem in #4462
- Migrate to the new downloader API after #4496 is merged.

cpp/open3d/data/Extract.h
reyanshsolis: Done. Thanks for the detailed suggestion.

reyanshsolis:
> One additional request:
> 
> * Migrate to use the new filesystem in [rm -rf folder, std::experimental::filesystem, macOS 10.15+ #4462](https://github.com/isl-org/Open3D/pull/4462)
> * Migrate to the new downloader API after [Refactor of downloader #4496](https://github.com/isl-org/Open3D/pull/4496) is merged.

Done.

This also allows us to create a `temp. folder` for each test, where we can dump data for that test, and delete the folder for that test at once.

cpp/open3d/data/ExtractZIP.cpp
```diff
@@ -0,0 +1,251 @@
...
+            uf = unzOpen64(file_path_try);
+        }
+    }
+
+    if (uf == nullptr) {
+        utility::LogError("Failed to open file {}.", file_path);
+    }
+
+    // Change working directory to the extraction directory.
+    if (chdir(extract_dir.c_str())) {
```
reyanshsolis, line 239: Issue: The working directory is getting changed here. So, all the unit-test data is getting dumped in this folder.
However, as we delete this directory in the cleanup step, the unit tests are failing as the functions are unable to find anything at this location. 

Need to get the current working directory before this function, and revert back to it after execution of this function.

reyanshsolis: Fixed.

yxlao: Can `ChangeWorkingDirectory()` have potential side effects?

E.g. if there is an exception occurred in between the two `ChangeWorkingDirectory()`, the program may exit and we have only executed the first `ChangeWorkingDirectory()`. 

Is it possible to achieve extraction, without changing the directory? e.g. by prefixing the path in the actual extraction function?

reyanshsolis: Thanks for pointing this out! Also, another concern was, that throwing exception may not ensure if the file was closed before exiting. 

Current Solution:
The easiest solution was to  `switch working dir, and close file before throwing error` in the `ExtractAll` function, as `ExtractCurrentFile` function is not expected to throw an error in the ideal case. 

It might still be an improvement if `prefixing the path` in `ExtractCurrentFile` works without changing dir. I will try that.

reyanshsolis: Done. `changing working dir` is avoided. Also further cleaned up the functions.

yxlao (approved):

PR #4502, opened 2021-12-27
Fix #4497

### LINKED ISSUE #4497: Limit case in `get_rotation_matrix_from_axis_angle`: rotation by angle 0
**Describe the bug**
The limit case of a rotation by angle 0 is not supported in `o3d.geometry.get_rotation_matrix_from_axis_angle` (i.e. `Geometry3D::GetRotationMatrixFromAxisAngle`): the rotation matrix returned is all `nan`'s.

**To Reproduce**
The following code will show the issue: a rotation by a very small angle, `1e-9` here, will yield a valid rotation matrix which is almost the identity matrix, whereas a rotation by angle 0 will yield a matrix of all `nan` values.

```python
import numpy as np
import open3d as o3d
rot_mat_almost_identity = o3d.geometry.get_rotation_matrix_from_axis_angle(np.array([0, 1, 0]) * 1e-9)
np.testing.assert_allclose(rot_mat_almost_identity, np.identity(3), atol=1e-8)  # OK
rot_mat_should_be_identity = o3d.geometry.get_rotation_matrix_from_axis_angle(np.array([0, 1, 0]) * 0)
np.testing.assert_allclose(rot_mat_should_be_identity, np.identity(3), atol=1e-8)  # AssertionError
```

**Expected behavior**
For a vector which is all zeros, `o3d.geometry.get_rotation_matrix_from_axis_angle` should return the identity matrix, or at least it should raise an error, rather returning silently a matrix of `nan`'s.

- Open3D version: 0.14.1

**Additional context**

The `o3d.geometry.get_rotation_matrix_from_axis_angle` function calls `Geometry3D::GetRotationMatrixFromAxisAngle`. In this function, the variable `phi` will be zero in the case of a vector which is all zeros:
https://github.com/isl-org/Open3D/blob/7c62640441e3da18bcbe146723ed83ff544b2fbb/cpp/open3d/geometry/Geometry3D.cpp#L210-L213

theNded:
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4502)

### DIFF
--- a/cpp/open3d/geometry/Geometry3D.cpp
+++ b/cpp/open3d/geometry/Geometry3D.cpp
@@ -210,7 +210,9 @@ Eigen::Matrix3d Geometry3D::GetRotationMatrixFromYXZ(
 Eigen::Matrix3d Geometry3D::GetRotationMatrixFromAxisAngle(
         const Eigen::Vector3d& rotation) {
     const double phi = rotation.norm();
-    return Eigen::AngleAxisd(phi, rotation / phi).toRotationMatrix();
+    Eigen::Vector3d axis =
+            phi > 0 ? rotation / phi : Eigen::Vector3d::Zero().eval();
+    return Eigen::AngleAxisd(phi, axis).toRotationMatrix();
 }
 
 Eigen::Matrix3d Geometry3D::GetRotationMatrixFromQuaternion(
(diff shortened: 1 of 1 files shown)


reyanshsolis (approved):

reyanshsolis:
nit: may we make axis `const` ?

PR #4546, opened 2022-01-06
Fix wrong coordinates

### LINKED ISSUE #4244: mouse-and-point-coord.py example giving different coordinates from different camera angle

ZhuoyueLyu:
Fixes #4244

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4546)

### DIFF
--- a/examples/python/gui/mouse-and-point-coord.py
+++ b/examples/python/gui/mouse-and-point-coord.py
@@ -103,7 +103,7 @@ class ExampleApp:
                     text = ""
                 else:
                     world = self.widget3d.scene.camera.unproject(
-                        event.x, event.y, depth, self.widget3d.frame.width,
+                        x, (self.widget3d.frame.height - y), depth, self.widget3d.frame.width,
                         self.widget3d.frame.height)
                     text = "({:.3f}, {:.3f}, {:.3f})".format(
                         world[0], world[1], world[2])


ZhuoyueLyu:
Hi @yxlao @errissa, the style checker returns an error, but it seems it doesn't say where the problem is. It's a one-line change, and I really don't know why it's not okay 😅

errissa:
@ZhuoyueLyu This change to `mouse-and-point-coord.py` is no longer necessary. Issue #4244 was resolved via #4431. I just tested the example with master and it works as expected.

If I am missing something, please let me know. Otherwise, I will close this PR.

Issue #4539, opened 2022-01-06
Issues in RANSCA registration using 0.13.0 (& 0.14.1), no issue with 0.10.0


chz31:
Dear all,

I originally used RANSAC and ICP in open3d 0.10.0 to register a group of mouse skulls, and it works fine. However, with the 0.13.0 and 0.14.1 version, the skulls are now aligned poorly (see the screenshots below). I have tested these versions of open3d on both Windows 10 and Linux with Python 3.6.7 and 3.8.10. I counter the same issues. I am pretty sure it is cause by RANSAC registration.
![Align_0 10 0](https://user-images.githubusercontent.com/80793828/148307951-6559eda3-6a37-43c4-90f7-49c2e9e05942.PNG)
![Align_0 13 0](https://user-images.githubusercontent.com/80793828/148307949-5ebe3bcf-f39d-45c2-8543-6d67bd164f97.PNG)

I submitted a similar question early on but replied late, so the issue was closed (#4154 https://github.com/isl-org/Open3D/issues/4154). Sorry for the delayed response. I did what you asked by increasing the iteration rounds to 4000,000 and confidence level to 0.99999, but it still does not work.

I've created python script with pilot data (posted on GitHub) to reproduce this issue. **You may clone or download the repository as zip file that contains the data (mouse ply files) and scripts here:** **https://github.com/chz31/o3d_test** . The "data" folder in the repository "o3d_test" contains the mouse models for you to test. 

The scripts are in "test_scripts" folder. The "test_0.10.0.py" contains the script using open3d 0.10.0. The "test_0.13.0.py" contains script using open3d 0.13.0. 

**For each script, please modify three variables for accessing and output data (line 129 to 133 for test_0.10.0.py and line 131 to 135 for test_0.13.0.py):
" sourceFileDir" points to the folder that contains the mouse models downloaded from the Github repo I provided above.
"targetFileDir" points to a specific model "129_S1" in the "sourceFileDir" that serves as the universal target for registration.
"outDir" points to the folder that stores the mouse models register to the target "129_S1"**

To run the script after specifying the above three directories and path, you can simply run in your python console by using the code:
1. >path = "Enter_Your_Directory_that_stores_the_script_for_o3d_0.10.0/test_0.10.0.py" #Specify the path to test o3d 0.10.0.py
    >exec(open(path).read())    #The registered models will be stored in "outDir"

[...]

Issue #4522, opened 2021-12-31
FastGlobalRegistrationBasedOnFeatureMatching sometime fail


hungmiu:
**IMPORTANT: Please use the following template to report the bug.**

Describe the bug

Sometimes, when I use FastGlobalRegistrationBasedOnFeatureMatching for registering two pointclouds. It fails. And a message box appears "vector out of range".

I check the code, I found that at the line 74 of file FastGlobalRegistration.cpp
if (corres_ji[ij.second] == ij.first) corres_cross.push_back(ij);
some time ij.second = -1.

Please add a condition to check valid ij.second.

Thank you.

ssheorey:
Hi @hungmiu thanks for reporting this. Can you provide the code and data that can produce this error? That will help us reproduce it and debug it faster.

hungmiu:
> Hi @hungmiu thanks for reporting this. Can you provide the code and data that can produce this error? That will help us reproduce it and debug it faster.

This is sample code and data.
Running the function 1 time, the error may not appear but run 1000 times, you can see the error.


[RegistrationRANSAC.zip](https://github.com/isl-org/Open3D/files/7797284/RegistrationRANSAC.zip)

Thanks.

PR #3849, opened 2021-08-01
Parallelize FastGlobalRegistration

gsakkis:
Parallelize with OpenMP the initial matching of `FastGlobalRegistration`, plus minor refactoring.

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/3849)

PR #3108, opened 2021-03-02
Introducing SLAC

theNded:
Introduces SLAC and (partially) fix #540.
```
Simultaneous Localization and Calibration: Self-Calibration of Consumer Depth Cameras
Qian-Yi Zhou and Vladlen Koltun
CVPR 2014
```
Now SLAC works fine with downsampled input at the voxel resolution 0.05. At this setup, the system is insensitive to parameters. Only tuning the `distance_threshold` parameter would fit most of the scenes. Further refinement remains to be the future work.

### Examples
Simulated scene with heavy noise:
Before SLAC (with rigid optimization):
image

After SLAC:
image

Real world large scale scene:
Before SLAC (with rigid optimization):
before

After SLAC:
after

---
This change is [Reviewable](https://reviewable.io/reviews/intel-isl/open3d/3108)

PR #4547, opened 2022-01-06
Multiple RANSAC/FGR fixes

### LINKED ISSUE #3494: where can I  find the fast_global_registration  Module?
when I run  
https://github.com/intel-isl/Open3D/blob/master/examples/python/benchmark/benchmark_ransac.py
  File "E:\open3d\src\IO\examples\python\benchmark\benchmark_ransac.py", line 20, in <module>
    from fast_global_registration import *
ModuleNotFoundError: No module named 'fast_global_registration'
where can I find the fast_global_registration  Module?  

python 3.6
os windows 7
open3D version 0.12.0

### LINKED ISSUE #4481: running open3d.pipelines.registration.registration_fgr_based_on_feature_matching in a loop crashes
Hi,

I'm trying to register a set of 100 pairs of images using RANSAC followed by ICP using Open3D library in Python (3.7).
For the RANSAC step, I'm using the function [_open3d.pipelines.registration.registration_fgr_based_on_feature_matching_](http://www.open3d.org/docs/release/python_api/open3d.pipelines.registration.registration_fgr_based_on_feature_matching.html) that is in the latest version of the Open3D package (0.14.1).

Running these steps over one pair of images isn't a problem. But, when I'm running it in a loop 100 times, it crashes during one of the iterations when calling the RANSAC function.

The error I get in Python is:

> Process finished with exit code 139 (interrupted by signal 11: SIGSEGV)

A code example that reproduces the issue is:
```
import numpy as np
import open3d as o3d

# generating the source points
source = o3d.geometry.PointCloud()
source.points = o3d.utility.Vector3dVector(np.arange(20000*3).reshape([20000, 3]))

# generating the target points
target = o3d.geometry.PointCloud()
target.points = o3d.utility.Vector3dVector(np.asarray(source.points) + 1)

# generating the source features
source_fpfh = o3d.pybind.pipelines.registration.Feature()
source_fpfh.data = np.arange(33*20000).reshape([33, 20000])

# generating the target features
target_fpfh = o3d.pybind.pipelines.registration.Feature()
target_fpfh.data = source_fpfh.data + 1

for i in range(100):
    print(i+1)
    o3d.pipelines.registration.registration_fgr_based_on_feature_matching(source, target, source_fpfh, target_fpfh,
    o3d.pipelines.registration.FastGlobalRegistrationOption(maximum_correspondence_distance=0.5))
```

The output, in this case, is:

> 1
> 2
> 3
> 
> Process finished with exit code 139 (interrupted by signal 11: SIGSEGV)

The issue didn't occur in previous versions (when using the function [_open3d.pipelines.registration.registration_fast_based_on_feature_matching_](http://www.open3d.org/docs/0.13.0/python_api/open3d.pipelines.registration.registration_fast_based_on_feature_matching.html)). I'm using the latest version (0.14.1) because of the option to give a seed parameter to [_open3d.pipelines.registration.FastGlobalRegistrationOption_](http://www.open3d.org/docs/release/python_api/open3d.pipelines.registration.FastGlobalRegistrationOption.html) to make the RANSAC algorithm determenistic.

My OS details are:
NAME="Debian GNU/Linux"
VERSION="10 (buster)"

theNded:
- Fix bug of evaluating RANSAC quality introduced in #2636
- Add docs to avoid early termination (consume all iterations) by assigning confidence to 1.0.
- Follow up #4539, provide a python example and more options to the cpp example.
- Fix #4481, #4522 caused by #3849
- Fix #3494 
- Revert several parameter choices in the reconstruction systems brought up in #3108 

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4547)


reyanshsolis:
Hi @theNded, 
Do you think, in output displaying the non-downsampled point cloud might look better?
or the down-sampled point cloud is more accurate to be displayed as output?

image

For example, we may use the `transformation` obtained from `registration_result`, and apply it to the `original input`, to be displayed as output. I don't have a strong opinion, just wanted to know your opinion on this.

reyanshsolis:
I have verified `cpp` and `python` examples for `fgr` and `ransac`. 
Python benchmark for `ransac` is working, `fgr` is not, I will push the fix soon (just requires `execute_fast_global_registration` and `preprocess_point_cloud` to be moved locally to `benchmark_fgr` file. 
Another concern is that these benchmark examples, does not seem to perform performance benchmark, they are more like examples, so we might want to change that.

reyanshsolis:
Fix benchmark_fgr.py #4566

theNded:
> I have verified `cpp` and `python` examples for `fgr` and `ransac`. Python benchmark for `ransac` is working, `fgr` is not, I will push the fix soon (just requires `execute_fast_global_registration` and `preprocess_point_cloud` to be moved locally to `benchmark_fgr` file. Another concern is that these benchmark examples, does not seem to perform performance benchmark, they are more like examples, so we might want to change that.

Yes, we should reorganize the examples in a separate PR. `benchmark` are not benchmarks, and `misc` should be reorganized into `pipelines` and `geometry`.

theNded:
> Hi @theNded, Do you think, in output displaying the non-downsampled point cloud might look better? or the down-sampled point cloud is more accurate to be displayed as output?
> 
> image
> 
> For example, we may use the `transformation` obtained from `registration_result`, and apply it to the `original input`, to be displayed as output. image
> 
> I don't have a strong opinion, just wanted to know your opinion on this.

Done

reyanshsolis (approved):

theNded:
@reyanshsolis please do not merge now, the performance of RANSAC seems already affecting the reconstruction system. Need to fix it before merging.

theNded:
Still do not merge -- reconstruction system still has problems.

yxlao (approved):
> Reviewed 4 of 7 files at r2, 2 of 4 files at r3, 2 of 2 files at r4, 2 of 4 files at r5, 1 of 4 files at r6, 3 of 3 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4547)* status: all files reviewed, 2 unresolved discussions (waiting on @reyanshsolis and @theNded)

___
*[examples/python/pipelines/registration_fgr.py, line 58 at r7](https://reviewable.io/reviews/isl-org/open3d/4547#-MtRnLfEAYh9MIAQ8lWx:-MtRmP0n0TQn-Xgfm580:b78xi5v) ([raw file](https://github.com/isl-org/open3d/blob/a30ee054a0e0d6a501af210bdc920f523c04702b/examples/python/pipelines/registration_fgr.py#L58)):*
> ```Python
>         default=1.5,
>         help=
>         'multipler used to compute distance threshold between correspondences. Threshold is computed by voxel_size * distance_multiplier.'
> ```

nit: split into multiple lines:
```
help='xxxxxxxxxxxxxxxxx'
     'xxxxxxxxxxxxxxxxx'
```
___
*[examples/python/pipelines/registration_ransac.py, line 58 at r7](https://reviewable.io/reviews/isl-org/open3d/4547#-MtRnU7yBceA6JOYbCCk:-MtRnU7z4WmFKqhxE1vO:b78xi5v) ([raw file](https://github.com/isl-org/open3d/blob/a30ee054a0e0d6a501af210bdc920f523c04702b/examples/python/pipelines/registration_ransac.py#L58)):*
> ```Python
>         default=1.5,
>         help=
>         'multipler used to compute distance threshold between correspondences. Threshold is computed by voxel_size * distance_multiplier.'
> ```

nit: split into multiple lines:
```
help='xxxxxxxxxxxxxxxxx'
     'xxxxxxxxxxxxxxxxx'
```

reyanshsolis (approved):

reyanshsolis:
TODO: We currently have `function level unit testing`, may need to come up with a mechanism for `pipeline/module level unit testing`, as sometimes the difference in function level performance is not much, but at pipeline level, it results in very poor performance.

theNded:
> TODO: We currently have `function level unit testing`, may need to come up with a mechanism for `pipeline/module level unit testing`, as sometimes the difference in function level performance is not much, but at pipeline level, it results in very poor performance.

Where is the performance drop? Did you use the same parameter set?

PR #4553, opened 2022-01-07
T::IO::PointCloud PLY Cleanups Fixes #4548

### LINKED ISSUE #4548: Tensor Pointcloud to PLY writes non-standard types
The header generated when writing using `o3d.t.io.write_point_cloud` uses data types like "float32" and "uint8". These don't seem to be standard types. There are tools which can open that, but as far as I understand the only valid types are:
```
name        type        number of bytes
---------------------------------------
char       character                 1
uchar      unsigned character        1
short      short integer             2
ushort     unsigned short integer    2
int        integer                   4
uint       unsigned integer          4
float      single-precision float    4
double     double-precision float    8
```
That comes from http://paulbourke.net/dataformats/ply/, https://se.mathworks.com/help/vision/ug/the-ply-format.html and many more. I haven't seen a source mentioning specific sized types. Maybe it's legal as there isn't really a strict "standard", but it's still unusual. No other tools that I have checked save like that, like PCL and Meshlab.

Also this is a different behavior from `o3d.io.write_point_cloud` (non-Tensor) variant as that writes standard names.

It makes sense that maybe a custom type not in these specified ones are written like this. Or if the user could name the type as required.

**To Reproduce**
Use `o3d.t.io.write_point_cloud()` to write a PLY file in either ASCII or Binary. Header will have sized types.

**Expected behavior**
All types should be as standard.

Example header when saving with `o3d.io.write_point_cloud` (non-Tensor):
```
ply
format binary_little_endian 1.0
comment Created by Open3D
element vertex 168230
property double x
property double y
property double z
property uchar red
property uchar green
property uchar blue
end_header
```

Header when saving with `o3d.t.io.write_point_cloud` (Tensor):
```
ply
format binary_little_endian 1.0
comment Created by Open3D
element vertex 168230
property float32 x
property float32 y
property float32 z
property uint8 red
property uint8 green
property uint8 blue
end_header
```

reyanshsolis:
Closes #4548 

This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4553)

### DIFF
--- a/cpp/open3d/t/io/file_format/FilePLY.cpp
+++ b/cpp/open3d/t/io/file_format/FilePLY.cpp
@@ -127,23 +127,15 @@ static core::Dtype GetDtype(e_ply_type type) {
     // Currently, we are not doing datatype conversions, so some of the ply
     // datatypes are not included.
 
-    if (type == PLY_UINT8) {
+    if (type == PLY_UINT8 || type == PLY_UCHAR) {
         return core::UInt8;
     } else if (type == PLY_UINT16) {
         return core::UInt16;
-    } else if (type == PLY_INT32) {
+    } else if (type == PLY_INT32 || type == PLY_INT) {
         return core::Int32;
-    } else if (type == PLY_FLOAT32) {
+    } else if (type == PLY_FLOAT32 || type == PLY_FLOAT) {
         return core::Float32;
-    } else if (type == PLY_FLOAT64) {
-        return core::Float64;
-    } else if (type == PLY_UCHAR) {
-        return core::UInt8;
-    } else if (type == PLY_INT) {
-        return core::Int32;
-    } else if (type == PLY_FLOAT) {
-        return core::Float32;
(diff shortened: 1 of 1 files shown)


cpp/open3d/t/io/file_format/FilePLY.cpp
```diff
@@ -289,23 +281,19 @@ bool ReadPointCloudFromPLY(const std::string &filename,
...
-        return PLY_INT32;
+        return PLY_INT;
     } else if (dtype == core::Float32) {
         return PLY_FLOAT;
-    } else {
+    } else if (dtype == core::Float64) {
         return PLY_DOUBLE;
+    } else {
+        utility::LogError(
+                "Data-type {} is not supported in WritePointCloudToPLY.",
```
yxlao, line 295: Let's give more info. Tell the user the available options to convert to.

reyanshsolis: Done.

PR #4556, opened 2022-01-07
Add mesh to virtual point cloud to ray casting tutorial

### LINKED ISSUE #3561: ICP between reconstructed cloud model and mesh CAD model
I want to align a mesh CAD model in a reconstructed scene containing the model. 
One problem is that the CAD model is  a mesh containing a lot of internal not visible surfaces.
Another problem is creating a uniform cloud representation of the mesh CAD model.

I was thinking something like a virtual reconstruction of the CAD model. The output should be a uniform cloud representing only the visible surfaces of the CAD model. It is also important to be able the set view points(virtual sensor poses).
It is important that it is relative fast. Something like 1.000.000 points in a second or faster.

Any suggestions or examples on how to best align a mesh CAD model in a live sensor cloud and/or reconstructed scene using Open3D?

ssheorey:
Resolves #3561

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4556)

### DIFF
--- a/docs/jupyter/geometry/ray_casting.ipynb
+++ b/docs/jupyter/geometry/ray_casting.ipynb
@@ -3,6 +3,11 @@
   {
    "cell_type": "code",
    "execution_count": null,
+   "metadata": {
+    "nbsphinx": "hidden",
+    "tags": []
+   },
+   "outputs": [],
    "source": [
     "import open3d as o3d\n",
     "import numpy as np\n",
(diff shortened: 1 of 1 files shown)


docs/jupyter/geometry/ray_casting.ipynb
```diff
@@ -3,6 +3,11 @@
   {
```
benjaminum, line 3: <pre class="ql-syntax" spellcheck="false">Line #4.    # Press Ctrl/Cmd-C in the visualization window to copy the current viewpoint
</pre><p>This comment is meant for people writing the tutorial notebooks, right?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/isl-org/Open3D/pull/4556/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='29'/>

ssheorey: <p>Also for users, since they may wonder if the author arrived at these precise viewpoint numbers by trial and error. It's not related to this tutorial, but I think its useful in general since its a pretty convenient feature for comparing different results, etc. [We should add it to the new filament GUI as well]</p>

PR #4573, opened 2022-01-10
Replace `SHA256` with `MD5` Sum Verification

reyanshsolis:
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4573)


yxlao (approved):
LGTM after test passing.

PR #4613, opened 2022-01-18
Create inspection mode for O3D Visualizer

errissa:
Normal rendering:
2022-01-17-223326_1024x768_scrot

Inspection mode rendering:
2022-01-17-223336_1024x768_scrot

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4613)


yxlao:
Comment 1: I noticed some aliasing when visualizing lines:

image

Is it possible to enable anti-aliasing in Filament?

yxlao:
Comment 2: Seems like the standalone app does not have the inspection mode option. Can we add that to the standalone app as well? 

Question for future work: can we share more code for both the `draw()` and the standalone app, such that the behavior can be consistent?

yxlao:
Comment 3: We can consider renaming "Inspection Mode" to "Basic Mode". "Basic" seems more commonly used. It's not clear what "Inspection" means for the user.

cpp/open3d/visualization/visualizer/O3DVisualizer.cpp
```diff
@@ -1131,6 +1144,88 @@ struct O3DVisualizer::Impl {
...
+            for (auto &o : inspection_objects_) {
+                auto gtype = o.geometry->GetGeometryType();
+                CreateInspectionModeMaterial(
+                        o.material,
+                        gtype == geometry::Geometry::GeometryType::PointCloud);
+                if (o.geometry) {
+                    // Need to compute triangle normals for triangle meshes
+                    if (gtype ==
+                        geometry::Geometry::GeometryType::TriangleMesh) {
+                        auto tmesh = std::dynamic_pointer_cast<
```
yxlao, line 1191: It shall work if do:

```cpp
if (auto tmesh = std::dynamic_pointer_cast<geometry::TriangleMesh>(o.geometry)) { }
```

This way, we don't need to check `gtype`.

yxlao (approved):

errissa (commented):
Comment 1: Disabling post-processing for basic mode also disable anti-aliasing, unfortunately. There's no workaround to that problem.

Comment 2: Done. Added to standalone viewer. Yes, the plan after the scene graph is to fully combine the standalone and O3D viewers into a single program.

Comment 3: Done.

> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4613#-:-Mufd3oe642hmaPpsSVx:bb8d6k)* status: 0 of 10 files reviewed, 1 unresolved discussion (waiting on @yxlao)

___
*[cpp/open3d/visualization/visualizer/O3DVisualizer.cpp, line 1191 at r1](https://reviewable.io/reviews/isl-org/open3d/4613#-MtfF2wXK0qSz1H20wL1-r1-1191:-MttnUg22HGKo9cozhH5:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/2c9daeec3c7ebd34063adb4c16d6f62fd04048dd/cpp/open3d/visualization/visualizer/O3DVisualizer.cpp#L1191)):*

It shall work if do:

```cpp
if (auto tmesh = std::dynamic_pointer_cast<geometry::TriangleMesh>(o.geometry)) { }
```

This way, we don't need to check `gtype`.
</blockquote></details>

Done.

PR #4621, opened 2022-01-19
build openblas with cmake instead of gnu make

yxlao:
Advantages:
- We pass `ExternalProject_CMAKE_ARGS`, such that the cache info can be used by OpenBLAS build. According to my local test, all C compilations are cached. Fortran compilations are not cached, but they compile relatively fast. 
- Potential cross-platform support for Windows (not verified).

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4621)

### DIFF
--- a/3rdparty/openblas/openblas.cmake
+++ b/3rdparty/openblas/openblas.cmake
@@ -8,7 +8,7 @@ else()
     set(OPENBLAS_TARGET "NEHALEM")
 endif()
 
-set(OPENBLAS_INCLUDE_DIR "${OPENBLAS_INSTALL_PREFIX}/include/") # The "/"" is critical, see open3d_import_3rdparty_library.
+set(OPENBLAS_INCLUDE_DIR "${OPENBLAS_INSTALL_PREFIX}/include/openblas/") # The "/"" is critical, see open3d_import_3rdparty_library.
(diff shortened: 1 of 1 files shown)


3rdparty/openblas/openblas.cmake
```diff
@@ -8,7 +8,7 @@ else()
     set(OPENBLAS_TARGET "NEHALEM")
```
ssheorey, line 8: [nit/future] We should sync this with the ISPC arch. Also, is there a way to set this to the arch of the current machine? That will likely give a boost in benchmarks.

3rdparty/openblas/openblas.cmake
```diff
@@ -18,12 +18,10 @@ ExternalProject_Add(
     URL https://github.com/xianyi/OpenBLAS/archive/refs/tags/v0.3.18.tar.gz
     URL_HASH SHA256=1632c1e8cca62d8bed064b37747e331a1796fc46f688626337362bf0d16aeadb
```
ssheorey, line 19: [nit] We can update to v0.3.19. Looks like there are some bug fixes and support for newer Intel CPUs.

ssheorey (approved):

PR #4562, opened 2022-01-08
Update paths in python examples

theNded:
- Replace `..` with `os.path.dirname` for all the relative import and test data paths
- Fixed minor incompatibility issues

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4562)

PR #4645, opened 2022-01-23
Refactor python examples directory

ShreyanshDarshan:
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4645)


examples/python/visualization/tensorboard_tensorflow.py
```diff
@@ -33,8 +33,9 @@
 import tensorflow as tf
 
 BASE_LOGDIR = "demo_logs/tf/"
-MODEL_DIR = os.path.join(os.path.dirname(os.path.realpath(__file__)), '..',
```
yxlao, line 36: Just a comment: I thought the original one is more concise, what is the advantage of using `os.path.dername`? If we use `pathlib`, it will be called `parent`, which is also good. It's fine to keep it as it is. We'll do another round after the dataset class is in place.

examples/python/visualization/render_to_image.py
```diff
@@ -74,6 +74,7 @@ def main():
 
     img = render.render_to_image()
     o3d.io.write_image("/tmp/test.png", img, 9)
+    print("check")
```
yxlao, line 77: remove after debugging?

yxlao (changes requested):

yxlao:
Mostly looks good to me after we fix the line numbers.

examples/python/visualization/tensorboard_tensorflow.py
ShreyanshDarshan: I'm not really sure what the advantage is of using `os.path.dirname`. I was trying to keep it consistent with #4562.

examples/python/reconstruction_system/optimize_posegraph.py
```diff
@@ -27,12 +27,7 @@
 # examples/python/reconstruction_system/optimize_posegraph.py
```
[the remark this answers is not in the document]
ShreyanshDarshan: Done.

examples/python/visualization/render_to_image.py
ShreyanshDarshan: Done.

PR #4660, opened 2022-01-27
Change Linking of OpenGL so Open3D works with Mesa/llvmpipe

errissa:
This PR changes the link order of the Filament and OpenGL libraries in order to ensure that Open3D can be used for both GPU-accelerated and CPU-based visualization using Mesa's LLVMpipe. By default, Open3D will use whatever OpenGL implementation it finds first in the library search path which is typically the installed GPU driver's OpenGL implementation. In order to switch to LLVMpipe you can use one of the following three options:

1) Set the library path to point first to the Mesa LLVMpipe OpenGL library. For example:
```
 export LD_LIBRARY_PATH=<path to directory with Mesa LLVMpipe OpenGL library>:$LD_LIBRARY_PATH
```
or:
```
LD_PRELOAD=<path to Mesa LLVMpipe libGL.so> python3 examples/python/visualization/draw.py
```

This will force Open3D to use the Mesa LLVMpipe implementation for CPU rendering.

2) In Python it is possible to force Open3D to use Mesa LLVMpipe by loading its `libGL.so` _before_ importing Open3d:
```
import ctypes
ctypes.cdll.LoadLibrary('<path to Mesa LLVMpipe libGL.so>')
import open3d as o3d
# Now Open3D will use Mesa's CPU OpenGL implementation
```

3) On Ubuntu 20.04, if you are using a Mesa-based graphics driver (e.g., Intel GPU) you can switch between GPU and CPU rendering using the following environment variable settings:
```
LIBGL_ALWAYS_SOFTWARE=true
```
For example:
```
LIBGL_ALWAYS_SOFTWARE=true python3 examples/python/visualization/draw.py
```

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4660)


3rdparty/find_dependencies.cmake
```diff
@@ -1040,7 +1040,6 @@ else()
         TARGETS OpenGL::GL
     )
 endif()
-list(APPEND Open3D_3RDPARTY_PRIVATE_TARGETS Open3D::3rdparty_opengl)
```
ssheorey, line 1043: Does it work if we just move the `Open3D::3rdparty_opengl` in the list `Open3D_3RDPARTY_PRIVATE_TARGETS` to the beginning (or is that end)? This will avoid manually adding this to each example. Also, I think user code that uses Open3D may also have the same issue.

errissa: Fixed. Found way to force OpenGL to be last on link line without burden of linking explicitly to `Open3D::3rdparty_opengl`

ssheorey (commented):
@errissa Could you update the docs with this procedure for CPU rendering (including mesa version requirements >=20.3)?

Also, how do we get the Mesa `libGL.so`? Is it the one that comes in the Ubuntu `libgl1` package? That's already in the system path, so I'm not sure if adding it to the `LD_LIBRARY_PATH` will change anything.

errissa:
> @errissa Could you update the docs with this procedure for CPU rendering (including mesa version requirements >=20.3)?
> 
I will update the documentation in a separate PR.

errissa:
> @errissa Could you update the docs with this procedure for CPU rendering (including mesa version requirements >=20.3)?
> 
> Also, how do we get the Mesa `libGL.so`? Is it the one that comes in the Ubuntu `libgl1` package? That's already in the system path, so I'm not sure if adding it to the `LD_LIBRARY_PATH` will change anything.

Option 1 above refers to running Open3D with a custom built Mesa having followed the instructions [here](https://docs.mesa3d.org/drivers/llvmpipe.html#building). In order to run against a custom Mesa you would `LD_LIBRARY_PATH` to point to your custom built Mesa. I will be sure that is clear in the documentation.

On a system running a Mesa-based driver you are correct that Mesa's `libGL.so` is already in the system path. In that case you would use Option 3 to enable CPU rendering since the default would be for Mesa to dispatch the OpenGL calls through a hardware accelerated driver. Of course, that only works on Ubuntu 20.04 which has a sufficiently new version of Mesa to be compatible with Filament. On 18.04 the only option is to use Option 1 and build a custom Mesa.

ssheorey (approved):
@errissa Option 1 (setting `LD_LIBRARY_PATH`) did not work for me. It still used the GPU.  Option 2 (`LD_PRELOAD`) worked. (Ubuntu 20.04, using Intel GPU + Mesa).

errissa:
> @errissa, on Ubuntu 20.04, using Intel GPU + Mesa:
> 
>     * Option 1 (setting `LD_LIBRARY_PATH`) did not work for me. It still used the GPU.

I tested it again on Ubuntu 18.04 with Nvidia as a sanity check and it does work there. I will test with Intel GPU + Mesa. To be clear, you used the `libGL.so` that I sent you and made sure it was first in the `LD_LIBRARY_PATH`?

ssheorey:
> > @errissa, on Ubuntu 20.04, using Intel GPU + Mesa:
> > ```
> > * Option 1 (setting `LD_LIBRARY_PATH`) did not work for me. It still used the GPU.
> > ```
> 
> I tested it again on Ubuntu 18.04 with Nvidia as a sanity check and it does work there. I will test with Intel GPU + Mesa. To be clear, you used the `libGL.so` that I sent you and made sure it was first in the `LD_LIBRARY_PATH`?

Yes, that's correct:
```
$ ls /home/ssheorey/Downloads/mesa-llvmpipe-libgl
libGL.so
LD_LIBRARY_PATH='/home/ssheorey/Downloads/mesa-llvmpipe-libgl':$LD_LIBRARY_PATH python examples/python/visualization/draw.py
```

yxlao (approved):

Issue #4104, opened 2021-09-26
Additional Features for void open3d::t::geometry::TSDFVoxelGrid::Integrate	() function


TheRealRolandDeschain:
**Is your feature request related to a problem? Please describe.**
Instead of using an RGBD camera, my pipeline consists of an Android application using Unity + Googles AR-Core to create RGBD Data without any depth sensor (https://developers.google.com/ar/develop/unity-arf/depth/introduction). The data is sent via LAN to a PC running a C++ application utilizing Open3D. There are currently two major issues with this approach:

1. Android RGB camera has a much higher resolution compared to the depth map calculated (which is only 160x90 in my case). So at the moment I have to scale down the RGB image (obviously upscaling/interpolation of the depth image would also be possible)
2. The calculated Depth Data is quite noisy. The algorithm to integrate is able to smooth out quite a lot of the noise (slow-camera movement, correct TSDFVoxelGrid parameters,...) but it is far from perfect. 

**Describe the solution you'd like**
1. Allow different resolutions of RGB image and Depth image for the integration function.
2. In one of the newer AR-Core version, there is the possibility to get Raw Depth data in combination with a confidence map instead of the full depth image (info here: https://developers.google.com/ar/develop/unity-arf/depth/raw-depth). The option to use this data in the integration algorithm (discarding depth pixels with low confidence) would increase the accuracy of the algorithm quite a lot I guess.

theNded:
Okay, these tasks are doable. I will put them in my todo list.

pranav-asthana:
+1 on having depth confidence as an input to TSDF. Even for RGBD camera depths, it is useful to have this feature since sensors typically have lower confidence at higher depth values.

Is this expected to be available anytime soon?

PR #4663, opened 2022-01-28
Separate intrinsic matrix for RGB-D integration

theNded:
Addresses #4104. Allows color and depth of different resolutions and intrinsic matrices. 
Works on ScanNet with depth resolution 640x480 and color resolution 1296x968:
Screenshot from 2022-01-27 19-28-28

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4663)


reyanshsolis:
LGTM.

[discussion]
do you think `intrinsic` and `color_intrinsic` might confuse users?
an alternate design might be to take both `depth_intrinsic` and `color_intrinsic` and let the user pass the same intrinsic if they want to.

reyanshsolis (approved):

theNded:
> LGTM.
> 
> [discussion] do you think `intrinsic` and `color_intrinsic` might confuse users? an alternate design might be to take both `depth_intrinsic` and `color_intrinsic` and let the user pass the same intrinsic if they want to.
> 
> although the description is clear, so the current design is also fine.

I have made it clear in config helper so it should be fine. It also keeps the API.

salmanmaq:
Thanks for this feature! We need this and were quite eager for this to be implemented.

Issue #2645, opened 2020-11-20
How to use cv::Mat create Open3d Image in C++ API ?


joy-yjl:
I want to create a Open3D Image (open3d::geometry::Image) with cv::Mat (or unchar* ).And the image is 16 bit depth image.  I did not find the interface in the c++ api doc. Could somebody help me?

ssheorey:
Hi @joy-yjl you can use this to convert from OpenCV Mat to open3d::t::geometry::Image:
```c++
int main(int argc, char *argv[]) {
    fmt::print("Reading image {}\n", argv[1]);
    // Read an image without an conversion with OpenCV
    cv::Mat im_cv = cv::imread(argv[1], cv::IMREAD_UNCHANGED);
    if (im_cv.empty()) {
        open3d::utility::LogError("Could not open image!");
    }
    fmt::print("rows={}, cols={}, channels={}, depth={}\n", im_cv.rows,
               im_cv.cols, im_cv.channels(), im_cv.depth());
    // Convert OpenCV depth to Open3D Dtype
    open3d::core::Dtype dt;
    switch (im_cv.depth()) {
        case CV_8U:
            dt = open3d::core::Dtype::UInt8;
            break;
        case CV_16U:
            dt = open3d::core::Dtype::UInt16;
            break;
        case CV_32F:
            dt = open3d::core::Dtype::Float32;
            break;
        case CV_64F:
            dt = open3d::core::Dtype::Float64;
            break;
        default:
            open3d::utility::LogError("See docs for other element Dtypes.");
    }
    // Create a blob for externally managed memory with no-op deleter
    auto pblob = std::make_shared<open3d::core::Blob>(
            open3d::core::Device(), im_cv.data, [](void *) {});
    // Create tensor
    open3d::core::Tensor data_o3d(
            /*shape=*/{im_cv.rows, im_cv.cols, im_cv.channels()},
            /*stride in elements (not bytes)*/
            {int64_t(im_cv.step[0] / im_cv.elemSize1()),
             int64_t(im_cv.step[1] / im_cv.elemSize1()), 1},
            im_cv.data, dt, pblob);
    open3d::t::geometry::Image im_o3d(data_o3d);
    // If you need the Eigen based legacy Image:
    // open3d::geometry::Image im_o3d_e = im_o3d.ToLegacy();
    fmt::print("rows={}, cols={}, channels={}, dtype={}\n", im_o3d.GetRows(),
[...]

PR #4672, opened 2022-01-30
Tensor wrapper ctor for externally managed data

ssheorey:
This simplifies interfacing between T APIs and legacy APIs, as well as between Open3D and other libraries such as OpenCV (e.g.: #2645).

image

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4672)

### DIFF
--- a/cpp/open3d/core/FunctionTraits.h
+++ b/cpp/open3d/core/FunctionTraits.h
@@ -36,18 +36,19 @@ struct FunctionTraits : public FunctionTraits<decltype(&T::operator())> {};
 
 /// Pointers to class members that are themselves functors.
 /// For example, in the following code:
-/// template <typename func_t>
-/// struct S {
-///     func_t f;
-/// };
-/// template <typename func_t>
-/// S<func_t> make_s(func_t f) {
-///     return S<func_t> { .f = f };
-/// }
 ///
-/// auto s = make_s([] (int, float) -> double { /* ... */ });
+///     template <typename func_t>
+///     struct S {
+///         func_t f;
+///     };
+///     template <typename func_t>
+///     S<func_t> make_s(func_t f) {
+///         return S<func_t> { .f = f };
+///     }
 ///
-/// FunctionTraits<decltype(&s::f)> traits;
+///     auto s = make_s([] (int, float) -> double { /* ... */ });
+///
+///     FunctionTraits<decltype(&s::f)> traits;
 template <typename ClassType, typename T>
 struct FunctionTraits<T ClassType::*> : public FunctionTraits<T> {};
 
--- a/cpp/open3d/core/Tensor.h
+++ b/cpp/open3d/core/Tensor.h
@@ -121,6 +121,21 @@ public:
           dtype_(dtype),
           blob_(blob) {}
 
+    /// Constructor from raw host buffer. This creates a Tensor wrapper for
+    /// externally managed memory.
+    Tensor(void* data_ptr,
+           Dtype dtype,
+           const SizeVector& shape,
+           const SizeVector& strides = {},
+           const Device& device = Device("CPU:0"))
+        : shape_(shape), strides_(strides), data_ptr_(data_ptr), dtype_(dtype) {
+        if (strides_.empty()) {
+            strides_ = shape_util::DefaultStrides(shape);
+        }
+        // blob with no-op deleter
+        blob_ = std::make_shared<Blob>(device, (void*)data_ptr_, [](void*) {});
+    }
+
     /// Copy constructor performs a "shallow" copy of the Tensor.
     /// This takes a lvalue input, e.g. `Tensor dst(src)`.
     Tensor(const Tensor& other) = default;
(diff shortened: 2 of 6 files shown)


yxlao (changes requested):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4672)* status: all files reviewed, 4 unresolved discussions (waiting on @ssheorey)

___
*[cpp/open3d/core/Tensor.h, line 125 at r1](https://reviewable.io/reviews/isl-org/open3d/4672#-MuhxPciBKT6OZn5v0dw:-Muhvt8f4IpsRH71W-BS:b-i0jisw) ([raw file](https://github.com/isl-org/open3d/blob/49204388fbfd00f0b46515c156b242166f746a11/cpp/open3d/core/Tensor.h#L125)):*
> ```objc
> 
>     /// Constructor from raw host buffer. This creates a Tensor wrapper for
>     /// externally managed memory.
> ```

In general, it is possible that `blob_data_ptr` is not the same as the tensor's `data_ptr`, e.g. when a stride is negative; or when Tensor B = A.Slice(xxx), B and A will share the same blob but they have different `data_ptr`. 

Here, we are making the assumption that `data_ptr` is the same as `blob_data_ptr`.  It shall be fine in this use case. I would:

- mention this assumption in the docstring and inform the user it is their responsibility to keep the data valid during the lifetime of the tensor
- add docstring for every argument
- emphasize that the strides are per-element, not per-byte
___
*[cpp/open3d/core/Tensor.h, line 135 at r1](https://reviewable.io/reviews/isl-org/open3d/4672#-Muhx932CxpoS27zWPCP:-Muhx932CxpoS27zWPCQ:bmwbkvj) ([raw file](https://github.com/isl-org/open3d/blob/49204388fbfd00f0b46515c156b242166f746a11/cpp/open3d/core/Tensor.h#L135)):*
> ```objc
>             strides_ = shape_util::DefaultStrides(shape);
>         }
>         // blob with no-op deleter
> ```

nit: `blob` -> `Blob`
___
*[cpp/tests/core/Tensor.cpp, line 117 at r1](https://reviewable.io/reviews/isl-org/open3d/4672#-Muhx1R2-c-4wFQ8GBJQ:-Muhx1R2-c-4wFQ8GBJR:b-tliw93) ([raw file](https://github.com/isl-org/open3d/blob/49204388fbfd00f0b46515c156b242166f746a11/cpp/tests/core/Tensor.cpp#L117)):*
> ```cpp
>         EXPECT_EQ(wrapper.ToFlatVector<float>(), vals);
>         // Updating original data updates wrapper
>         t.SetItem({core::TensorKey::Index(1), core::TensorKey::Index(1)},
> ```

Simplify to: `t[1][1] = 0;`
___
*[cpp/tests/core/Tensor.cpp, line 122 at r1](https://reviewable.io/reviews/isl-org/open3d/4672#-Muhx5_tEmY1KJs6JcAA:-Muhx5_tEmY1KJs6JcAB:bwj1iiv) ([raw file](https://github.com/isl-org/open3d/blob/49204388fbfd00f0b46515c156b242166f746a11/cpp/tests/core/Tensor.cpp#L122)):*
> ```cpp
>         EXPECT_EQ(wrapper.ToFlatVector<float>(), vals);
>     }
>     // original data is present after wrapper is destructed.
> ```

nit: `original` -> `Original`

ssheorey (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4672)* status: 0 of 6 files reviewed, 3 unresolved discussions (waiting on @ssheorey and @yxlao)

___
*[cpp/open3d/core/Tensor.h, line 125 at r1](https://reviewable.io/reviews/isl-org/open3d/4672#-MuhxPciBKT6OZn5v0dw:-MuiHtrd3_FOwZ4Jj91Z:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/49204388fbfd00f0b46515c156b242166f746a11/cpp/open3d/core/Tensor.h#L125)):*

In general, it is possible that `blob_data_ptr` is not the same as the tensor's `data_ptr`, e.g. when a stride is negative; or when Tensor B = A.Slice(xxx), B and A will share the same blob but they have different `data_ptr`. 

Here, we are making the assumption that `data_ptr` is the same as `blob_data_ptr`.  It shall be fine in this use case. I would:

- mention this assumption in the docstring and inform the user it is their responsibility to keep the data valid during the lifetime of the tensor
- add docstring for every argument
- emphasize that the strides are per-element, not per-byte
</blockquote></details>

Done.
___
*[cpp/tests/core/Tensor.cpp, line 117 at r1](https://reviewable.io/reviews/isl-org/open3d/4672#-Muhx1R2-c-4wFQ8GBJQ:-MuiIG3S5ICt4duN4nwU:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/49204388fbfd00f0b46515c156b242166f746a11/cpp/tests/core/Tensor.cpp#L117)):*

Simplify to: `t[1][1] = 0;`
</blockquote></details>

Done.

yxlao (approved):

PR #4677, opened 2022-01-31
Update to ASSIMP 5.1.6

errissa:
Updates to ASSIMP 5.1.6 which incorporates our OBJ PBR patches so the patching step is no longer necessary.

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4677)


yxlao (approved):
Tested on ARM macs. LGTM after CI.

PR #4680, opened 2022-02-01
Fix Crash With Legacy Visualizer Examples

errissa:
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4680)


yxlao (approved):
Tested on Apple ARM.

PR #4684, opened 2022-02-02
automatically cancel concurrent workflows

yxlao:
- This works on PR
- This shall also work on master, as `github.event.pull_request.number` is undefined, and we'll fall back to `github.ref`
- `github.workflow` is needed since we have multiple workflows PR

Ref:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency
https://github.com/TeamAmaze/AmazeFileManager/blob/release/3.7/.github/workflows/android-feature.yml

This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4684)

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -14,6 +14,10 @@ on:
   pull_request:
     types: [opened, reopened, synchronize]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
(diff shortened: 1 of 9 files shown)


yxlao:
image

yxlao:
image

yxlao:
Seems to work well!

ssheorey (approved):
Looks good. Thanks for adding this in!

PR #4685, opened 2022-02-02
Fix sample code in docstring for tensor pc, mesh, & lineset

ceroytres:
- The examples in o3d.t.geometry [PointCloud](http://www.open3d.org/docs/release/python_api/open3d.t.geometry.PointCloud.html#open3d.t.geometry.PointCloud), [LineSet](http://www.open3d.org/docs/release/python_api/open3d.t.geometry.LineSet.html#open3d.t.geometry.LineSet), [TriangleMesh](http://www.open3d.org/docs/release/python_api/open3d.t.geometry.TriangleMesh.html#open3d.t.geometry.TriangleMesh) wrote `o3c.core` instead of `o3d.core`.

- Also it seems that in the example in PointCloud the comment seems to suggest:
```python
# Common attributes: "normals", "colors".
# Common attributes are used in built-in point cloud operations. The
# spellings must be correct. For example, if "normal" is used instead of
# "normals", some internal operations that expects "normals" will not work.
# "normals" and "colors" must have shape (N, 3) and must be on the same
# device as the point cloud.
pcd.point["normals"] = o3d.core.Tensor([[0, 0, 1],
                                        [0, 1, 0],
                                        [1, 0, 0]], dtype, device)
pcd.point["colors"] = o3d.core.Tensor([[0.0, 0.0, 0.0],
                                        [0.1, 0.1, 0.1],
                                        [0.2, 0.2, 0.2]], dtype, device)
```
instead of:
```python
pcd.point["normals"] = o3d.core.Tensor([[0, 0, 1],
                                        [0, 1, 0],
                                        [1, 0, 0]], dtype, device)
pcd.point["normals"] = o3d.core.Tensor([[0.0, 0.0, 0.0],
                                        [0.1, 0.1, 0.1],
                                        [0.2, 0.2, 0.2]], dtype, device)
```
- Also label is misspelled on the PointCloud example

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4685)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 * Python 3.9 support. Tensorflow bump 2.4.1 -> 2.5.0. PyTorch bump 1.7.1 -> 1.8.1 (LTS)
 * Fix undefined names: docstr and VisibleDeprecationWarning (PR #3844)
+* Corrected documentation for Tensor based PointClound, LineSet, TriangleMesh (PR #4685)
 
 ## 0.13
 
--- a/cpp/pybind/t/geometry/lineset.cpp
+++ b/cpp/pybind/t/geometry/lineset.cpp
@@ -77,7 +77,7 @@ The attributes of the line set have different levels::
     # "color", some internal operations that expects "colors" will not work.
     # "colors" must have shape (N, 3) and must be on the same device as the
     # line set.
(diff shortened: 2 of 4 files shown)


ssheorey (approved):
Thanks for fixing the docstring, @ceroytres !

PR #4692, opened 2022-02-03
use fmt library in compiled (non-header-only) mode

yxlao:


errissa (changes requested):
This PR introduces another 3rdparty library configuration and build step which (slightly) increases the Open3D build time. Given that `fmt` is available as header-only I think there has to be a good reason to prefer building it. Is there no alternative to avoiding the `windows.h` dependency?

> Reviewed 4 of 8 files at r1, 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4692#-:-Mv-ZN3k5GMC7GUru79y:b-hpqw2c)* status: all files reviewed, 3 unresolved discussions (waiting on @yxlao)

___
*[cpp/open3d/core/SizeVector.cpp, line 29 at r2](https://reviewable.io/reviews/isl-org/open3d/4692#-Mv-ZH8p9ewfJBSc_rxA:-Mv-ZH8q8N37HWhNAjhR:bny1btx) ([raw file](https://github.com/isl-org/open3d/blob/6792a00a73bf34f7be7ef0b4a938b6ec4d4f8324/cpp/open3d/core/SizeVector.cpp#L29)):*
> ```cpp
> #include "open3d/core/SizeVector.h"
> 
> #include <algorithm>
> ```

why did this become necessary?
___
*[cpp/open3d/core/Tensor.h, line 29 at r2](https://reviewable.io/reviews/isl-org/open3d/4692#-Mv-ZJJwD7Q4XKQ6AP7X:-Mv-ZJJwD7Q4XKQ6AP7Y:bu6xv6u) ([raw file](https://github.com/isl-org/open3d/blob/6792a00a73bf34f7be7ef0b4a938b6ec4d4f8324/cpp/open3d/core/Tensor.h#L29)):*
> ```objc
> #pragma once
> 
> #include <algorithm>
> ```

See previous comment.
___
*[cpp/open3d/utility/Logging.cpp, line 31 at r2](https://reviewable.io/reviews/isl-org/open3d/4692#-Mv-YvnM6oVMzcqhbjBb:-Mv-YvnN7DbE2B5f5yna:bc3mu4p) ([raw file](https://github.com/isl-org/open3d/blob/6792a00a73bf34f7be7ef0b4a938b6ec4d4f8324/cpp/open3d/utility/Logging.cpp#L31)):*
> ```cpp
> #include <fmt/core.h>
> #include <fmt/printf.h>
> #include <fmt/ranges.h>
> ```

Aren't these redundant? They are included in `Logging.h`

yxlao (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4692)* status: all files reviewed, 3 unresolved discussions (waiting on @errissa)

___
*[cpp/open3d/core/SizeVector.cpp, line 29 at r2](https://reviewable.io/reviews/isl-org/open3d/4692#-Mv-ZH8p9ewfJBSc_rxA:-Mv-aDIBCU5ay54rkvWB:bmqhiep) ([raw file](https://github.com/isl-org/open3d/blob/6792a00a73bf34f7be7ef0b4a938b6ec4d4f8324/cpp/open3d/core/SizeVector.cpp#L29)):*
During debugging, when I removed all `fmt` headers, the compiler complains about missing the `<algorithm>` header here.

`<algorithm>` is brought in by the `fmt` headers. If `Logging.h` does not use `fmt` headers at all, we will need to include `<algorithm>` here. 

I was following this documentation: https://google.github.io/styleguide/cppguide.html#Include_What_You_Use:
> Do not rely on transitive inclusions. This allows people to remove no-longer-needed #include statements from their headers without breaking clients. This also applies to related headers - foo.cc should include bar.h if it uses a symbol from it even if foo.h includes bar.h.
___
*[cpp/open3d/core/Tensor.h, line 29 at r2](https://reviewable.io/reviews/isl-org/open3d/4692#-Mv-ZJJwD7Q4XKQ6AP7X:-Mv-blZJ3l-kHlR4BrBa:bbfhety) ([raw file](https://github.com/isl-org/open3d/blob/6792a00a73bf34f7be7ef0b4a938b6ec4d4f8324/cpp/open3d/core/Tensor.h#L29)):*

See previous comment.
</blockquote></details>

Same as above. 
___
*[cpp/open3d/utility/Logging.cpp, line 31 at r2](https://reviewable.io/reviews/isl-org/open3d/4692#-Mv-YvnM6oVMzcqhbjBb:-Mv-bmpn2MBs8hS5IZHk:b-plyua) ([raw file](https://github.com/isl-org/open3d/blob/6792a00a73bf34f7be7ef0b4a938b6ec4d4f8324/cpp/open3d/utility/Logging.cpp#L31)):*
Same as above. 

In https://google.github.io/styleguide/cppguide.html#Include_What_You_Use :
> This also applies to related headers - foo.cc should include bar.h if it uses a symbol from it even if foo.h includes bar.h.

errissa (approved):
> Reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4692)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @yxlao)

JackBoosY:
To anyone who met this issue again:
I was posted this issue in nvidia side: https://developer.nvidia.com/nvidia_bug/3597169
And the member told me this issue was fixed by https://github.com/NVIDIA/cub/pull/423, it should be fine in cub 1.16.0.

PR #4697, opened 2022-02-03
Raw mode fixes

errissa:
- Renames "basic" mode to "raw" mode.
- Starts visualizer with sun-follows-camera mode by default
- Fixes various UI inconsistencies related to raw mode

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4697)


yxlao:
notes from the sync:
1. back side is dark for point cloud
2. loading a new model when normal is on
3. support switching among lit, unlit, normal, depth in raw mode
4. under raw mode with unlit, point cloud shall show the original color (compare that with the old visualizer)

yxlao:
Checklist:
- [x] Raw mode + unlit: point cloud shall show its original color
- [x] Load a new geometry while in raw mode
- [ ] When HDR map is off, the backside is black

yxlao (approved):

PR #4698, opened 2022-02-04
Enable C++17

junha-l:
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4698)

PR #4704, opened 2022-02-04
Enforce CXX14, print cxx standard and compiler id/version

yxlao:
This is printed every time we run unit tests or benchmarks:

```
make tests -j && ./bin/tests --gtest_filter=""
[Open3D INFO] CompilerInfo: C++14, clang 10.0.0, nvcc 11.6.55.
[Open3D INFO] CPUInfo: 18 cores, 36 threads.
[Open3D INFO] ISAInfo: DISABLED instruction set used in ISPC code.
```

This is needed to verify https://github.com/isl-org/Open3D/pull/4698. In the future, it could also be helpful for detecting DPC++.

BTW, after the Faiss removal, clang-10 + cuda-11 works under ubuntu. This is the first time that we can compile Open3D with Clang and CUDA.

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4704)


ssheorey (commented):
We are repeating work that CMake has already done for us. We can get this info from CMake, and add compiler definitions while building the code directly:
e.g.
```cmake
get_property(CxxStandard  VARIABLE PROPERTY CMAKE_CXX_STANDARD)
add_compile_definitions(OPEN3D_CXX_STANDARD=${CxxStandard})
```
Other CMake properties: 
```
CMAKE_CXX_COMPILER_ID
CMAKE_CXX_COMPILER_VERSION
CMAKE_CUDA_COMPILER_ID
CMAKE_CUDA_COMPILER_VERSION
```

yxlao:
> We are repeating work that CMake has already done for us. We can get this info from CMake, and add compiler definitions while building the code directly: e.g.
> 
> ```cmake
> add_compile_definitions(OPEN3D_CXX_STANDARD=${CMAKE_CXX_STANDARD})
> ```
> 
> Other CMake properties:
> 
> ```
> CMAKE_CXX_COMPILER_ID
> CMAKE_CXX_COMPILER_VERSION
> CMAKE_CUDA_COMPILER_ID
> CMAKE_CUDA_COMPILER_VERSION
> ```

Yes, we can do that check as well. 

The main motivation for this came from the NNS meeting and #4698, in order to check CXX standard version. In CMake, there are multiple ways to set CXX standards: e.g. 1) set `CXX_STANDARD` for targets, 2) set `CMAKE_CXX_STANDARD` globally, and 3) directly set compiler flags. We could be setting a global flag in one way, but some specific target in another way. It was unclear to us which one to use. This PR serves as verification for #4698. 

It's true that the compiler ID and version are less of a problem. But still, the same argument can be made for `CMAKE_CXX_COMPILER_ID`. For instance, even though we set `CMAKE_CXX_COMPILER_ID` globally, some specific targets (e.g. Filament) might be compiled with a different compiler. The check inside the source code makes sure that at least that particular source code is indeed compiled with this compiler.

yxlao (commented):
Updated according to comments. The new approach now uses cmake to get compiler info. 

> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4704#-:-MvNHwfi0JDzBqhIXqRg:b-luey63)* status: 0 of 10 files reviewed, all discussions resolved

yxlao:
I've added a comment in the code. Currently, it is just used for consistency with `utility::CPUInfo`. We'll revisit this later.

PR #4708, opened 2022-02-05
Add instructions to preview built docs

ShreyanshDarshan:
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4708)


yxlao (approved):
I also added similar contents in #4704, you may merge the two together after merging.

PR #4721, opened 2022-02-08
Check in install_deps_ubuntu.sh if the variable APT_CONFIRM is set outside the script

meyerjo:
Checking whether ENV Variable APT_CONFIRM is set outside the script and if so we do not check the assume-yes argument

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4721)


ssheorey:
@meyerjo what's your use case? You can simply call `install_deps_ubuntu.sh asume-yes` instead of setting `APT_CONFIRM` outside. Why is this change needed?

Issue #4725, opened 2022-02-08
Using visualization.Visualizer().add_geometry(None) results a crash without error message


soyszala:
### Checklist

- [X] I have searched for [similar issues](https://github.com/isl-org/Open3D/issues).
- [X] For Python issues, I have tested with the [latest development wheel](http://www.open3d.org/docs/latest/getting_started.html#development-version-pip).
- [X] I have checked the [release documentation](http://www.open3d.org/docs/release/) and the [latest documentation](http://www.open3d.org/docs/latest/) (for `master` branch).

### Describe the issue

I was doing some geometric calculations and tried to render various geometries in the process.

One mistake caused one of my functions to accidentally return a "None" and it gave that to the "add_geometry()" function. And this caused the Visualizer window to show a black screen and then instantly disappear without any message (form Visual Studio Code). Running from the terminal it also resulted a black screen, what just hanged there forever.

I think this can result difficulties in finding the bug if you didn't pay attention. I would expect a warning or an error message. So I thought I will post this issue here.

### Steps to reproduce the bug

```python
import open3d as o3d

vis = o3d.visualization.Visualizer()
vis.create_window(width=1920, height=1080)

# Creating a sphere or whatever, this is just here,
# so I (you) can see if it works beforehand
mesh_sphere = o3d.geometry.TriangleMesh.create_sphere(1)
mesh_sphere.compute_vertex_normals()
mesh_sphere.paint_uniform_color((1, 0, 0))

vis.add_geometry(mesh_sphere)

# The problematic line
vis.add_geometry(None)

vis.run()
vis.destroy_window()
```

### Open3D, Python and System information

```markdown
- Operating system: Ubuntu 18.04
- Python version: Python 3.6.9
- Open3D version: 0.14.1
```

PR #4728, opened 2022-02-08
Check for invalid geometry input to DrawGeometries

errissa:
Addresses issue #4725

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4728)


ssheorey (approved):
Thanks for the quickfix @errissa !

PR #4739, opened 2022-02-09
Add dataset functionality to cpp unit tests and benchmarks

ShreyanshDarshan:
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4739)


reyanshsolis:
It would be cleaner to put data in a different data_root for Unit tests, and benchmarks, and cleanup the same inside the test.

PR #4748, opened 2022-02-11
Skip building cpp examples on Documentation CI

reyanshsolis:
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4748)


ssheorey (approved):
Thanks @reyanshsolis for figuring this out and making the change!

Issue #4759, opened 2022-02-14
How to reset a VoxelBlockGrid?


martinakos:
### Checklist

- [X] I have searched for [similar issues](https://github.com/isl-org/Open3D/issues).
- [X] For Python issues, I have tested with the [latest development wheel](http://www.open3d.org/docs/latest/getting_started.html#development-version-pip).
- [X] I have checked the [release documentation](http://www.open3d.org/docs/release/) and the [latest documentation](http://www.open3d.org/docs/latest/) (for `master` branch).

### My Question

I've been using the new VoxelBlockGrid for a reconstruction system. It works well.  But  I'm not sure what's the best way to reset it?  ScalableTSDFVolume or UniformTSDFVolume have a reset() function to do that but VoxelBlockGrid doesn't.

In my system I'm repeatedly reseting the TSDF because I'm just interested in the integration of the last few frames. In fact what I'd like is to be able to deintegrate pointclouds so as to have a sort of FIFO integration of the last N pointclouds. I see this was requested in #2613 but I don't see it's available yet. So as a compromise I reset the TSDF periodically. 

For the moment I'm recreating the VoxelBlockGrid every time. Any better way of doing this?

theNded:
Thanks for the feedback, I will develop this feature. Please stay tuned.

PR #4774, opened 2022-02-17
VoxelBlockGrid update

theNded:
- Provide `Reset` to address #4759 
- Provide `Prune` to recycle outlier blocks
- Minor parameter adjustment for loading to avoid stdgpu hashmap overflow

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4774)


yxlao (approved):

yxlao:
LGTM after fixing unit tests.

theNded:
Unit tests failed on CPU with ISPC turned on.
Reason: when converting the bool result of element-wise boolean operations (`eq`, `le`, `gt`, etc) to scalars, `True` becomes `255` or equivalent while the expected result should be `1`. Awaiting fix before merging.

yxlao:
```bash
python -c "import open3d as o3d; print(o3d.core.Tensor.ones((3), o3d.core.int32).le(1).to(o3d.core.int32))"
```

When computing `Tensor::Le()` with ISPC, ISPC assigns `0b11111111` for the `true` values. If we initialize a fresh tensor, the `true` value will always be `0b00000001`.

reyanshsolis:
> If we initialize a fresh tensor, the true value will always be 0b00000001.

Is this even the case when using ISPC for initializing a new tensor?

ssheorey:
Hi @theNded , do you have an update on this PR?

theNded:
Let me close it for now and reopen after some upgrades.

PR #4779, opened 2022-02-19
privately link fmt, but propagate its header

yxlao:
This issue was introduced in #4692.

### After installation
```bash
➜  ~ tree open3d_install/include/open3d/3rdparty/fmt  
open3d_install/include/open3d/3rdparty/fmt
├── chrono.h
├── color.h
├── compile.h
├── core.h
├── format.h
├── format-inl.h
├── locale.h
├── ostream.h
├── posix.h
├── printf.h
├── ranges.h
└── safe-duration-cast.h

0 directories, 12 files
➜  ~ tree open3d_install/lib                        
open3d_install/lib
├── cmake
│   └── Open3D
│       ├── Open3DConfig.cmake
│       ├── Open3DConfigVersion.cmake
│       ├── Open3DTargets.cmake
│       └── Open3DTargets-relwithdebinfo.cmake
└── libOpen3D.so

2 directories, 5 files
```

This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4779)

### DIFF
--- a/3rdparty/find_dependencies.cmake
+++ b/3rdparty/find_dependencies.cmake
@@ -994,7 +994,6 @@ if(USE_SYSTEM_FMT)
     endif()
 endif()
 if(NOT USE_SYSTEM_FMT)
-    # We set the FMT_HEADER_ONLY macro, so no need to actually compile the source
     include(${Open3D_3RDPARTY_DIR}/fmt/fmt.cmake)
     open3d_import_3rdparty_library(3rdparty_fmt
         HEADER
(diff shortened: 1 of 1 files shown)


3rdparty/find_dependencies.cmake
```diff
@@ -997,7 +997,7 @@ if(NOT USE_SYSTEM_FMT)
     # We set the FMT_HEADER_ONLY macro, so no need to actually compile the source
```
ssheorey, line 997: This comment can be removed now, since we compile `fmt` now.

ssheorey (approved):

3rdparty/find_dependencies.cmake
yxlao: Done,

PR #4784, opened 2022-02-20
auto-clean google cloud vm

yxlao:
### Setup

Auto-clean is triggered at 10 hours (36000s). The timer starts when the instance is started.

The instance is stopped and deleted after the timeout.

### Experiments

[Setting time off to 5 mins (300s)](https://github.com/isl-org/Open3D/pull/4784/commits/6896f59a50a1c28b82e27b25c27df4df57dc96a5). Works as expected.

image
image

### Reference
https://cloud.google.com/community/tutorials/create-a-self-deleting-virtual-machine

This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4784)


ssheorey (approved):
Looks good. Thanks for implementing this.
May be simplified since we already know the `INSTANCE_ZONE` and `INSTANCE_NAME` during launch, so querying from Google is not necessary.

PR #4785, opened 2022-02-21
Ensure entire macOS CI workflow has access to DEVELOPER_BUILD env var

ssheorey:
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4785)


yxlao (approved):

.github/workflows/macos.yml
```diff
@@ -67,7 +68,7 @@ jobs:
           DEVEL_PKG_NAME="$(basename package/open3d-devel-*.tar.xz)"
           echo "DEVEL_PKG_NAME=$DEVEL_PKG_NAME" >> $GITHUB_ENV
       - name: Build Open3D viewer app
-        if: ${{ matrix.CONFIG == 'OFF' }}
+        if: ${{ matrix.SHARED == 'OFF' }}
```
yxlao, line 71: @ssheorey seems like we only have `matrix.CONFIG` and `env.SHARED`. Did you mean `env.SHARED`?

yxlao (changes requested):

.github/workflows/macos.yml
ssheorey: Good catch. Done.

yxlao (approved):

PR #4795, opened 2022-02-22
Update example to load triangle meshes as models

errissa:
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4795)


ssheorey (approved):
Thanks, looks good. Tested on a few sampled meshes with textures.

PR #4819, opened 2022-02-28
Corrected documentation for KeyEvent

equant:
Looks like the KeyEvent was copy/pasted from MouseEvent and the doc string wasn't changed.  I believe I've fixed that issue with this PR.

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4819)


yxlao (approved):

yxlao:
Thanks @equant !

PR #4825, opened 2022-03-01
Dataset and Examples Update

reyanshsolis:
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4825)


reyanshsolis:
Done:
- [x] Add datasets, modify unit-tests and examples to use dataset module. 
- [x] Delete `test_data` dir and associated utility functions.

TODO:
- [ ] Add docs for new datasets.
- [ ] Add missing pybinds and py unit tests.
- [ ] Improve docs for `RGBD` datasets, to add `Camera Intrinsic IO`.
- [ ] Improve `TICPSequential` example, and other cpp examples.

python/test/data/test_data.py
```diff
@@ -100,6 +100,10 @@ def test_demo_icp_pointclouds():
         assert Path(gt_path) == Path(demo_icp_path)
         assert Path(gt_path).exists()
 
+    assert Path(demo_icp.transformation_log_path) == Path(gt_extract_dir /
+                                                          "init.log")
+    assert Path(demo_icp.transformation_log_path).exists()
```
yxlao, line 105: `is_file()` and `is_dir()` is better than `exists()`

cpp/tests/data/Dataset.cpp
```diff
@@ -330,6 +335,38 @@ TEST(Dataset, PLYPointCloud) {
     utility::filesystem::DeleteDirectory(extract_dir);
 }
 
+TEST(Dataset, PTSPointCloud) {
+    const std::string prefix = "PTSPointCloud";
+    const std::string data_root = utility::filesystem::GetHomeDirectory() +
```
yxlao, line 340: Future work: use https://en.cppreference.com/w/cpp/filesystem/temp_directory_path

yxlao (approved):

cpp/tests/data/Dataset.cpp
yxlao: Actually, what about something different: don't create an explicit test directory, but allow local unit tests to have the files cached **If the downloaded file belongs to one of the standard datasets**. 

- On the local machine: cached download files can make the unit test run faster.
- On the remote machine: CI will always start fresh. The download will always be tested.

PR #4863, opened 2022-03-08
Multithread Plane Segmentation

yuecideng:
I implement a multi-thread version of `SegmentPlane` function in `PointCloud`. I also compare the time cost with original implementation using the python example `point_cloud_plane_segmentation.py`, and the results are as follow:

cpu info: intel@core i7-1165G7 2.8Gz 8 threads 
point size: 113662
original time cost: 0.95s
multi-thread version: 0.16s

**Update**:
I use this equation: `iter = log(1 - probability) / log(1 - fitness ^ minimal_sample)` to update the number of iteration, such that the best result can be found without taken up to `num_iterations`. Then the time cost is reduced to about 0.03-0.05s.

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4863)

### DIFF
--- a/cpp/open3d/geometry/PointCloud.h
+++ b/cpp/open3d/geometry/PointCloud.h
@@ -343,6 +343,7 @@ public:
     /// \param ransac_n Number of initial points to be considered inliers in
     /// each iteration.
     /// \param num_iterations Number of iterations.
+    /// \param min_iterations Minimum number of iterations to be executed.
     /// \param seed Sets the seed value used in the random
     /// generator, set to nullopt to use a random seed value with each function
     /// call.
@@ -352,6 +353,7 @@ public:
             const double distance_threshold = 0.01,
             const int ransac_n = 3,
             const int num_iterations = 100,
+            const int min_iterations = 1,
             utility::optional<int> seed = utility::nullopt) const;
 
     /// \brief Factory function to create a pointcloud from a depth image and a
--- a/cpp/open3d/geometry/PointCloud.h
+++ b/cpp/open3d/geometry/PointCloud.h
@@ -342,8 +342,8 @@ public:
     /// model, and still be considered an inlier.
     /// \param ransac_n Number of initial points to be considered inliers in
     /// each iteration.
-    /// \param num_iterations Number of iterations.
-    /// \param min_iterations Minimum number of iterations to be executed.
+    /// \param num_iterations Maximum number of iterations.
+    /// \param probability Expected probability of finding the optimal plane.
     /// \param seed Sets the seed value used in the random
     /// generator, set to nullopt to use a random seed value with each function
     /// call.
@@ -353,7 +353,7 @@ public:
             const double distance_threshold = 0.01,
             const int ransac_n = 3,
             const int num_iterations = 100,
-            const int min_iterations = 1,
+            const double probability = 0.99999999,
             utility::optional<int> seed = utility::nullopt) const;
 
     /// \brief Factory function to create a pointcloud from a depth image and a
--- a/cpp/open3d/geometry/PointCloudSegmentation.cpp
+++ b/cpp/open3d/geometry/PointCloudSegmentation.cpp
@@ -138,15 +138,9 @@ std::tuple<Eigen::Vector4d, std::vector<size_t>> PointCloud::SegmentPlane(
         const int num_iterations /* = 100 */,
         utility::optional<int> seed /* = utility::nullopt */) const {
     RANSACResult result;
-    double error = 0;
 
-    // Initialize the plane model ax + by + cz + d = 0.
-    Eigen::Vector4d plane_model = Eigen::Vector4d(0, 0, 0, 0);
     // Initialize the best plane model.
     Eigen::Vector4d best_plane_model = Eigen::Vector4d(0, 0, 0, 0);
-
-    // Initialize consensus set.
-    std::vector<size_t> inliers;
 
     size_t num_points = points_.size();
     std::vector<size_t> indices(num_points);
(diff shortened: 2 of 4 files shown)


benjaminum (changes requested):
@yuecideng Thank you for this PR. Please see the code comments.

> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4863#-:-MyHQLs90XBQnKohqVXr:b-ajgh19)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @theNded and @yuecideng)

___
*[cpp/open3d/geometry/PointCloudSegmentation.cpp, line 138 at r2](https://reviewable.io/reviews/isl-org/open3d/4863#-MyHOxO83HjaiKmSMFGO:-MyHOxO83HjaiKmSMFGP:b-ffywks) ([raw file](https://github.com/isl-org/open3d/blob/8045725eb050cd1fc509bdba622e52b4b5c682d5/cpp/open3d/geometry/PointCloudSegmentation.cpp#L138)):*
> ```cpp
>         const double distance_threshold /* = 0.01 */,
>         const int ransac_n /* = 3 */,
>         const int num_iterations /* = 100 */,
> ```

I wonder if there is an easy and backwards compatible way to redefine the signature to allow users to get the old behavior that always runs for `num_iterations`.
Maybe it makes sense to add an additional `min_iterations` argument, which users can set to the same value. I will need to get more feedback first before changing the signature.
What are your thoughts on this?
___
*[cpp/open3d/geometry/PointCloudSegmentation.cpp, line 176 at r2](https://reviewable.io/reviews/isl-org/open3d/4863#-MyHKvmTAlqLX5X0nbPL:-MyHKvmTAlqLX5X0nbPM:b-onrb75) ([raw file](https://github.com/isl-org/open3d/blob/8045725eb050cd1fc509bdba622e52b4b5c682d5/cpp/open3d/geometry/PointCloudSegmentation.cpp#L176)):*
> ```cpp
>     for (int itr = 0; itr < num_iterations; itr++) {
>         if ((size_t)iteration_count > break_iteration) {
>             continue;
> ```

This should be `break` right?
___
*[cpp/open3d/geometry/PointCloudSegmentation.cpp, line 179 at r2](https://reviewable.io/reviews/isl-org/open3d/4863#-MyHNJfz33ApTWnEB5rc:-MyHNJfz33ApTWnEB5rd:b-emv5jl) ([raw file](https://github.com/isl-org/open3d/blob/8045725eb050cd1fc509bdba622e52b4b5c682d5/cpp/open3d/geometry/PointCloudSegmentation.cpp#L179)):*
> ```cpp
>         }
> 
>         for (int i = 0; i < ransac_n; ++i) {
> ```

Multiple threads manipulating the `indices` array seems to be a problem here.
I think it is best to entirely remove it and create a small thread local vector with the selected indices and a function for sampling without duplicates.

yuecideng:
Thanks for your reply.

1. I think it makes sense, and the break condition can be modified as:
```c++
        if ((size_t)iteration_count > break_iteration &&
            iteration_count > min_iterations) {
            continue;
        }
```
2. `break` syntax seems cannot be used in openmp for loop, so I use `continue` instead. Maybe there is another way to do the same operation.
3. Consider this modification:
add a rondomsampling function
```cpp
template <typename T>
std::vector<T> RandomSampling(const std::vector<T> &indices,
                              size_t sample_size) {
    std::random_device rd;
    std::mt19937 rng(rd());
    const size_t num = indices.size();
    std::vector<T> indices_ = indices;

    for (size_t i = 0; i < sample_size; ++i) {
        std::swap(indices_[i], indices_[rng() % num]);
    }

    std::vector<T> sample;
    sample.reserve(sample_size);
    for (size_t idx = 0; idx < sample_size; ++idx) {
        sample.push_back(indices_[idx]);
    }

    return sample;
}
```
   In for loop:
```cpp
        const std::vector<size_t> sampled_indices =
                RandomSampling(indices, ransac_n);
        std::vector<size_t> inliers = sampled_indices;
```
Remove original random sampling code.
@benjaminum

benjaminum (commented):
I think the RandomSampler class makes sense. This way each thread can have its own `rng` object.
The `indices` array is not needed. Knowing the size is enough.
The operator needs to check for duplicates to make sure that we have 3 points for estimating the plane.

> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4863#-:-MyNn9IQ2Wrj6SGs-e8W:b-njxntr)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @theNded and @yuecideng)

yuecideng:
I push a new change relative to the above problems. I also add a new arg `min_iterations` in `SegmentPlane` function as well as its python binding. 
If there exist a better way to do the backwards compatible of original behavior, feel free to point out and make change to this PR.

yuecideng:
How about this improvement:
Add a `std::mutux` in `RandomSampler` and call lock at the beginning of the operator to make synchronization of `rng_`
```cpp
private:
    size_t size_;
    std::mt19937 rng_;
    std::mutex mutex_;

-----------------------------------------------------------------------
    std::vector<T> operator()(size_t sample_size) {
        std::lock_guard<std::mutex> guard(mutex_);

```

benjaminum:
> How about this improvement: Add a `std::mutux` in `RandomSampler` and call lock at the beginning of the operator to make synchronization of `rng_`

Yes that will work.

yuecideng:
> probability is set to 1 which should the algorithm force to use num_iterations (unless all points are inliers)

 I agree with the idea of using `probability` instead of `min_iteration`, which makes it more flexible and reasonable to control the behaviour of executed iterations. I will keep make changes to this PR based on what you have mentioned.

benjaminum:
@yuecideng Thanks for improving the unit test.

> > probability is set to 1 which should the algorithm force to use num_iterations (unless all points are inliers)
> 
> Another problem is that how can I test this condition in UnitTest?

You are right the number of iterations cannot be easily checked in the unit test. I think it is enough to manually check this behavior once.
What the test should capture is that there are no problems with special values for `probability`. Special values are 1 which could give problems with the `log` functions. Other special values are <=0 or >1 which should result in an exception thrown which we can check.

yuecideng:
I push a new commit and test the special case of `probability` which is `1`, `<= 0` and `>1`.
The case when `probability` is `1` in python example `point_cloud_plane_segmentation.py` shows as below:
The algorithm force iteration to 1000 (equal to the num_iterations)

image

And this line of code will throw exception when `probability` is illegal:
```cpp
    if (probability <= 0 || probability > 1) {
        utility::LogError("Probability must be > 0 or <= 1.0");
    }
```

benjaminum (changes requested):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4863)* status: 0 of 4 files reviewed, 8 unresolved discussions (waiting on @theNded and @yuecideng)

___
*[cpp/open3d/geometry/PointCloudSegmentation.cpp, line 44 at r5](https://reviewable.io/reviews/isl-org/open3d/4863#-MzLNWHK3HSejG6D3EtK:-MzOyyKZBi6jwSxPXwM7:b-kr35kn) ([raw file](https://github.com/isl-org/open3d/blob/ee47828caa22c93d106be6c186d9db9b996b3c24/cpp/open3d/geometry/PointCloudSegmentation.cpp#L44)):*

Can you put the RandomSampler in an anonymous namespace to indicate it is just a helper class.
</blockquote></details>

Can you add a namespace without name around the RandomSampler? Like this
```
namespace {

// RandomSampler
...

}
```

yuecideng:
> Can you put the RandomSampler in an anonymous namespace to indicate it is just a helper class.

Ok, I just misunderstand this info.

benjaminum (approved):
:lgtm: Thanks a lot for your contribution!

> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4863#-:-MzQb1kqEIfDYTFqNwgg:buqkcj5)* status: 0 of 4 files reviewed, 8 unresolved discussions (waiting on @theNded and @yuecideng)

PR #4872, opened 2022-03-09
docker args rename

yxlao:
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4872)


ssheorey:
Hi @yxlao do you want to work on this PR, or should we close it?

PR #4874, opened 2022-03-10
Fix windows CI wheel artifact download on master

yxlao:
1. When uploading only one artifact, the file is put in root's `py36.whl`. This works with PR CI.
2. When uploading multiple artifacts, github puts them in a folder: `pip_wheel/py36.whl, pip_wheel/py37.whl, ...`. This does not work with master CI.

This PR address enables both 1. and 2. 
1. https://github.com/isl-org/Open3D/runs/5599331175?check_suite_focus=true
2. https://github.com/isl-org/Open3D/runs/5599343802?check_suite_focus=true

This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4874)


yxlao:
minor fix, mrege directly

PR #4880, opened 2022-03-11
Easy to use CPU rendering for Linux

ssheorey:
+ Prebuilt Mesa CPU rendering library binary uploaded to Open3D_downloads.
+ Build script + docker instructions.
+ Auto-download during build.
+ Packaged with Python wheel, autoload if `OPEN3D_CPU_RENDERING` env var is set to `true`. Also obeys `LIBGL_ALWAYS_SOFTWARE=true` (mesa setting).
+ Tutorial for Python and C++
+ Python GUI unit test with CPU rendering

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4880)


ssheorey:
`xvfb` is required for testing, since offscreen rendering (`render_to_image()`) also needs an X server.

errissa (changes requested):
> Reviewed 2 of 6 files at r1, 5 of 8 files at r2, 1 of 5 files at r3, 5 of 5 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4880)* status: all files reviewed, 2 unresolved discussions (waiting on @ssheorey and @yxlao)

___
*[3rdparty/find_dependencies.cmake, line 1248 at r4](https://reviewable.io/reviews/isl-org/open3d/4880#-MylmMHQ0ksG8sF7A-LI:-MylmMHQ0ksG8sF7A-LJ:b-io18h1) ([raw file](https://github.com/isl-org/open3d/blob/a82dd1c7f3c1c6c95e91ca4203fd5b3eb5db796e/3rdparty/find_dependencies.cmake#L1248)):*
> ```CMake
> if(BUILD_GUI AND UNIX AND NOT APPLE)
>     include(FetchContent)
>     FetchContent_Declare(
> ```

1) This builds Mesa unconditionally on Linux which makes our build larger and take longer. Is this necessary? Could we make it optional?

2) This hard codes the current latest Mesa. Unlike most third party libraries we use, libGL has a well-defined and stable API so we are not dependent on the version of Mesa. I would prefer we build the latest  Mesa if there is a way to download the latest.
___
*[cpp/pybind/CMakeLists.txt, line 66 at r4](https://reviewable.io/reviews/isl-org/open3d/4880#-MylnY1c-oUM1Ci8khAC:-MylnY1d9aM13_DOAPsX:bizvbj6) ([raw file](https://github.com/isl-org/open3d/blob/a82dd1c7f3c1c6c95e91ca4203fd5b3eb5db796e/cpp/pybind/CMakeLists.txt#L66)):*
> ```CMake
> if (BUILD_GUI AND CMAKE_SYSTEM_NAME STREQUAL "Linux")
>     list(APPEND PYTHON_EXTRA_LIBRARIES ${CPP_LIBRARY}.1 ${CPPABI_LIBRARY}
>         ${MESA_CPU_GL_LIBRARY})
> ```

I don't think this should be necessary. The `target_link_libraries(pybind PRIVATE Open3D::Open3D)` above should be enough to resolve the symbols.

errissa (commented):
Did you try `headless=True` with `render_to_image()`? That uses EGL which doesn't require an X server. I'm not sure how that would interact with CPU rendering but it might just work.

> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4880#-:-MylqY1U6y1FB0WuUDII:b-z6qqqi)* status: all files reviewed, 2 unresolved discussions (waiting on @ssheorey and @yxlao)

ssheorey (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4880)* status: all files reviewed, 2 unresolved discussions (waiting on @errissa and @yxlao)

___
*[3rdparty/find_dependencies.cmake, line 1248 at r4](https://reviewable.io/reviews/isl-org/open3d/4880#-MylmMHQ0ksG8sF7A-LI:-MymCX9s7Rb89Dgror0h:bm52swu) ([raw file](https://github.com/isl-org/open3d/blob/a82dd1c7f3c1c6c95e91ca4203fd5b3eb5db796e/3rdparty/find_dependencies.cmake#L1248)):*

1) This builds Mesa unconditionally on Linux which makes our build larger and take longer. Is this necessary? Could we make it optional?

2) This hard codes the current latest Mesa. Unlike most third party libraries we use, libGL has a well-defined and stable API so we are not dependent on the version of Mesa. I would prefer we build the latest  Mesa if there is a way to download the latest.
</blockquote></details>

1. We only download a pre-built `libGL.so` - there's a separate build script `3rdparty/mesa/build-mesa-cpu.sh`, so build time is only minimally affected. I didn't make it optional for this reason.
2. Building mesa from source requires sudo access to install build dependencies -- it may be possible to avoid sudo, but the build docs don't mention it. I don't think it's a good idea to build mesa from source in Open3D due to this. 
3. This does make keeping mesa up-to-date tricky. The `master` branch sometimes fails CI, so I'd prefer the latest stable tag. We can discuss this in the meeting.
___
*[cpp/pybind/CMakeLists.txt, line 66 at r4](https://reviewable.io/reviews/isl-org/open3d/4880#-MylnY1c-oUM1Ci8khAC:-MymBray2HOOqBZ3-6bU:bls8scx) ([raw file](https://github.com/isl-org/open3d/blob/a82dd1c7f3c1c6c95e91ca4203fd5b3eb5db796e/cpp/pybind/CMakeLists.txt#L66)):*

I don't think this should be necessary. The `target_link_libraries(pybind PRIVATE Open3D::Open3D)` above should be enough to resolve the symbols.
</blockquote></details>

This adds `libGL.so` to the wheel - it's for run time. Ubuntu 18.04 doesn't have the min required version of mesa, so we need to package `libGL.so` with the wheel for it to work on Ubuntu 18.04. Also, mesa may not be installed if there is only an Nvidia GPU (no Intel GPU).

docs/tutorial/visualization/cpu_rendering.rst
```diff
@@ -0,0 +1,111 @@
...
+
+        LIBGL_ALWAYS_SOFTWARE=true python examples/python/visualization/draw.py
+
+.. note:: Mesa drivers must be in use for this method to work; just having
+   them installed is not sufficient. You can check the drivers in use with the
+   ``glxinfo`` command.
+
+2.  **You use Nvidia or AMD drivers or old Mesa drivers (< v20.2).**  We provide
+    the Mesa software rendering library binary for download `here
+    <https://github.com/isl-org/open3d_downloads/releases/download/mesa-libgl/mesa_libGL_22.0.tar.xz>`__.
```
ssheorey, line 89: Does the URL need to be updated?

docs/tutorial/visualization/cpu_rendering.rst
```diff
@@ -0,0 +1,111 @@
...
+    This is automatically downloaded to
+    `build/_deps/download_mesa_libgl-src/libGL.so.1.5.0` when you build Open3D
```
ssheorey, line 91: Update since we have more libraries now(?)

docs/tutorial/visualization/cpu_rendering.rst
```diff
@@ -0,0 +1,111 @@
...
+    .. code:: bash
+
+        export LD_LIBRARY_PATH=$HOME/.local/lib:$LD_LIBRARY_PATH
+
+    For occasional use, you can instead launch a program with CPU rendering with:
+
+    .. code:: bash
+
+        LD_PRELOAD=$HOME/.local/lib/libGL.so.1.5.0 Open3D
+
```
ssheorey, line 105: Do we need to LD_PRELOAD any other libraries? (libEGL, etc.)

errissa (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4880)* status: 3 of 20 files reviewed, 4 unresolved discussions (waiting on @errissa, @ssheorey, and @yxlao)

___
*[`docs/tutorial/visualization/cpu_rendering.rst` line 89 at r12](https://reviewable.io/reviews/isl-org/open3d/4880#-MxvAHfINQtz_2vJlzKd-r12-89:-N9I6HsA2JLSTxAb21zl:b-oznanb) ([raw file](https://github.com/isl-org/open3d/blob/f52a84c9dd335b2d8f19053296b25b8a00c082fd/docs/tutorial/visualization/cpu_rendering.rst#L89)):*

Does the URL need to be updated?
</blockquote></details>

There are now two Mesa builds, one for CPU Headless and one for CPU Interactive. This link is still correct for CPU Interactive Mesa build.
___
*[`docs/tutorial/visualization/cpu_rendering.rst` line 91 at r12](https://reviewable.io/reviews/isl-org/open3d/4880#-MxvAHfINQtz_2vJlzKd-r12-91:-N9I7RmGEZ6Xch8VW2Rn:b976nmb) ([raw file](https://github.com/isl-org/open3d/blob/f52a84c9dd335b2d8f19053296b25b8a00c082fd/docs/tutorial/visualization/cpu_rendering.rst#L91)):*

Update since we have more libraries now(?)
</blockquote></details>

Not for CPU Interactive rendering. For that use case, these instructions are still correct - it's just a single `libGL`
___
*[`docs/tutorial/visualization/cpu_rendering.rst` line 105 at r12](https://reviewable.io/reviews/isl-org/open3d/4880#-MxvAHfINQtz_2vJlzKd-r12-105:-N9I7Zzr1LFFIeupfTyM:bbip6rd) ([raw file](https://github.com/isl-org/open3d/blob/f52a84c9dd335b2d8f19053296b25b8a00c082fd/docs/tutorial/visualization/cpu_rendering.rst#L105)):*

Do we need to LD_PRELOAD any other libraries? (libEGL, etc.)
</blockquote></details>

Not for interactive CPU rendering which is described here. It's still just a single `libGL`

errissa (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/4880#-:-N9IT8he25EUjVirzzMd:bnfp4nl)* status: 3 of 20 files reviewed, 4 unresolved discussions (waiting on @errissa, @ssheorey, and @yxlao)

PR #4920, opened 2022-03-24
Refactoring unit test to use fs::temp_directory

reyanshsolis:
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4920)

### DIFF
--- a/cpp/tests/io/OctreeIO.cpp
+++ b/cpp/tests/io/OctreeIO.cpp
@@ -34,18 +34,9 @@
 #include "open3d/geometry/Octree.h"
 #include "open3d/geometry/PointCloud.h"
 #include "open3d/io/PointCloudIO.h"
+#include "open3d/utility/FileSystem.h"
 #include "open3d/utility/IJsonConvertible.h"
 #include "tests/Tests.h"
-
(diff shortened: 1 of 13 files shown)


cpp/tests/io/OctreeIO.cpp
```diff
@@ -37,22 +37,29 @@
 #include "open3d/utility/IJsonConvertible.h"
 #include "tests/Tests.h"
 
+#ifdef WIN32
```
yxlao, line 40: Include this in `cpp/tests/Tests.h`. Typically we don't include in `.h` file in this way, but this can be an exception case as we are back-porting CXX17.

yxlao (approved):
LGTM after moving the include header.

cpp/tests/io/OctreeIO.cpp
yxlao: The better solution is to add `utility::filesystem::GetTempDirectoryPath()`, which I already pushed.

reyanshsolis: Thanks for adding the changes @yxlao.

PR #4931, opened 2022-03-27
fix: O3DVisualizer::RemoveGeometry() function vector subscript out of…

UnadXiao:
After `objects_.erase()`, `objects_[i].name` has out-of-bounds access risk

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4931)


ssheorey (approved):
Thanks @UnadXiao !

PR #4942, opened 2022-03-28
Fix mesh naming in mesh docs

pmokeev:
Changed:
- `mesh_in` -> `mesh` on [Mesh page](http://www.open3d.org/docs/release/tutorial/geometry/mesh.html#Vertex-clustering)

This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4942)


yxlao:
Looks like the name `mesh_in` corresponds to the name `mesh_out`. Probably we should keep the original name. Any other reasons to update the name besides cosmetics @pmokeev ?

pmokeev:
> Any other reasons to update the name besides cosmetics?

In this example, a mesh is created with the name `mesh` (a few lines above, I did not change them), and then the name `mesh_in` is used. Therefore, this example cannot be reproduced.
I can replace the named BunnyMesh initialization with `mesh_in` if you think that looks better.

reyanshsolis:
> > Any other reasons to update the name besides cosmetics?
> 
> In this example, a mesh is created with the name `mesh` (a few lines above, I did not change them), and then the name `mesh_in` is used. Therefore, this example cannot be reproduced. I can replace the named BunnyMesh initialization with `mesh_in` if you think that looks better.

Hi @pmokeev, Yes I think that will look better (init. with `mesh_in`). Thanks for your valuable contribution.

pmokeev:
> Hi @pmokeev, Yes I think that will look better (init. with `mesh_in`). Thanks for your valuable contribution.
It's done :)

reyanshsolis (approved):

PR #4955, opened 2022-03-30
Fix link error during build due to invalid LIBDIR (#4945)

### LINKED ISSUE #4945: Cannot compile on Ubuntu 20.04: libassimp error
### Checklist

#### I first cloned Open3D by:
```
git clone https://github.com/isl-org/Open3D.git
cd Open3D
util/install_deps_ubuntu.sh

```

#### Then, I build Open3D (on Ubuntu 20.04, with CUDA 11.4) with:
```
mkdir build
cd build
cmake -DBUILD_SHARED_LIBS=OFF -DBUILD_CUDA_MODULE=ON -DUSE_BLAS=ON -DBUILD_WEBRTC=OFF ..
make -j$(nproc)
```

### Error message

```shell
make[2]: *** No rule to make target 'assimp/lib/x86_64-linux-gnu/libassimp.a', needed by 'lib/Release/libOpen3D.so'.  Stop.
```

```markdown
- Operating system: Ubuntu 20.04 
- Python version: Python 3.8.13 
- Open3D version: output from python: `print(open3d.__version__)`
- System architecture: x86
- Is this a remote workstation?: yes or no
- Compiler version (if built from source): gcc 9.4
```

### Additional information

I installed it once and it was working. I had to rebuild it and I get this error. I also tried a fresh clone of the repository but the error is still there

bchretien:
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4955)

This PR fixes #4945: Open3D could not be compiled on recent Debian-based distributions. The problem was that the proper LIBDIR was not always used for 3rd party dependencies.

I also added a fix for the DenseSLAM example that requires GUI (else `open3d::visualization::Draw` is not available).


3rdparty/stdgpu/stdgpu.cmake
```diff
@@ -28,5 +29,5 @@ ExternalProject_Add(
 
 ExternalProject_Get_Property(ext_stdgpu INSTALL_DIR)
 set(STDGPU_INCLUDE_DIRS ${INSTALL_DIR}/include/) # "/" is critical.
-set(STDGPU_LIB_DIR ${INSTALL_DIR}/lib)
+set(STDGPU_LIB_DIR ${INSTALL_DIR}/${Open3D_INSTALL_LIB_DIR})
```
bchretien: Probably invalid: the project does not respect conventions (see [here](https://github.com/stotko/stdgpu/blob/e10f6f3ccc9902d693af4380c3bcd188ec34a2e6/CMakeLists.txt#L139)).

3rdparty/assimp/assimp.cmake
```diff
@@ -16,6 +16,7 @@ ExternalProject_Add(
     CMAKE_ARGS
         -DBUILD_SHARED_LIBS=OFF
         -DCMAKE_INSTALL_PREFIX=<INSTALL_DIR>
+        -DCMAKE_INSTALL_LIBDIR=${Open3D_INSTALL_LIB_DIR}
```
yxlao, line 19: Lets add this flag to:

```diff
diff --git a/3rdparty/find_dependencies.cmake b/3rdparty/find_dependencies.cmake
index dd00fd6b3..f06029d63 100644
--- a/3rdparty/find_dependencies.cmake
+++ b/3rdparty/find_dependencies.cmake
@@ -164,6 +164,7 @@ set(ExternalProject_CMAKE_ARGS
     -DCMAKE_CXX_COMPILER_LAUNCHER=${CMAKE_CXX_COMPILER_LAUNCHER}
     -DCMAKE_CUDA_COMPILER_LAUNCHER=${CMAKE_CUDA_COMPILER_LAUNCHER}
     -DCMAKE_OSX_DEPLOYMENT_TARGET=${CMAKE_OSX_DEPLOYMENT_TARGET}
+    -DCMAKE_INSTALL_LIBDIR=${Open3D_INSTALL_LIB_DIR}
     # Always build 3rd party code in Release mode. Ignored by multi-config
     # generators (XCode, MSVC). MSVC needs matching config anyway.
     -DCMAKE_BUILD_TYPE=Release
```

Since most third-party libs uses either `ExternalProject_CMAKE_ARGS` or `ExternalProject_CMAKE_ARGS_hidden`.

yxlao (changes requested):
LGTM in general. Let's try if we can move the flag to the centralized location.

3rdparty/assimp/assimp.cmake
bchretien: I'll need to check if that works for `libjpeg-turbo`, as using it led to errors and I had to change it to `-DCMAKE_INSTALL_DEFAULT_LIBDIR=${Open3D_INSTALL_LIB_DIR}` for that one.

bchretien: Ok, I've got a new version using `ExternalProject_CMAKE_ARGS` that also works for `libjpeg-turbo`. Pushing it soon.

bchretien:
> However, my concern is that the PR sets `-DCMAKE_INSTALL_LIBDIR=${Open3D_INSTALL_LIB_DIR}` in `find_dependencies.cmake`, but `Open3D_INSTALL_LIB_DIR` itself is set to `"${CMAKE_INSTALL_LIBDIR}"` in `CMakeLists.txt`, so I am not sure about the cyclic dependency of the same.

@reyanshsolis Normally this only impacts third-party dependencies, i.e. the `LIBDIR` is properly propagated to external projects built by CMake, but that should not impact the build/config of Open3D itself (unless I'm missing something in the way `ExternalProject_Add` operates). The whole thing smells like a bug in CMake that might be fixed in some obscure policy.

yxlao (approved):

PR #5020, opened 2022-04-19
Remove setuptools and wheel as end-user requirements

### LINKED ISSUE #5019: setuptools and wheel incorrectly defined as runtime dependencies
### Checklist

### Describe the issue

`setup.py` gets its `install_requires` requirements from `requirements.txt`:

https://github.com/isl-org/Open3D/blob/88efd40edd2a9e3a77bd9d035c38989ff58988f0/python/setup.py#L117-L119

And `requirements.txt` includes `setuptools` and `wheel`:

https://github.com/isl-org/Open3D/blob/88efd40edd2a9e3a77bd9d035c38989ff58988f0/python/requirements.txt#L1-L2

These are _build_ dependencies, but should not be _runtime_ dependencies that are installed into each end user of `open3d` environment. This adds extra dependencies and size into things like Docker images that depend on `open3d`.

The source build dependencies are already properly defined in `pyproject.toml`:

https://github.com/isl-org/Open3D/blob/b449b58a1ee2920f5ec4ef351fca25431f63ea13/python/pyproject.toml#L1-L3

### Expected behavior

`open3d` does not have an `install_requires` dependency on `setuptools` or `wheel`.

johnthagen:
Closes #5019

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/5020)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 * Fix undefined names: docstr and VisibleDeprecationWarning (PR #3844)
 * Corrected documentation for Tensor based PointClound, LineSet, TriangleMesh (PR #4685)
 * Corrected documentation for KDTree (typo in Notebook) (PR #4744)
+* Remove `setuptools` and `wheel` from requirements for end users (PR #5020)
 
 ## 0.13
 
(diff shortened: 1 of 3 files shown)


yxlao:
Agreed that these are build dependencies. We should probably create a `python/requirements_build.txt` and use that in CI.

yxlao (approved):

PR #4898, opened 2022-03-17
Open3D Visualizer Tutorials

AlexZakIntel:
Yixing, Sameer, Rene, Rishabh,

This is a Pull Request for the **Beginner and Advanced Open3D Visualizer Tutorials** beta version. There might be some more alterations to it based on our further work with Rene and your collective input.

Please feel free to suggest changes.

Thanks,

Alex.

P.S.: I have altered the **Datasets Reference** (`Open3D/docs/tutorial/data/index.rst`). @reyanshsolis RIshabh, could you take a look at the changes?

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/4898)

PR #5029, opened 2022-04-22
add `--parallel` option to `make_docs.py`

yxlao:
This PR:
- Adds `--parallel` option to `make_docs.py`
- Standardizes empty lines in docs:`\n\n\n` -> `\n\n`

Example usage:

```bash
python make_docs.py --parallel --clean_notebooks --execute_notebooks=never --sphinx
```

Merge this PR before merging https://github.com/isl-org/Open3D/pull/4898. 

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/5029)

### DIFF
--- a/docs/make_docs.py
+++ b/docs/make_docs.py
@@ -43,6 +43,7 @@ import nbconvert
 import ssl
 import certifi
 import urllib.request
+import multiprocessing
 
 
 def _create_or_clear_dir(dir_path):
@@ -385,11 +386,12 @@ class SphinxDocsBuilder:
     """
 
     def __init__(self, current_file_dir, html_output_dir, is_release,
-                 skip_notebooks):
+                 skip_notebooks, parallel):
         self.current_file_dir = current_file_dir
         self.html_output_dir = html_output_dir
         self.is_release = is_release
         self.skip_notebooks = skip_notebooks
+        self.parallel = parallel
 
     def run(self):
         """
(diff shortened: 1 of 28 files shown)


reyanshsolis:
LGTM. Testing locally.

reyanshsolis:
I can verify that it is using all cores when using `--parallel`.

With `--execute_notebooks=never`, with `--parallel` it takes about `25sec.`, and without `--parallel` it takes about `36sec.`, on an 8 core / 16 thread system. 

I have added the `--parallel` flag in the `Building the documentation` doc.

docs/make_docs.py
```diff
@@ -631,6 +639,13 @@ def run(self):
         default=False,
         help="Show Open3D version number rather than git hash.",
     )
+    parser.add_argument(
+        "--parallel",
+        action="store_true",
+        default=False,
```
reyanshsolis, line 645: Do we want to make the default `True`?

reyanshsolis (approved):

docs/make_docs.py
yxlao: By default, we may not need the `--parallel` flag, as the error message will be hard to locate. The `--parallel` flag is only used when you know that the compilation will not have errors and you'll recompile the docs again and again.

yxlao: I added several examples instead

```bash
    # Run `python make_docs.py --help` to usage of the flags.
    python make_docs.py --help

    # Example: build .rst and C++ docs only, skip notebooks.
    python make_docs.py --execute_notebooks=never --sphinx --doxygen

    # Example: build .rst and C++ docs only, skip notebooks, with parallel build.
    python make_docs.py --execute_notebooks=never --sphinx --doxygen --parallel

    # Example: build .rst and c++ docs, execute notebooks when it has not been executed.
    python make_docs.py --execute_notebooks=auto --sphinx --doxygen
```

PR #5047, opened 2022-04-24
Implement RemoveDuplicatedPoints on geometry::PointCloud

scimad:
This PR includes
* Implementation of RemoveDuplicatePoints for geometry::PointCloud
* Pybind for above function
* A minimal test for above function

Open to feedback or if anything more is required.

Also, I believe that copying `covariances_` from their corresponding index doesn't do good because changing points will change their covariances. Correct me if wrong.

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/5047)


cpp/open3d/geometry/PointCloud.cpp
```diff
@@ -158,6 +158,44 @@ std::vector<double> PointCloud::ComputePointCloudDistance(
...
+            points_[k] = points_[i];
+            if (has_normals) normals_[k] = normals_[i];
+            if (has_colors) colors_[k] = colors_[i];
+            // estimate covariance later (copying covariances won't work)
+            k++;
+        }
+    }
+
+    points_.resize(k);
+    if (has_normals) normals_.resize(k);
```
reyanshsolis, line 186: Normals are computed using covariance, so it might change as well. We should compute the same again. 

Also, check `has_covariances` before `has_normals`. As it's almost free to compute normals when we have covariances.

cpp/open3d/geometry/PointCloud.cpp
```diff
@@ -158,6 +158,44 @@ std::vector<double> PointCloud::ComputePointCloudDistance(
...
 
+PointCloud &PointCloud::RemoveDuplicatedPoints() {
+    bool has_normals = HasNormals();
+    bool has_colors = HasColors();
+    bool has_covariances = HasCovariances();
+    size_t old_points_num = points_.size();
+    size_t k = 0;
+
+    typedef std::tuple<double, double, double> Coordinate3;
+    std::unordered_map<Coordinate3, size_t, utility::hash_tuple<Coordinate3>>
```
reyanshsolis, line 169: Is there any particular reason to use `unordered_map` instead of `unordered_set` here? 

If you want to return `std::tuple<PointCloud, std::vector<size_t>>&` i.e. 
`pointcloud, indices_of_removed_points = pcd.remove_duplicate_points()`, then we can keep using the `unordered_map`.

reyanshsolis (changes requested):

cpp/open3d/geometry/PointCloud.cpp
hernot: Having a variant which in addition to Point-cloud the indices cleared or indices kept would be desireable for adjusting application data attached to each point. Currently i do this (using TriangleMesh.remove_duplicate_vertices) by temporarily abusing the vertex/point colors as they are float values between [0,1] is just replace them by a color triplet of the kind [point_index/num_points,point_index/num_points,point_index/num_points] and than after removal simply take one color channel and multiply by num_points and round to closest integer. Having the information directly from the method would mean a great simplification.

scimad: @reyanshsolis, My inspiration to use hashmap is originally from [`removeDuplicatedVertices` in `TriangleMesh`](https://github.com/isl-org/Open3D/blob/master/cpp/open3d/geometry/TriangleMesh.cpp#L681). I guess the point is simply that, it would be costly to find and store each unique element in `vertices_`  (and hence `colors_` into a separate array). Finding duplicated vertices otherwise would require comparing vertices (each three coordinates separately) with all other element in vertices_.

Also, using `unordered_set` to store unique vertices would not preserve the order of index creating a mess to copy `colors_` for example.

reyanshsolis: Follow up: For legacy, I am keeping the current implementation with minor changes. Attribute re-computation must be avoided (read the last comment).

The return is just the PointCloud (no indices) (reason: i. consistent with other legacies pre-processing remove-* function, ii. as it is used in pre-processing, in a common case the indices might not be required).

The `t::PointCloud::RemoveDuplicatedPoints` may be implemented using `Tensor::Unique` op (to be added), and to be consistent with the `numpy / torch unique` op, it may return the removed / selected indices.

reyanshsolis (approved):

PR #5114, opened 2022-05-21
Support for sharing memory with vtk structures

benjaminum:
This PR adds support for sharing memory with vtk to avoid data copies

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/5114)


cpp/open3d/t/geometry/VtkUtils.cpp
```diff
@@ -0,0 +1,333 @@
...
+                        *data = array->GetTypedComponent(i, j);
+                        ++data;
+                    }
+                }
+            }
+        } else {
+            auto sp = data_array;
+            auto blob = std::make_shared<core::Blob>(core::Device(),
+                                                     array->GetVoidPointer(0),
+                                                     [sp](void*) { (void)sp; });
```
ssheorey, line 224: Capturing `data_array` by value directly in the deleter lambda should work. We can skip `sp`.

benjaminum (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/5114)* status: 0 of 10 files reviewed, 7 unresolved discussions (waiting on @ssheorey and @yxlao)

___
*[`cpp/tests/t/geometry/VtkUtils.cpp` line 65 at r2](https://reviewable.io/reviews/isl-org/open3d/5114#-N2_QwIQKn8Br4_Ui5SV-r2-65:-N4CkRdk6MF_Ps-cWfwD:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/a8340dbd3426b283ddc40e3729d841c472ea9970/cpp/tests/t/geometry/VtkUtils.cpp#L65)):*

Perhaps also check that the data values match if copy=True?
</blockquote></details>

Done.
___
*[`cpp/tests/t/geometry/VtkUtils.cpp` line 87 at r2](https://reviewable.io/reviews/isl-org/open3d/5114#-N2_QwIQKn8Br4_Ui5SV-r2-87:-N4CkSN6Dgbh-73QnIlz:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/a8340dbd3426b283ddc40e3729d841c472ea9970/cpp/tests/t/geometry/VtkUtils.cpp#L87)):*

Check data values here too if copy=True.
</blockquote></details>

Done.
___
*[`cpp/tests/t/geometry/VtkUtils.cpp` line 129 at r2](https://reviewable.io/reviews/isl-org/open3d/5114#-N2_QwIQKn8Br4_Ui5SV-r2-129:-N4CkTOU8pA7SN7SugvM:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/a8340dbd3426b283ddc40e3729d841c472ea9970/cpp/tests/t/geometry/VtkUtils.cpp#L129)):*

Check if data values match if copy=True.
</blockquote></details>

Done.
___
*[`cpp/open3d/t/geometry/VtkUtils.cpp` line 102 at r2](https://reviewable.io/reviews/isl-org/open3d/5114#-N2amn_MBvMIo1sKdxHF-r2-102:-N4ChzCZ1jjfLTJcmECG:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/a8340dbd3426b283ddc40e3729d841c472ea9970/cpp/open3d/t/geometry/VtkUtils.cpp#L102)):*

If tensor is on CUDA and copy is True, we will do 2 data copies. Can we create tensor_cpu only if copy is not needed? And use device to host copy here.
</blockquote></details>

Done.
___
*[`cpp/open3d/t/geometry/VtkUtils.cpp` line 177 at r2](https://reviewable.io/reviews/isl-org/open3d/5114#-N2amn_MBvMIo1sKdxHF-r2-177:-N4CiFiI5cwFjxPd7xXJ:bbtbe5b) ([raw file](https://github.com/isl-org/open3d/blob/a8340dbd3426b283ddc40e3729d841c472ea9970/cpp/open3d/t/geometry/VtkUtils.cpp#L177)):*

So if copy=True, does this mean we get a reference to the tensor data or we get a copy?
</blockquote></details>

Unfortunately, the triangle connectivity data will always be a copy, i.e., the `copy` argument has no effect with the current version of vtk. It is unclear to me why the vtk code uses internally an old array api which does not implement the shallow copy. Maybe this is an oversight. Nightly has a new function for setting the connectivity (see comment) and maybe it will be fixed in the future and then this code will reuse the memory without changes needed.
___
*[`cpp/open3d/t/geometry/VtkUtils.cpp` line 224 at r2](https://reviewable.io/reviews/isl-org/open3d/5114#-N2amn_MBvMIo1sKdxHF-r2-224:-N4Cjyhw5QnQFGpnpSEJ:b-cx6ctv) ([raw file](https://github.com/isl-org/open3d/blob/a8340dbd3426b283ddc40e3729d841c472ea9970/cpp/open3d/t/geometry/VtkUtils.cpp#L224)):*

Capturing `data_array` by value directly in the deleter lambda should work. We can skip `sp`. 
</blockquote></details>

There is a problem with the implicit capture of the this pointer. I moved the reference count increment now inside of the worker struct, which is a cleaner solution and avoids the problem.
___
*[`cpp/open3d/t/geometry/VtkUtils.cpp` line 315 at r2](https://reviewable.io/reviews/isl-org/open3d/5114#-N2amn_MBvMIo1sKdxHF-r2-315:-N4CkQeXEmeYdyOoD-6t:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/a8340dbd3426b283ddc40e3729d841c472ea9970/cpp/open3d/t/geometry/VtkUtils.cpp#L315)):*

Issue a warning if other data is present, but not  converted.
</blockquote></details>

Done.

ssheorey (approved):

yxlao (approved):

PR #5119, opened 2022-05-23
fix: dead link in documentation

chunibyo-wly:
I have checked in release [0.15](https://github.com/isl-org/Open3D/tree/ed30e3b61fbe031e106fa64030bec3f698b316b4/examples/python), **gui** folder has already been renamed to **visualization**. But In the documentation, the path to the example file hasn't been updated simultaneously and causing dead link.

For example [this one](http://www.open3d.org/docs/0.15.1/tutorial/visualization/web_visualizer.html#python-server).

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/5119)

### DIFF
--- a/docs/tutorial/visualization/web_visualizer.rst
+++ b/docs/tutorial/visualization/web_visualizer.rst
@@ -102,10 +102,10 @@ Example:
 
 .. code-block:: sh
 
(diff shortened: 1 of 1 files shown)


yxlao:
Thank you @chunibyo-wly

PR #5150, opened 2022-06-06
Add SelectByIndex and improve minor of SelectByMask

yuecideng:
1. Change SelectPoints -> SelectByMask in tensor based PointCloud.
2. Remove tensor clone inside SelectByMask.
3. Add SelectByIndex (both unit test and pybind) in tensor based PointCloud.

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/5150)

### DIFF
--- a/cpp/benchmarks/t/geometry/PointCloud.cpp
+++ b/cpp/benchmarks/t/geometry/PointCloud.cpp
@@ -37,7 +37,6 @@
 #include "open3d/t/io/PointCloudIO.h"
 #include "open3d/visualization/utility/DrawGeometry.h"
 
-
 namespace open3d {
 namespace t {
 namespace geometry {
--- a/cpp/benchmarks/t/geometry/PointCloud.cpp
+++ b/cpp/benchmarks/t/geometry/PointCloud.cpp
@@ -28,8 +28,6 @@
 
 #include <benchmark/benchmark.h>
 
-#include <algorithm>
-
 #include "open3d/core/CUDAUtils.h"
 #include "open3d/core/Tensor.h"
 #include "open3d/data/Dataset.h"
@@ -151,7 +149,7 @@ void SelectByIndex(benchmark::State& state, const core::Device& device) {
 
     const int64_t num_points = pcd.GetPointPositions().GetLength();
     core::Tensor indices =
-            core::Tensor::Arange(0, num_points, 1, core::Dtype::Int64, device);
+            core::Tensor::Arange(0, num_points, 1, core::Int64, device);
 
     // Warm Up.
     PointCloud pcd_selected = pcd.SelectByIndex(indices);
--- a/cpp/benchmarks/t/geometry/PointCloud.cpp
+++ b/cpp/benchmarks/t/geometry/PointCloud.cpp
@@ -317,11 +317,12 @@ BENCHMARK_CAPTURE(SelectByIndex,
 #ifdef BUILD_CUDA_MODULE
 BENCHMARK_CAPTURE(Transform, CUDA, core::Device("CUDA:0"))
         ->Unit(benchmark::kMillisecond);
-BENCHMARK_CAPTURE(SelectByIndex, CUDA, false core::Device("CUDA:0"))
+BENCHMARK_CAPTURE(SelectByIndex, CUDA, false, core::Device("CUDA:0"))
         ->Unit(benchmark::kMillisecond);
 BENCHMARK_CAPTURE(SelectByIndex,
                   CUDA(remove duplicates),
-                  true core::Device("CUDA:0"))
+                  true,
+                  core::Device("CUDA:0"))
         ->Unit(benchmark::kMillisecond);
 #endif
 
--- a/cpp/benchmarks/t/geometry/PointCloud.cpp
+++ b/cpp/benchmarks/t/geometry/PointCloud.cpp
@@ -28,12 +28,15 @@
 
 #include <benchmark/benchmark.h>
 
+#include <algorithm>
+
 #include "open3d/core/CUDAUtils.h"
 #include "open3d/core/Tensor.h"
 #include "open3d/data/Dataset.h"
 #include "open3d/io/PointCloudIO.h"
 #include "open3d/t/io/PointCloudIO.h"
 #include "open3d/visualization/utility/DrawGeometry.h"
+
 
 namespace open3d {
 namespace t {
@@ -104,6 +107,22 @@ void VoxelDownSample(benchmark::State& state,
     }
 }
 
+void LegacyTransform(benchmark::State& state, const int no_use) {
+    open3d::geometry::PointCloud pcd;
+    open3d::io::ReadPointCloud(path, pcd, {"auto", false, false, false});
+
+    Eigen::Matrix4d transformation = Eigen::Matrix4d::Identity();
+    transformation.block<3, 1>(0, 3) = Eigen::Vector3d(1, 2, 3);
+
+    // Warm Up.
+    open3d::geometry::PointCloud pcd_transformed =
+            pcd.Transform(transformation);
+
+    for (auto _ : state) {
+        pcd_transformed = pcd.Transform(transformation);
+    }
+}
+
 void Transform(benchmark::State& state, const core::Device& device) {
     PointCloud pcd;
     t::io::ReadPointCloud(path, pcd, {"auto", false, false, false});
@@ -123,6 +142,40 @@ void Transform(benchmark::State& state, const core::Device& device) {
     for (auto _ : state) {
         pcd_transformed = pcd.Transform(transformation);
         core::cuda::Synchronize(device);
+    }
+}
+
+void SelectByIndex(benchmark::State& state, const core::Device& device) {
+    PointCloud pcd;
+    t::io::ReadPointCloud(path, pcd, {"auto", false, false, false});
+    pcd = pcd.To(device);
+
+    const int64_t num_points = pcd.GetPointPositions().GetLength();
+    core::Tensor indices =
+            core::Tensor::Arange(0, num_points, 1, core::Dtype::Int64, device);
+
+    // Warm Up.
+    PointCloud pcd_selected = pcd.SelectByIndex(indices);
+
+    for (auto _ : state) {
+        pcd_selected = pcd.SelectByIndex(indices);
+        core::cuda::Synchronize(device);
+    }
+}
+
+void LegacySelectByIndex(benchmark::State& state, const int no_use) {
+    open3d::geometry::PointCloud pcd;
+    open3d::io::ReadPointCloud(path, pcd, {"auto", false, false, false});
+
+    const size_t num_points = pcd.points_.size();
+    std::vector<size_t> indices(num_points);
+    std::iota(indices.begin(), indices.end(), 0);
+
+    // Warm Up.
+    open3d::geometry::PointCloud pcd_selected = *pcd.SelectByIndex(indices);
+
+    for (auto _ : state) {
+        pcd_selected = *pcd.SelectByIndex(indices);
     }
 }
 
@@ -252,8 +305,13 @@ ENUM_VOXELDOWNSAMPLE_BACKEND()
 BENCHMARK_CAPTURE(Transform, CPU, core::Device("CPU:0"))
         ->Unit(benchmark::kMillisecond);
 
+BENCHMARK_CAPTURE(SelectByIndex, CPU, core::Device("CPU:0"))
+        ->Unit(benchmark::kMillisecond);
+
 #ifdef BUILD_CUDA_MODULE
 BENCHMARK_CAPTURE(Transform, CUDA, core::Device("CUDA:0"))
+        ->Unit(benchmark::kMillisecond);
+BENCHMARK_CAPTURE(SelectByIndex, CUDA, core::Device("CUDA:0"))
         ->Unit(benchmark::kMillisecond);
 #endif
 
@@ -323,6 +381,9 @@ BENCHMARK_CAPTURE(EstimateNormals,
                   utility::nullopt)
         ->Unit(benchmark::kMillisecond);
 #endif
+
+BENCHMARK_CAPTURE(LegacyTransform, CPU, 1)->Unit(benchmark::kMillisecond);
+BENCHMARK_CAPTURE(LegacySelectByIndex, CPU, 1)->Unit(benchmark::kMillisecond);
 
 BENCHMARK_CAPTURE(LegacyEstimateNormals,
                   Legacy Hybrid[0.02 | 30 | 0.06],
--- a/cpp/benchmarks/t/geometry/PointCloud.cpp
+++ b/cpp/benchmarks/t/geometry/PointCloud.cpp
@@ -142,7 +142,9 @@ void Transform(benchmark::State& state, const core::Device& device) {
     }
 }
 
-void SelectByIndex(benchmark::State& state, const core::Device& device) {
+void SelectByIndex(benchmark::State& state,
+                   bool remove_duplicates,
+                   const core::Device& device) {
     PointCloud pcd;
     t::io::ReadPointCloud(path, pcd, {"auto", false, false, false});
     pcd = pcd.To(device);
@@ -152,7 +154,8 @@ void SelectByIndex(benchmark::State& state, const core::Device& device) {
             core::Tensor::Arange(0, num_points, 1, core::Int64, device);
 
     // Warm Up.
-    PointCloud pcd_selected = pcd.SelectByIndex(indices);
+    PointCloud pcd_selected =
+            pcd.SelectByIndex(indices, false, remove_duplicates);
 
     for (auto _ : state) {
         pcd_selected = pcd.SelectByIndex(indices);
@@ -302,13 +305,23 @@ ENUM_VOXELDOWNSAMPLE_BACKEND()
 BENCHMARK_CAPTURE(Transform, CPU, core::Device("CPU:0"))
         ->Unit(benchmark::kMillisecond);
 
-BENCHMARK_CAPTURE(SelectByIndex, CPU, core::Device("CPU:0"))
+BENCHMARK_CAPTURE(SelectByIndex, CPU, false, core::Device("CPU:0"))
+        ->Unit(benchmark::kMillisecond);
+
+BENCHMARK_CAPTURE(SelectByIndex,
+                  CPU(remove duplicates),
+                  true,
+                  core::Device("CPU:0"))
         ->Unit(benchmark::kMillisecond);
 
 #ifdef BUILD_CUDA_MODULE
 BENCHMARK_CAPTURE(Transform, CUDA, core::Device("CUDA:0"))
         ->Unit(benchmark::kMillisecond);
-BENCHMARK_CAPTURE(SelectByIndex, CUDA, core::Device("CUDA:0"))
+BENCHMARK_CAPTURE(SelectByIndex, CUDA, false core::Device("CUDA:0"))
+        ->Unit(benchmark::kMillisecond);
+BENCHMARK_CAPTURE(SelectByIndex,
+                  CUDA(remove duplicates),
+                  true core::Device("CUDA:0"))
         ->Unit(benchmark::kMillisecond);
 #endif
 
--- a/cpp/open3d/t/geometry/PointCloud.cpp
+++ b/cpp/open3d/t/geometry/PointCloud.cpp
@@ -253,19 +253,19 @@ PointCloud PointCloud::SelectByMask(const core::Tensor &boolean_mask,
 PointCloud PointCloud::SelectByIndex(const core::Tensor &indices,
                                      bool invert /* = false */) const {
     const int64_t length = GetPointPositions().GetLength();
-    core::AssertTensorDtype(indices, core::Dtype::Int64);
+    core::AssertTensorDtype(indices, core::Int64);
     core::AssertTensorDevice(indices, GetDevice());
 
     // The indices may have duplicate index value and will result in identity
-    // point cloud attributes. We convert indices Tensor into mask Tensor to and
+    // point cloud attributes. We convert indices Tensor into mask Tensor and
     // call SelectByMask to avoid this situation.
     core::Tensor mask =
-            core::Tensor::Zeros({length}, core::Dtype::Int16, GetDevice());
+            core::Tensor::Zeros({length}, core::Int16, GetDevice());
     mask.SetItem(core::TensorKey::IndexTensor(indices),
-                 core::Tensor(std::vector<int16_t>{1}, {}, core::Dtype::Int16,
+                 core::Tensor(std::vector<int16_t>{1}, {}, core::Int16,
                               GetDevice()));
 
-    PointCloud pcd = SelectByMask(mask.To(core::Dtype::Bool), invert);
+    PointCloud pcd = SelectByMask(mask.To(core::Bool), invert);
 
     utility::LogDebug("Pointcloud down sampled from {} points to {} points.",
                       length, pcd.GetPointPositions().GetLength());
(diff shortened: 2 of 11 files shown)


yuecideng:
I also add legacy `Transform` and `SelectByIndex` in benchmark. 
The `Transform` works fine for tensor based. (The legacy version can also be parallel using openmp)
image

The `SelectByIndex` is slow compared with legacy version (tensor based version is wrapped from `SelectByMask`, original `SelectPoints`)
image

yxlao (changes requested):
Looks good in general with some comments. After the boolean type change, the tensor-based can have some performance gains.

> Reviewed 3 of 4 files at r1, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/5150#-:-N41f1rr8m6k9qFv-9e3:b-rdrhxi)* status: 4 of 5 files reviewed, 6 unresolved discussions (waiting on @yuecideng and @yxlao)

___
*[`cpp/open3d/t/geometry/PointCloud.h` line 327 at r3](https://reviewable.io/reviews/isl-org/open3d/5150#-N41d21iC_6wCkSoPZcI:-N41d21iC_6wCkSoPZcJ:bco1vef) ([raw file](https://github.com/isl-org/open3d/blob/8a755e81bd9dd3cbc33617bf37d207bb51af833c/cpp/open3d/t/geometry/PointCloud.h#L327)):*
> ```objc
>     /// output point cloud.
>     ///
> ```

Remove blank line.
___
*[`cpp/open3d/t/geometry/PointCloud.h` line 328 at r3](https://reviewable.io/reviews/isl-org/open3d/5150#-N41d4IZ9uJJw739eE54:-N41d4IZ9uJJw739eE55:b-60kowe) ([raw file](https://github.com/isl-org/open3d/blob/8a755e81bd9dd3cbc33617bf37d207bb51af833c/cpp/open3d/t/geometry/PointCloud.h#L328)):*
> ```objc
>     ///
> 
>     /// \param indices Int32 indexing tensor of shape {n,} containing
> ```

Int64? Also mention how we handle duplicated indices.
___
*[`cpp/open3d/t/geometry/PointCloud.h` line 332 at r3](https://reviewable.io/reviews/isl-org/open3d/5150#-N41dhWB9-ZZvolH41iX:-N41dhWB9-ZZvolH41iY:bgeipu6) ([raw file](https://github.com/isl-org/open3d/blob/8a755e81bd9dd3cbc33617bf37d207bb51af833c/cpp/open3d/t/geometry/PointCloud.h#L332)):*
> ```objc
>     /// \param invert Set to `True` to invert the selection of indices.
>     PointCloud SelectByIndex(const core::Tensor &indices,
>                              bool invert = false) const;
> ```

It might not always be desirable to remove duplicated indices automatically, e.g., people may expect the output point cloud length to be the same as the input point cloud length. 

Shall we add an additional parameter called `remove_duplicates = false`, and by default, we do not remove the duplicates? This can also improve the default performance.
___
*[`cpp/open3d/t/geometry/PointCloud.cpp` line 256 at r3](https://reviewable.io/reviews/isl-org/open3d/5150#-N41eSlV5NajAx_pkmeU:-N41eSlV5NajAx_pkmeV:b-991m4u) ([raw file](https://github.com/isl-org/open3d/blob/8a755e81bd9dd3cbc33617bf37d207bb51af833c/cpp/open3d/t/geometry/PointCloud.cpp#L256)):*
> ```cpp
>                                      bool invert /* = false */) const {
>     const int64_t length = GetPointPositions().GetLength();
>     core::AssertTensorDtype(indices, core::Dtype::Int64);
> ```

Use `core::Int64` instead of `core::Dtype::xxx`. Same for the rest.
___
*[`cpp/open3d/t/geometry/PointCloud.cpp` line 265 at r3](https://reviewable.io/reviews/isl-org/open3d/5150#-N41eQW899D77XlMlG4h:-N41eQW899D77XlMlG4i:b-5wkz3l) ([raw file](https://github.com/isl-org/open3d/blob/8a755e81bd9dd3cbc33617bf37d207bb51af833c/cpp/open3d/t/geometry/PointCloud.cpp#L265)):*
> ```cpp
>             core::Tensor::Zeros({length}, core::Dtype::Int16, GetDevice());
>     mask.SetItem(core::TensorKey::IndexTensor(indices),
>                  core::Tensor(std::vector<int16_t>{1}, {}, core::Dtype::Int16,
> ```

We need some code change to allow `Bool` dtype. 

```diff
diff --git a/cpp/open3d/core/kernel/IndexGetSetCPU.cpp b/cpp/open3d/core/kernel/IndexGetSetCPU.cpp
index 329f78091..ca35e3aff 100644
--- a/cpp/open3d/core/kernel/IndexGetSetCPU.cpp
+++ b/cpp/open3d/core/kernel/IndexGetSetCPU.cpp
@@ -71,7 +71,7 @@ void IndexGetCPU(const Tensor& src,
             CPUCopyObjectElementKernel(src, dst, object_byte_size);
         });
     } else {
-        DISPATCH_DTYPE_TO_TEMPLATE(dtype, [&]() {
+        DISPATCH_DTYPE_TO_TEMPLATE_WITH_BOOL(dtype, [&]() {
             LaunchAdvancedIndexerKernel(ai, CPUCopyElementKernel<scalar_t>);
         });
     }
@@ -91,7 +91,7 @@ void IndexSetCPU(const Tensor& src,
             CPUCopyObjectElementKernel(src, dst, object_byte_size);
         });
     } else {
-        DISPATCH_DTYPE_TO_TEMPLATE(dtype, [&]() {
+        DISPATCH_DTYPE_TO_TEMPLATE_WITH_BOOL(dtype, [&]() {
             LaunchAdvancedIndexerKernel(ai, CPUCopyElementKernel<scalar_t>);
         });
     }
diff --git a/cpp/open3d/core/kernel/IndexGetSetCUDA.cu b/cpp/open3d/core/kernel/IndexGetSetCUDA.cu
index ef0de0c1a..547e943ca 100644
--- a/cpp/open3d/core/kernel/IndexGetSetCUDA.cu
+++ b/cpp/open3d/core/kernel/IndexGetSetCUDA.cu
@@ -80,7 +80,7 @@ void IndexGetCUDA(const Tensor& src,
                     CUDACopyObjectElementKernel(src, dst, object_byte_size);
                 });
     } else {
-        DISPATCH_DTYPE_TO_TEMPLATE(dtype, [&]() {
+        DISPATCH_DTYPE_TO_TEMPLATE_WITH_BOOL(dtype, [&]() {
             LaunchAdvancedIndexerKernel(
                     src.GetDevice(), ai,
                     // Need to wrap as extended CUDA lambda function
@@ -108,7 +108,7 @@ void IndexSetCUDA(const Tensor& src,
                     CUDACopyObjectElementKernel(src, dst, object_byte_size);
                 });
     } else {
-        DISPATCH_DTYPE_TO_TEMPLATE(dtype, [&]() {
+        DISPATCH_DTYPE_TO_TEMPLATE_WITH_BOOL(dtype, [&]() {
             LaunchAdvancedIndexerKernel(
                     src.GetDevice(), ai,
                     // Need to wrap as extended CUDA lambda function
diff --git a/cpp/open3d/t/geometry/PointCloud.cpp b/cpp/open3d/t/geometry/PointCloud.cpp
index 96547c01a..04fa4f158 100644
--- a/cpp/open3d/t/geometry/PointCloud.cpp
+++ b/cpp/open3d/t/geometry/PointCloud.cpp
@@ -259,13 +259,11 @@ PointCloud PointCloud::SelectByIndex(const core::Tensor &indices,
     // The indices may have duplicate index value and will result in identity
     // point cloud attributes. We convert indices Tensor into mask Tensor to and
     // call SelectByMask to avoid this situation.
-    core::Tensor mask =
-            core::Tensor::Zeros({length}, core::Dtype::Int16, GetDevice());
+    core::Tensor mask = core::Tensor::Zeros({length}, core::Bool, GetDevice());
     mask.SetItem(core::TensorKey::IndexTensor(indices),
-                 core::Tensor(std::vector<int16_t>{1}, {}, core::Dtype::Int16,
-                              GetDevice()));
+                 core::Tensor::Init<bool>(true, GetDevice()));
 
-    PointCloud pcd = SelectByMask(mask.To(core::Dtype::Bool), invert);
+    PointCloud pcd = SelectByMask(mask, invert);
 
     utility::LogDebug("Pointcloud down sampled from {} points to {} points.",
                       length, pcd.GetPointPositions().GetLength());
```

After this, we avoid some copies and type conversions. Although, it is still 2x slower than the legacy on my machine.
___
*[`cpp/tests/t/geometry/PointCloud.cpp` line 707 at r3](https://reviewable.io/reviews/isl-org/open3d/5150#-N41etx3CITyB8TtLRjY:-N41etx3CITyB8TtLRjZ:b-4bevhs) ([raw file](https://github.com/isl-org/open3d/blob/8a755e81bd9dd3cbc33617bf37d207bb51af833c/cpp/tests/t/geometry/PointCloud.cpp#L707)):*
> ```cpp
>                                        {0.2, 0.4, 0.2}},
>                                       device));
>     const core::Tensor indices = core::Tensor::Init<int64_t>({0, 3}, device);
> ```

Add another test case to test duplicated indices, and test the `remove_duplicates=xxx` flag.

yuecideng (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/5150)* status: 4 of 5 files reviewed, 6 unresolved discussions (waiting on @yuecideng and @yxlao)

___
*[`cpp/open3d/t/geometry/PointCloud.h` line 327 at r3](https://reviewable.io/reviews/isl-org/open3d/5150#-N41d21iC_6wCkSoPZcI:-N42MdMb31z3-997xu_f:bb6un1j) ([raw file](https://github.com/isl-org/open3d/blob/8a755e81bd9dd3cbc33617bf37d207bb51af833c/cpp/open3d/t/geometry/PointCloud.h#L327)):*

Remove blank line.
</blockquote></details>

Done
___
*[`cpp/open3d/t/geometry/PointCloud.h` line 332 at r3](https://reviewable.io/reviews/isl-org/open3d/5150#-N41dhWB9-ZZvolH41iX:-N42NatZ1_-zo-sKDTnS:b8k8d8y) ([raw file](https://github.com/isl-org/open3d/blob/8a755e81bd9dd3cbc33617bf37d207bb51af833c/cpp/open3d/t/geometry/PointCloud.h#L332)):*

It might not always be desirable to remove duplicated indices automatically, e.g., people may expect the output point cloud length to be the same as the input point cloud length. 

Shall we add an additional parameter called `remove_duplicates = false`, and by default, we do not remove the duplicates? This can also improve the default performance.
</blockquote></details>

Agree with that because sometimes we need duplicated values. 

I have two considerations: 
1. Adding `remove_duplicates` will make it different with legacy version of `SelectByIndex`, which the `remove_duplicates`  is not suported. Should we make them compatible? 
2. I think `remove_duplicates` and `invert` would not be used together. That is , if user want to keep duplicates, he will never turn `invert` to true. So now that we have `SelectByMask`, which is also supprt `invert`, could we remove `invert` from `SelectByIndex` and obly keep `remove_duplicates` as arg. We can use `GetItem` of Tensor, which is faster than `IndexGet`, and use `HashSet` for duplicated indices removal. But one thing should be consider, if points size is larger than 1M and the the number of duplicated value is very small (eg. only 1 duplicated index), the performance of `Inset` is bad (I have ran the benchmark, it abaout 0.25 - 0.3s). Should we look at other hashing method?
___
*[`cpp/open3d/t/geometry/PointCloud.cpp` line 256 at r3](https://reviewable.io/reviews/isl-org/open3d/5150#-N41eSlV5NajAx_pkmeU:-N42MNZz1zeNKSPt9V5a:bb6un1j) ([raw file](https://github.com/isl-org/open3d/blob/8a755e81bd9dd3cbc33617bf37d207bb51af833c/cpp/open3d/t/geometry/PointCloud.cpp#L256)):*

Use `core::Int64` instead of `core::Dtype::xxx`. Same for the rest.
</blockquote></details>

Done
___
*[`cpp/open3d/t/geometry/PointCloud.cpp` line 265 at r3](https://reviewable.io/reviews/isl-org/open3d/5150#-N41eQW899D77XlMlG4h:-N42MmRC9ITHX4LPYEs5:bl7tc1w) ([raw file](https://github.com/isl-org/open3d/blob/8a755e81bd9dd3cbc33617bf37d207bb51af833c/cpp/open3d/t/geometry/PointCloud.cpp#L265)):*

We need some code change to allow `Bool` dtype. 

```diff
diff --git a/cpp/open3d/core/kernel/IndexGetSetCPU.cpp b/cpp/open3d/core/kernel/IndexGetSetCPU.cpp
index 329f78091..ca35e3aff 100644
--- a/cpp/open3d/core/kernel/IndexGetSetCPU.cpp
+++ b/cpp/open3d/core/kernel/IndexGetSetCPU.cpp
@@ -71,7 +71,7 @@ void IndexGetCPU(const Tensor& src,
             CPUCopyObjectElementKernel(src, dst, object_byte_size);
         });
     } else {
-        DISPATCH_DTYPE_TO_TEMPLATE(dtype, [&]() {
+        DISPATCH_DTYPE_TO_TEMPLATE_WITH_BOOL(dtype, [&]() {
             LaunchAdvancedIndexerKernel(ai, CPUCopyElementKernel<scalar_t>);
         });
     }
@@ -91,7 +91,7 @@ void IndexSetCPU(const Tensor& src,
             CPUCopyObjectElementKernel(src, dst, object_byte_size);
         });
     } else {
-        DISPATCH_DTYPE_TO_TEMPLATE(dtype, [&]() {
+        DISPATCH_DTYPE_TO_TEMPLATE_WITH_BOOL(dtype, [&]() {
             LaunchAdvancedIndexerKernel(ai, CPUCopyElementKernel<scalar_t>);
         });
     }
diff --git a/cpp/open3d/core/kernel/IndexGetSetCUDA.cu b/cpp/open3d/core/kernel/IndexGetSetCUDA.cu
index ef0de0c1a..547e943ca 100644
--- a/cpp/open3d/core/kernel/IndexGetSetCUDA.cu
+++ b/cpp/open3d/core/kernel/IndexGetSetCUDA.cu
@@ -80,7 +80,7 @@ void IndexGetCUDA(const Tensor& src,
                     CUDACopyObjectElementKernel(src, dst, object_byte_size);
                 });
     } else {
-        DISPATCH_DTYPE_TO_TEMPLATE(dtype, [&]() {
+        DISPATCH_DTYPE_TO_TEMPLATE_WITH_BOOL(dtype, [&]() {
             LaunchAdvancedIndexerKernel(
                     src.GetDevice(), ai,
                     // Need to wrap as extended CUDA lambda function
@@ -108,7 +108,7 @@ void IndexSetCUDA(const Tensor& src,
                     CUDACopyObjectElementKernel(src, dst, object_byte_size);
                 });
     } else {
-        DISPATCH_DTYPE_TO_TEMPLATE(dtype, [&]() {
+        DISPATCH_DTYPE_TO_TEMPLATE_WITH_BOOL(dtype, [&]() {
             LaunchAdvancedIndexerKernel(
                     src.GetDevice(), ai,
                     // Need to wrap as extended CUDA lambda function
diff --git a/cpp/open3d/t/geometry/PointCloud.cpp b/cpp/open3d/t/geometry/PointCloud.cpp
index 96547c01a..04fa4f158 100644
--- a/cpp/open3d/t/geometry/PointCloud.cpp
+++ b/cpp/open3d/t/geometry/PointCloud.cpp
@@ -259,13 +259,11 @@ PointCloud PointCloud::SelectByIndex(const core::Tensor &indices,
     // The indices may have duplicate index value and will result in identity
     // point cloud attributes. We convert indices Tensor into mask Tensor to and
     // call SelectByMask to avoid this situation.
-    core::Tensor mask =
-            core::Tensor::Zeros({length}, core::Dtype::Int16, GetDevice());
+    core::Tensor mask = core::Tensor::Zeros({length}, core::Bool, GetDevice());
     mask.SetItem(core::TensorKey::IndexTensor(indices),
-                 core::Tensor(std::vector<int16_t>{1}, {}, core::Dtype::Int16,
-                              GetDevice()));
+                 core::Tensor::Init<bool>(true, GetDevice()));
 
-    PointCloud pcd = SelectByMask(mask.To(core::Dtype::Bool), invert);
+    PointCloud pcd = SelectByMask(mask, invert);
 
     utility::LogDebug("Pointcloud down sampled from {} points to {} points.",
                       length, pcd.GetPointPositions().GetLength());
```

After this, we avoid some copies and type conversions. Although, it is still 2x slower than the legacy on my machine.
</blockquote></details>

Yes, I also see this code because originally I want to used `SetItem` with `Bool` Dtype, which is not supported yet.
___
*[`cpp/tests/t/geometry/PointCloud.cpp` line 707 at r3](https://reviewable.io/reviews/isl-org/open3d/5150#-N41etx3CITyB8TtLRjY:-N42VUR77jOXpmgPrm-c:b-kqs72u) ([raw file](https://github.com/isl-org/open3d/blob/8a755e81bd9dd3cbc33617bf37d207bb51af833c/cpp/tests/t/geometry/PointCloud.cpp#L707)):*

Add another test case to test duplicated indices, and test the `remove_duplicates=xxx` flag.
</blockquote></details>

Will add after determine the modification.

yxlao (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/5150)* status: 4 of 5 files reviewed, 6 unresolved discussions (waiting on @yuecideng and @yxlao)

___
*[`cpp/open3d/t/geometry/PointCloud.h` line 332 at r3](https://reviewable.io/reviews/isl-org/open3d/5150#-N41dhWB9-ZZvolH41iX:-N478S8KDqQDHNYZBW3-:bggjkhz) ([raw file](https://github.com/isl-org/open3d/blob/8a755e81bd9dd3cbc33617bf37d207bb51af833c/cpp/open3d/t/geometry/PointCloud.h#L332)):*
<details><summary><i>Previously, yuecideng (Yueci Deng) wrote…</i></summary><blockquote>

Agree with that because sometimes we need duplicated values. 

I have two considerations: 
1. Adding `remove_duplicates` will make it different with legacy version of `SelectByIndex`, which the `remove_duplicates`  is not suported. Should we make them compatible? 
2. I think `remove_duplicates` and `invert` would not be used together. That is , if user want to keep duplicates, he will never turn `invert` to true. So now that we have `SelectByMask`, which is also supprt `invert`, could we remove `invert` from `SelectByIndex` and obly keep `remove_duplicates` as arg. We can use `GetItem` of Tensor, which is faster than `IndexGet`, and use `HashSet` for duplicated indices removal. But one thing should be consider, if points size is larger than 1M and the the number of duplicated value is very small (eg. only 1 duplicated index), the performance of `Inset` is bad (I have ran the benchmark, it abaout 0.25 - 0.3s). Should we look at other hashing method?
</blockquote></details>

1. It is okay to be different than the legacy.
2. `invert` should precedence over `remove_duplicates`, since `invert` is a more important property than `remove_duplicates`.
   - When `invert=true`, the value of `remove_duplicates` is ignored.
   - When `invert=false`, we value of `remove_duplicates` is honored.

In this way, we can keep both parameters in the `SelectByIndex` function. 

We can investigate the hashing problem later. For now, it is good to just use the approach that you are using.
___
*[`cpp/open3d/t/geometry/PointCloud.cpp` line 265 at r3](https://reviewable.io/reviews/isl-org/open3d/5150#-N41eQW899D77XlMlG4h:-N47A83JEV3SuHFYJLYy:bzgmmj3) ([raw file](https://github.com/isl-org/open3d/blob/8a755e81bd9dd3cbc33617bf37d207bb51af833c/cpp/open3d/t/geometry/PointCloud.cpp#L265)):*
<details><summary><i>Previously, yuecideng (Yueci Deng) wrote…</i></summary><blockquote>

Yes, I also see this code because originally I want to used `SetItem` with `Bool` Dtype, which is not supported yet.
</blockquote></details>

Could you apply the changes above to the PR?

cpp/open3d/t/geometry/PointCloud.cpp
```diff
@@ -241,8 +241,7 @@ PointCloud PointCloud::SelectPoints(const core::Tensor &boolean_mask,
     PointCloud pcd(GetDevice());
     for (auto &kv : GetPointAttr()) {
         if (HasPointAttr(kv.first)) {
-            pcd.SetPointAttr(kv.first,
-                             kv.second.IndexGet({indices_local}).Clone());
+            pcd.SetPointAttr(kv.first, kv.second.IndexGet({indices_local}));
```
reyanshsolis, line 244: Not sure if `not cloning` is safe here. 

- Legacy : Eigen always makes a copy. 

- This is a `const` function, returning a `new` point cloud, so we might want to clone instead of sharing memory. 

- Although sharing memory offers some flexibility and performance gain, it is also potentially dangerous for user applications, where one is not careful about the shared memory. So, we are sharing memory, we should always mention it clearly in the documentation/docstrings.

@yxlao any suggestions here ?

cpp/benchmarks/t/geometry/PointCloud.cpp
```diff
@@ -252,9 +305,24 @@ ENUM_VOXELDOWNSAMPLE_BACKEND()
...
+BENCHMARK_CAPTURE(SelectByIndex,
+                  CPU(remove duplicates),
+                  true,
+                  core::Device("CPU:0"))
+        ->Unit(benchmark::kMillisecond);
+
 #ifdef BUILD_CUDA_MODULE
 BENCHMARK_CAPTURE(Transform, CUDA, core::Device("CUDA:0"))
         ->Unit(benchmark::kMillisecond);
+BENCHMARK_CAPTURE(SelectByIndex, CUDA, false core::Device("CUDA:0"))
```
reyanshsolis, line 320: add `,` after false.

cpp/benchmarks/t/geometry/PointCloud.cpp
```diff
@@ -252,9 +305,24 @@ ENUM_VOXELDOWNSAMPLE_BACKEND()
...
+        ->Unit(benchmark::kMillisecond);
+BENCHMARK_CAPTURE(SelectByIndex,
+                  CUDA(remove duplicates),
+                  true core::Device("CUDA:0"))
```
reyanshsolis, line 324: add `,` after true.

reyanshsolis (changes requested):

cpp/open3d/t/geometry/PointCloud.cpp
yuecideng: The `Tensor::IndexGet` will return a new tensor except when it is a 0-D tensor with nonzero indexer. And it is a `const` function as well. 
```cpp
Tensor Tensor::IndexGet(const std::vector<Tensor>& index_tensors) const {
    if (NumDims() == 0) {
        if (index_tensors.size() != 1) {
            utility::LogError(
                    "A 0-D tensor can only be indexed by a 0-D boolean tensor, "
                    "but got {} index tensors.",
                    index_tensors.size());
        }
        Tensor index_tensor = index_tensors[0];
        core::AssertTensorShape(index_tensor, {});
        core::AssertTensorDtype(index_tensor, core::Bool);

        if (index_tensor.IsNonZero()) {
            // E.g. np.array(5)[np.array(True)].
            return *this;
        } else {
            // E.g. np.array(5)[np.array(False)].
            // The output tensor becomes 1D of 0 element.
            return Tensor(/*shape=*/{0}, GetDtype(), GetDevice());
        }
    }

    AdvancedIndexPreprocessor aip(*this, index_tensors);
    Tensor dst = Tensor(aip.GetOutputShape(), dtype_, GetDevice());

    kernel::IndexGet(aip.GetTensor(), dst, aip.GetIndexTensors(),
                     aip.GetIndexedShape(), aip.GetIndexedStrides());

    return dst;
}
```

yxlao: I pushed a commit to make `Tensor::IndexGet` always return a new tensor, as it should have consistent behavior for all cases.

yxlao (approved):

yxlao:
Merging, as all issues have been addressed.

PR #5167, opened 2022-06-08
Offline SLAM Small Bug Fix

yuecideng:
1. Change `depth_max` -> `max_depth`.
2. Add `--vis` to control whether to visualize the result.

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/5167)

### DIFF
--- a/cpp/open3d/pipelines/odometry/Odometry.cpp
+++ b/cpp/open3d/pipelines/odometry/Odometry.cpp
@@ -154,7 +154,7 @@ static CorrespondenceSetPixelWise ComputeCorrespondence(
                         double d_t = *depth_t.PointerAt<float>(u_t, v_t);
                         if (!std::isnan(d_t) &&
                             std::abs(transformed_d_s - d_t) <=
-                                    option.max_depth_diff_) {
+                                    option.depth_diff_max_) {
                             AddElementToCorrespondenceMap(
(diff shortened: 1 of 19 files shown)


yxlao (changes requested):
image
image

There are some inconsistencies across the library regarding whether to use `depth_max` or `max_depth`. A quick search shows that `depth_max` is more often used than `max_depth`. I think `depth_max` is preferred, as it is consistent with `depth_scale`. 

Could you do a search-replace to convert all `max_depth` to `depth_max` for everything related to the reconstruction system?

yuecideng:
> Could you do a search-replace to convert all max_depth to depth_max for everything related to the reconstruction system? For other use cases, e.g. the octree, we don't need to change

No problem, but should we change them in `OdometryOption`? Because I notice that there are both `max_depth`, `min_depth` and `max_depth_diff`. Should we also change `min_depth` -> `depth_min` and `max_depth_diff` -> `depth_diff_min`?

yxlao:
> No problem, but should we change them in OdometryOption? Because I notice that there are both max_depth, min_depth and max_depth_diff. Should we also change min_depth -> depth_min and max_depth_diff -> depth_diff_min?

Yes, that sounds good to me.

yxlao (approved):
LGTM after CI.

PR #5182, opened 2022-06-11
Add SYCL device and memory manager

yxlao:
Supported classes:
- `MemoryManager`
- `Blob`
- `Device`
- `core::sycl_utils` namespace

Refer to `3rdparty/README_SYCL.md` on how to setup the SYCL environment. To use SYCL GPU, you also need to install the proprietary Intel GPGPU driver: https://dgpu-docs.intel.com/installation-guides/ubuntu/ubuntu-focal.html.

This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/5182)


3rdparty/README_SYCL.md
```diff
@@ -56,6 +57,29 @@ referring to "DPC++".
...
+[opencl:cpu:1] Intel(R) OpenCL, 12th Gen Intel(R) Core(TM) i9-12900K
+[opencl:gpu:2] Intel(R) OpenCL HD Graphics, Intel(R) UHD Graphics 770
+[ext_oneapi_level_zero:gpu:0] Intel(R) Level-Zero, Intel(R) UHD Graphics 770
+[host:host:0] SYCL host platform, SYCL host device
+```
+
+Open3D is designed to make use of the SYCL GPU devices.
+- `core::Device("SYCL:0")` maps the default SYCL GPU returned by
+  `sycl::gpu_selector()`.
+  - For Intel GPUs, the "level-zero GPU" (`ext_oneapi_level_zero:gpu:0` in the
```
reyanshsolis, line 77: is this intended to be a sub-point?

CMakeLists.txt
```diff
@@ -70,6 +70,7 @@ if(BUILD_SYCL_MODULE)
 else()
     option(GLIBCXX_USE_CXX11_ABI  "Set -D_GLIBCXX_USE_CXX11_ABI=1"           OFF)
 endif()
+option(ENABLE_SYCL_UNIFIED_SHARED_MEMORY "Enable SYCL unified shared memory" OFF)
```
reyanshsolis, line 73: Will the use see an error if:
1. Both CUDA and SYCL flags are ON.
2. CUDA is OFF but CACHED_CUDA_MANAGER is ON.
3. SYCL is off but SYCL_UNIFIED_SHARED_MEMORY is ON.

reyanshsolis:
Please add `SYCL Support` in `open3d_print_configuration_summary`. 
Also, we may show the status of sycl unified memory or cuda cached mem manager in the build summary.

3rdparty/README_SYCL.md
```diff
@@ -27,19 +28,19 @@ referring to "DPC++".
...
    ```bash
    which icx        # /opt/intel/oneapi/compiler/<version>/linux/bin/icx
    which icpx       # /opt/intel/oneapi/compiler/<version>/linux/bin/icpx
    which python     # ${HOME}/.conda/envs/sycl/bin/python
    python --version # Python 3.8.12 :: Intel Corporation
    ```
-4. Config and build
+5. Config and build
    ```bash
    cmake -DBUILD_SYCL_MODULE=ON -DCMAKE_C_COMPILER=icx -DCMAKE_CXX_COMPILER=icpx ..
```
reyanshsolis, line 40: we may add `-DBUILD_UNIT_TESTS=ON`, as it is required to build tests (used for demo).

cpp/open3d/core/SYCLUtils.h
```diff
@@ -24,14 +24,41 @@
...
+
+#include "open3d/core/Device.h"
+
 namespace open3d {
 namespace core {
+namespace sycl_utils {
 
 /// Runs simple SYCL test program for sanity checks.
 /// \return Retuns 0 if successful.
 int SYCLDemo();
```
reyanshsolis, line 45: While this seems to be an important function for testing during development. This might not belong to core functionality. 
`IsSYCLWorkingAsExpected` or something like that might be better. Also, return True (boolean) if successful. 
Or can we move this implementation to our `test`.

cpp/pybind/core/sycl_utils.cpp
```diff
@@ -31,7 +31,9 @@
 namespace open3d {
 namespace core {
 
-void pybind_sycl_utils(py::module& m) { m.def("sycl_demo", &SYCLDemo); }
+void pybind_sycl_utils(py::module& m) {
```
reyanshsolis, line 34: same.

examples/cpp/SYCLDemo.cpp
```diff
@@ -24,9 +24,9 @@
 // IN THE SOFTWARE.
 // ----------------------------------------------------------------------------
 
-#include "open3d/core/TensorFunction.h"
+#include "open3d/core/SYCLUtils.h"
 
 int main() {
-    open3d::core::SYCLDemo();
+    open3d::core::sycl_utils::SYCLDemo();
```
reyanshsolis, line 30: It seems this example is more of a test than an example. We may remove this if `test` is covering this. 
Since Tensor will cover the implementation, an SYCL example on its own might not make much sense. 

We may add `sycl` device in core::Tensor documentation/python example showing tensor copy between CPU, CUDA and SYCL.

cpp/tests/core/SYCLUtils.cpp
```diff
@@ -34,12 +37,55 @@
 namespace open3d {
 namespace tests {
 
-class SYCLUtilsPermuteDevices : public PermuteDevices {};
-INSTANTIATE_TEST_SUITE_P(SYCLUtils,
-                         SYCLUtilsPermuteDevices,
-                         testing::ValuesIn(PermuteDevices::TestCases()));
+TEST(SYCLUtils, SYCLDemo) { core::sycl_utils::SYCLDemo(); }
+
+TEST(SYCLUtils, PrintAllSYCLDevices) {
```
reyanshsolis, line 42: running test with `"*SYCL*"` only runs a single test for me. 
other tests are not running. 
Also, I have an integrated gpu, so should I expect SYCL to run on both CPU and integrated GPU ?
```
build(master)$ ./bin/tests --gtest_filter="*SYCL*"
[Open3D INFO] CompilerInfo: C++ 17, IntelLLVM 2022.1.0, CUDA disabled.
[Open3D INFO] CPUInfo: 8 cores, 16 threads.
[Open3D INFO] ISAInfo: AVX512SKX instruction set used in ISPC code.
Note: Google Test filter = *SYCLUtils*
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from SYCLUtils/SYCLUtilsPermuteDevices
[ RUN      ] SYCLUtils/SYCLUtilsPermuteDevices.SYCLDemo/0
[Open3D INFO] SYCLDemo passed!
[       OK ] SYCLUtils/SYCLUtilsPermuteDevices.SYCLDemo/0 (116 ms)
[----------] 1 test from SYCLUtils/SYCLUtilsPermuteDevices (116 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (116 ms total)
[  PASSED  ] 1 test.
```

cpp/open3d/core/Device.cpp
```diff
@@ -110,16 +113,22 @@ std::string Device::ToString() const {
...
+            str += "SYCL";
+            break;
         default:
             utility::LogError("Unsupported device type");
     }
     str += ":" + std::to_string(device_id_);
     return str;
 }
 
+Device::DeviceType Device::GetType() const { return device_type_; }
```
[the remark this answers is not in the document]
yxlao: Done. Updated GetType() and GetID\().

cpp/open3d/core/MemoryManager.cpp
```diff
@@ -63,18 +63,30 @@ void MemoryManager::Memcpy(void* dst_ptr,
...
+    else if (src_device.IsCPU() && dst_device.IsSYCL()) {
         device_mm = GetMemoryManagerDevice(dst_device);
+    } else if (src_device.IsSYCL() && dst_device.IsCPU()) {
+        device_mm = GetMemoryManagerDevice(src_device);
+    } else if (src_device.IsSYCL() && dst_device.IsSYCL()) {
+        device_mm = GetMemoryManagerDevice(src_device);
+    }
+    // Not supporting other combinations at the moment, e.g. SYCL->CUDA.
+    else {
+        utility::LogError("Unsupported device type.");
```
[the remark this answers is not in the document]
yxlao: Done, updated gettype and getid.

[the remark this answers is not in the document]
yxlao: Done.

cpp/open3d/core/MemoryManagerSYCL.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+void MemoryManagerSYCL::Memcpy(void* dst_ptr,
+                               const Device& dst_device,
+                               const void* src_ptr,
+                               const Device& src_device,
+                               size_t num_bytes) {
+    Device device_with_queue;
+
+    if (src_device.IsCPU() && dst_device.IsCPU()) {
+        utility::LogError("Wrong device {}->{}.", src_device.ToString(),
```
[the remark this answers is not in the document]
yxlao: Updated, comment as "internal error".

CMakeLists.txt
yxlao: 1. Added the check. 
2. It doesn't matter. CACHED_CUDA_MANAGER will have no effect in this case.
3. It doesn't matter. SYCL_UNIFIED_SHARED_MEMORY will have no effect in this case.

yxlao:
> Please add SYCL Support in open3d_print_configuration_summary.
Also, we may show the status of sycl unified memory or cuda cached mem manager in the build summary.

Both added.

3rdparty/README_SYCL.md
yxlao: Addeded.

examples/cpp/SYCLDemo.cpp
yxlao: We need the example since the cmake is different between example and test. This is used to ensure that the SYCL build system is working as expected. The pybind is the same idea.

Once we have actual SYCL implementation, e.g. SYCL Tensor support, then we can remove this example and replace it with the SYCL tensor example.

cpp/open3d/core/SYCLUtils.h
yxlao: > While this seems to be an important function for testing during development. This might not belong to core functionality.

This is true. We can rename it to have a better name. For CUDA, we are using `open3d::core::cuda`. However, the official SYCL library is already using the `sycl::` namespace, so it will cause confusion if we call it `open3d::core::sycl`. Do you have a suggestion for the namespace name?

> `IsSYCLWorkingAsExpected`. 

This is the `IsAvailable()` function. We are following the Torch's convention: [torch.cuda.is_available](https://pytorch.org/docs/stable/generated/torch.cuda.is_available.html) .

yxlao: For the utils, to be consistent to our user, i think it is the best to use:
```cpp
open3d::core::cuda::xxx();
open3d::core::sycl::xxx();
```

Then, internally, we refer to the official SYCL namespace as:
```cpp
using sy = cl::sycl;
```

yxlao: image

Intel's SYCL implementation is using `namespace s = cl::sycl;`. We can do two letters `sy::` to be more clear.

reyanshsolis (approved):
> Reviewed 9 of 22 files at r1, 1 of 7 files at r2, 1 of 1 files at r3, 5 of 7 files at r4, 9 of 9 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/5182)* status: all files reviewed, 3 unresolved discussions (waiting on @ssheorey and @yxlao)

___
*[`CMakeLists.txt` line 73 at r2](https://reviewable.io/reviews/isl-org/open3d/5182#-N4WhlVU56K4Q21pJ-jv-r2-73:-N6GxMvgDv0UhfpQ_xol:b-eapff3) ([raw file](https://github.com/isl-org/open3d/blob/8df220311bb6f77aed9112fd472c0126326e13bb/CMakeLists.txt#L73)):*

1. Added the check. 
2. It doesn't matter. CACHED_CUDA_MANAGER will have no effect in this case.
3. It doesn't matter. SYCL_UNIFIED_SHARED_MEMORY will have no effect in this case.
</blockquote></details>

Resolved.
___
*[`cpp/open3d/core/SYCLUtils.h` line 45 at r2](https://reviewable.io/reviews/isl-org/open3d/5182#-N4H0jYU0X8p9ba43lmz-r2-45:-N6GzaMP0A9lMvjpn5Lk:bipfekd) ([raw file](https://github.com/isl-org/open3d/blob/8df220311bb6f77aed9112fd472c0126326e13bb/cpp/open3d/core/SYCLUtils.h#L45)):*

image

Intel's SYCL implementation is using `namespace s = cl::sycl;`. We can do two letters `sy::` to be more clear.
</blockquote></details>

This `open3d::core::sycl::xxx();` sounds consistent to me. I agree regarding the namespace convention. Thanks for presenting the detailed explanation.
___
*[`cpp/tests/core/SYCLUtils.cpp` line 81 at r5](https://reviewable.io/reviews/isl-org/open3d/5182#-N6H0CqO680PJ86XNbK3:-N6H0CqO680PJ86XNbK4:bi4m5jq) ([raw file](https://github.com/isl-org/open3d/blob/13d5f1060b0ba6fb3abb89905aca5438d678f4d1/cpp/tests/core/SYCLUtils.cpp#L81)):*
> ```cpp
>     core::MemoryManager::Memcpy(sycl_ptr, sycl_device, host_ptr, host_device,
>                                 byte_size);
>     for (int i = 0; i < 4; i++) {
> ```

Is it for a demo explanation regarding where it will fail?
We can even put it in EXPECT_THROW or something similar. I am good with keeping this, just wanted to confirm.

yxlao (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/5182)* status: all files reviewed, 3 unresolved discussions (waiting on @reyanshsolis, @ssheorey, and @yxlao)

___
*[`cpp/open3d/core/SYCLUtils.h` line 45 at r2](https://reviewable.io/reviews/isl-org/open3d/5182#-N4H0jYU0X8p9ba43lmz-r2-45:-N6JDQkO9hTbi_BFKkB3:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/8df220311bb6f77aed9112fd472c0126326e13bb/cpp/open3d/core/SYCLUtils.h#L45)):*

This `open3d::core::sycl::xxx();` sounds consistent to me. I agree regarding the namespace convention. Thanks for presenting the detailed explanation.
</blockquote></details>

Done.
___
*[`cpp/tests/core/SYCLUtils.cpp` line 81 at r5](https://reviewable.io/reviews/isl-org/open3d/5182#-N6H0CqO680PJ86XNbK3:-N6JDRkj11ppIVUvoxh-:blb5xl4) ([raw file](https://github.com/isl-org/open3d/blob/13d5f1060b0ba6fb3abb89905aca5438d678f4d1/cpp/tests/core/SYCLUtils.cpp#L81)):*

Is it for a demo explanation regarding where it will fail?
We can even put it in EXPECT_THROW or something similar. I am good with keeping this, just wanted to confirm. 
</blockquote></details>

`EXPECT_THROW` does not protect against segfault.

PR #5186, opened 2022-06-13
Legacy offline reconstruction C++ Runner

yuecideng:
initialize the workflow for the whole system.

Will make it ready for review when complete the migration of python version and performance test.

Update:
The whole system is nearly identical with python version of legacy reconstruction system, except the `PoseEstiamtion` part, where the python version uses opencv orb to find correspodences, and the cpp version uses fpfh to do this.

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/5186)


yuecideng (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/5186)* status: all files reviewed, 10 unresolved discussions (waiting on @theNded, @yuecideng, and @yxlao)

___
*[`examples/cpp/CMakeLists.txt` line 64 at r2](https://reviewable.io/reviews/isl-org/open3d/5186#-N54MHme1AUDqVWZ5z_F:-N54sTNf90KF161sVMQI:bmq3s07) ([raw file](https://github.com/isl-org/open3d/blob/c40795e146241405672e9e59ac208fb817f82470/examples/cpp/CMakeLists.txt#L64)):*

Not needing `if (TARGET Open3D::3rdparty_openmp)` seems to work for your case. Could you remove this in `TrimMeshBasedOnPointCloud` as well and see if the CI passes?
</blockquote></details>

Do you mean only keeping this line `open3d_add_example(TrimMeshBasedOnPointCloud Open3D::3rdparty_openmp)`?
___
*[`examples/cpp/LegacyReconstructionUtil.h` line 50 at r2](https://reviewable.io/reviews/isl-org/open3d/5186#-N54RpdnBAicp5_harMG:-N54tY9j7-OjfG1f_jcW:b-6gcjvq) ([raw file](https://github.com/isl-org/open3d/blob/c40795e146241405672e9e59ac208fb817f82470/examples/cpp/LegacyReconstructionUtil.h#L50)):*

We can move this to `cpp/open3d/utility/Timer.h`. 

For example:
```
double Timer::GetDurationInMillisecond() const; // Replace the old one. The old one is not clear about the unit.
double Timer::GetDurationInSecond() const; // Seconds are useful sometimes.
std::tuple<int, int, int> Timer::GetDurationInHMS() const; // Shall these be int?
```

</blockquote></details>

Maybe `std::tuple<int, int, double> Timer::GetDurationInHMS() const;`? Because somtimes we would like to see second in more precisely.

yuecideng (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/5186)* status: all files reviewed, 10 unresolved discussions (waiting on @theNded, @yuecideng, and @yxlao)

___
*[`examples/cpp/LegacyReconstructionUtil.h` line 42 at r2](https://reviewable.io/reviews/isl-org/open3d/5186#-N54ReO3BCiXIJ6_zLTI:-N564wM61CQo0i9XMF3E:bb6un1j) ([raw file](https://github.com/isl-org/open3d/blob/c40795e146241405672e9e59ac208fb817f82470/examples/cpp/LegacyReconstructionUtil.h#L42)):*

`fmt::` can do this, e.g. https://stackoverflow.com/q/64305897/1255535. 
</blockquote></details>

Done
___
*[`examples/cpp/LegacyReconstructionUtil.h` line 50 at r2](https://reviewable.io/reviews/isl-org/open3d/5186#-N54RpdnBAicp5_harMG:-N564z0_31qkC9sGxVU2:bb6un1j) ([raw file](https://github.com/isl-org/open3d/blob/c40795e146241405672e9e59ac208fb817f82470/examples/cpp/LegacyReconstructionUtil.h#L50)):*
<details><summary><i>Previously, yuecideng (Yueci Deng) wrote…</i></summary><blockquote>

Maybe `std::tuple<int, int, double> Timer::GetDurationInHMS() const;`? Because somtimes we would like to see second in more precisely. 
</blockquote></details>

Done
___
*[`examples/cpp/LegacyReconstructionUtil.h` line 71 at r2](https://reviewable.io/reviews/isl-org/open3d/5186#-N54ljxU6hWgZ-hgWUAH:-N5650XY2WaYyYW8uQI_:bb6un1j) ([raw file](https://github.com/isl-org/open3d/blob/c40795e146241405672e9e59ac208fb817f82470/examples/cpp/LegacyReconstructionUtil.h#L71)):*

Same, `fmt` can do that. 

We use fmt library globally in Open3D, even in headers (since `Logging.h` is using it).
</blockquote></details>

Done
___
*[`examples/cpp/LegacyReconstructionUtil.h` line 77 at r2](https://reviewable.io/reviews/isl-org/open3d/5186#-N54m-m-D6268ETMu1ER:-N5653t03Gp5Di8-4pMw:bb6un1j) ([raw file](https://github.com/isl-org/open3d/blob/c40795e146241405672e9e59ac208fb817f82470/examples/cpp/LegacyReconstructionUtil.h#L77)):*

Move to `cpp/open3d/utility/FileSystem.h` and implement with https://en.cppreference.com/w/cpp/filesystem/path/append
</blockquote></details>

Done
___
*[`examples/cpp/LegacyReconstructionUtil.h` line 86 at r2](https://reviewable.io/reviews/isl-org/open3d/5186#-N54m5Iq2dbzSctKM47t:-N5654x28c9Yrd03rOeQ:bb6un1j) ([raw file](https://github.com/isl-org/open3d/blob/c40795e146241405672e9e59ac208fb817f82470/examples/cpp/LegacyReconstructionUtil.h#L86)):*

Same, move to FileSystem.h/.cpp.
</blockquote></details>

Done
___
*[`examples/cpp/LegacyReconstructionUtil.h` line 94 at r2](https://reviewable.io/reviews/isl-org/open3d/5186#-N54mOoN79YBkq-kYZQ6:-N5656JZ3pcgqusMopRG:bpmh3yw) ([raw file](https://github.com/isl-org/open3d/blob/c40795e146241405672e9e59ac208fb817f82470/examples/cpp/LegacyReconstructionUtil.h#L94)):*

This can be potentially implemented with https://stackoverflow.com/a/612176/1255535. Also move to FileSystem.h.
</blockquote></details>

It seems that the original implementation is more straightforward.

yxlao:
We have some separate discussions today:

1. We can move the example to the `cpp/apps` folder, as this is larger scale than an "example". We are providing an "application" for people to use with many config options. When we move to the `cpp/apps` folder, it is possible to create multiple files, multiple headers, making the code more structured. 
2. We can move most functionality to `open3d::pipelines::reconstruction` namespace. The main function in the `cpp/apps` folder only serves as argument parsing and driver for the code. The main processing is done inside `open3d::pipelines::reconstruction`.
3. After 2. is done, we shall be ready to pybind `open3d::pipelines::reconstruction` and make the python-based reconstruction pipeline to use the same code. This will reduce code duplication.

We can achieve these step by step. For this PR, we may do step 1 first. We can discuss the design of 2. and 3. in future meetings to design it, before actual implementation.

theNded:
> We have some separate discussions today:
> 
> 1. We can move the example to the `cpp/apps` folder, as this is larger scale than an "example". We are providing an "application" for people to use with many config options. When we move to the `cpp/apps` folder, it is possible to create multiple files, multiple headers, making the code more structured.
> 2. We can move most functionality to `open3d::pipelines::reconstruction` namespace. The main function in the `cpp/apps` folder only serves as argument parsing and driver for the code. The main processing is done inside `open3d::pipelines::reconstruction`.
> 3. After 2. is done, we shall be ready to pybind `open3d::pipelines::reconstruction` and make the python-based reconstruction pipeline to use the same code. This will reduce code duplication.
> 
> We can achieve these step by step. For this PR, we may do step 1 first. We can discuss the design of 2. and 3. in future meetings to design it, before actual implementation.

I disagree with 2 and 3. Python itself has a lot of flexible packages, from `configargparse` to torch and numpy. They are good for fast prototyping and debugging.

If we ship yet another C++ pybind layer upon the existing pipeline modules, to maintain the same usability, we have to spend 10x time on C++, while the users will probably still stick to native python packages. We have to composite docs which are already lacking.

IMO it is better to spend time on refining modularized pipeline components (odometry, integrate, slac) than adding another layer to the existing layers.

yuecideng:
1. Move `LegacyReconstruction` to `cpp/apps` folder and change name into `OfflineReconstruction`, which make it compatible wit h tensor based offline reconstruction in the future.
2. Seperate some helper functions into several header files to make the code more structured.
3. Currently the `OfflineReconstruction` command line executable can be found in `/bin/OfflineReconstruction/`. The output path I think should be designed for a better place.
4. Currently the installation is not supported.

cpp/apps/CMakeLists.txt
```diff
@@ -119,6 +119,46 @@ macro(open3d_add_app SRC_DIR APP_NAME TARGET_NAME)
     endif()
 endmacro()
 
+macro(open3d_add_app_cmd SRC_DIR APP_NAME)
```
yxlao, line 122:
```
macro(open3d_add_app_common SRC_DIR APP_NAME TARGET_NAME)

endmacro()

target_sources(xxx PRIVATE
    xxx.cpp
    xxx.cpp
)
```

yuecideng: It seems that we must add source file in `add_executable`, so I use the original method (`GLOB`) to sum up the source files inside `open3d_add_app_common `

yxlao (approved):

yxlao (approved):

Issue #4849, opened 2022-03-05
Implement UniformDownSample for t::geometry::PointCloud


heethesh:
### Checklist

- [X] I have searched for [similar issues](https://github.com/isl-org/Open3D/issues).
- [X] For Python issues, I have tested with the [latest development wheel](http://www.open3d.org/docs/latest/getting_started.html#development-version-pip).
- [X] I have checked the [release documentation](http://www.open3d.org/docs/release/) and the [latest documentation](http://www.open3d.org/docs/latest/) (for `master` branch).

### Proposed new feature or change

Which one of these tensor operations would be useful to sample every-k points and the most efficient (more interested with CUDA over CPU) here: `Slice`, `IndexGet`, or `AsStrided`?

PR #5202, opened 2022-06-15
Add uniform and random down sample methods in Tensor PointCloud

yuecideng:
**Update**
Use tensor slicing stead of indexing for uniform sampling, which makes it more faster.

This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/5202)


yuecideng:
image

cpp/open3d/t/geometry/PointCloud.cpp
```diff
@@ -314,6 +314,44 @@ PointCloud PointCloud::VoxelDownSample(
...
+    for (auto &kv : GetPointAttr()) {
+        pcd_down.SetPointAttr(
+                kv.first,
+                kv.second.Slice(0, 0, length, (int64_t)every_k_points));
+    }
+
+    return pcd_down;
+}
+
+PointCloud PointCloud::RandomDownSample(double sampling_ratio) const {
```
ssheorey, line 335: 1. For reproducible experiments, we need the ability to set the random seed for all RNG calls by Open3D. One way to achieve this is with a singleton RNG, with some care for concurrent access. 
2. CUDA has an RNG that will likely be faster for CUDA tensors.

reyanshsolis:
I appreciate the effort, but I was wondering if there is any use case for this functionality?

Here `uniform` is misleading, as the point cloud is unorganized data, so removing every k-th point, will not downsample the point cloud uniformly.

Do we have a reference implementation/insipiration of such functionality in PCL or other 3D data processing packages?

reyanshsolis:
Voxel down-sampling is actually a uniform down-sample in a way. 
However, it lacks the option of `averaging or sampling`, i.e. control if the new point attributes are the mean of the points and their attributes within a voxel, or it's one of the existing points (sampled). 

I guess users might appreciate an improved voxel-down-sample functionality over other filtering options.

yuecideng:
The `uniform` here means taking a subset of point cloud with the same interval w.r.t original order. The use case is to provide a simple interface for user to fast down sample a very large pcd (eg. 10M points, particularly for industrial scenario, this can be commomly seen). Sometimes it is no worth to use `VoxelDownSample` when consider time comsuming. Beside, we also provide this function in legacy version.

> However, it lacks the option of averaging or sampling, i.e. control if the new point attributes are the mean of the points and their attributes within a voxel, or it's one of the existing points (sampled).

I agree with this because currenly the voxel down sample method in tensor based pcd only return coordinates of voxels, leading to the inconsistent with legacy version and other library's behaviour.

cpp/open3d/t/geometry/PointCloud.h
```diff
@@ -334,11 +334,24 @@ class PointCloud : public Geometry, public DrawableGeometry {
                              bool remove_duplicates = false) const;
 
     /// \brief Downsamples a point cloud with a specified voxel size.
+    ///
     /// \param voxel_size Voxel size. A positive number.
     PointCloud VoxelDownSample(double voxel_size,
                                const core::HashBackendType &backend =
                                        core::HashBackendType::Default) const;
 
+    /// \brief Downsamples a point cloud uniformly
```
reyanshsolis, line 343: please elaborate on the description such as:
Downsamples a point cloud, by selecting every kth index point and its attributes.

Similarly for RandomDownSample

cpp/pybind/t/geometry/pointcloud.cpp
```diff
@@ -198,6 +198,18 @@ The attributes of the point cloud have different levels::
             },
             "Downsamples a point cloud with a specified voxel size.",
             "voxel_size"_a);
+    pointcloud.def(
+            "uniform_down_sample",
```
reyanshsolis, line 202: You may use the following:
Instead of 
```cpp
pointcloud.def(
            "uniform_down_sample",
            [](const PointCloud& pointcloud, const size_t every_k_points) {
                return pointcloud.UniformDownSample(every_k_points);
            },
            "Downsamples a point cloud uniformly.", "every_k_points"_a);
  ```
  you may try:
  ```cpp
 pointcloud.def(
            "uniform_down_sample", &UniformDownSample,
            "Downsamples a point cloud uniformly.", "every_k_points"_a);
  ```
  
  similarly to other functions. 
  
  Also, test the same if it is working, you may find similar use in pybind folder for reference.

reyanshsolis (changes requested):

reyanshsolis (changes requested):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/5202)* status: 0 of 5 files reviewed, 5 unresolved discussions (waiting on @yuecideng and @yxlao)

___
*[`cpp/tests/t/geometry/PointCloud.cpp` line 770 at r2](https://reviewable.io/reviews/isl-org/open3d/5202#-N4wkdzR9ngvdOKKXZjb:-N4wkdzSB2_y8Pp17trF:b-5tuvu9) ([raw file](https://github.com/isl-org/open3d/blob/710a8a76dfdfa73c4766c2e0ab48df7757047697/cpp/tests/t/geometry/PointCloud.cpp#L770)):*
> ```cpp
>     core::Device device = GetParam();
> 
>     // Value test
> ```

Add a `.`
___
*[`cpp/tests/t/geometry/PointCloud.cpp` line 789 at r2](https://reviewable.io/reviews/isl-org/open3d/5202#-N4wkhWsBQbXq1hOed4y:-N4wkhWsBQbXq1hOed4z:bb7od0r) ([raw file](https://github.com/isl-org/open3d/blob/710a8a76dfdfa73c4766c2e0ab48df7757047697/cpp/tests/t/geometry/PointCloud.cpp#L789)):*
> ```cpp
>     core::Device device = GetParam();
> 
>     // Value test
> ```

same

cpp/open3d/t/geometry/PointCloud.h
yuecideng: Done

cpp/pybind/t/geometry/pointcloud.cpp
yuecideng: Down

yuecideng (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/5202)* status: 0 of 5 files reviewed, 5 unresolved discussions (waiting on @reyanshsolis, @yuecideng, and @yxlao)

___
*[`cpp/tests/t/geometry/PointCloud.cpp` line 770 at r2](https://reviewable.io/reviews/isl-org/open3d/5202#-N4wkdzR9ngvdOKKXZjb:-N4ycaRo8p38hx0KnPux:bb6un1j) ([raw file](https://github.com/isl-org/open3d/blob/710a8a76dfdfa73c4766c2e0ab48df7757047697/cpp/tests/t/geometry/PointCloud.cpp#L770)):*

Add a `.`
</blockquote></details>

Done
___
*[`cpp/tests/t/geometry/PointCloud.cpp` line 789 at r2](https://reviewable.io/reviews/isl-org/open3d/5202#-N4wkhWsBQbXq1hOed4y:-N4yccHt7YiB0Fd7pIr0:bb6un1j) ([raw file](https://github.com/isl-org/open3d/blob/710a8a76dfdfa73c4766c2e0ab48df7757047697/cpp/tests/t/geometry/PointCloud.cpp#L789)):*

same
</blockquote></details>

Done

heethesh:
#4849

yuecideng:
Update with using open3d random singleton. CUDA implementation will be added in the future PR.

yuecideng (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/5202)* status: 4 of 5 files reviewed, 8 unresolved discussions (waiting on @reyanshsolis, @ssheorey, @yuecideng, and @yxlao)

___
*[`cpp/pybind/t/geometry/pointcloud.cpp` line 208 at r5](https://reviewable.io/reviews/isl-org/open3d/5202#-N5Zfc2F1jU-gRTs1CDz:-N5ZqF6N2CCcKBkBKX8P:b8j0ai4) ([raw file](https://github.com/isl-org/open3d/blob/929a374c829dd88db230ee20e306316eb5169928/cpp/pybind/t/geometry/pointcloud.cpp#L208)):*

We need to call `docstring::ClassMethodDocInject` to enable the docs. In our case, we need to document `sampling_ratio` by passing in a dictionary. 
</blockquote></details>

Should we also add `docstring::ClassMethodDocInject` to the recently added functions, like `SelectByMask`, `SelectByIndex`, `RemoveRadiusOutliers`? I think it is better to add them all.

yxlao (changes requested):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/5202)* status: 3 of 5 files reviewed, 9 unresolved discussions (waiting on @reyanshsolis, @ssheorey, @yuecideng, and @yxlao)

___
*[`cpp/open3d/t/geometry/PointCloud.cpp` line 359 at r6](https://reviewable.io/reviews/isl-org/open3d/5202#-N5c0vwHEPVFRJ67_6zm:-N5c0vwHEPVFRJ67_6zn:bar5vgn) ([raw file](https://github.com/isl-org/open3d/blob/67c315a5733b88aa9031c2997020a4789852232c/cpp/open3d/t/geometry/PointCloud.cpp#L359)):*
> ```cpp
>                 false, false);
>     } else {
>         utility::LogError("RandomDownSample is not supported on GPU.");
> ```

LGTM, after fixing this. We can copy the indices from CPU to CUDA for now. This is required to pass the CUDA CI test.

yxlao (changes requested):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/5202)* status: 3 of 5 files reviewed, 9 unresolved discussions (waiting on @reyanshsolis, @ssheorey, @yuecideng, and @yxlao)

yxlao (approved):

PR #5211, opened 2022-06-17
Fix data path and termination in headless rendering python example

ShreyanshDarshan:
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/5211)


reyanshsolis (changes requested):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/5211)* status: 0 of 1 files reviewed, 4 unresolved discussions (waiting on @reyanshsolis and @ShreyanshDarshan)

___
*[`examples/python/visualization/headless_rendering.py` line 33 at r2](https://reviewable.io/reviews/isl-org/open3d/5211#-N4wh8uUDVc4Bi75na6H:-N4wh8uUDVc4Bi75na6I:b-ydgra7) ([raw file](https://github.com/isl-org/open3d/blob/52d5710eae4f1399ca0127902251fedb05231f13/examples/python/visualization/headless_rendering.py#L33)):*
> ```Python
> 
> pyexample_path = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
> test_data_path = os.path.join(os.path.dirname(pyexample_path), 'test_data')
> ```

don't use the `test_data` path. 

I would suggest to get `open3d_data` path, and create a `open3d_data` / `example_results` / `example_name` / `image` and `depth`.
`example_name` here might be `visualization_headless_rendering`.

Also, print the path of output dirs.
___
*[`examples/python/visualization/headless_rendering.py` line 62 at r2](https://reviewable.io/reviews/isl-org/open3d/5211#-N4wgJTBDTDWW_T598BJ:-N4wgJTBDTDWW_T598BK:b-omov4g) ([raw file](https://github.com/isl-org/open3d/blob/52d5710eae4f1399ca0127902251fedb05231f13/examples/python/visualization/headless_rendering.py#L62)):*
> ```Python
>         if glb.index >= 0:
>             print("Capture image {:05d}".format(glb.index))
>             depth = vis.capture_depth_float_buffer(False)
> ```

` depth = vis.capture_depth_float_buffer(False)`

I would suggest mentioning the param. name in the examples, such as:
` depth = vis.capture_depth_float_buffer(do_render = False)`

Also even when I tried with True, here, I couldn't see any rendering. Can you verify the same on your system?
Is it expected to see any rendering/viz? anything other than results being saved in the test_data dir?
___
*[`examples/python/visualization/headless_rendering.py` line 70 at r2](https://reviewable.io/reviews/isl-org/open3d/5211#-N4wfq9s-KBCHfTuZ1_8:-N4wfq9s-KBCHfTuZ1_9:b9zqyd6) ([raw file](https://github.com/isl-org/open3d/blob/52d5710eae4f1399ca0127902251fedb05231f13/examples/python/visualization/headless_rendering.py#L70)):*
> ```Python
>                        np.asarray(image),
>                        dpi=1)
>             #vis.capture_depth_image("depth/{:05d}.png".format(glb.index), False)
> ```

why are these lines commented out?
___
*[`examples/python/visualization/headless_rendering.py` line 80 at r2](https://reviewable.io/reviews/isl-org/open3d/5211#-N4wfshK-y2-NgpA8dhl:-N4wfshK-y2-NgpA8dhm:b-aw9j1q) ([raw file](https://github.com/isl-org/open3d/blob/52d5710eae4f1399ca0127902251fedb05231f13/examples/python/visualization/headless_rendering.py#L80)):*
> ```Python
>                 register_animation_callback(None)
>             vis.destroy_window()
>         return False
> ```

what is the behavior here? why are we returning `False`? Also add a comment there to explain the same, as it is not intuitive. 
Does the callback function expects `False` at the end of the sequence?

reyanshsolis:
TODO: Fix a design for storing example outputs if any.
Suggestion: `open3d_data / example_results / example_name / `

This will also be followed later for the reconstruction system, where there are multiple outputs and storing output in the input dataset dir. does not look clean (we should not write anything to dataset dir).

reyanshsolis:
let's fix this up quickly and merge the pr

ShreyanshDarshan (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/5211)* status: 0 of 1 files reviewed, 4 unresolved discussions (waiting on @reyanshsolis)

___
*[`examples/python/visualization/headless_rendering.py` line 33 at r2](https://reviewable.io/reviews/isl-org/open3d/5211#-N4wh8uUDVc4Bi75na6H:-N6AF35NDAWX6ME8I87P:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/52d5710eae4f1399ca0127902251fedb05231f13/examples/python/visualization/headless_rendering.py#L33)):*

don't use the `test_data` path. 

I would suggest to get `open3d_data` path, and create a `open3d_data` / `example_results` / `example_name` / `image` and `depth`.
`example_name` here might be `visualization_headless_rendering`.

Also, print the path of output dirs.
</blockquote></details>

Done.
___
*[`examples/python/visualization/headless_rendering.py` line 62 at r2](https://reviewable.io/reviews/isl-org/open3d/5211#-N4wgJTBDTDWW_T598BJ:-N6DAvUf0lW4NGGKqPgf:b-yoa1k5) ([raw file](https://github.com/isl-org/open3d/blob/52d5710eae4f1399ca0127902251fedb05231f13/examples/python/visualization/headless_rendering.py#L62)):*

` depth = vis.capture_depth_float_buffer(False)`

I would suggest mentioning the param. name in the examples, such as:
` depth = vis.capture_depth_float_buffer(do_render = False)`

Also even when I tried with True, here, I couldn't see any rendering. Can you verify the same on your system?
Is it expected to see any rendering/viz? anything other than results being saved in the test_data dir?
</blockquote></details>

Changing the `do_render` parameter to `True` didn't perform any rendering on my end as well. Enabling headless rendering while building Open3D disables the GUI module so this should be expected. I'm not sure what the parameter do_render does, it might need more investigation.
___
*[`examples/python/visualization/headless_rendering.py` line 70 at r2](https://reviewable.io/reviews/isl-org/open3d/5211#-N4wfq9s-KBCHfTuZ1_8:-N6DC1xO051aFwllCjxE:b-tyo9zt) ([raw file](https://github.com/isl-org/open3d/blob/52d5710eae4f1399ca0127902251fedb05231f13/examples/python/visualization/headless_rendering.py#L70)):*

why are these lines commented out?
</blockquote></details>

The depth and color images can also be captured and saved directly (without the use of `plt.imsave`) using the commented lines. Perhaps they were being used earlier in this example but then were replaced by `vis.capture_screen_float_buffer(False)` and `vis.capture_depth_float_buffer(False)` in order to use the depth colormaps of `plt.imsave`.

Do you suggest we remove these lines from the example?

output of depth image from `plt.imsave` :


output of depth image from `vis.capture_depth_image` :

___
*[`examples/python/visualization/headless_rendering.py` line 80 at r2](https://reviewable.io/reviews/isl-org/open3d/5211#-N4wfshK-y2-NgpA8dhl:-N6DEvxb7OPyTo4qzmqB:b-2tguqm) ([raw file](https://github.com/isl-org/open3d/blob/52d5710eae4f1399ca0127902251fedb05231f13/examples/python/visualization/headless_rendering.py#L80)):*

what is the behavior here? why are we returning `False`? Also add a comment there to explain the same, as it is not intuitive. 
Does the callback function expects `False` at the end of the sequence?
</blockquote></details>

The callback function returns true if `UpdateGeometry()` needs to be run which is not required for the static geometry being used in the example.

Added comments for the same.

reyanshsolis (approved):

reyanshsolis:
- added review changes.
- added `data.output_dir` [like `data.extract_dir`], to return the absolute path to an output dir corresponding to the dataset to store results. i.e. `open3d_data/output/dataset_name/`. Currently, the reconstruction systems save the output in the dataset folder itself (by default design that will be the `open3d_data/extract/dataset_name/` folder for us), which does not seem to be a very clean design to me. Would like to check with @yxlao regarding the design.

yxlao (approved):

PR #5219, opened 2022-06-19
Change 'depth_diff_max' to 'max_depth_diff'

Techassi:
When using the reconstruction system located in `examples/python/reconstruction_system` the first step *Make fragments* failes to run. I tested the latest version of the master branch (c43eabebc2b14b1ae3ee5371146027b887c41eee).

```shell
python run_system.py --config ./config/tutorial.json --make
```

After fixing the error

```
AttributeError: 'open3d.cpu.pybind.pipelines.odometry.OdometryOption' object has no attribute 'depth_diff_max'
```

with the in this PR provided change, the script ran without any issues.

It seems like the commit c8f8f401c1e7ef6efd728a10e0e25895fb9d8ff1 has introduced this change in the C++ source.

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/5219)

### DIFF
--- a/examples/python/reconstruction_system/config/tutorial.json
+++ b/examples/python/reconstruction_system/config/tutorial.json
@@ -4,7 +4,7 @@
     "path_intrinsic": "",
     "depth_max": 3.0,
     "voxel_size": 0.05,
-    "depth_diff_max": 0.07,
+    "max_depth_diff": 0.07,
     "preference_loop_closure_odometry": 0.1,
     "preference_loop_closure_registration": 5.0,
     "tsdf_cubic_size": 3.0,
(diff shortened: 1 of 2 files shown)


yuecideng:
If you want to use the latest master of open3d, please download the latest [python package](http://www.open3d.org/docs/release/getting_started.html) or build open3d from source. Otherwise you should use the latest release `0.15.1` branch to run the example. @Techassi

yxlao:
Thanks @Techassi . This has been addressed in https://github.com/isl-org/Open3D/pull/5167. 

If you are using the example code from `mater`, you may compile from source, or download the `master` wheel from http://www.open3d.org/docs/latest/getting_started.html#pip-pypi

Techassi:
Okay thx. Good to know!

PR #5247, opened 2022-06-24
Random seed Open3D globally

yxlao:
## Summary

This PR allows all of Open3D's random number generator and random samplers to be globally seeded. If you plan to add a new random number or random sampling feature to Open3D, you should always use one of these helper functions/classes, instead of seeding your own random numbers.

## Changes

- Global singleton and mutex for random number generation.
- Several helper functions and helper classes to generate random numbers or use the random number engine.
- Adapt all code to use the new random number mechanism, and removed all other code for seeding.

## Usage

```cpp
#include <algorithm>
#include <iostream>

#include "open3d/utility/Random.h"

int main() {
    using namespace open3d;

    // Globally seed.
    utility::random::Seed(0);

    // Simply get a random number.
    std::cout << utility::random::RandUint32() << std::endl;

    // Using the random engine.
    std::vector<int> vals{0, 1, 2, 3, 4, 5, 6, 7, 8, 9};
    {
        std::lock_guard<std::mutex> lock(*utility::random::GetMutex());
        std::shuffle(vals.begin(), vals.end(), *utility::random::GetEngine());
    }

    // Create a generator: integer in [0, 10).
    utility::random::UniformIntGenerator gen_int(0, 10);
    for (int i = 0; i < 5; i++) {
        std::cout << gen_int() << std::endl;
    }

    // Create a generator: double in [-1.0, 1.0).
    utility::random::UniformDoubleGenerator gen_double(-1.0, 1.0);
    for (int i = 0; i < 5; i++) {
        std::cout << gen_double() << std::endl;
    }

    return 0;
}
```

```python
import open3d as o3d

o3d.utility.random.seed(xxx);
```

## Known limitations
- The random number is not platform-independent.
  - `std::mt19937` is platform-independent.
  - `std::uniform_int_distribution` is **not** platform-independent. 
  - `uniform_real_distribution` is **not** platform-independent.
- No CUDA support yet.

## Future PR
1. Port or write our own `uniform_int_distribution` and `uniform_real_distribution`, such that everything becomes OS/compiler independent.
2. Add CUDA support. We don't require CUDA to generate the same random number as the CPU.
3. Add Tenosr rand operation to generate randomized Tensor.

This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/5247)

### DIFF
--- a/cpp/open3d/pipelines/registration/Registration.cpp
+++ b/cpp/open3d/pipelines/registration/Registration.cpp
@@ -302,11 +302,13 @@ RegistrationResult RegistrationRANSACBasedOnFeatureMatching(
         const Feature &target_feature,
         bool mutual_filter,
         double max_correspondence_distance,
-        const TransformationEstimation &estimation,
-        int ransac_n,
+        const TransformationEstimation
+                &estimation /* = TransformationEstimationPointToPoint(false)*/,
+        int ransac_n /* = 3*/,
         const std::vector<std::reference_wrapper<const CorrespondenceChecker>>
-                &checkers,
-        const RANSACConvergenceCriteria &criteria) {
+                &checkers /* = {}*/,
+        const RANSACConvergenceCriteria
+                &criteria /* = RANSACConvergenceCriteria()*/) {
     if (ransac_n < 3 || max_correspondence_distance <= 0.0) {
         return RegistrationResult();
     }
--- a/cpp/open3d/pipelines/registration/Registration.cpp
+++ b/cpp/open3d/pipelines/registration/Registration.cpp
@@ -29,9 +29,9 @@
 #include "open3d/geometry/KDTreeFlann.h"
 #include "open3d/geometry/PointCloud.h"
 #include "open3d/pipelines/registration/Feature.h"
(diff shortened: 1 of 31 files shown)


cpp/open3d/pipelines/registration/Registration.cpp
```diff
@@ -306,14 +302,11 @@ RegistrationResult RegistrationRANSACBasedOnFeatureMatching(
...
-        int ransac_n /* = 3*/,
+        const TransformationEstimation &estimation,
+        int ransac_n,
         const std::vector<std::reference_wrapper<const CorrespondenceChecker>>
-                &checkers /* = {}*/,
-        const RANSACConvergenceCriteria &criteria,
-        /* = RANSACConvergenceCriteria()*/
-        utility::optional<unsigned int> seed /* = utility::nullopt*/) {
+                &checkers,
+        const RANSACConvergenceCriteria &criteria) {
```
ssheorey, line 309: Keep comments with default values?

ssheorey (approved):
Thanks, this looks very useful!

yuecideng (approved):

PR #5257, opened 2022-06-28
add boolean mesh operations to Tensor TriangleMesh

benjaminum:
This PR adds 3 new functions to t::geometry::TriangleMesh for computing the boolean union, intersection, and difference

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/5257)

### DIFF
--- a/cpp/open3d/t/geometry/TriangleMesh.h
+++ b/cpp/open3d/t/geometry/TriangleMesh.h
@@ -476,6 +476,9 @@ public:
 
     /// Function to simplify mesh using Quadric Error Metric Decimation by
     /// Garland and Heckbert.
+    ///
+    /// This function always uses the CPU device.
+    ///
     /// \param target_reduction The factor of triangles to delete, i.e.,
     /// setting this to 0.9 will return a mesh with about 10% of the original
     /// triangle count.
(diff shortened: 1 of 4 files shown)


python/test/t/geometry/test_trianglemesh.py
```diff
@@ -48,3 +48,25 @@ def test_simplify_quadric_decimation():
...
+    assert ans.vertex['positions'].shape == (730, 3)
+    assert ans.triangle['indices'].shape == (1384, 3)
+
+    ans = box.boolean_intersection(sphere)
+    assert ans.vertex['positions'].shape == (154, 3)
+    assert ans.triangle['indices'].shape == (232, 3)
+
+    ans = box.boolean_difference(sphere)
+    assert ans.vertex['positions'].shape == (160, 3)
+    assert ans.triangle['indices'].shape == (244, 3)
```
ssheorey, line 72: [nit] Can we check any statistical property of the triangle mesh for a more definite test?
Maybe position covariance?

ssheorey (approved):

python/test/t/geometry/test_trianglemesh.py
benjaminum: Since we use functionality from VTK we rely mostly on their tests. I think our test just detects if something has changed (VTK version update).

cpp/open3d/t/geometry/TriangleMesh.h
```diff
@@ -487,6 +490,46 @@ class TriangleMesh : public Geometry, public DrawableGeometry {
...
+    /// Both meshes should be manifold.
+    ///
+    /// This function always uses the CPU device.
+    ///
+    /// \param mesh This is the second operand for the boolean operation.
+    /// \param tolerance Threshold which determines when point distances are
+    /// considered to be 0.
+    ///
+    /// \return The mesh describing the union volume.
+    TriangleMesh BooleanUnion(const TriangleMesh &mesh,
```
yxlao, line 504: LGTM.

Is it reasonable to call it `Union` instead of `BooleanUnion`? To put it in another way, will we ever have a union operation that is not "boolean"?

yxlao (approved):

cpp/open3d/t/geometry/TriangleMesh.h
benjaminum: Probably we won't have another union operation. I think it makes sense to add the prefix to make the function easier to find and it will also make all three operations appear next to each other.

PR #5259, opened 2022-06-28
fix windows 3.8 dll loading for cuda

yxlao:
- CUDA pybind fails to load for Python 3.8. Somehow it worked for 3.9. This PR fixes that. 
- Windows CI does not trigger this error so it was hidden from us.
- Changed from `warnings` to `logging` package, as the warnings package does not print on Windows by default. 

This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/5259)


python/open3d/__init__.py
```diff
@@ -34,57 +34,86 @@
...
+
+def _load_cdll(path):
+    """
+    Wrapper around ctypes.CDLL to take care of Windows compatibility.
+    """
+    path = _Path(path)
+    if not path.is_file():
+        raise FileNotFoundError(f"Shared library file not found: {path}.")
+
+    if sys.platform == 'win32' and sys.version_info >= (3, 8):
```
ssheorey, line 55: From the [linked Python issue](https://bugs.python.org/issue42114), looks like this is fixed in 3.9, 3.10 and onwards. Should we only check for 3.8 here?

ssheorey (changes requested):
The warnings are actually not shown by default on any platform, since the CUDA devices not found or CUDA libraries not found warning will always show up when loading CUDA Open3D package on a CPU only machine.

http://www.open3d.org/docs/latest/getting_started.in.html?highlight=warnings#troubleshooting

They should show up only with `python -Wdefault "import open3d"`. If this is not the desired behavior, we can enable them all the time with warning filter settings.

https://docs.python.org/3/library/warnings.html#default-warning-filter
`warnings.filterwarnings('default', message='.*', category=ImportWarning,module='open3d')`
(or more fine grained control is possible with lineno)

ssheorey:
Main change: Add winmode=0 for CDLL function

ssheorey (approved):
Tested on Win 10 with Python 3.8.

PR #5263, opened 2022-06-29
add create_box in tensor-based module

ZhengyuDiao:
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/5263)

### DIFF
--- a/cpp/open3d/t/geometry/TriangleMeshFactory.cpp
+++ b/cpp/open3d/t/geometry/TriangleMeshFactory.cpp
@@ -1,0 +1,100 @@
+// ----------------------------------------------------------------------------
+// -                        Open3D: www.open3d.org                            -
+// ----------------------------------------------------------------------------
+// The MIT License (MIT)
+//
+// Copyright (c) 2018-2021 www.open3d.org
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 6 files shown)


cpp/open3d/t/geometry/TriangleMeshFactory.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+    // Mesh.
+    TriangleMesh mesh(vertex_positions, triangle_indices);
+
+    return mesh;
+}
+
+}  // namespace geometry
+}  // namespace t
+}  // namespace open3d
```
reyanshsolis, line 100: always leave a newline at the end of the code. There is also some setting to do that in vs-code. https://stackoverflow.com/a/44704969

reyanshsolis (commented):
Are you planning to add the uv_map attribute as well?

yxlao (approved):

yxlao:
LGTM after CI.

yxlao:
> Are you planning to add the uv_map attribute as well?

Have we designed the data structure for UV map for t-mesh? Do we have an example for that?

For this PR, it looks good, and we can add UV map support to this later.

cpp/open3d/t/geometry/TriangleMeshFactory.cpp
ZhengyuDiao: ok, got it

PR #5294, opened 2022-07-08
Add tensor-based creation APIs

ZhengyuDiao:
Add tensor-based creation APIs that reach feature parity with the legacy APIs.

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/5294)


reyanshsolis (approved):

reyanshsolis:
These creation APIs might need to be implemented using native tensor ops in the future. In interest of tensor-geometry op migration timeline, let's merge this as a placeholder, complete some other important ops, and revisit later if we have time.

PR #5299, opened 2022-07-11
Add normal related functions to PointCloud and TriangleMesh

yuecideng:
1. Add `NormalizeNormals` for both Tensor PointCloud and TriangleMesh.
2. Add normal orientation functions for Tensor PointCloud.
3. Add triangle and vertex normals computation for Tensor TriangleMesh.

This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/5299)


cpp/open3d/t/geometry/TriangleMesh.h
```diff
@@ -446,6 +446,17 @@ class TriangleMesh : public Geometry, public DrawableGeometry {
     /// \return Rotated TriangleMesh
     TriangleMesh &Rotate(const core::Tensor &R, const core::Tensor &center);
 
+    /// Normalize both triangle normals and vertex normals to length 1.
+    TriangleMesh &NormalizeNormals();
```
reyanshsolis, line 450: We might not need kernels for normalizing normals. 
P.S. The goal is to reduce the usage of kernels without taking much performance loss. So, we must always use and improve tensor ops, while avoiding kernel implementation.

The following might work instead of kernel (or might take some fixing):
```cpp
TriangleMesh &TriangleMesh::NormalizeNormals() {
    if (HasVertexNormals()) {
        SetVertexNormals(GetVertexNormals().Contiguous());
        core::Tensor &vertex_normals = GetVertexNormals();
        core::Tensor vertex_normals_norm = vertex_normals.Mul(vertex_normals).Sum({1}).Sqrt();
        vertex_normals = vertex_normals.Div(vertex_normals_norm);
    } else {
        utility::LogWarning("TriangleMesh has no vertex normals.");
    }

    if (HasTriangleNormals()) {
        core::Tensor &triangle_normals = GetTriangleNormals();
        core::Tensor triangle_normals_norm = triangle_normals.Mul(triangle_normals).Sum({1}).Sqrt();
        triangle_normals = triangle_normals.Div(triangle_normals_norm);
    }

    return *this;
}
```

yuecideng: The situation when `norm` is 0 will make the normal equal to `nan` if using tensor ops to do `NormalizeNormals`, while we want the normal to be `0` as original. So we may use one more operation like (in python):
```python
triangle_normals[triangle_normals == float('nan')] = 0
```
But currently, the tensor indexing is no very well and this extra operation will cost some time.

cpp/open3d/t/geometry/PointCloud.cpp
```diff
@@ -512,6 +532,81 @@ void PointCloud::EstimateNormals(
...
+    const core::Tensor equal_0 = normals_norm.Eq(0);
+
+    normals.IndexSet({less_0}, normals.IndexGet({less_0}).Mul(-1));
+
+    core::Tensor reference_assign_to_0 = reference.IndexGet({equal_0});
+    // Normalize reference vector which will be assigned to normals.
+    const core::Tensor reference_assign_to_0_norm =
+            reference_assign_to_0.Mul(reference_assign_to_0).Sum({1}).Sqrt();
+    reference_assign_to_0 /= reference_assign_to_0_norm.Reshape({-1, 1});
+    // Set to (0, 0, 1) if the reference vector norm is 0.
```
reyanshsolis, line 591: is this behavior same as legacy?

reyanshsolis (commented):
Can you please try to compare the performance of Compute Normals functions between Calling the `NormalizeNormals` function v/s doing the same in the ComputeNormal kernel?

`NormalizeNormal` function using Tensor makes sense, but if it is causing the `ComputeNormal` function to slow, which is very frequently used, then we might consider adding the optional normalize_normal functionality within the kernel.

yuecideng:
Sure, I will test it and update.

cpp/open3d/t/geometry/PointCloud.cpp
yuecideng: Yes, the unit test is the same as legacy.

yuecideng:
By implement these fucntions in kernel, the performance has lot of gain:
With normalization:
image

PointCloud functions:
image

cpp/open3d/t/geometry/kernel/TriangleMeshCPU.cpp
```diff
@@ -0,0 +1,75 @@
...
+            int64_t idx = 3 * i;
+            int64_t triangle_id1 = triangle_ptr[idx];
+            int64_t triangle_id2 = triangle_ptr[idx + 1];
+            int64_t triangle_id3 = triangle_ptr[idx + 2];
+
+            scalar_t n1 = triangle_normals_ptr[idx];
+            scalar_t n2 = triangle_normals_ptr[idx + 1];
+            scalar_t n3 = triangle_normals_ptr[idx + 2];
+
+            vertex_normals_ptr[3 * triangle_id1] += n1;
```
yxlao, line 58: we can use critical section or atomic add similar to CUDA

yxlao (approved):

PR #5304, opened 2022-07-11
Add default parameters in slam::Model cpp.

jdavidberger:
This seemed like an active todo and it would be useful for me to have these levers in the python code.

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/5304)


cpp/open3d/t/pipelines/slam/Model.h
```diff
@@ -75,6 +75,27 @@ class Model {
...
+    /// In multiscale odometry the order is from coarse to fine. Inputting a
+    /// vector of iterations by default triggers the implicit conversion.
+
+    odometry::OdometryResult TrackFrameToModel(
+            const Frame& input_frame,
+            const Frame& raycast_frame,
+            float depth_scale,
+            float depth_max,
+            float depth_diff,
+            odometry::Method method,
```
reyanshsolis, line 95: also overload method here to `odometry::Method::PointToPlane`

reyanshsolis: we can remove the later `TrackFrameToModel`. i.e. we can just pass the default argument and remove function overloading.

reyanshsolis (changes requested):

cpp/open3d/t/pipelines/slam/Model.cpp
```diff
@@ -78,24 +78,33 @@ void Model::SynthesizeModelFrame(Frame& raycast_frame,
...
             t::geometry::RGBDImage(raycast_frame.GetDataAsImage("color"),
                                    raycast_frame.GetDataAsImage("depth")),
             raycast_frame.GetIntrinsics(), identity, depth_scale, depth_max,
-            std::vector<odometry::OdometryConvergenceCriteria>{6, 3, 1},
-            odometry::Method::PointToPlane,
-            odometry::OdometryLossParams(depth_diff));
+            criteria, method, odometry::OdometryLossParams(depth_diff));
+}
+
+odometry::OdometryResult Model::TrackFrameToModel(const Frame& input_frame,
```
reyanshsolis, line 101: remove the overloaded function.

reyanshsolis (changes requested):

cpp/open3d/t/pipelines/slam/Model.h
jdavidberger: How worried are are you about version compatibility details? The only reason I didn't do it this way from the start is that it'll no longer have the original symbol in the shared object and could break applications if they try to do a drop in replacement shared library update. If that isn't a concern I agree this is the cleaner approach

reyanshsolis: Both parameters have default values (after adding default for odometry_method), therefore existing applications based on this won't be affected.

jdavidberger: This is true only if the application is rebuilt against the this version or later. If an application has been built against the released v0.15 and wants to try dropping in a shared object from this commit onward, that application might crash due to the original function signature not being present in the newer shared object. 

This might not be a thing you care about, especially since it's still in the major version 0 beta stage, and if that is the case  I'll resolve these as you put them, but it is something I wanted to call out in case there was some desire to keep backwards compatibility on binaries.

jdavidberger:
I've added similar customization parameters to SynthesizeModelFrame. Depending on the feedback on whether to keep binary compatibility vs simplify the code I'll make it match whatever is wanted for that.

reyanshsolis:
Thanks for your contributions @jdavidberger, I will try to check this and get back to you by this weekend.

reyanshsolis:
The functionality you wanted to achieve, could be done using default params instead of overloading functions. Previously, the default values were present in pybind but missing in cpp, so I have added the changes accordingly to this PR and tested the cpp slam system example. 

May you please try the update and check if your code is working as well?

jdavidberger:
The function declarations in the pybind were missing; I'm guessing just got typoed out. I added those back in and it builds and works.

reyanshsolis (approved):
LGTM.

jdavidberger:
Is there anything more I need to do for this pr?

reyanshsolis:
> Is there anything more I need to do for this pr?
I don't think so. The PR looks good to merge. Thanks for your contribution.

PR #5308, opened 2022-07-12
Tensor Ops - Any, All, core::RemoveNonFinite, t-PointCloud::RemoveNonFinitePoints

reyanshsolis:


yuecideng (changes requested):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/5308)* status: 0 of 12 files reviewed, 1 unresolved discussion (waiting on @reyanshsolis and @yuecideng)

___
*[`cpp/open3d/core/TensorFunction.h` line 138 at r2](https://reviewable.io/reviews/isl-org/open3d/5308#-N6pUAcv7swfHKH04NxS:-N6pUAcv7swfHKH04NxT:b-vy7obu) ([raw file](https://github.com/isl-org/open3d/blob/53711f1bde97badc7899e44aaa2466a59cb73faf/cpp/open3d/core/TensorFunction.h#L138)):*
> ```objc
> /// \param dim dimension along which to remove non-finite values.
> /// \param remove_nan Remove NaN values from the PointCloud.
> /// \param remove_infinite Remove infinite values from the PointCloud.
> ```

We should use `tensor` here instead of `PointCloud`

yuecideng (approved):

PR #5311, opened 2022-07-13
Tensor PointCloud RemoveDuplicatedPoints

reyanshsolis:
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/5311)


cpp/tests/t/geometry/PointCloud.cpp
```diff
@@ -826,6 +826,35 @@ TEST_P(PointCloudPermuteDevices, RemoveRadiusOutliers) {
...
+            pcd_small.RemoveDuplicatedPoints();
+
+    EXPECT_TRUE(output_pcd.GetPointPositions().AllClose(
+            core::Tensor::Init<float>({{1.0, 1.0, 1.0},
+                                       {1.1, 1.1, 1.1},
+                                       {1.2, 1.2, 1.2},
+                                       {1.3, 1.3, 1.3},
+                                       {5.0, 5.0, 5.0},
+                                       {5.1, 5.1, 5.1}},
+                                      device)));
```
yuecideng, line 855: We may add one more case to test `floating_precision`.

ShreyanshDarshan (changes requested):
> Reviewed 4 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/5311)* status: all files reviewed, 4 unresolved discussions (waiting on @reyanshsolis)

___
*[`cpp/open3d/t/geometry/PointCloud.h` line 373 at r1](https://reviewable.io/reviews/isl-org/open3d/5311#-N7IkVux2UWENiK8FACo:-N7IkVuy1bGH2zXV-gVe:bunqsgj) ([raw file](https://github.com/isl-org/open3d/blob/822ecd8bd803ec4738409295f107eac8607ff002/cpp/open3d/t/geometry/PointCloud.h#L373)):*
> ```objc
>     /// treated as different positions, however the floating precision is not
>     /// very accurate on GPU, therefore use higher floating precision than
>     /// required (higher floating_precision refers to lower value).
> ```

Nit: Sounds a bit confusing, can be rephrased to: 
```
... therefore use higher precision than required (higher precision refers to a lower value for floating_precision).
```



_Code quote:_
```objc
    /// very accurate on GPU, therefore use higher floating precision than
    /// required (higher floating_precision refers to lower value).
```
___
*[`cpp/open3d/t/geometry/PointCloud.cpp` line 394 at r1](https://reviewable.io/reviews/isl-org/open3d/5311#-N7IpZvoCGl-fvY1sV-k:-N7IpZvoCGl-fvY1sV-l:b-9uzhk8) ([raw file](https://github.com/isl-org/open3d/blob/822ecd8bd803ec4738409295f107eac8607ff002/cpp/open3d/t/geometry/PointCloud.cpp#L394)):*
> ```cpp
>         double floating_precision) const {
>     core::Tensor points_voxeld = GetPointPositions() / floating_precision;
>     core::Tensor points_voxeli = points_voxeld.To(core::Int64);
> ```

Will there be potential overflow issues while converting floating point value to `Int64`?
___
*[`cpp/pybind/t/geometry/pointcloud.cpp` line 361 at r1](https://reviewable.io/reviews/isl-org/open3d/5311#-N7IlcjF2jDbKY2YO35q:-N7IlcjF2jDbKY2YO35r:b20gi4e) ([raw file](https://github.com/isl-org/open3d/blob/822ecd8bd803ec4738409295f107eac8607ff002/cpp/pybind/t/geometry/pointcloud.cpp#L361)):*
> ```cpp
>               "1.00001] and [1.00001, 1.00002, 1.00002] must be treated as "
>               "different positions, however the floating precision is not very "
>               "accurate on GPU, therefore use higher floating precision than "
> ```

Nit: Same comment as above.

reyanshsolis:
Removed `floating_precision`. Using `ReinterpretCast`.

ShreyanshDarshan (approved):
> Reviewed 6 of 6 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/5311)* status: all files reviewed, 4 unresolved discussions (waiting on @reyanshsolis)

cpp/open3d/core/Tensor.cpp
```diff
@@ -371,6 +371,15 @@ Tensor& Tensor::operator=(Tensor&& other) && {
     return *this;
 }
 
+Tensor Tensor::ReinterpretCast(const core::Dtype& dtype) const {
+    if (dtype_.ByteSize() != dtype.ByteSize()) {
+        utility::LogError(
+                "Cannot reinterpret cast between data-types of different "
```
yxlao, line 377: Print the name and byte size of the dtypes in the error message.

cpp/open3d/core/Tensor.h
```diff
@@ -225,6 +225,9 @@ class Tensor : public IsDevice {
         return *this;
     }
 
+    /// Tensor reinterpret cast operator.
```
yxlao, line 228: - Descirbe what does reinterpret cast do for tensor.
- Describe that the size should match.

yxlao (approved):

cpp/open3d/core/Tensor.cpp
reyanshsolis: Done.

cpp/open3d/core/Tensor.h
reyanshsolis: Done.

PR #5312, opened 2022-07-13
Tensor PointCloud PaintUniformColor Op

reyanshsolis:
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/5312)


cpp/open3d/t/geometry/PointCloud.cpp
```diff
@@ -388,6 +388,23 @@ std::tuple<PointCloud, core::Tensor> PointCloud::RemoveRadiusOutliers(
     return std::make_tuple(pcd, valid);
 }
 
+PointCloud PointCloud::PaintUniformColor(const core::Tensor &color) const {
+    core::AssertTensorShape(color, {3});
+    core::Tensor clipped_color = color.To(GetDevice());
+    if (color.GetDtype() == core::Float32 ||
+        color.GetDtype() == core::Float64) {
+        clipped_color = clipped_color.Clip(0.0f, 1.0f);
+    }
```
yuecideng, line 397: Should we consider the situation when color is `int` or `bool`?

cpp/open3d/t/geometry/PointCloud.h
```diff
@@ -364,6 +364,12 @@ class PointCloud : public Geometry, public DrawableGeometry {
     std::tuple<PointCloud, core::Tensor> RemoveRadiusOutliers(
             size_t nb_points, double search_radius) const;
 
+    /// \brief Assigns uniform color to the point cloud.
+    ///
+    /// \param color  RGB color for the point cloud. {3,} shaped Tensor.
+    /// Floating color values are clipped between 0.0 and 1.0.
+    PointCloud PaintUniformColor(const core::Tensor &color) const;
+
```
yuecideng, line 372: What is the consideration to return the copy of the painted point cloud? Do we want to keep the original unpainted pointcloud?

cpp/open3d/t/geometry/PointCloud.cpp
reyanshsolis: The visualizer supports Float32 data-type with range [0,1]. However, currently, we are allowing users to init. colors with any data-type. This is consistence with that (See SetPointColors).
If we want to add constraints or warnings in the future, they can be added in the `PointCloud::SetPointColors` function.

ShreyanshDarshan (changes requested):
> Reviewed 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/5312)* status: all files reviewed, 3 unresolved discussions (waiting on @reyanshsolis and @yuecideng)

___
*[`cpp/open3d/t/geometry/PointCloud.cpp` line 419 at r2](https://reviewable.io/reviews/isl-org/open3d/5312#-N7ohi9C50LSpHTVEudm:-N7ohi9C50LSpHTVEudn:bl0efwy) ([raw file](https://github.com/isl-org/open3d/blob/a989f3788bd53a8eed264618103876faef89d5ae/cpp/open3d/t/geometry/PointCloud.cpp#L419)):*
> ```cpp
>     if (color.GetDtype() == core::Float32 ||
>         color.GetDtype() == core::Float64) {
>         clipped_color = clipped_color.Clip(0.0f, 1.0f);
> ```

Should we move the clipping check to to `SetPointColors`?
Also maybe we can add a warning like:
```cpp
utility::LogWarning(
        "invalid color in PaintUniformColor, clipping to [0, 1]");
```
similar to the one in the legacy function.

yuecideng (approved):

ShreyanshDarshan (approved):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/5312)* status: all files reviewed, 2 unresolved discussions (waiting on @yuecideng)

Issue #5322, opened 2022-07-16
AttributeError about 'create_mesh_coordinate_frame'


where2go947:
### Checklist

- [X] I have searched for [similar issues](https://github.com/isl-org/Open3D/issues).
- [X] For Python issues, I have tested with the [latest development wheel](http://www.open3d.org/docs/latest/getting_started.html#development-version-pip).
- [X] I have checked the [release documentation](http://www.open3d.org/docs/release/) and the [latest documentation](http://www.open3d.org/docs/latest/) (for `master` branch).

### My Question

I follow the latest [documentation](http://www.open3d.org/docs/release/tutorial/visualization/visualization.html?highlight=create_mesh_coordinate_frame#:~:text=mesh_frame%20%3D%20o3d.geometry.TriangleMesh.create_coordinate_frame(%0A%20%20%20%20size%3D0.6%2C%20origin%3D%5B%2D2%2C%20%2D2%2C%20%2D2%5D))  with `open3d=0.15.2`  to use `create_mesh_coordinate_frame` as follow:
```
coordinate_frame = open3d.geometry.TriangleMesh.create_mesh_coordinate_frame()
```
but AttributeError raises:
```
File "build_pointcloud.py", line 161, in <module>
    coordinate_frame = open3d.geometry.TriangleMesh.create_mesh_coordinate_frame()
AttributeError: type object 'open3d.cuda.pybind.geometry.TriangleMesh' has no attribute 'create_mesh_coordinate_frame'
```
why `cuda.pybind` appears and how to use it right?

Could you please give me some advice?
Thanks in advance!

yuecideng:
Hi, this is typo and you should use `create_coordinate_frame` instead.

where2go947:
Thank you!
The issue has been solved.

PR #5323, opened 2022-07-16
fix doc typo 

yuecideng:
Fix documentation typo in geometry primitives section, which related to issue #5322.

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/5323)


ssheorey (approved):
Thanks Yueci!

Issue #5321, opened 2022-07-16
ImportError: cannot import name 'get_default_testdata' from 'common' (/home/user/source_builds/Open3D/examples/python/t_reconstruction_system/common.py)


sandyflute:
### Checklist

- [X] I have searched for [similar issues](https://github.com/isl-org/Open3D/issues).
- [X] For Python issues, I have tested with the [latest development wheel](http://www.open3d.org/docs/latest/getting_started.html#development-version-pip).
- [X] I have checked the [release documentation](http://www.open3d.org/docs/release/) and the [latest documentation](http://www.open3d.org/docs/latest/) (for `master` branch).

### Describe the issue

While running integrate.py(dense_slam.py as well) examples from t_reconstruction_system, a function, get_default_testdata, needs to be imported from common.py located in the same folder. Looks like this function is missing from the file.

### Steps to reproduce the bug

```python
Enter the python examples from open3d root dir:

`cd examples/python/t_reconstruction_system/`

run integrate.py

`python integrate.py`
```

### Error message

```
Traceback (most recent call last):
  File "integrate.py", line 40, in <module>
    from common import load_rgbd_file_names, load_depth_file_names, save_poses, load_intrinsic, load_extrinsics, get_default_testdata
ImportError: cannot import name 'get_default_testdata' from 'common' (/home/username/source_builds/Open3D/examples/python/t_reconstruction_system/common.py)
```

### Open3D, Python and System information

```markdown
- Operating system: Ubuntu 18.04 64-bit
- Python version: Python 3.8.13 
- Open3D version: output from python: 0.15.2+393bbce6 `print(open3d.__version__)`
- System architecture: x86
- Is this a remote workstation?: no
- How did you install Open3D?: build from source
- Compiler version (if built from source): gcc 7.5.0
```

yuecideng:
Hi, this error can be fixed by PR #5326. You can try to use `integrate_custom.py` currently.

PR #5326, opened 2022-07-17
Fix error in python t_reconstruction example

yuecideng:
This PR fix the issue #5321

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/5326)


reyanshsolis (approved):
Thanks.

PR #5347, opened 2022-07-21
Enable attributes in vtk wrapper functions

benjaminum:
This PR
- enables attributes in vtk wrapper functions allowing to preserve (interpolate) vertex and triangle attributes
- fixes a segfault when converting empty vtkPolyData to a TriangleMesh

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/5347)

### DIFF
--- a/cpp/open3d/t/geometry/PointCloud.h
+++ b/cpp/open3d/t/geometry/PointCloud.h
@@ -124,6 +124,9 @@ public:
     /// Getter for point_attr_ TensorMap. Used in Pybind.
     const TensorMap &GetPointAttr() const { return point_attr_; }
 
+    /// Getter for point_attr_ TensorMap.
+    TensorMap &GetPointAttr() { return point_attr_; }
+
     /// Get attributes. Throws exception if the attribute does not exist.
     ///
     /// \param key Attribute name.
--- a/cpp/open3d/t/geometry/TensorMap.h
+++ b/cpp/open3d/t/geometry/TensorMap.h
@@ -28,6 +28,7 @@
 
 #include <string>
 #include <unordered_map>
+#include <unordered_set>
 
 #include "open3d/core/Tensor.h"
 
@@ -113,6 +114,15 @@ public:
 
     /// Returns the primary key of the TensorMap.
     std::string GetPrimaryKey() const { return primary_key_; }
+
+    /// Returns a set with all keys.
+    std::unordered_set<std::string> GetKeySet() const {
+        std::unordered_set<std::string> keys;
+        for (const auto& item : *this) {
+            keys.insert(item.first);
+        }
+        return keys;
+    }
 
     /// Returns true if all tensors in the map have the same size.
     bool IsSizeSynchronized() const;
--- a/cpp/open3d/t/geometry/TriangleMesh.cpp
+++ b/cpp/open3d/t/geometry/TriangleMesh.cpp
@@ -324,12 +324,22 @@ TriangleMesh TriangleMesh::SimplifyQuadricDecimation(
                 target_reduction);
     }
 
-    auto polydata = CreateVtkPolyDataFromGeometry(*this);
+    // exclude triangle attributes because they will not be preserved
+    auto polydata =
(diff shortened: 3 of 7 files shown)


ssheorey (approved):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/5347)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @benjaminum and @yxlao)

___
*[`cpp/open3d/t/geometry/VtkUtils.cpp` line 306 at r2](https://reviewable.io/reviews/isl-org/open3d/5347#-N7rmyG5AF20uX17sxoE:-N7rmyG5AF20uX17sxoF:b-jyoj3h) ([raw file](https://github.com/isl-org/open3d/blob/22e05ddcd3dae5855a19608517f4b4b5cad016e5/cpp/open3d/t/geometry/VtkUtils.cpp#L306)):*
> ```cpp
> /// \param copy If true always create a copy for attribute arrays.
> /// \param include A set of keys to select which attributes should be added.
> ///                The special key "*" includes all attributes.
> ```

"*" may hint that globbing / regexp is supported. Perhaps use a special constant (e.g.: TensorMap::ALL_KEYS) instead? 
Also. document precedence of include and exclude. From code below, looks like exclude has precedence over include. i.e. if include = {"colors"} and exclude = {"colors'} then "colors" is excluded. Or should this be an error?
___
*[`cpp/open3d/t/geometry/VtkUtils.cpp` line 340 at r2](https://reviewable.io/reviews/isl-org/open3d/5347#-N7rqviBEQpxgz0XmlOC:-N7rqviC1PXCKpQTocwK:b-2lre2o) ([raw file](https://github.com/isl-org/open3d/blob/22e05ddcd3dae5855a19608517f4b4b5cad016e5/cpp/open3d/t/geometry/VtkUtils.cpp#L340)):*
> ```cpp
>                     "Ignoring attribute '{}' for TensorMap with primary key "
>                     "'{}'",
>                     key_tensor.first, tmap.GetPrimaryKey());
> ```

We shouldn't need a `LogWarning` for doing what the user asked us to do, i.e. if they explicitly asked to exclude "colors", no need to warn them about this.

benjaminum (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/5347)* status: 0 of 7 files reviewed, 3 unresolved discussions (waiting on @ssheorey and @yxlao)

___
*[`cpp/open3d/t/geometry/VtkUtils.cpp` line 306 at r2](https://reviewable.io/reviews/isl-org/open3d/5347#-N7rmyG5AF20uX17sxoE:-N7u_xnl7E0a-3dsrdWN:b-vmze9a) ([raw file](https://github.com/isl-org/open3d/blob/22e05ddcd3dae5855a19608517f4b4b5cad016e5/cpp/open3d/t/geometry/VtkUtils.cpp#L306)):*

"*" may hint that globbing / regexp is supported. Perhaps use a special constant (e.g.: TensorMap::ALL_KEYS) instead? 
Also. document precedence of include and exclude. From code below, looks like exclude has precedence over include. i.e. if include = {"colors"} and exclude = {"colors'} then "colors" is excluded. Or should this be an error?
</blockquote></details>

Done.  I changed to signature to avoid the special "*" element and improved the docstring.
___
*[`cpp/open3d/t/geometry/VtkUtils.cpp` line 340 at r2](https://reviewable.io/reviews/isl-org/open3d/5347#-N7rqviBEQpxgz0XmlOC:-N7uilCz-ok2iBoc0qnS:b3566au) ([raw file](https://github.com/isl-org/open3d/blob/22e05ddcd3dae5855a19608517f4b4b5cad016e5/cpp/open3d/t/geometry/VtkUtils.cpp#L340)):*

We shouldn't need a `LogWarning` for doing what the user asked us to do, i.e. if they explicitly asked to exclude "colors", no need to warn them about this.
</blockquote></details>

The function is not really meant to be directly used by the user and we decide which attributes to keep/exclude based on the algorithm we want to use.
However, if a user wants to combine o3d and vtk in C++ they can use the function; we don't hide the functionality.

Issue #3713, opened 2021-07-08
Return the original index of cropped point cloud by SelectionPolygonVolume


HowcanoeWang:
**Is your feature request related to a problem? Please describe.**
Either `open3d.visualization.SelectionPolygonVolume.crop_point_cloud()` or `open3d.geometry.PointCloud.crop()` return the point cloud type directly. However, sometimes the cropped points index are also useful.

And I check the source code of cpp (I am a python user and very new to cpp), it seems these functions utilized by
```cpp
return input.SelectByIndex(CropInPolygon(input.points_));
```
which get the point index first, and then `selected_by_index` to get the cropped point cloud.

**Describe the solution you'd like**
Add a control parameter of `return_index` in the function:
```python
cropped_pcd = .SelectionPolygonVolume.crop_point_cloud(raw_pcd, return_index=False)

cropped_pcd, cropped_index = .SelectionPolygonVolume.crop_point_cloud(raw_pcd, return_index=True)
```

or add an new function `crop_point_cloud_and_trace()`:
```python
cropped_pcd, cropped_index = .SelectionPolygonVolume.crop_point_cloud_and_trace(raw_pcd)
```

**Additional context**
Then reason why I want to get the cropping point index, is willing to accelerate the batch cropping of large amount of polygons in a very large point cloud

For example, I have a large point cloud with 8000,0000 points, and I need to extract and save 1000 sub point clouds from corresponding 1000 polygon boundries. Current processing time is around 1.5 hours by for loop of each polygon boundary, and it is really necessary to accelerate this process.

One idea is after cropping one polygon, remove those cropped points in raw point cloud, it will gradually decrease the points that need to sorted in the next iteration. So getting the index of previous cropped point is really important.

Or any other suggestions to solve current dilemma?

PR #5354, opened 2022-07-24
Make SelectPolygonVolume.CropInPolygon return indices

jamesdi1993:
Following issue created at https://github.com/isl-org/Open3D/issues/3713, in some circumstance it would be really great to have the indices of the cropped point clouds in SelectionVolume.CropInPolygon() function, for example, when the point cloud size is very large, and it's expensive to make a copy of the cropped point clouds. 

This PR proposes to make the SelectionVolume.CropInPolygon() a public method, which directly returns the indices for the cropped point cloud. Python bindings function is added to make it callable from python.

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/5354)

### DIFF
--- a/cpp/open3d/visualization/utility/SelectionPolygonVolume.h
+++ b/cpp/open3d/visualization/utility/SelectionPolygonVolume.h
@@ -63,15 +63,18 @@ public:
     /// \param input The input triangle mesh.
     std::shared_ptr<geometry::TriangleMesh> CropTriangleMesh(
             const geometry::TriangleMesh &input) const;
+    /// Function to crop point cloud with polygon boundaries
+    ///
+    /// \param input The input point Cloud.
+    std::vector<size_t> CropInPolygon(const geometry::PointCloud &input) const;
 
-    std::vector<size_t> CropInPolygon(
-            const std::vector<Eigen::Vector3d> &input) const;
 private:
     std::shared_ptr<geometry::PointCloud> CropPointCloudInPolygon(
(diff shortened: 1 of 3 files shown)


cpp/open3d/visualization/utility/SelectionPolygonVolume.h
```diff
@@ -64,13 +64,14 @@ class SelectionPolygonVolume : public utility::IJsonConvertible {
     std::shared_ptr<geometry::TriangleMesh> CropTriangleMesh(
             const geometry::TriangleMesh &input) const;
 
+    std::vector<size_t> CropInPolygon(
+            const std::vector<Eigen::Vector3d> &input) const;
```
yuecideng, line 68: If this method is set public, I think it would be better to change input from `std::vector<Eigen::Vector3d>` to `geometry::PointCloud`.

yuecideng:
Hi, thanks for this PR. Please see [style guide](http://www.open3d.org/docs/release/contribute/styleguide.html#style-guide) and apply it to this PR.

cpp/open3d/visualization/utility/SelectionPolygonVolume.h
jamesdi1993: Sounds good, added a public api to take `geometry::PointCloud` as input, which internally uses the private api to crop.

yuecideng (approved):

yxlao (approved):

Issue #5344, opened 2022-07-21
Blocky / incomplete raycast output


jdavidberger:
### Checklist

- [X] I have searched for [similar issues](https://github.com/isl-org/Open3D/issues).
- [X] For Python issues, I have tested with the [latest development wheel](http://www.open3d.org/docs/latest/getting_started.html#development-version-pip).
- [X] I have checked the [release documentation](http://www.open3d.org/docs/release/) and the [latest documentation](http://www.open3d.org/docs/latest/) (for `master` branch).

### Describe the issue

I'm working on code derived from the new SLAM reconstruction model and so far it works very well and is very fast, with the exception of situations in which the raycasting to generate synthetic frames produces bad output. This always seems to be the case for larger voxel grid sizes -- say, .010 but I have seen it even on .002 voxel grid sizes.  

### .01 output

The artifacting is worse the larger the voxel grid is.

Raycast:
![raycast_vs01](https://user-images.githubusercontent.com/8194827/180270025-5dccad23-a8fd-49c3-88de-5641ddd09975.png)

Pointcloud: 
![pointcloud_vs01](https://user-images.githubusercontent.com/8194827/180270176-2bd64a05-b7b3-4761-827d-8fef6f1bcd02.png)

Frustrum (plotted voxels from -.5 to .5 tsdf, same frustrum given to raycast)
![frustrum_v01](https://user-images.githubusercontent.com/8194827/180270396-44a7b06b-9424-4075-b2c2-fe7420422f27.png)

Raycast superimposed on PC:
![raycast_over_pc_vs01](https://user-images.githubusercontent.com/8194827/180270569-148a6047-dbca-4290-aa24-4dea0b7fe110.png)

Raycast superimposed over frustrum
![raycast_over_frustrum_vs01](https://user-images.githubusercontent.com/8194827/180270809-94868351-42f7-4784-9821-acea074e3095.png)

Note that my tooling regenerates everything to turn on frustrum visibility; each call to raycast generates different splotchy output seemingly at random

### .003 output

 Same artifacts at .003 on same dataset

![raycast_vs003](https://user-images.githubusercontent.com/8194827/180271206-08d38c7d-bb60-4b88-8868-29cb9c2c124f.png)
![pointcloud_vs003](https://user-images.githubusercontent.com/8194827/180271203-b8188f9d-21e6-4154-99fe-3a759e334dae.png)

[...]

PR #5358, opened 2022-07-25
Fixed an issue in EstimateRange

jdavidberger:
This also gracefully handles the preallocated buffer being too small and adds a heuristic which estimates the number of fragments needed. 

This is the bug responsible for #5344 and fixes that issue absolutely:

From GPU:
image

The reason it didn't trigger on lower voxel sizes is that with a small enough voxel size, each reprojection only took 1 fragment anyway.

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/5358)


jdavidberger:
I also added a commit to store the allocated buffer to avoid a re-allocation for that fragment buffer each call into raycast

theNded (changes requested):
Thanks for fixing the issue! Please see my review comments, any discussion is welcome. After reaching a consensus we can merge this PR.

> Reviewed 6 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/5358#-:-N7vEQWb2eKSzNSfns1N:b76evfl)* status: all files reviewed, 2 unresolved discussions (waiting on @benjaminum and @jdavidberger)

___
*[`cpp/open3d/t/geometry/VoxelBlockGrid.h` line 256 at r1](https://reviewable.io/reviews/isl-org/open3d/5358#-N7vDtob8OrgWFNQjrc_:-N7vDtob8OrgWFNQjrca:bsfvvxi) ([raw file](https://github.com/isl-org/open3d/blob/d093c8b805108de6a90b31ac07b7c9f1643650c2/cpp/open3d/t/geometry/VoxelBlockGrid.h#L256)):*
> ```objc
> 
>     // Allocated fragment buffer for reuse in depth estimation
>     core::Tensor fragment_buffer_;
> ```

This seems a little bit ad hoc. Would be better to remove it as a member function
___
*[`cpp/open3d/t/geometry/kernel/VoxelBlockGridImpl.h` line 331 at r1](https://reviewable.io/reviews/isl-org/open3d/5358#-N7vE6HIAeaKDQsMk4Uo:-N7vE6HIAeaKDQsMk4Up:bgb91gr) ([raw file](https://github.com/isl-org/open3d/blob/d093c8b805108de6a90b31ac07b7c9f1643650c2/cpp/open3d/t/geometry/kernel/VoxelBlockGridImpl.h#L331)):*
> ```objc
>          float depth_min,
>          float depth_max,
>          core::Tensor& fragment_buffer) {
> ```

In general I would prefer to keep it as a static variable only presented in the function scope, like
```
static core::Tensor fragment_buffer;
if (fragment_buffer.GetDataPtr() == 0) ...
```

jdavidberger:
The main issue with this is thread safety. As is it's not particularly thread safe but if you lock properly around the VoxelGrid it would be; here if you have two voxelgrids churning it'll have issues. So I'd say this is workable with ```thread_local``` and just let each thread have one. 

Are there CI checks for memory leaks? Because this would flag that there since it technically is a memory leak.

theNded (approved):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/5358)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @benjaminum)

jdavidberger:
Is there anything more I need to do in this and #5304 to get them merged or is it still in review?

theNded:
Could you please fix the style? `make apply-style` will automatically do the trick.

theNded:
@yuecideng could you please take a look at what may have triggered the issue in unit test on windows, if you have the environment and bandwidth?

yuecideng:
> @yuecideng could you please take a look at what may have triggered the issue in unit test on windows, if you have the environment and bandwidth?

Sure, I will try it.

nfrankisrg:
Hi I'm facing a similar issue. will this solution (or any other) will be merged into the latest version?

yuecideng:
@nfrankisrg I forget to submit my review. Sorry for the lately response. Now the ci is no problem, and I think it would be merged to master soom.

nfrankisrg:
after the recent change I'm unable to create a VoxelBlockGrid with device =  CUDA:0 or transfer the CPU one to the GPU using .cuda or .to

PR #5367, opened 2022-07-27
Slice mesh along plane

benjaminum:
This PR adds a method for slicing a mesh along planes creating line sets.

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/5367)


ssheorey (approved):
Hey, could you confirm that we are not leaking any Vtk symbols [I didn't check this]? Also, can we add OPEN3D_LOCAL (from Macro.h) attributes to the non-static functions in vtkutils.cpp, so that they are not available outside the Open3D DSO.

> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/5367#-:-N84mpaBAnt-oGt-8-TM:b3rici)* status: 0 of 10 files reviewed, 4 unresolved discussions (waiting on @benjaminum and @yxlao)

___
*[`cpp/open3d/t/geometry/TriangleMesh.h` line 680 at r2](https://reviewable.io/reviews/isl-org/open3d/5367#-N84dqOyCiM3b1KqND9F:-N84dqOyCiM3b1KqND9G:bzebmgm) ([raw file](https://github.com/isl-org/open3d/blob/635a59bd9d35b59110163aab93b573210cc6c988/cpp/open3d/t/geometry/TriangleMesh.h#L680)):*
> ```objc
>     LineSet SlicePlane(const core::Tensor &point,
>                        const core::Tensor &normal,
>                        const std::list<double> contour_values = {0.0}) const;
> ```

Use initializer_list<double> as a generic container, or use vector.
What do the contour_values represent? Are they distances from the given point? Clarify docs.
___
*[`cpp/pybind/t/geometry/trianglemesh.cpp` line 232 at r2](https://reviewable.io/reviews/isl-org/open3d/5367#-N84la4Q5YeB8Gi80YUB:-N84la4Q5YeB8Gi80YUC:bqbe684) ([raw file](https://github.com/isl-org/open3d/blob/635a59bd9d35b59110163aab93b573210cc6c988/cpp/pybind/t/geometry/trianglemesh.cpp#L232)):*
> ```cpp
> 
> This method generates slices as LineSet from the mesh at specific contour 
> values with respect to a plane.
> ```

Clarify meaning of contour_values.
___
*[`cpp/pybind/t/geometry/trianglemesh.cpp` line 241 at r2](https://reviewable.io/reviews/isl-org/open3d/5367#-N84khRQ5NmzsxYyk1f9:-N84khRQ5NmzsxYyk1fA:b8u313l) ([raw file](https://github.com/isl-org/open3d/blob/635a59bd9d35b59110163aab93b573210cc6c988/cpp/pybind/t/geometry/trianglemesh.cpp#L241)):*
> ```cpp
>     contour_values (list): A list of contour values at which slices will be
>         generated.
> ```

[nit] : extra newlines between args descriptions.
___
*[`python/test/t/geometry/test_trianglemesh.py` line 52 at r2](https://reviewable.io/reviews/isl-org/open3d/5367#-N84lFfgBPqPaf9ORJp3:-N84lFfgBPqPaf9ORJp4:b-l5u42z) ([raw file](https://github.com/isl-org/open3d/blob/635a59bd9d35b59110163aab93b573210cc6c988/python/test/t/geometry/test_trianglemesh.py#L52)):*
> ```Python
>     slices = box.slice_plane([0, 0.5, 0], [1, 1, 1], [-0.1, 0, 0.1])
>     assert slices.point['positions'].shape == (9, 3)
>     assert slices.line['indices'].shape == (9, 2)
> ```

Test lineset values as well.

benjaminum (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/5367)* status: 0 of 10 files reviewed, 4 unresolved discussions (waiting on @benjaminum, @ssheorey, and @yxlao)

___
*[`cpp/open3d/t/geometry/TriangleMesh.h` line 680 at r2](https://reviewable.io/reviews/isl-org/open3d/5367#-N84dqOyCiM3b1KqND9F:-N88ZGqS16xt-xoJFRfO:bos4wwf) ([raw file](https://github.com/isl-org/open3d/blob/635a59bd9d35b59110163aab93b573210cc6c988/cpp/open3d/t/geometry/TriangleMesh.h#L680)):*

Use initializer_list<double> as a generic container, or use vector.
What do the contour_values represent? Are they distances from the given point? Clarify docs.
</blockquote></details>

I wanted to use `std::vector<double>` but I noticed that it will be mapped to `o3d.utility.DoubleVector` by pybind which makes specifying contour values less comfortable. Do you think we can change this behaviour without breaking the old api?

yxlao (approved):
> Reviewed 1 of 10 files at r1, 8 of 8 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/5367)* status: all files reviewed, 4 unresolved discussions (waiting on @ssheorey)

PR #5391, opened 2022-08-02
changed output of colored icp to be identity if no correpondence is found

bkhanal-11:
When correspondence between source and target point cloud is not found, generalized icp returns their transformation to be identity. Updated this idea to colored icp.

---
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/5391)


scimad:
@bkhanal-11 , I think the verbosity level WARNING might be more appropriate.

reyanshsolis:
I would highly recommend using the new API. The legacy colored-ICP API has some issues, but it is going to be depreciated soon anyways. 
Documentation: 
http://www.open3d.org/docs/latest/tutorial/t_pipelines/t_icp_registration.html#Case-of-no-correspondences.

Thanks

bkhanal-11:
Thank you for letting me know. I will close this then.

PR #5400, opened 2022-08-04
sleep for 90 secs after VM create

yxlao:
An attempt to fix the following issue:

image

This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/5400)


ssheorey (approved):
Thanks, looks like it worked.

PR #5424, opened 2022-08-10
Enable radius search for Tensor Pointcloud 

yuecideng:
1. Improve some docs in `nns` module.
2. Eaable radius search for `EstimateNormals` and `EstimateColorGradients` of tensor PointCloud.

This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/5424)


cpp/open3d/t/geometry/PointCloud.h
```diff
@@ -422,24 +422,32 @@ class PointCloud : public Geometry, public DrawableGeometry {
...
-            const int max_nn = 30,
+            const utility::optional<int> max_nn = 30,
             const utility::optional<double> radius = utility::nullopt);
 
     /// \brief Function to compute point color gradients. If radius is provided,
     /// then HybridSearch is used, otherwise KNN-Search is used.
     /// Reference: Park, Q.-Y. Zhou, and V. Koltun,
     /// Colored Point Cloud Registration Revisited, ICCV, 2017.
-    /// \param max_nn Neighbor search max neighbors parameter [Default = 30].
+    /// It uses KNN search if only max_nn parameter is provided, Radius search
```
reyanshsolis, line 441: same as above.

cpp/open3d/t/geometry/PointCloud.h
```diff
@@ -422,24 +422,32 @@ class PointCloud : public Geometry, public DrawableGeometry {
 public:
     /// \brief Function to estimate point normals. If the point cloud normals
     /// exist, the estimated normals are oriented with respect to the same.
-    /// It uses KNN search if only max_nn parameter is provided, and
-    /// HybridSearch if radius parameter is also provided.
-    /// \param max_nn Neighbor search max neighbors parameter [Default = 30].
-    /// \param radius [optional] Neighbor search radius parameter to use
-    /// HybridSearch. [Recommended ~1.4x voxel size].
+    /// It uses KNN search if only max_nn parameter is provided, Radius search
```
reyanshsolis, line 425: Mention `(Not recommended to use on GPU)` next to KNN search, and `(Recommended)` next to HybridSearch. To caution users about the performance of using KNN search on GPU.

reyanshsolis (approved):
LGTM. My only concern is that by default it will use KNN search for the functionality, which is not the recommended search on GPU due to performance, however, the advantage is at least a number here (like max_knn = 30) as a default makes sense unlike a default param for radius for a general point cloud. I would highly recommend adding a tensor-point cloud tutorial as soon as possible and notifying the users about the recommended configs and issues.

yuecideng:
> I would highly recommend adding a tensor-point cloud tutorial as soon as possible and notifying the users about the recommended configs and issues.

@reyanshsolis Sure, we may discuss about this later.

yuecideng:
image

It seems that the `RadiusSearch`  in CUDA is also not very well. Should we also point it out in docs? @reyanshsolis

yxlao (approved):
> Reviewed 3 of 10 files at r1, 12 of 12 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/5424)* status: all files reviewed, 2 unresolved discussions (waiting on @junha-l and @yuecideng)

PR #5444, opened 2022-08-17
A slight efficiency improvement

Xiang-Zeng:
A slight efficiency improvement of InitializePointCloudForColoredICP() method. 
The $dp^T (f(p')-p)$ in Eq. 10 of (Park 2017) can be simplified as $dp^T (p'-p)$, since $dp^T f(p')=dp^T (p'-n_p (s-p')^T n_p) = dp^T p'$. 
(Note that $dp^T n_p =0$) 
Hence there is no need to calculate f(p').

- - -
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/5444)


Xiang-Zeng:
Further, we can discover that Cp( f(s) ) = Cp (s). Hence the equations (18) and (28) in Park (2017) can also be simplified, which may or may not improve the efficiency. I haven't try such modification yet.

yxlao:
LGTM if the test passes.

yxlao:
I pushed a commit to check before and after values. They are not exactly matching. To run the comparison:

```sh
make install-pip-package -j
open3d example pipelines/colored_pointcloud_registration
```

I am getting the following outputs:

```
...
[Open3D WARNING] A_old != A_new
[Open3D WARNING] A_old: {0.00238263, -0.0105446, 0.0104479}
[Open3D WARNING] A_new: {-0.00144744, -0.0176024, 0.00419813}
[Open3D WARNING] A_old != A_new
[Open3D WARNING] A_old: {0.00378348, 0.00951644, -0.0130656}
[Open3D WARNING] A_new: {0.0, 0.00254440, -0.0192394}
...
```

@Xiang-Zeng Can you do some debugging regarding this?

Xiang-Zeng:
@yxlao Sure.
Before submitting the pull request, I did some comparison on the "color_gradient_" of each point calculated before and after the modification, they had little difference.
I'll do more comparison on the output gradients and the final Rt based on your commit.

Xiang-Zeng:
The matrix A will no doubt be different after the modification, but the solution of Ax=b will remain the same. 
Here are comparisons of gradients and final Rt (see the images or attached files). They are almost identical.
compare_Rt
compare_gradients

[comparison-open3d.zip](https://github.com/isl-org/Open3D/files/9391226/comparison-open3d.zip)

@yxlao

yxlao (approved):

yxlao:
@Xiang-Zeng thanks for the confirmation. LGTM.

PR #5495, opened 2022-08-31
Update libc++ and libc++abi linking

yxlao:
- Fixes issue for ubuntu 18.04 ARM CI: https://bugs.launchpad.net/ubuntu/+source/apt/+bug/1988563
- Update libc++/libc++abi linking
- Allow `OPEN3D_THIRD_PARTY_DOWNLOAD_DIR` to be customized
- Use miniconda instead of pyenv

image

This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/5495)


3rdparty/find_dependencies.cmake
```diff
@@ -1193,16 +1193,18 @@ if(BUILD_GUI)
                                  clang-10
                                  clang-9
                                  clang-8
-                                 clang-7)
+                                 clang-7
```
ssheorey, line 1196: Finding libc++ location for all versions of llvm / clang is difficult to maintain. Perhaps we could assume that clang is in the path and run it to check it's version and query it for the location of libc++ and libc++abi? This would also let it work in Redhat based Linux, where the library folder is different.

3rdparty/find_dependencies.cmake
```diff
@@ -1267,16 +1298,22 @@ if(BUILD_GUI)
...
-                             REQUIRED)
+                )
                 get_filename_component(CLANG_LIBDIR ${CPPABI_LIBRARY} DIRECTORY)
             endif()
+
             # Find clang libraries at the exact path ${CLANG_LIBDIR}.
+            if (CLANG_LIBDIR)
+                message(STATUS "Using CLANG_LIBDIR: ${CLANG_LIBDIR}")
+            else()
+                message(FATAL_ERROR "Cannot find matchig libc++ and libc++abi libraries with version >=7.")
```
ssheorey, line 1309: typo: matching

util/install_deps_ubuntu.sh
```diff
@@ -32,20 +33,33 @@ deps=(
     libtool
 )
 
-# Ubuntu ARM64
-if [ "$(uname -m)" == "aarch64" ]; then
-    # For LAPACK
-    deps+=("gfortran")
+source /etc/lsb-release
```
ssheorey, line 36: [nit] We can use eval to prevent polluting with unwanted variables from lsb-release.
https://unix.stackexchange.com/a/266630

ssheorey (approved):

3rdparty/find_dependencies.cmake
yxlao: I tried `clang -print-search-dirs` on both Fedora and Ubuntu. They have slightly different behaviors. We can revisit this sometime later.

3rdparty/find_dependencies.cmake
yxlao: fixed

util/install_deps_ubuntu.sh
yxlao: done.

PR #5498, opened 2022-09-02
Add wireframe mode to visualizer

errissa:
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/5498)

### DIFF
--- a/cpp/open3d/visualization/gui/Materials/unlitLine.mat
+++ b/cpp/open3d/visualization/gui/Materials/unlitLine.mat
@@ -9,7 +9,8 @@ material {
     vertexDomain: world,
 
     parameters : [
-        { type : float3,    name : baseColor },
+        { type : float4,    name : baseColor },
+        { type : float4,    name : emissiveColor },
         { type : float,     name : lineWidth }
     ],
     requires: [
@@ -53,6 +54,9 @@ fragment {
     void material(inout MaterialInputs material) {
         prepareMaterial(material);
 
-        material.baseColor.rgb = materialParams.baseColor*getColor().rgb;
+        float combined_alpha = getColor().a * materialParams.baseColor.a;
+        material.baseColor.rgb = materialParams.baseColor.rgb * getColor().rgb * combined_alpha;
+        material.baseColor.a = combined_alpha;
+        material.emissive = materialParams.emissiveColor;
 }
 }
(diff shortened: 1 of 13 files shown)


errissa (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/5498)* status: 0 of 13 files reviewed, 1 unresolved discussion (waiting on @benjaminum)

___
*[`cpp/open3d/visualization/visualizer/GuiVisualizer.cpp` line 515 at r1](https://reviewable.io/reviews/isl-org/open3d/5498#-NAxH7JN8NZ6_JD_HyOu:-NAyGF-qCqkf97CBvn07:b-4aat07) ([raw file](https://github.com/isl-org/open3d/blob/5ffef87f03d1612ec819fcd24d8e7487347b2da5/cpp/open3d/visualization/visualizer/GuiVisualizer.cpp#L515)):*

Just a general question, would it be possible to render as wireframe without creating a LineSet?
I noticed some problems with the visualizer confusing its state (switch shader or enable raw mode before or after using wireframe mode)
</blockquote></details>

Unfortunately, not for a couple of reasons. 1) the 'easy' way to do that would be to use the OpenGL`PolygonMode` setting but it isn't support by Filament; 2) we need the line set to create the necessary per-vertex data for the wide lines shader. Since Filament doesn't support geometry shaders we create the wide lines geometry in the vertex shader which requires augmenting the per-vertex data of the lines with some additional information.

I had noticed some issues which I thought I had fixed. I will continue to investigate.

errissa (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/5498)* status: 0 of 13 files reviewed, 1 unresolved discussion (waiting on @benjaminum)

___
*[`cpp/open3d/visualization/visualizer/GuiVisualizer.cpp` line 515 at r1](https://reviewable.io/reviews/isl-org/open3d/5498#-NAxH7JN8NZ6_JD_HyOu:-NAydDyW588AT6xBDF2B:b-896fix) ([raw file](https://github.com/isl-org/open3d/blob/5ffef87f03d1612ec819fcd24d8e7487347b2da5/cpp/open3d/visualization/visualizer/GuiVisualizer.cpp#L515)):*

Unfortunately, not for a couple of reasons. 1) the 'easy' way to do that would be to use the OpenGL`PolygonMode` setting but it isn't support by Filament; 2) we need the line set to create the necessary per-vertex data for the wide lines shader. Since Filament doesn't support geometry shaders we create the wide lines geometry in the vertex shader which requires augmenting the per-vertex data of the lines with some additional information.

I had noticed some issues which I thought I had fixed. I will continue to investigate.
</blockquote></details>

Done.

benjaminum (approved):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/5498)* status: 0 of 13 files reviewed, all discussions resolved (waiting on @benjaminum)

PR #5509, opened 2022-09-07
Enable pickling for tensor and tensor based geometry

yuecideng:
This PR makes `Tensor`, `TensorMap` and tensor based geometry pickling.

Current behaviours:
- Tensor will be on the same device after de-serialization.
- Non contiguous tensor will be converted to contiguous tensor after de-serilization.

- - -
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/5509)


yxlao:
LGTM after adding python unit tests.

yuecideng:
- Add python unit tests for pickling support.
- Minor change `PointCloud::ToString()` output to be consistent with `TriangleMesh`:
```
PointCloud on CPU:0 [2 points (Float32)].
Attributes: normals (dtype = Float32, shape = {2, 3}).
```
@yxlao

.gitignore
```diff
@@ -90,3 +90,6 @@ docs/Doxyfile
 docs/getting_started.rst
 docs/docker.rst
 docs/tensorboard.md
+
+# test
+*.pkl
```
yxlao, line 95: remove this

cpp/pybind/core/tensor.cpp
```diff
@@ -356,6 +356,34 @@ void pybind_core_tensor(py::module& m) {
...
+                return py::make_tuple(t.GetDevice(),
+                                      TensorToPyArray(t.To(Device("CPU:0"))));
+            },
+            [](py::tuple t) {
+                // __setstate__
+                if (t.size() != 2) {
+                    utility::LogError(
+                            "Invalid state! Expecting a tuple of size 2.");
+                }
+                const Device& device = t[0].cast<Device>();
```
yxlao, line 375: Use `Device::IsAvailable()`. For example, when pickling, we might store `CUDA:1`, but another computer may only have `CUDA:0`. Same for others.

cpp/pybind/t/geometry/image.cpp
```diff
@@ -101,6 +101,22 @@ void pybind_image(py::module &m) {
                  "tensor"_a);
     docstring::ClassMethodDocInject(m, "Image", "__init__",
                                     map_shared_argument_docstrings);
+
+    // Pickle support.
+    image.def(py::pickle(
+            [](const Image &image) {
+                // __getstate__
+                return py::make_tuple(image.AsTensor());
```
yxlao, line 109: Do we need to transfer to CPU?

cpp/pybind/t/geometry/image.cpp
```diff
@@ -282,6 +298,26 @@ void pybind_image(py::module &m) {
...
+
+            // Pickling support.
+            .def(py::pickle(
+                    [](const RGBDImage &rgbd) {
+                        // __getstate__
+                        return py::make_tuple(rgbd.color_, rgbd.depth_,
+                                              rgbd.aligned_);
+                    },
+                    [](py::tuple t) {
+                        // __setstate__
```
yxlao, line 310: Same, do we need to check device as well?

cpp/pybind/core/device.cpp
```diff
@@ -46,7 +46,20 @@ void pybind_core_device(py::module &m) {
...
-            .def("get_id", &Device::GetID);
+            .def("get_id", &Device::GetID)
+            .def(py::pickle(
+                    [](const Device &d) {
+                        return py::make_tuple(d.GetType(), d.GetID());
+                    },
+                    [](py::tuple t) {
+                        if (t.size() != 2) {
+                            utility::LogError(
+                                    "Invalid state! Expecting a tuple of size "
```
yxlao, line 57: Add a comment: "Cannot unpickle Device." It is useful for the user to know which class went wrong. Same for others.

cpp/pybind/core/tensor.cpp
```diff
@@ -356,6 +356,34 @@ void pybind_core_tensor(py::module& m) {
...
+                if (device.IsCUDA() && !core::cuda::IsAvailable()) {
+                    utility::LogWarning(
+                            "CUDA is not available, tensor will be "
```
yxlao, line 378: Change this comment as well. Same for others.

python/test/t/geometry/test_image.py
```diff
@@ -76,3 +77,12 @@ def test_buffer_protocol_cpu(device):
         im = im.to(device=device)
         dst_t = np.asarray(im.cpu())
         np.testing.assert_array_equal(src_t, dst_t)
+
+
+@pytest.mark.parametrize("device", list_devices())
+def test_pickle(device):
+    img = o3d.t.geometry.Image(o3c.Tensor.ones((10, 10, 3), o3c.uint8, device))
+    pickle.dump(img, open("img.pkl", "wb"))
```
yxlao, line 85: Use 
```python
import tempfile

with tempfile.TemporaryDirectory() as temp_dir:
    pickle_path = xxx
```

Same for others.

yxlao (changes requested):

cpp/pybind/t/geometry/image.cpp
yuecideng: The conversion to CPU is performed in `Tensor` side, so we don't need here.

cpp/pybind/t/geometry/image.cpp
yuecideng: Same, the device will be checked by the tensor inside Image.

.gitignore
yuecideng: Done.

cpp/pybind/core/tensor.cpp
yuecideng: Done.

cpp/pybind/core/device.cpp
yuecideng: Done.

python/test/t/geometry/test_image.py
yuecideng: Done.

yxlao (approved):

reyanshsolis:
Great work. It would be very helpful to add a brief documentation example on the Tensor page.
TODO: Check newly added features that are missing documentation and make a priority based checklist for the same.

yuecideng:
> It would be very helpful to add a brief documentation example on the Tensor page

Sounds good, I will add it later.

PR #5546, opened 2022-09-23
Noise model for depth images

yxlao:
## Usage

```python
import open3d as o3d

# Redwood Indoor LivingRoom1 (Augmented ICL-NUIM)
# http://redwood-data.org/indoor/
data = o3d.data.RedwoodIndoorLivingRoom1()
noise_model_path = data.noise_model_path
im_src_path = data.depth_paths[0]
depth_scale = 1000.0

# Read clean depth image (uint16)
im_src = o3d.t.io.read_image(im_src_path)

# Run noise model simulation
simulator = o3d.t.geometry.DepthNoiseSimulator(noise_model_path)
im_dst = simulator.simulate(im_src, depth_scale=depth_scale)

# Save noisy depth image (uint16)
o3d.t.io.write_image("noisy_depth.png", im_dst)
```

## Numerically matching the original implementation 

The new implementation is designed to match [the original Python implementation](http://redwood-data.org/indoor/data/simdepth.py) numerically.  

To verify the exact match, call `simulator.enable_deterministic_debug_mode()`. In the original implementation (in `python/test/t/geometry/test_image.py`), also set `deterministic=True` for the original implementation. Python unit test is included.

## Performance 
~1000x faster than the original Python implementation, tested under a 16 core CPU.
```bash
# Generate 640x480 depth image
Old run time: 1.0949831008911133
New run time: 0.0009479522705078125
```

This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/5546)


ssheorey (changes requested):
I think `open3d.t.io` is a better place for the depth noise simulator compared to `open3d.t.geometry`. This will typically be used as part of data loading code.

> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/5546#-:-ND6O1Gx8Mio9KTmnw2x:b-j2gj0d)* status: 0 of 21 files reviewed, 5 unresolved discussions (waiting on @ssheorey and @yxlao)

___
*[`cpp/open3d/t/geometry/Image.h` line 377 at r1](https://reviewable.io/reviews/isl-org/open3d/5546#-ND6MdGZ1VNyiU7cSctQ:-ND6MdGZ1VNyiU7cSctR:b-tc20zx) ([raw file](https://github.com/isl-org/open3d/blob/54a759576b10e26bdc962e6df2c3c30b524b3f23/cpp/open3d/t/geometry/Image.h#L377)):*
> ```objc
>     /// is UInt16, the depth_scale is typically larger than 1.0, e.g. it can be
>     /// 1000.0.
>     /// \return Noisy depth image with the same shape and dtype as \p im_src.
> ```

We don't normally have examples in C++, but I think we should start adding C++ example snippets to all new code, as we have started doing with Python. The C++ version of the PR description example would be pretty good.
___
*[`cpp/open3d/t/geometry/Image.cpp` line 602 at r1](https://reviewable.io/reviews/isl-org/open3d/5546#-ND6JKs823UqZVAK5Sgm:-ND6JKs97OnEUDA8ws2S:bo703a5) ([raw file](https://github.com/isl-org/open3d/blob/54a759576b10e26bdc962e6df2c3c30b524b3f23/cpp/open3d/t/geometry/Image.cpp#L602)):*
> ```cpp
>         if (line.empty()) {
>             continue;
>         } else {
> ```

nit: if (!line.empty()) {... }
___
*[`cpp/open3d/t/geometry/Image.cpp` line 663 at r1](https://reviewable.io/reviews/isl-org/open3d/5546#-ND6KXFh1a-05tGnv_QW:-ND6KXFh1a-05tGnv_QX:b-4qz6jt) ([raw file](https://github.com/isl-org/open3d/blob/54a759576b10e26bdc962e6df2c3c30b524b3f23/cpp/open3d/t/geometry/Image.cpp#L663)):*
> ```cpp
>     int height = im_src.GetRows();
> 
>     im_src_tensor = im_src_tensor.To(core::Device("CPU:0")).Contiguous();
> ```

[future] Is a CUDA PRNG the only thing preventing this from running on CUDA? Similar question for OneAPI / OpenCL as well. At some point we should integrate the PRNG from CUDA / MKL.
___
*[`cpp/open3d/t/geometry/Image.cpp` line 730 at r1](https://reviewable.io/reviews/isl-org/open3d/5546#-ND6KL-k-bc78K66x1bI:-ND6KL-k-bc78K66x1bJ:b-2869l0) ([raw file](https://github.com/isl-org/open3d/blob/54a759576b10e26bdc962e6df2c3c30b524b3f23/cpp/open3d/t/geometry/Image.cpp#L730)):*
> ```cpp
>         im_dst_tensor = (im_dst_tensor * depth_scale).To(core::UInt16);
>     }
>     assert(im_dst_tensor.GetDtype() == original_dtype);
> ```

assert can be removed if leftover from debugging.
___
*[`cpp/pybind/t/geometry/image.cpp` line 410 at r1](https://reviewable.io/reviews/isl-org/open3d/5546#-ND6MHr-2pt_Q9aWTufm:-ND6MHr-2pt_Q9aWTufn:b-80fmjs) ([raw file](https://github.com/isl-org/open3d/blob/54a759576b10e26bdc962e6df2c3c30b524b3f23/cpp/pybind/t/geometry/image.cpp#L410)):*
> ```cpp
>               "dtype is Float32, the depth_scale must be 1.0. If the dtype is "
>               "is UInt16, the depth_scale is typically larger than 1.0, e.g. "
>               "it can be 1000.0."}});
> ```

Add the example from the PR description here.

ssheorey (approved):

PR #5572, opened 2022-10-01
Add RemoveStatisticalOutliers to tensor PointCloud

yuecideng:
The benchmark is as follow:
image

- - -
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/5572)


cpp/open3d/t/geometry/PointCloud.h
```diff
@@ -379,6 +379,14 @@ class PointCloud : public Geometry, public DrawableGeometry {
     std::tuple<PointCloud, core::Tensor> RemoveRadiusOutliers(
             size_t nb_points, double search_radius) const;
 
+    /// \brief Remove points that have less than \p nb_points in a sphere of a
+    /// given radius. This function is not recommended to use on GPU.
+    ///
+    /// \param nb_neighbors Number of neighbors around the target point.
+    /// \param std_ratio Standard deviation ratio.
```
yxlao, line 386: Specify `\return` to show what are being returned.

cpp/pybind/t/geometry/pointcloud.cpp
```diff
@@ -253,6 +253,12 @@ The attributes of the point cloud have different levels::
                    "nb_points"_a, "search_radius"_a,
                    "Remove points that have less than nb_points neighbors in a "
                    "sphere of a given search radius.");
+    pointcloud.def(
+            "remove_statistical_outliers",
+            &PointCloud::RemoveStatisticalOutliers, "nb_neighbors"_a,
+            "std_ratio"_a,
+            "Remove points that have less than nb_points in a sphere of "
+            "a given radius. This function is not recommended to use on GPU.");
```
yxlao, line 261: Same, talk about what is being returned here.

yxlao (approved):

cpp/pybind/t/geometry/pointcloud.cpp
yuecideng: I notice that the other functions in pybind do not mention the return as well. Should we apply the same modification to them? @yxlao

Issue #5594, opened 2022-10-12
Cannot compile on Ubuntu 20.04 using the latest Eigen3 (3.3.7-2)


sebastienmascha:
### Checklist

- [X] I have searched for [similar issues](https://github.com/isl-org/Open3D/issues).
- [X] For Python issues, I have tested with the [latest development wheel](http://www.open3d.org/docs/latest/getting_started.html#development-version-pip).
- [X] I have checked the [release documentation](http://www.open3d.org/docs/release/) and the [latest documentation](http://www.open3d.org/docs/latest/) (for `master` branch).

#### I first cloned Open3D by:
```
git clone https://github.com/isl-org/Open3D.git
git clone --recursive https://github.com/isl-org/Open3D
cd Open3D
```

#### Then, I build Open3D (on Ubuntu 20.04, with CUDA 11.4.2) with:
```
sudo rm -rf build &&  sudo mkdir build && \
sudo cmake -S . -B build \
  -DBUILD_SHARED_LIBS=ON -DBUILD_PYTHON_MODULE=ON \
  -DPython3_ROOT=/usr/bin/python3 \
  -DUSE_SYSTEM_EIGEN3=ON && \
sudo cmake --build build -j10
```

### Error message

```shell
In file included from /home/seb/dev/contributions/Open3D/cpp/apps/OfflineReconstruction/LegacyReconstructionUtil.h:31,
                 from /home/seb/dev/contributions/Open3D/cpp/apps/OfflineReconstruction/OfflineReconstruction.cpp:28:
/home/seb/dev/contributions/Open3D/cpp/apps/OfflineReconstruction/DebugUtil.h:40:2: error: no matching function for call to 'Eigen::Matrix<double, 4, 4>::Matrix(<brace-enclosed initializer list>)'
   40 | });
      |  ^
In file included from /usr/include/eigen3/Eigen/Core:458,
                 from /home/seb/dev/contributions/Open3D/cpp/open3d/camera/PinholeCameraIntrinsic.h:29,
                 from /home/seb/dev/contributions/Open3D/cpp/open3d/Open3D.h:31,
                 from /home/seb/dev/contributions/Open3D/cpp/apps/OfflineReconstruction/FileSystemUtil.h:37,
                 from /home/seb/dev/contributions/Open3D/cpp/apps/OfflineReconstruction/Config.h:31,
[...]

PR #5595, opened 2022-10-12
Cannot compile on Ubuntu 20.04 using the latest Eigen3 (3.3.7-2) #5594

sebastienmascha:
https://github.com/isl-org/Open3D/issues/5594

- - -
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/5595)


sebastienmascha:
Cf. https://github.com/isl-org/Open3D/issues/5594
This is a bug using the default `libeigen3-dev` version from APT.
Currently APT install  the version `3.3.7` on Ubuntu 20.04.
Therefore, you cannot use some new features like _initializer_list constructors to Matrix and Array_ etc...

Removing this version and compiling Eigen version `3.4` works perfectly. I will close the issue.

PR #5610, opened 2022-10-21
Fix release docs jupyter size issue

yxlao:
- delete `*.ipynb` in output folder
- do not execute `draw_plotly.ipynb`, as WebGL widgets does not render well on our docs website yet

- - -
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/5610)


ssheorey (approved):
This fixes the large file size issue and not rendering images issue. Remaining issues:
- draw_plotly tutorial is not linked into the index, so not visible to users.
- There is no title.

Will that be in a separate PR?

Issue #5612, opened 2022-10-22
Many Linux & macOS wheels missing on PyPI for 0.16.0 release


johnthagen:
### Checklist

- [X] I have searched for [similar issues](https://github.com/isl-org/Open3D/issues).
- [X] For Python issues, I have tested with the [latest development wheel](http://www.open3d.org/docs/latest/getting_started.html#development-version-pip).
- [X] I have checked the [release documentation](http://www.open3d.org/docs/release/) and the [latest documentation](http://www.open3d.org/docs/latest/) (for `master` branch).

### Steps to reproduce the issue

The 0.16.0 release to PyPI:

- https://pypi.org/project/open3d/0.16.0/#files

 Is missing wheels for the following platforms:

- Python 3.8 & 3.9 Linux x86 (were provided in previous 0.15.2 release)
- Python 3.7-3.10 Linux ARM (3.7-3.9 were provided in previous 0.15.1 release)
- Python 3.8-3.10 Mac ARM (3.8-3.9 were provided in previous 0.15.1 release)

Please build and upload these wheels to the 0.16.0 release so that users can pin to a single release and deploy this to multiple OS/Arch combinations.

Related:

- #4535 

Maybe related historically to:

- #4796

### Open3D, Python and System information

```markdown
- Operating system: Linux & macOS
- Python version: Python 3.8 & 3.9
- Open3D version: 0.16.0
- System architecture: x86 & ARM
- Is this a remote workstation?: no
- How did you install Open3D?: pip
- Compiler version (if built from source): N/A
```

ssheorey:
We ran into PyPI project size limit while uploading v0.16.0 wheels. Here is the request to increase our storage limit to enable us to upload the remaining wheels:

https://github.com/pypa/pypi-support/issues/2311

We will upload them as soon as the request is granted. In the mean time, please download the wheels from the [github release page](https://github.com/isl-org/Open3D/releases/tag/v0.16.0)

johnthagen:
All missing wheels have been uploaded 🎉 :

- https://pypi.org/project/open3d/0.16.0/#files

PR #5614, opened 2022-10-23
Document support for Python 3.10 wheels

johnthagen:
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/5614)

3.10 wheels are included for 0.16.0:

- https://pypi.org/project/open3d/0.16.0/#files

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ For more, please visit the [Open3D documentation](http://www.open3d.org/docs).
 ## Python quick start
 
 Pre-built pip packages support Ubuntu 18.04+, macOS 10.15+ and Windows 10+
-(64-bit) with Python 3.6-3.9.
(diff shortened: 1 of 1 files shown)


johnthagen:
3.6 is published as supported, but there are no wheels:

https://github.com/isl-org/Open3D/blob/927c4e5bb10d5c260e5f04c80127ca9634e161a2/python/setup.py#L139

Are 3.6 wheels planned as part of:

- #5612

Or should setup.py be changed to remove 3.6 support?

yxlao:
LGTM for this PR. 

We don't provide Python 3.6 pre-compiled wheels anymore starting from 0.16 release. But, we still allow people to build from source with Python 3.6. Until we completely disable 3.6, we can still keep the description as it is.

yxlao (approved):

PR #5643, opened 2022-11-02
Add options to use system libraries for new dependencies

roehling:
This is a maintenance PR to ease Debian packaging of Open3D.

- - -
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/5643)


ssheorey (commented):
Thanks @roehling ! Looks good to me.
@yxlao Is it OK to use the system OpenSSL instead of BoringSSL?

cielavenir:
May I ask a question? What happens if the system have libssl-dev but the curl is built for gnutls (or nss)?

My version of cmake does not care about ssl library installation if system curl is used.

roehling:
You're right, it makes sense to not include (Open)SSL at all if CURL is used. I'll update my PR accordingly.

cielavenir:
@ssheorey How is this pull request going?

reyanshsolis:
BoringSSL was required by WebRTC and Curl was required for the dataset download manager. May you please check if WebRTC is working properly when BoringSSL is not present?

ssheorey:
Hi @roehling I had to make a change to `Download.cpp` to get it to build without warnings as errors on Ubuntu 22.04. The other change is informational.

[pr5643.txt](https://github.com/isl-org/Open3D/files/10516110/pr5643.txt)

Verified that everything works (https downloads, webrtc and VTK functionality) as usual with these system libraries.

ssheorey:
Closing this, since the updated PR #5893 is merged.

PR #5672, opened 2022-11-10
Fix build issue on Apple M1

errissa:
Xcode CommandLine Tools 14.1 includes SDK13.0 as the default SDK. Numerous APIs are deprecated in 13.0 which causes warnings. On Apple Silicon systems Open3D builds Filament from scratch and one of its subprojects unconditionally adds `-Werror` to the compiler flags which causes a handful for deprecation warnings to become errors. A quick workaround is to build against SDK12. This PR ensures that if Open3D is configured with a user set `OSX_SYSROOT` that Filament is configured with the same value.

### DIFF
--- a/3rdparty/filament/filament_build.cmake
+++ b/3rdparty/filament/filament_build.cmake
@@ -66,5 +66,7 @@ ExternalProject_Add(
         -DFILAMENT_SUPPORTS_VULKAN=OFF
         -DFILAMENT_SKIP_SAMPLES=ON
         -DFILAMENT_OPENGL_HANDLE_ARENA_SIZE_IN_MB=20 # to support many small entities
+        -DCMAKE_OSX_SYSROOT=${CMAKE_OSX_SYSROOT}
+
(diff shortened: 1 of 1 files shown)


yxlao (approved):
1. On my M1 macs, it works with or without this fix.
2. However, I have to use https://github.com/isl-org/Open3D/pull/5673 to build Open3D fully. @errissa could you review this as well?
3. If `CMAKE_OSX_SYSROOT` is a common flag that we should use, we could put it in `ExternalProject_CMAKE_ARGS` in `3rdparty/find_dependencies.cmake`. This way, it will also be used in other third-party libs.

errissa:
@yxlao #5673 fixes the compilation issue I was having and I think it's a better workaround than this PR. I will close this PR and consider whether it makes sense to support passing `CMAKE_OSX_SYSROOT` to third party libraries. I believe Apple recommends always building against the latest SDK even when deploying on older systems.

PR #5679, opened 2022-11-12
Build with libomp 11.1 and package it in the macOS wheel

ssheorey:
Package libomp v11.1.0, if it is not installed. Later version cause crash on x86_64 if PyTorch is already imported.
https://github.com/microsoft/LightGBM/issues/4229

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -59,7 +59,11 @@ jobs:
       - name: Install dependencies
         run: |
           brew install ccache pkg-config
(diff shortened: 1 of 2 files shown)


errissa (approved):
> Reviewed 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/5679)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @yxlao)

PR #5708, opened 2022-11-19
Fix bug for open3d cuda call

yuecideng:
Use `CUDA_CALL` without open3d namespace will result error during compilation.

- - -
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/5708)


ssheorey (approved):
Thanks for catching this!

PR #5709, opened 2022-11-19
Enable python copy&deepcopy for tensor geometry

yuecideng:
Use case:
```python
import open3d as o3d
import copy

Class Foo:
    def __init__(self, name='foo', pcd=None):
        self.name = name
        self.pcd = pcd

a = Foo('test', pcd=o3d.t.geometry.PointCloud())
b = copy.deepcopy(b)
```

Error will occur if no adding `bind_copy_functions` in pybind:
image

- - -
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/5709)


yuecideng:
Seems that one Ubuntu CUDA CI fails many times, although I rerun it several times.

ssheorey (approved):

PR #5761, opened 2022-12-06
remove normals from assimp importer

cielavenir:
## symptom

https://www.dropbox.com/s/7f0g474xf9nb5wm/teapot.stl

with this patch:

```
python3 -c 'import open3d,sys; mesh=open3d.io.read_triangle_mesh(sys.argv[1], enable_post_processing=True); print("triangles=%d, vertices=%d"%(len(mesh.triangles),len(mesh.vertices))); mesh=mesh.compute_vertex_normals(); mesh=mesh.simplify_quadric_decimation(target_number_of_triangles=400); open3d.io.write_triangle_mesh("out.stl", mesh)' ~/Downloads/teapot.stl 
triangles=894, vertices=480
```

https://www.dropbox.com/s/ad75o4vegilsvsb/out_patch.stl

without this patch:

```
python3 -c 'import open3d,sys; mesh=open3d.io.read_triangle_mesh(sys.argv[1], enable_post_processing=True); print("triangles=%d, vertices=%d"%(len(mesh.triangles),len(mesh.vertices))); mesh=mesh.compute_vertex_normals(); mesh=mesh.simplify_quadric_decimation(target_number_of_triangles=400); open3d.io.write_triangle_mesh("out.stl", mesh)' ~/Downloads/teapot.stl 
triangles=894, vertices=2310
```

https://www.dropbox.com/s/5xv1odt7bd610d4/out_nopatch.stl

In latter state, simplify_quadric_decimation gives broken mesh.

## method

The core issue is that aiProcess_JoinIdenticalVertices does not work well if the stl has normals because JoinIdenticalVertices considers the normals value as vertex normals while assimp STLLoader loads triangle normals. According to https://github.com/assimp/assimp/issues/407 , we need to remove normals component during import.

Also aiProcess_GenNormals (from aiProcessPreset_TargetRealtime_Fast) saves normal before running aiProcess_JoinIdenticalVertices, so it has to be removed as well.

As we dropped aiProcess_GenNormals, we ensure normals by calling ComputeVertexNormals.

/cc @jerasman @RunqiuBao

- - -
This change is [Reviewable](https://reviewable.io/reviews/isl-org/open3d/5761)

### DIFF
--- a/cpp/open3d/io/file_format/FileASSIMP.cpp
+++ b/cpp/open3d/io/file_format/FileASSIMP.cpp
@@ -257,6 +257,9 @@ bool ReadTriangleMeshUsingASSIMP(
         current_vidx += assimp_mesh->mNumVertices;
     }
 
+    // Compute triangle normals independent from assimp
+    mesh.ComputeTriangleNormals();
+
     // Now load the materials
     for (size_t i = 0; i < scene->mNumMaterials; ++i) {
         auto* mat = scene->mMaterials[i];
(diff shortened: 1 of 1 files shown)


cielavenir:
@errissa Our team decided to use `merge_close_vertices(0.0001)`. Thank you so much for the advice.

Issue #5856, opened 2023-01-22
Request for minimal oriented bounding box for a given point cloud


ligerlac:
### Checklist

- [X] I have searched for [similar issues](https://github.com/isl-org/Open3D/issues).
- [X] For Python issues, I have tested with the [latest development wheel](http://www.open3d.org/docs/latest/getting_started.html#development-version-pip).
- [X] I have checked the [release documentation](http://www.open3d.org/docs/release/) and the [latest documentation](http://www.open3d.org/docs/latest/) (for `master` branch).

### Proposed new feature or change

There are currently two functions that calculate a bounding box for a given point cloud:
`PointCloud.get_axis_aligned_bounding_box()` calculates the minimum axis-aligned bounding box w.r.t. to the lab frame, and `PointCloud.get_oriented_bounding_box()` calculates the minimum axis-aligned-bounding box w.r.t. to the coordinate system of the point cloud. Here, the coordinate system of the point cloud is determined via PCA. However, this is not necessarily the smallest bounding box of the point cloud. Other libraries already provide the smallest bounding box of a point cloud and it seems some Open3D users request it as well.

### References

previous discussion: [https://github.com/isl-org/Open3D/issues/5417](https://github.com/isl-org/Open3D/issues/5417)
Wikipedia about bounding boxes: [https://en.wikipedia.org/wiki/Minimum_bounding_box](https://en.wikipedia.org/wiki/Minimum_bounding_box)
implementation in different library: [https://github.com/compas-dev/compas/blob/main/src/compas/geometry/bbox/bbox_numpy.py](https://github.com/compas-dev/compas/blob/main/src/compas/geometry/bbox/bbox_numpy.py)

ssheorey:
Closed by PR #5857 
Thanks @ligerlac !

elun15:
Thanks so much @ligerlac !!! I tried the new version of o3d with the new feature and it works perfectly :D

PR #5857, opened 2023-01-22
New feature: minimal bounding box for a given point cloud

ligerlac:
Introduced new function that calculates the minimal bounding box for a given point cloud. So far, only the minimal axis aligned and the object-oriented bounding box are implemented.
 I have created a feature request here: [https://github.com/isl-org/Open3D/issues/5856](https://github.com/isl-org/Open3D/issues/5856).
I used the same algorithm as the library 'compas' [https://github.com/compas-dev/compas/blob/main/src/compas/geometry/bbox/bbox_numpy.py](https://github.com/compas-dev/compas/blob/main/src/compas/geometry/bbox/bbox_numpy.py)

- - -
This change is [Reviewable](https://reviewable.io/reviews/isl-org/Open3D/5857)

### DIFF
--- a/cpp/open3d/geometry/BoundingVolume.cpp
+++ b/cpp/open3d/geometry/BoundingVolume.cpp
@@ -34,6 +34,8 @@
 #include "open3d/geometry/TriangleMesh.h"
 #include "open3d/utility/Logging.h"
 
+#include<iostream>
+
 namespace open3d {
 namespace geometry {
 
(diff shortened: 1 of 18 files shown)


ssheorey:
Hi @ligerlac thanks for adding this new feature! 
- You can fix the style errors by running: `python util/check_style.py --apply`
- Please add documentation (in both C++ and Python bindings) for the new function in each of lineset, pointcloud, mesh, etc.

ligerlac:
> Hi @ligerlac thanks for adding this new feature!
> 
> * You can fix the style errors by running: `python util/check_style.py --apply`
> * Please add documentation (in both C++ and Python bindings) for the new function in each of lineset, pointcloud, mesh, etc.

Hi @ssheorey , I fixed the style errors and wrote additional documentation. Do we have to worry about the failed tests?

ssheorey:
Great, thanks @ligerlac for useful contribution! Merging...

PR #5866, opened 2023-01-26
Adding in memory support for xyz files

### LINKED ISSUE #1560: Support reading and writing file-like objects
I would like to write .ply bytes to a file-like object without a filename. I can't do that with the current API. I'm trying to avoid using the disk when unnecessary - my particular use case is sending a ply file over HTTP.

Adjusting the API to work with either filenames or file-like objects, like so:
```python
point_cloud = open3d.geometry.PointCloud()
point_cloud_bytes = io.BytesIO()
open3d.io.write_point_cloud(point_cloud_bytes, point_cloud)
decoded_point_cloud = open3d.io.read_point_cloud(point_cloud_bytes)
# point_cloud and decoded_point_cloud should be identical
```

I tried `tempfile.NamedTemporaryFile`, but was unable to get it to work. I also could write it to disk and read the bytes directly, but I'd prefer to keep things in memory.

**Additional context**
None.

samypr100:
## Type

-   [x] New feature (non-breaking change which adds functionality). Resolves #1560 (partially)

## Motivation and Context

The ability to read/write formats in memory is a typical ask due to a variety of use cases. Open3D has some in memory options for file formats such as PNG/JPG (e.g. [ReadPNGFromMemory](https://github.com/isl-org/Open3D/blob/master/cpp/open3d/io/file_format/FilePNG.cpp#L119)) so it's useful if other file formats had this kind of support.

## Checklist:

    to my code.
        updated accordingly.
    
## Description

The changes in this PR doesn't fully close #1560 since I believe there's other formats to still support such as `xyzn`, `xyzrgb,` `pts`, `pcd`, and maybe others.

This PR only adds full "in memory" support for `xyz` file format in order to keep the PR small, simple, and reviewable. It's also early so if somethings needs to change with the approach it's easier rather than trying to retro-fit to all formats.

Since we're reading contents from memory, `"auto"` would not quite work as there's no magic headers for all these files, so some changes were made so that `format` is specified for both Reading/Writing Point Clouds. I'm using a prefix of `mem::{format}` for the new in memory formats.

It simplifies the Python API when temporary files are used to achieve similar behavior like shown below.

```python
>>> import open3d as o3d
>>> xyz_bytes = b"x y z\nx y z\n"
>>> pcr = o3d.io.read_point_cloud_from_bytes(xyz_bytes, "mem::xyz")
>>> pcb = o3d.io.write_point_cloud_to_bytes(pc, "mem::xyz")
>>> pcb.decode("utf8")
"x y z\nx y z\n"
```

The issue asked to mimic something like the file-like interface in `BytesIO()`, but Pybind currently doesn't "officially" offer that functionality despite being discussed in pybind/pybind11#1477.

There's a solution to it from that issue, but that would mean bundling a custom [pybind extension](https://github.com/CadQuery/OCP/blob/master/pystreambuf.h) into the build process, which we could explore. The benefit of it is that we'd be able to retain a file-like interface rather than sending a buffer around, although I'm not sure if it will work with `FILE` though and some of the other more advanced File System IO that's done through the codebase.

Laslty, with the current approach it's possible there'll be a decent amount of code duplication (similar to [FileJPG.cpp](https://github.com/isl-org/Open3D/blob/master/cpp/open3d/io/file_format/FileJPG.cpp), and [FilePNG.cpp](https://github.com/isl-org/Open3D/blob/master/cpp/open3d/io/file_format/FilePNG.cpp)). The existing code seems tightly coupled with the File IO, so I'm open to suggestions/recommendations/alternatives.

- - -
This change is [Reviewable](https://reviewable.io/reviews/isl-org/Open3D/5866)


samypr100:
Please retrigger workflows when possible, thank you 😅

cpp/pybind/io/class_io.cpp
```diff
@@ -194,22 +194,65 @@ void pybind_class_io(py::module &m_io) {
     docstring::FunctionDocInject(m_io, "read_point_cloud",
                                  map_shared_argument_docstrings);
 
+    m_io.def(
+            "read_point_cloud_from_bytes",
```
yxlao, line 198: Please add python unit tests as well. You may put it in `python/test/io`. This will show the user how to prepare `py::bytes` in python and test the pybind functions.

yxlao (changes requested):

cpp/open3d/io/PointCloudIO.cpp
```diff
@@ -51,6 +51,15 @@ static const std::unordered_map<
                 {"pts", ReadPointCloudFromPTS},
         };
 
+static const std::unordered_map<
+        std::string,
+        std::function<bool(const std::vector<char> &,
```
yxlao, line 56: I would recommend `const unsigned char*` rather than `const std::vector<char>&`. Raw buffers are typically more flexible in lower-level APIs. With raw buffers, we will need to specify the buffer size as an input.

yxlao (changes requested):

cpp/pybind/io/class_io.cpp
```diff
@@ -194,22 +194,65 @@ void pybind_class_io(py::module &m_io) {
...
+            [](const py::bytes &bytes, const std::string &format,
+               bool remove_nan_points, bool remove_infinite_points,
+               bool print_progress) {
+                const char *dataptr = PYBIND11_BYTES_AS_STRING(bytes.ptr());
+                auto length = PYBIND11_BYTES_SIZE(bytes.ptr());
+                auto data = std::vector<char>(dataptr, dataptr + length);
```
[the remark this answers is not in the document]
samypr100: I think using py::bytes requires holding the GIL. I can go with a raw buffer but we'd still have to copy to avoid issues in concurrent code

cpp/tests/io/PointCloudIO.cpp
```diff
@@ -470,7 +470,27 @@ TEST_P(ReadWritePC, UpdateProgressCallback) {
     }
 }
 
+// Test In Memory XYZ
+TEST(ReadWritePC, InMemoryXYZ) {
+    geometry::PointCloud pc_start;
+    Eigen::Vector3d one(1, 1, 1);
```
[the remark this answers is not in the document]
samypr100: Understood suggestion 1, a bit unclear about suggestion 2. Do you mean casting dv.data() to unsigned char without any conversions since I assume that might not read unless we convert the numbers to string first and then cast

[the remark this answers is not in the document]
samypr100: Added suggestion 1 in 1984a20

cpp/pybind/io/class_io.cpp
samypr100: Done in 1984a20

cpp/pybind/io/class_io.cpp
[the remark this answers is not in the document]
samypr100: I made the switch in 1984a20, but there will still be a copy :(

cpp/open3d/io/PointCloudIO.cpp
samypr100: Done in 1984a20

cpp/pybind/io/class_io.cpp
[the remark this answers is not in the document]
samypr100: From my own testing, most of the time spent was actually in `utility::FormatString` (for write) and `std::getline` and `sscanf` (for read). It seems sscanf will be slow on glibc calls strlen on every input (musl libc doesn't have this issue). I added an `utility::FastFormatString` which showed a remarkable speed improvement using `asprintf` instead of `snprintf`, but the downside is that it won't be available on Windows. I didn't attempt to get `std::getline` to be faster, but I'm sure there are things we could do.

On a 30mb xyz file:
```
Read From Bytes took: 1.59355s. <-- ~1s spent in std::getline
Read To File took: 0.595679s.
Write From Bytes took: 0.811818s.
Write To File took: 0.49677s.
```

On a 4k xyz file:
```
Read From Bytes took: 0.00023s.
Read To File took: 0.000512s.
Write From Bytes took: 0.000118s.
Write To File took: 0.000271s.
```

samypr100:
@yxlao I've made some updates, thanks

cpp/pybind/io/class_io.cpp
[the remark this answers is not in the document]
samypr100: Forgot to mention that the File I/O up there was w/ an SSD. Results might be very different on HDD's.

samypr100:
@ssheorey Please re-run workflows, I've pushed a change that hopefully fixes windows build, thanks

samypr100:
Any pointers as to why I'd be getting these errors on the windows pipelines are welcome. Short of trying to setup a development environment on windows myself, I'm not sure 100% how this is working on osx/linux but not on Windows unless I missed something obvious during building. I haven't tried to touch the code under `t` and maybe that's the issue?

```
D:\a\Open3D\Open3D\cpp\tests\t\io\PointCloudIO.cpp(423,1): error C2440: '<function-style-cast>': cannot convert from 'initializer list' to 'open3d::io::WritePointCloudOption' [C:\Open3D\build\cpp\tests\tests.vcxproj]
D:\a\Open3D\Open3D\cpp\tests\t\io\PointCloudIO.cpp(426,1): message : No constructor could take the source type, or constructor overload resolution was ambiguous [C:\Open3D\build\cpp\tests\tests.vcxproj]
D:\a\Open3D\Open3D\cpp\tests\t\io\PointCloudIO.cpp(440,1): error C2440: '<function-style-cast>': cannot convert from 'initializer list' to 'open3d::io::WritePointCloudOption' [C:\Open3D\build\cpp\tests\tests.vcxproj]
D:\a\Open3D\Open3D\cpp\tests\t\io\PointCloudIO.cpp(443,1): message : No constructor could take the source type, or constructor overload resolution was ambiguous [C:\Open3D\build\cpp\tests\tests.vcxproj]
```

samypr100:
@ssheorey workflow run please :), I tested on windows seems to build now 🙌

ssheorey:
Hi @yxlao can you take another look? Thanks.

cpp/open3d/io/file_format/FileXYZ.cpp
```diff
@@ -106,5 +139,40 @@ bool WritePointCloudToXYZ(const std::string &filename,
...
+            }
+        }
+        // nothing to report...
+        if (content.length() == 0) {
+            reporter.Finish();
+            return false;
+        }
+        length = content.length();
+        buffer = new unsigned char[length];  // we do this for the caller
+        memcpy(buffer, content.c_str(), length);
```
yxlao, line 167: Since we're using `std::free`, we can also use `std::memcpy` here for consistency.

cpp/pybind/io/class_io.cpp
```diff
@@ -194,22 +194,73 @@ void pybind_class_io(py::module &m_io) {
...
 
+    m_io.def(
+            "read_point_cloud_from_bytes",
+            [](const py::bytes &bytes, const std::string &format,
+               bool remove_nan_points, bool remove_infinite_points,
+               bool print_progress) {
+                const char *dataptr = PYBIND11_BYTES_AS_STRING(bytes.ptr());
+                auto length = PYBIND11_BYTES_SIZE(bytes.ptr());
+                auto buffer = new unsigned char[length];
+                memcpy(buffer, dataptr, length);  // copy before releasing GIL
```
yxlao, line 205: same, `std::memcpy`

cpp/open3d/io/file_format/FileXYZ.cpp
```diff
@@ -106,5 +139,40 @@ bool WritePointCloudToXYZ(const std::string &filename,
...
+    try {
+        utility::CountingProgressReporter reporter(params.update_progress);
+        reporter.SetTotal(pointcloud.points_.size());
+
+        std::string content;
+        for (size_t i = 0; i < pointcloud.points_.size(); i++) {
+            const Eigen::Vector3d &point = pointcloud.points_[i];
+            std::string line = utility::FastFormatString(
+                    "%.10f %.10f %.10f\n", point(0), point(1), point(2));
+            content.append(line);
```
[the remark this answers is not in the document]
samypr100: Thanks, I had tried `istringstream` but it ended up being slower on my local testing. I can go back and try get the numbers to compare

cpp/pybind/io/class_io.cpp
samypr100: @yxlao We have no choice but to copy before releasing the GIL, we can't release the gil and also operate on a `Py_buffer` as far as I know.

samypr100: Thanks, for a minute I thought you were talking about something else 😆

cpp/open3d/io/file_format/FileXYZ.cpp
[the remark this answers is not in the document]
samypr100: Fixed in e64217f254a3446818d5b51a7a8f549033b80e2f

cpp/open3d/io/file_format/FileXYZ.cpp
samypr100: Fixed in e64217f254a3446818d5b51a7a8f549033b80e2f

cpp/pybind/io/class_io.cpp
samypr100: Fixed in e64217f254a3446818d5b51a7a8f549033b80e2f

cpp/open3d/io/file_format/FileXYZ.cpp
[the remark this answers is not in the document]
samypr100: @yxlao 

Using `ostringstream` with formatting:

- Via `oss << std::fixed << std::setprecision(10) << point(0) << " " << point(1) << " " << point(2)`

Read From Bytes took: 1.58005s.
Read To File took: 0.602676s.
Write From Bytes took: 1.06527s.
Write To File took: 0.614322s.

Using `FastFormatString` with an `ostringstream`:

Read From Bytes took: 1.53005s.
Read To File took: 0.53583s.
Write From Bytes took: 0.933503s.
Write To File took: 0.538835s.

Using `FastFormatString`:

Read From Bytes took: 1.59355s.
Read To File took: 0.595679s.
Write From Bytes took: 0.811818s.
Write To File took: 0.49677s.

Using `FormatString`:

Read From Bytes took: 1.54758s.
Read To File took: 0.595833s.
Write From Bytes took: 1.31834s.
Write To File took: 0.531619s.

### Result

| Approach                            | Time      |
|-------------------------------------|-----------|
| FastFormatString with string        | 0.811818s |
| FastFormatString with ostringstream | 0.933503s |
| ostringstream                       | 1.06527s  |
| FormatString                        | 1.31834s  |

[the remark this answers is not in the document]
samypr100: Additional note: On windows testing, FastFormatString gave me about 1.3s as well as it falls back to FormatString on windows, which is expected. Windows might benefit more off of a pure ostringstream implementation, although I'm not sure yet if we can use a pure ostringstream implementation on the other file formats.

ssheorey:
Hi @yxlao can you take another look at this PR? Thanks.

Chen-Suyi:
Thanks for your contributions!

Is this feature accessible now? It would be really helpful for me, but I cannot find any API descriptions in the documents.

charangodwin:
I could help review this PR.

charangodwin (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/open3d/5866)* status: 0 of 11 files reviewed, 9 unresolved discussions (waiting on @samypr100, @ssheorey, and @yxlao)

___
*[`cpp/open3d/io/PointCloudIO.h` line 95 at r4](https://reviewable.io/reviews/isl-org/open3d/5866#-NjvYde983bT6T5pTv9w:-NjvYde983bT6T5pTv9x:b5slmq8) ([raw file](https://github.com/isl-org/open3d/blob/ff5868f79239d6ad19a0fbf916baf763a5caeb2f/cpp/open3d/io/PointCloudIO.h#L95)):*
> ```objc
>             // Attention: when you update the defaults, update the docstrings in
>             // pybind/io/class_io.cpp
>             std::string format = "auto",
> ```

[ES.23: Prefer the `{}`\-initializer syntax](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Res-list)
___
*[`cpp/open3d/io/file_format/FileXYZ.cpp` line 67 at r5](https://reviewable.io/reviews/isl-org/open3d/5866#-NjvtASz28vvjcPKZ8uo:-NjvtASz28vvjcPKZ8up:b-4ckr82) ([raw file](https://github.com/isl-org/open3d/blob/04c74079b46bbbae2cec364379b5c15e893e56c9/cpp/open3d/io/file_format/FileXYZ.cpp#L67)):*
> ```cpp
>         reporter.SetTotal(static_cast<int64_t>(length));
> 
>         std::string content(reinterpret_cast<const char *>(buffer), length);
> ```

This create another copy of the buffer as std::string deep copy the input buffer. Also in the class_io.cpp we use std::memcpy. 
Is there a way to process the buffer directly using char*? Maybe using std::istrstream.

samypr100:
@charangodwin re istratream, see https://github.com/isl-org/Open3D/pull/5866#discussion_r1110037293 about some history on how changing may affect benchmarks significantly.

python/test/io/rpc/test_io.py
```diff
@@ -0,0 +1,48 @@
...
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+# FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+# IN THE SOFTWARE.
+# ----------------------------------------------------------------------------
```
nsaiapova, line 25: Should not the header be in this format?
```suggestion
 # ---------------------------------------------------------------------------- 
 # -                        Open3D: www.open3d.org                            - 
 # ---------------------------------------------------------------------------- 
 # Copyright (c) 2023 www.open3d.org 
 # SPDX-License-Identifier: MIT 
 # ---------------------------------------------------------------------------- 
```
Also for a new file I think the year should be just 2023.

nsaiapova (commented):
The PR looks good to me with just a couple of minor suggestions.

ssheorey (approved):

cpp/open3d/io/file_format/FileXYZ.cpp
```diff
@@ -6,10 +6,12 @@
 // ----------------------------------------------------------------------------
 
 #include <cstdio>
+#include <iostream>
```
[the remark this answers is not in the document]
ssheorey: Done

python/test/io/rpc/test_io.py
ssheorey: Done

ssheorey:
Thanks @samypr100 for this very useful contribution and sorry for the delay in getting this merged - we'll try to merge faster in the future....

Minor comment about other formats (xyzn, xyzi, xyzrgb) - these are better supported in the more generic Tensor based datatypes (open3d::t::io namespace). Functions for reading from a file are already available there.

Issue #5904, opened 2023-02-13
Release `open3d` wheel with shared library (libOpen3D.so).


yuecideng:
### Checklist

- [X] I have searched for [similar issues](https://github.com/isl-org/Open3D/issues).
- [X] For Python issues, I have tested with the [latest development wheel](http://www.open3d.org/docs/latest/getting_started.html#development-version-pip).
- [X] I have checked the [release documentation](http://www.open3d.org/docs/release/) and the [latest documentation](http://www.open3d.org/docs/latest/) (for `master` branch).

### Proposed new feature or change

Currently, the official `open3d` package is released using `-DBUILD_SHARED_LIBS=OFF`, which makes C++ users unable to use the library directly. Although the pre-built C++ shared library is provided, it is inconvenient for users who want to use both C++ and Python API of Open3D at the same time (eg, via just simple `pip3 install open3d`).  

The proprosed feature will allow Open3D to be used for both Python and C++, and makes it possible for external project which uses Open3D as dependency (use shared 3D data structure, tensor, etc. as function interfaces.) with both Python and C++ API.

Expected outcome:
- The shared library, header files and cmake files are stored in the open3d python package folder (eg, `open3d/lib`, `open3d/include`, `open3d/share`)
-  Add `cmake_prefix_path` to open3d such that the cmake prefix path can be automatically detected by python.

PR #5931, opened 2023-02-18
Enable shared library linking for open3d*.wheel

yuecideng:
## Type

-   [x] Breaking change (fix or feature that would cause existing functionality to not work as expected) Resolves #

## Motivation and Context

This PR aims at solving the issue mentioned at #5904.

steps:
```bash
cd Open3D
mkdir build && cd build
cmake .. -DBUILD_SHARED_LIBS=ON
make 
make install-pip-package
```
The c++ lib and include will be installed inside the open3d package , which is in the current python environment. 
The `open3d/cpu/pybind*.so` (eg. ubuntu, cpu version) will is linked with the `open3d/lib/libOpen3D.so` 

I have tested the change on macos and ubuntu 20.04. Windows is not yet tested. I will follow up if the CI fail.

TODO:
- [ ] Add changes to CI, such that the wheels are built with `-DBUILD_SHARD_LIBS=ON`.
- [ ] Add variable to `open3d` to indicate the path of cmake config.

## Checklist:

    to my code.
        updated accordingly.
-   [x ] I will follow up and update the code if CI fails.
    
## Description

- - -
This change is [Reviewable](https://reviewable.io/reviews/isl-org/Open3D/5931)


yuecideng:
Previously, Open3D builds wheel with both cpu and cuda, so there are two folder `cpu` and `cuda` inside the wheel produced in CI. For static library, it is ok for user to import (with or without cuda in the machine). Now we want to use shared library, what is the best way to organize the folder structure such that we can achieve the same behaviour? @yxlao @ssheorey

ssheorey:
Closing in favor of #7355

PR #5940, opened 2023-02-21
Basic support for Python Optional Dependencies

### LINKED ISSUE #2991: pip: Why does Open3D pull in Open3D-ML/requirements.txt?

### LINKED ISSUE #5740: Make dash an optional dependency

samypr100:
## Type

-   [x] Breaking change (fix or feature that would cause existing functionality to not work as expected) Resolves #5740

## Motivation and Context

This PR is a more conservative approach to what's described in #5853. Using only `[standard]` came up as an option in the discussion so I've decided to adopt it in this alternate PR.

I've only included the `gui` dependencies as part of standard, but we can include the ML dependencies (from open3d-ml) as well as discussed in https://github.com/isl-org/Open3D/pull/5853#issuecomment-1423560428 if we desire to.

Note, Including the `ml` dependencies would help close #2991 in addition to #5740.

## Checklist:

    to my code.
        updated accordingly.
    
## Description

As discussed in #5853, below is a table of the changes:

| Optional Dependency | Contents |
| --- | --- |
| **standard** | Includes new (this PR) `requirements_gui.txt` file from https://github.com/isl-org/Open3D |

Strictly speaking from a python dependencies size, below are the sizes with the new optional dependencies calculated on OSX (transitive sizes may vary between platforms). 

| Package        | Transitive Dependencies Size (MB) |
|----------------|------------------------------|
| open3d         | 350.13                      |
| open3d[standard]    | 459.13                         |

If we decide to include `ml` as part of standard, the above table becomes

| Package        | Transitive Dependencies Size (MB) |
|----------------|------------------------------|
| open3d         | 107.69                      |
| open3d[standard] | 459.13                   |

### Total Savings

For table <span>#</span>1, the total savings would be about 109mb.

For table <span>#</span>2, the total savings would be about 351mb.

- - -
This change is [Reviewable](https://reviewable.io/reviews/isl-org/Open3D/5940)


theNded:
I am not an expert on CI. I had a conversation with the CI expert on our team, and we think your solution would be viable as long as we keep 
- workflows stable, 
- artifacts correctly uploaded
- docs consistent

I think we can start with this minimal PR first and check those items one by one until reach somewhere that looks promising. It's my pleasure to work with you!

samypr100:
@theNded 👍 Can you re-run the CI on both PR's again with the latest commit? Thanks

theNded:
> 

Nah, it is just a general comment for future reference. Docs look great now, although I haven't gone through them thoroughly.

samypr100:
@theNded Thanks! Looks like CI is passing on both PR's 🙂

theNded:
@samypr100 thanks for the update! 
Here are two of my thoughts before a thorough review:
1. Would it be possible that we keep the original installation behavior, i.e. `pip install open3d` installs the complete package, and `pip install open3d[nogui]` installs the partial package? That will probably make the transition smoother.
2. Would you be able to upgrade the workflows as well, such that both wheels are produced in the CI? For instance, right now, only [one artifact](https://github.com/isl-org/Open3D/actions/runs/5878710618?pr=5940) is produced. We will need to [upload to github](https://github.com/isl-org/Open3D/blob/master/.github/workflows/macos.yml#L205), [upload to gcloud](https://github.com/isl-org/Open3D/blob/master/.github/workflows/macos.yml#L226), and distribute them with pypi correspondingly.

Please correct me if I'm wrong in those comments; otherwise, I will be glad to help with the workflows.

johnthagen:
@theNded Like @samypr100 said, the reasons for why extras can only be additive are explained in the link @samypr100 posted above.

But at the same time, many popular Python packages have gone through this transition due to how useful it is for end users to be able to better control/trim transitive dependencies. This can speed up development workflows, CI workflows, container image sizes, etc for all downstream users of Open3D who only need a subset of Open3D's functionality.

theNded:
@samypr100 @johnthagen Thanks for the update, I have read the explanation and it looks very reasonable. I have a heavy workload right now at my full-time job, but hopefully I will be back to review this PR at the end of next week.

python/open3d/__init__.py
```diff
@@ -109,7 +109,14 @@ def _insert_pybind_names(skip_names=()):
     sys.modules.update(submodules)
 
 
-import open3d.visualization
+try:
```
theNded, line 112: We should implement systematic unit tests starting from this point. 
For example, in the pytest files, expect this exception conditioned on the extra package (open3d or open3d[standard]). This can help us make sure that the core behavior is preserved without the gui requirements. Might need to tweak some of the CIs, but a summary of some local tests would be also helpful.

3rdparty/README_SYCL.md
```diff
@@ -98,7 +98,7 @@ Open3D is designed to make use of the SYCL GPU devices.
 
 ## List of oneAPI Python packages
 
-To make `pip install open3d` works out-of-the box on SYCL-enabled platforms,
+To make `pip install open3d[standard]` works out-of-the box on SYCL-enabled platforms,
```
theNded, line 101: Ideally, also add doc for the behavior of `pip install open3d`. Same below.

.github/workflows/windows.yml
```diff
@@ -399,7 +400,7 @@ jobs:
             $PIP_PKG_NAME=(Get-ChildItem open3d*-$py_tag-*.whl).Name
           }
           echo "Installing Open3D wheel $PIP_PKG_NAME in virtual environment..."
-          python -m pip install "$PIP_PKG_NAME"
+          python -m pip install "$PIP_PKG_NAME[standard]"
```
theNded, line 403: I feel like `[all], [full], [complete]`, or even `[gui]` might be more informative. Standard gives me the impression that it is a standard minimal set. But this is possibly subjective. @ssheorey what's your opinion?

README.md
```diff
@@ -49,8 +49,8 @@ Pre-built pip packages support Ubuntu 18.04+, macOS 10.15+ and Windows 10+
 
 ```bash
 # Install
-pip install open3d       # or
-pip install open3d-cpu   # Smaller CPU only wheel on x86_64 Linux (v0.17+)
+pip install open3d[standard]       # or
+pip install open3d-cpu[standard]   # Smaller CPU only wheel on x86_64 Linux (v0.17+)
```
theNded, line 53: Same, please add the behavior of `pip install open3d` and `pip install open3d-cpu`. Ideally, elaborate on the design choice and the expected usage of both options.

cpp/pybind/make_install_pip_package.cmake
```diff
@@ -5,4 +5,4 @@
 #       it is guaranteed that there is only one wheel in ${PYTHON_PACKAGE_DST_DIR}/pip_package/*.whl
 file(GLOB WHEEL_FILE "${PYTHON_PACKAGE_DST_DIR}/pip_package/*.whl")
 execute_process(COMMAND ${Python3_EXECUTABLE} -m pip uninstall open3d open3d-cpu --yes)
-execute_process(COMMAND ${Python3_EXECUTABLE} -m pip install ${WHEEL_FILE} -U)
+execute_process(COMMAND ${Python3_EXECUTABLE} -m pip install "${WHEEL_FILE}[standard]" -U)
```
theNded, line 8: Can we expose this extra option to cmake/make as well?

cpp/pybind/make_python_package.cmake
```diff
@@ -113,6 +113,7 @@ if (BUILD_JUPYTER_EXTENSION)
     # These will be installed when `pip install open3d`.
     execute_process(COMMAND ${CMAKE_COMMAND} -E cat
         ${PYTHON_PACKAGE_SRC_DIR}/requirements.txt
+        ${PYTHON_PACKAGE_SRC_DIR}/requirements_gui.txt
```
theNded, line 116: Same, can we expose this extra option to cmake/make as well?

docs/arm.rst
```diff
@@ -9,7 +9,7 @@ we provide pre-compiled ARM64 wheels for Linux and macOS. Install the wheel by:
 
 .. code-block:: bash
 
-    pip install open3d
+    pip install open3d[standard]
```
theNded, line 12: Same, explain how `pip install open3d` would work.

docs/arm.rst
```diff
@@ -107,7 +107,7 @@ TBB, Parallel STL, BLAS, LAPACK) may cause compatibility issues if they are not
 the same version as the one used by Open3D.
 
 If you only need the Python wheel, consider using the Docker build method or
-install Open3D via ``pip install open3d`` directly.
+install Open3D via ``pip install open3d[standard]`` directly.
```
theNded, line 110: Same, try to provide something like "if you only need the core functionalities and do not need visualization, you may install the minimal wheel via `pip install open3d`".

docs/getting_started.in.rst
```diff
@@ -35,9 +35,12 @@ Pip (PyPI)
 
 .. code-block:: bash
 
-    pip install open3d        # or
-    pip install open3d-cpu    # Smaller CPU only wheel on x86_64 Linux (since v0.17+)
+    pip install open3d[standard]        # or
```
theNded, line 38: Same, could use some explanation.

util/ci_utils.sh
```diff
@@ -83,6 +83,7 @@ install_python_dependencies() {
 
     # TODO: modify other locations to use requirements.txt
     python -m pip install -r "${OPEN3D_SOURCE_ROOT}/python/requirements.txt"
+    python -m pip install -r "${OPEN3D_SOURCE_ROOT}/python/requirements_gui.txt"
```
theNded, line 86: Make this requirement optional (controlled by the script input flags).

python/setup.py
```diff
@@ -93,6 +94,15 @@ def finalize_options(self):
 with open("requirements.txt", "r") as f:
     install_requires = [line.strip() for line in f.readlines() if line]
 
+# For extra requirements
+extra_requires = defaultdict(list)
+
+# Standard Extras
```
theNded, line 100: Could also put a link to the relevant PEP docs for reference.

python/open3d/__init__.py
```diff
@@ -109,7 +109,14 @@ def _insert_pybind_names(skip_names=()):
...
+    import open3d.visualization
+except ModuleNotFoundError:
+    warnings.warn(
+        "Open3D Python GUI Libraries not found. "
```
theNded, line 116: NIT: GUI libraries

.github/workflows/windows.yml
```diff
@@ -276,6 +276,7 @@ jobs:
         run: |
           $ErrorActionPreference = 'Stop'
           python -m pip install -r python/requirements.txt
+          python -m pip install -r python/requirements_gui.txt
```
theNded, line 279: In practice the difference between `open3d` and `open3d[standard]` is gui packages, so naming feels fine.
But semantically this feels a bit inconsistent: open3d + gui = open3d[standard].

python/open3d/__init__.py
theNded: This may not be quite critical for this PR, but if we plan on more advanced dependency management I think it would be required.

theNded (changes requested):

.github/workflows/windows.yml
johnthagen: As an FYI, `standard` is used in another popular Python package:

- https://www.uvicorn.org/#quickstart

Others use `all`:

- https://github.com/tiangolo/fastapi/blob/a3f1689d78abef3d5d54ba655ec87cef81bf7384/pyproject.toml#L56

PR #5978, opened 2023-03-08
Enable picking points from TriangleMeshModels in O3D Visualizer

errissa:
## Description

Add ability to select points from a `TriangleMeshModel` in O3D Visualizer (e.g., draw.py). Previously, attempting to do so caused a crash with a cryptic message.

The following script demonstrates the new capability:
```py
import open3d as o3d
monkey = o3d.data.MonkeyModel()
m = o3d.io.read_triangle_model(monkey.path)
o3d.visualization.draw(m, show_ui=True, actions=[["Print selection", lambda o3dvis: print(o3dvis.get_selection_sets())]])
```
Output:
```py
[Open3D WARNING] Geometry Object 1 has already been added to scene graph.
[{'Object 1': {{ index: 47335, order: 5, point: (5.42036, -0.955666, -2.12278) }, { index: 49813, order: 3, point: (-1.15616, -0.0953989, -1.86053) }, { index: 53731, order: 4, point: (-0.625896, -0.964231, -6.8371) }, { index: 40037, order: 1, point: (-4.25987, -0.402343, -1.95247) }, { index: 57889, order: 2, point: (-4.27602, -0.404048, -1.47792) }, { index: 40561, order: 0, point: (-1.6692, -0.834914, -2.62355) }}}]
```
Select the "Selection" tab at the top right. Previously, selection the tab would cause a crash. Now, you can Ctrl-click to select points from the Monkey model.

- - -
This change is [Reviewable](https://reviewable.io/reviews/isl-org/Open3D/5978)

### DIFF
--- a/cpp/open3d/visualization/visualizer/O3DVisualizer.cpp
+++ b/cpp/open3d/visualization/visualizer/O3DVisualizer.cpp
@@ -2029,7 +2029,7 @@ struct O3DVisualizer::Impl {
         // Count number of meshes stored in TriangleMeshModels
         size_t model_mesh_count = 0;
         size_t model_count = 0;
-        for (auto &o: objects_) {
+        for (auto &o : objects_) {
             if (!IsGeometryVisible(o)) {
                 continue;
             }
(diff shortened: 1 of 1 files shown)


ssheorey (approved):
Looks good. Tested with a different mesh and added output to PR description.

Question: Is the point index in the selection set a single index over all meshes? So to recover the index for a specific mesh, we would need the mesh number, right?

ssheorey:
Warning is gone. LGTM.

PR #5991, opened 2023-03-14
Add new Mitsuba Python example + update CPU rendering docs

errissa:
This change is [Reviewable](https://reviewable.io/reviews/isl-org/Open3D/5991)


ssheorey (approved):
LGTM after adding test data

PR #6012, opened 2023-03-20
Version bump to 0.17.1

errissa:
This change is [Reviewable](https://reviewable.io/reviews/isl-org/Open3D/6012)


docs/versions.js
```diff
@@ -18,8 +18,8 @@ document.write('\
             <td><a href="http://www.open3d.org/docs/latest/cpp_api">master C++</a></td>\
         </tr>\
         <tr>\
-            <td><a href="http://www.open3d.org/docs/release">0.17.0 (release)</a></td>\
-            <td><a href="http://www.open3d.org/docs/release/cpp_api">0.17.0 C++ (release)</a></td>\
+            <td><a href="http://www.open3d.org/docs/release">0.17.1 (release)</a></td>\
+            <td><a href="http://www.open3d.org/docs/release/cpp_api">0.17.1 C++ (release)</a></td>\
```
ssheorey, line 22: We can skip recreating the documentation for v0.17.1 since it will be identical to v0.17.0, and v0.17.1 is only for macOS on Apple Si. We don't have separate documentation for v0.16.1 either.

Let's not change this file, so that `release` continues to point at `v0.17.0` for docs.

ssheorey (changes requested):

docs/versions.js
errissa: Done. Reverted to previous version.

errissa:
Didn't need to create a 0.17.1 release after all.

PR #6079, opened 2023-04-14
Fix and improve ISS keypoint

### LINKED ISSUE #4847: Bug with compute_iss_keypoints (failing randomly)
### Checklist

### Describe the issue

I'm trying to extract the keypoints from a point cloud with the method open3d.geometry.keypoint.compute_iss_keypoints(). I'm looping thru a lot of files in directory and sub-directories.  It looks like : 

`` 

    import open3d as o3d
    from pathlib import Path

    directory_pcd = Path("directory_to_pcd_files")
    for file in directory_pcd.iterdir():
        pcd = o3d.io.read_point_cloud(str(file))
        print(f"Compute keypoints of {file}")
        keypoint = o3d.geometry.keypoint.compute_iss_keypoints(
            pcd, gamma_21=10, gamma_32=10, min_neighbors=1
        )
        print(f"Finished computing keypoints of {file}")
``

The problem is the compute_iss_keypoints fails randomly. Like I can see with my print that's exactly at this line that it fails, but it will never fails at the same file, never files after processing the same number of files. Yes, the method .iterdir() process the files in the same order each time.

I've tried this with Python 3.8.7, 3.9.10. open3d version 0.14 and version 0.15. It still fails. 

```python
1. Take a directory with some .pcd files in it and replace the first line of code with the valid directory
2. Run and look at the print statements. It will never fails at the same file (while still looping them in the same order each time)
```

### Error message

Process finished with exit code -1073741819 (0xC0000005)

Error in the console from PyCharm

### Expected behavior

The method should work for each file and not fail randomly

```markdown
- Python version: Python 3.9.10
- Open3D version: output from python: 0.15.1
- System architecture: x64
- How did you install Open3D?: From Poetry
- Compiler version (if built from source): not applicable
```

### Additional information

_No response_

DNKpp:
## Type

-   [x] Bug fix (non-breaking change which fixes an issue): Fixes #4847

## Motivation and Context

During my usage of the ISS Keypoint functionality I encountered some random crashes, which I tried to find the reason for. I explained it here: https://github.com/isl-org/Open3D/issues/4847#issuecomment-1508390527

Secondly there seems to be a more subtile issue in ``ComputeModelResolution``, as the original code only checks for "more than 0" neighbors, but as the query point itself will always be present, this condition will never fail. I adjusted the condition here to ``>= 2``.

Eventually I applied some std algorithms instead of manually written loops, which makes the intent of the code more obvious.

## Checklist:

    to my code.
   *Yes, but it crashes due to some strange errors. But resharper is applying the clang-format file anyways.*
        updated accordingly.
    
## Description

This is the error popup
image

Those are some of the entries of kp_indices
image

- - -
This change is [Reviewable](https://reviewable.io/reviews/isl-org/Open3D/6079)

### DIFF
--- a/cpp/open3d/geometry/ISSKeypoints.cpp
+++ b/cpp/open3d/geometry/ISSKeypoints.cpp
@@ -118,6 +118,7 @@ std::shared_ptr<PointCloud> ComputeISSKeypoints(
 
             if (nb_neighbors >= min_neighbors &&
                 IsLocalMaxima(i, nn_indices, third_eigen_values)) {
+#pragma omp critical
                 kp_indices.emplace_back(i);
             }
         }
--- a/cpp/open3d/geometry/ISSKeypoints.cpp
+++ b/cpp/open3d/geometry/ISSKeypoints.cpp
@@ -32,12 +32,9 @@ bool IsLocalMaxima(int query_idx,
                    const std::vector<int>& indices,
                    const std::vector<double>& third_eigen_values) {
     return std::none_of(
-        indices.begin(),
-        indices.end(),
-        [&third_eigen_values, value = third_eigen_values[query_idx]](const int idx) {
-            return value < third_eigen_values[idx];
-        }
-    );
+            indices.begin(), indices.end(),
+            [&third_eigen_values, value = third_eigen_values[query_idx]](
+                    const int idx) { return value < third_eigen_values[idx]; });
 }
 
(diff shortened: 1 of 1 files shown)


ssheorey:
@benjaminum Is this PR ready to merge? We can ignore the Ubuntu wheel CI failure for now.

benjaminum:
yes, please merge

PR #5938, opened 2023-02-20
Fix for Issue #5924:

dbs4261:
# Fix for Issue #5924:

Changed TriangleMesh::materials_ to be a std::vector<std::pair<std::string, Material>> so that the order of materials when loading a mesh is respected. Using the unordered map caused issues when a Default Material and a texture were loaded and the texture would end up as the second material in the iterator. Now TriangleMesh::triangle_material_ids_ will indicate what material index should be used. A warning is given when passing a TriangleMesh to the visualizer and more than one material index is found, and the minimum value found in TriangleMesh::triangle_material_ids_ is used. If no triangle_material_ids_ exist then the first material in the order they are loaded will be used.

## Type

-   [ ] Bug fix (non-breaking change which fixes an issue): Fixes #
-   [ ] New feature (non-breaking change which adds functionality). Resolves #
-   [x] Breaking change (fix or feature that would cause existing functionality to not work as expected) Resolves #5924

## Motivation and Context

When loading a textured mesh with Assimp or TinyOBJ the materials are loaded in order, and a material index is provided for each triangle. However these materials are stored in an unordered map. A common occurrence is loaded a textured mesh along with a default material. The texture might be loaded as index 1 which can be seen in the TriangleMesh::triangle_material_ids_. By instead loading to a vector, the order is maintained.

This change breaks any access to the TriangleMesh::materials_ parameter. Though no instances of this access occur within the Open3d C++ library, it is exposed publicly in the Python API. Instead of a dictionary, the user would get a list of pairs.

## Checklist:

-   [x] I have run `python util/check_style.py --apply` to apply Open3D **code style**
    to my code.
-   [ ] This PR changes Open3D behavior or adds new functionality.
    -   [ ] Both C++ (Doxygen) and Python (Sphinx / Google style) **documentation** is
        updated accordingly.
    -   [ ] I have added or updated C++ and / or Python **unit tests** OR included **test
        results** (e.g. screenshots or numbers) here.
-   [x] I will follow up and update the code if CI fails.
    
-   [x] For fork PRs, I have selected **Allow edits from maintainers**.

## Description

- - -
This change is [Reviewable](https://reviewable.io/reviews/isl-org/Open3D/5938)

PR #6101, opened 2023-04-21
Update the GLTF writer to support writing textured models

dbs4261:
The GLTF exporter does not currently support textured meshes. This is because Open3d's per triangle UV coordinates makes the task non-trivial. This change set adds that feature.
Note: this depends on #5938 which changes how textures are stored in an `open3d::geometry::TriangleMesh`

## Type

## Motivation and Context

The current GLTF exporter does not support textured meshes. This PR adds an exporter for TriangleMeshModels. Currently there is only a reader. It does not add support to the existing writer.

## Checklist:

    to my code.
        updated accordingly.
    
## Description

Code was added to create TriangleMeshModels from a TriangleMesh. This isn't a simple conversion because TriangleMesh's support multiple materials, assigned on a per triangle basis. However TriangleMeshModels are supposed to only have one material per mesh. There is a similar issue for the export of the meshes. TriangleMesh's store texture coordinates on a per triangle basis. This is to support seams where the same vertex has different texture coordinates. However GLTF requires texture coordinates to be stored on a per vertex basis. So this PR adds a function to convert a mesh to have per-vertex texture coordinates. Currently this function is in a detail namespace but could be moved to a public API.

- - -
This change is [Reviewable](https://reviewable.io/reviews/isl-org/Open3D/6101)

### DIFF
--- a/3rdparty/find_dependencies.cmake
+++ b/3rdparty/find_dependencies.cmake
@@ -1889,6 +1889,14 @@ else()
     set(BUILD_WEBRTC_COMMENT "//")
 endif()
 
+# tcb::span
+include(${Open3D_3RDPARTY_DIR}/tcbspan/tcbspan.cmake)
+open3d_import_3rdparty_library(3rdparty_tcbspan
+    INCLUDE_DIRS ${TCB_SPAN_INCLUDE_DIRS}
+    LIBRARIES    ${TCB_SPAN_LIBRARIES}
+)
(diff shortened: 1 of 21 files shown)


TheExDeus:
Can this be deconflicted and merged? The 2020 version of tinygltf causes issues if you have your own. And you cannot disable GLTF support inside open3d, so that is not an option either.

ssheorey:
Closing since texture map writing support was added to gltf, fbx, obj in #7508

dbs4261:
Yes

Issue #4747, opened 2022-02-11
Link Time Optimization (-flto) triggers static initialization bug in Open3D


roehling:
### Checklist

- [X] I have searched for [similar issues](https://github.com/isl-org/Open3D/issues).
- [X] For Python issues, I have tested with the [latest development wheel](http://www.open3d.org/docs/latest/getting_started.html#development-version-pip).
- [X] I have checked the [release documentation](http://www.open3d.org/docs/release/) and the [latest documentation](http://www.open3d.org/docs/latest/) (for `master` branch).

#### I first cloned Open3D by:
```
git clone https://github.com/isl-org/Open3D.git
cd Open3D
```

#### Then, I build Open3D (on Debian Sid) with:
```
mkdir build
cd build
cmake -DCMAKE_INTERPROCEDURAL_OPTIMIZATION=ON ..
make -j$(nproc)
```

### Error message

```shell
Build succeeds without error, but the Open3D GUI will fail to load materials at runtime.
```

### Open3D, Python and System information

```markdown
- Operating system: Debian unstable ("sid")
- Python version: Python 3.9
- Open3D version: 0.14.1
- System architecture: x86
- Is this a remote workstation?: no
- How did you install Open3D?: build from source
- Compiler version (if built from source): gcc 11
```

### Additional information

I tracked the bug to a corrupted `defaults_mapping::shader_mappings` map. The map has static storage and depends on members of `FilamentResourceManager`, again with stattic storage. This creates a initialization order dependency across translation units, also known as the [static initialization order fiasco](https://en.cppreference.com/w/cpp/language/siof), and will break Open3D if the static variables from `FilamentResourceManager.cpp` are not initialized before those in `FilamentScene.cpp`.

As initialization normally happens in the order the object files are linked, and (thanks to the alphabetic sorting) `FilamentResourceManager.cpp` comes before `FilamentScene.cpp`, this bug has not surfaced before. However, LTO seems to mess with the initialization order, thereby exposing the bug.

The static initialization order problem is usually avoided using the Construct On First Use idiom. For now, the bug can be mitigated by disabling/not enabling Link Time Optimization for the Open3D main library. The Python module is not affected and *should* be built with LTO for performance reasons, actually.

PR #6181, opened 2023-05-27
Enable multi-core link time optimization

roehling:
## Type

## Motivation and Context

GCC runs LTO on a single core by default, making the final build step for the Python module needlessly long.

## Checklist:

    to my code.
        updated accordingly.
    
## Description

The PR sets the optional `auto` argument for the `-flto` option, which will distribute the workload among the available jobs from GNU make (or, as fallback if GNU make is not used, the number of detected CPU cores).

- - -
This change is [Reviewable](https://reviewable.io/reviews/isl-org/Open3D/6181)

### DIFF
--- a/cpp/pybind/CMakeLists.txt
+++ b/cpp/pybind/CMakeLists.txt
@@ -66,6 +66,7 @@ elseif (UNIX)   # Linux
 };]=])
     target_link_options(pybind PRIVATE $<$<CONFIG:Release>:
         "-Wl,--version-script=${CMAKE_CURRENT_BINARY_DIR}/pybind.map" >)
+    target_link_options(pybind PRIVATE "-flto=auto")
 endif()
 
 


cpp/pybind/CMakeLists.txt
```diff
@@ -66,6 +66,7 @@ elseif (UNIX)   # Linux
 };]=])
     target_link_options(pybind PRIVATE $<$<CONFIG:Release>:
         "-Wl,--version-script=${CMAKE_CURRENT_BINARY_DIR}/pybind.map" >)
+    target_link_options(pybind PRIVATE "-flto=auto")
```
ssheorey, line 69: Hi @roehling Is this something we can have for the Open3D DSO as well? (in `cpp/open3d/CMakeLists.txt`)

roehling: Yes, but only after #4747 has been resolved.

Issue #5647, opened 2022-11-04
Random seed works at 0.15.1 segment_plane  but not works at 0.16.0


mcmingchang:
### Checklist

- [X] I have searched for [similar issues](https://github.com/isl-org/Open3D/issues).
- [X] For Python issues, I have tested with the [latest development wheel](http://www.open3d.org/docs/latest/getting_started.html#development-version-pip).
- [X] I have checked the [release documentation](http://www.open3d.org/docs/release/) and the [latest documentation](http://www.open3d.org/docs/latest/) (for `master` branch).

### Describe the issue

Random seed works at 0.15.1 segment_plane  but not works at 0.16.0

### Steps to reproduce the bug

```python
#### in 0.15.1
import numpy as np
import open3d as o3d
pcd = o3d.io.read_point_cloud('1.ply')
pcd_down = pcd.voxel_down_sample(voxel_size=0.01)
plane_model, inliers = pcd_down.segment_plane(distance_threshold=0.02,
                                              ransac_n=5,
                                              num_iterations=5000, seed=1)
plane_model, inliers2 = pcd_down.segment_plane(distance_threshold=0.02,
                                              ransac_n=5,
                                              num_iterations=5000, seed=1)
print(np.all(np.array(inliers) == np.array(inliers2)))


##### in 0.16.0
import numpy as np
import open3d as o3d
o3d.utility.random.seed(1)
pcd = o3d.io.read_point_cloud('1.ply')
pcd_down = pcd.voxel_down_sample(voxel_size=0.01)
plane_model, inliers = pcd_down.segment_plane(distance_threshold=0.02,
                                              ransac_n=5,
                                              num_iterations=5000)
plane_model, inliers2 = pcd_down.segment_plane(distance_threshold=0.02,
                                              ransac_n=5,
                                              num_iterations=5000)
print(np.all(np.array(inliers) == np.array(inliers2)))
```

### Error message

None

### Expected behavior

The result of 0.16 should be the same as that of 0.15

### Open3D, Python and System information

```markdown
None
```

### Additional information

Random seed works at 0.15.1 segment_plane  but not works at 0.16.0

yuecideng:
Since 0.16.0, the `segment_plane` function is parallel and the iteration is changed with the estimated inliers ratio, which may lead to inconsistent results under the same random seed.

mcmingchang:
You mean it's not a bug？But setting seed is to achieve a stable result.

yuecideng:
[...]

PR #6308, opened 2023-08-13
Fix segment plane non deterministic issue 

yuecideng:
## Type

## Motivation and Context

This PR fix issue #5647 and add a unit test for deterministic check of `SegmentPlane`. 

The reason why the function is non deterministic is that, the `RandomSampler` is not guarded by the mutex, which may result in the order of the sampled indices mismatch among multithreading workload.

## Checklist:

    to my code.
        updated accordingly.
    
## Description

- - -
This change is [Reviewable](https://reviewable.io/reviews/isl-org/Open3D/6308)

### DIFF
--- a/cpp/open3d/geometry/PointCloudSegmentation.cpp
+++ b/cpp/open3d/geometry/PointCloudSegmentation.cpp
@@ -29,6 +29,7 @@ public:
     explicit RandomSampler(const size_t total_size) : total_size_(total_size) {}
 
     std::vector<T> operator()(size_t sample_size) {
+        std::lock_guard<std::mutex> lock(mutex_);
         std::vector<T> samples;
         samples.reserve(sample_size);
 
@@ -48,6 +49,7 @@ public:
 
 private:
     size_t total_size_;
+    std::mutex mutex_;
 };
 
 /// \class RANSACResult
--- a/cpp/tests/geometry/PointCloud.cpp
+++ b/cpp/tests/geometry/PointCloud.cpp
@@ -18,6 +18,7 @@
 #include "open3d/io/ImageIO.h"
 #include "open3d/io/PinholeCameraTrajectoryIO.h"
 #include "open3d/io/PointCloudIO.h"
+#include "open3d/utility/Random.h"
 #include "open3d/visualization/utility/DrawGeometry.h"
 #include "tests/Tests.h"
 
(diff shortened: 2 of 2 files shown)


cpp/open3d/geometry/PointCloudSegmentation.cpp
```diff
@@ -29,6 +29,7 @@ class RandomSampler {
     explicit RandomSampler(const size_t total_size) : total_size_(total_size) {}
 
     std::vector<T> operator()(size_t sample_size) {
+        std::lock_guard<std::mutex> lock(mutex_);
```
theNded, line 32: I think it is sacrificing performance for reproducibility. The omp's performance gain will probably be negated by a mutex. 
If you want to keep this change, would you please show benchmark result (before and after) and show there is no degrade in speed? Otherwise, it may be better to use an option to toggle omp for reproducibility instead of introducing mutex.

theNded (changes requested):

cpp/open3d/geometry/PointCloudSegmentation.cpp
yuecideng: Hi @theNded , I have used the following code to benchmark the performace under macos m1 and linux intel i7-11700 :
```python
import numpy as np
import open3d as o3d
import time 

data = o3d.data.PLYPointCloud()
pcd = o3d.io.read_point_cloud(data.path)

num_iter = 1000
ts = []
for _ in range(100):
    o3d.utility.random.seed(1)
    pcd_down = pcd.voxel_down_sample(voxel_size=0.01)

    t0 = time.time()
    plane_model, inliers = pcd_down.segment_plane(distance_threshold=0.02,
                                                  ransac_n=3,
                                                  num_iterations=num_iter,
                                                  probability=1)
    duration = time.time() - t0
    ts.append(duration)

print(f'mean time {np.mean(np.array(ts))}')
```

The results (before and after) are very closed:
macos m1 : 0.030s - 0.032s
linux intel i7-11700: 0.090s - 0.093s

I think it is suitable to adopt this modification.

theNded: Thank you for the clarification. I will merge this PR, but we still need to bear in mind potential problems. Please refer to this PR if you observe relevant issues.

theNded (approved):

ssheorey:
Thanks @yuecideng, @theNded !

PR #6312, opened 2023-08-14
Removed the rudementary operations of taking the arc cosine from the function of comparing pairs

PiatrouskiIM:
…

Since the arccosine function is always decreasing on its domain, the expression acos(x) > acos(y) is equivalent to the expression x < y.

## Type

## Checklist:

    to my code.
        updated accordingly.
    
## Description

- - -
This change is [Reviewable](https://reviewable.io/reviews/isl-org/Open3D/6312)

### DIFF
--- a/cpp/open3d/pipelines/registration/Feature.cpp
+++ b/cpp/open3d/pipelines/registration/Feature.cpp
@@ -32,7 +32,7 @@ static Eigen::Vector4d ComputePairFeatures(const Eigen::Vector3d &p1,
     auto n2_copy = n2;
     double angle1 = n1_copy.dot(dp2p1) / result(3);
     double angle2 = n2_copy.dot(dp2p1) / result(3);
-    if (acos(fabs(angle1)) > acos(fabs(angle2))) {
+    if (fabs(angle1) < fabs(angle2)) {
         n1_copy = n2;
         n2_copy = n1;
         dp2p1 *= -1.0;


theNded:
Thanks for the catch!
 Since this is more like a typo fix (of course, from the math perspective) with no performance/correctness impact, I would suggest accumulating various instances before making a PR. CI is a relatively valuable resource. I will close this PR now, but feel free to reopen it when you have more findings.

cpp/open3d/pipelines/registration/Feature.cpp
```diff
@@ -32,7 +32,7 @@ static Eigen::Vector4d ComputePairFeatures(const Eigen::Vector3d &p1,
     auto n2_copy = n2;
     double angle1 = n1_copy.dot(dp2p1) / result(3);
     double angle2 = n2_copy.dot(dp2p1) / result(3);
-    if (acos(fabs(angle1)) > acos(fabs(angle2))) {
+    if (fabs(angle1) < fabs(angle2)) {
```
saurabheights, line 35: @theNded Could you explain why you said this has no performance impact and its a "typo fix"?

theNded: This is definetely a sub-millisecond operation, so there would not be a performance impact at all.
It is a "typo fix" in the math sense. Nothing is really wrong, it just becomes slightly neater.

Of course, I encourage such minor fixes! But as said, it might be better to accumulate multiple such changes to make a PR altogether. This will help maintain a neat git history and stable CI usage.

saurabheights: Thanks. Makes sense.

PR #6407, opened 2023-10-04
Add missing keyword arg names to visualization bindings

djkeyes:
## Type

## Motivation and Context

I was using the python visualization API, and I noticed that some of the keyword arguments in the python bindings show up as "arg0" "arg1" etc. This made it confusing, because I had to keep switching between the python API and the source code to see what the arguments really meant. This adds the missing argument names.

## Checklist:

    to my code.
        updated accordingly.
    
## Description

Before, if I ran something like `help(open3d.visualization.Visualizer.reset_view_point)`, I would get a message that looks like this (notice the `arg0`):

```
Help on instancemethod in module open3d.cpu.pybind.visualization:

reset_view_point(...)
    reset_view_point(self, arg0)
    Function to reset view point
    
    Args:
        arg0 (bool)
    
    Returns:
        None
```

Now it looks like this:

```
Help on instancemethod in module open3d.cpu.pybind.visualization:

reset_view_point(...)
    reset_view_point(self, reset_bounding_box=False)
    Function to reset view point
    
    Args:
        reset_bounding_box (bool, optional, default=False): Set to ``False`` to keep current viewpoint
    
    Returns:
        None
```

I have added similar fixes for other functions in `open3d.visualization`. I believe this is the only module where the keyword arguments were missing in the python docs.  In `O3DVisualizer`, some of the documentation style was inconsistent, so I cleaned that up, too.

I checked the generated html docs and verified that the results look okay. Let me know if you'd like any other changes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/isl-org/Open3D/6407)

### DIFF
--- a/cpp/pybind/visualization/o3dvisualizer.cpp
+++ b/cpp/pybind/visualization/o3dvisualizer.cpp
@@ -158,16 +158,17 @@ void pybind_o3dvisualizer(py::module& m) {
                  "close",
                  "callback"_a)
             .def("show_menu", &O3DVisualizer::ShowMenu,
-                 "show_menu(show): shows or hides the menu in the window, "
-                 "except on macOS since the menubar is not in the window "
-                 "and all applications must have a menubar.")
+                 "Shows or hides the menu in the window, except on macOS since "
+                 "the menubar is not in the window and all applications must "
+                 "have a menubar.",
+                 "show"_a)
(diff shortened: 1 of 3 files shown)


ssheorey (approved):
Thanks @djkeyes for this very useful update!

PR #6409, opened 2023-10-05
Add RemoveDuplicatedTriangles feature using tensor API

PierreIntel:
## Type

-   [ x] New feature (non-breaking change which adds functionality). Resolves #

## Motivation and Context

This PR is an implementation of the RemoveDuplicatedTriangles but using the tensor API to be able to run it either on CPU or GPU (only CPU for now in this PR).
This was also an opportunity to get familiar with the Open3D code.

## Checklist:

-   [x ] I have run `python util/check_style.py --apply` to apply Open3D **code style**
    to my code.
-   [x ] This PR changes Open3D behavior or adds new functionality.
    -   [x ] Both C++ (Doxygen) and Python (Sphinx / Google style) **documentation** is
        updated accordingly.
    -   [x ] I have added or updated C++ and / or Python **unit tests** OR included **test
-   [x ] I will follow up and update the code if CI fails.
    
-   [x ] For fork PRs, I have selected **Allow edits from maintainers**.

## Description

Here is the note of my PR for the RemoveDuplicatedTriangles feture using Tensor API.

The first problem of this feature is how to identify duplicated triangles.
I keep the same solution than the Eigen based API RemoveDuplicatedTriangles that is to use a unordered_map that has a complexity of search in O(1) due to the use of a hash function.

This unordered_map is implemented with a tuple as key instead of a tensor directly is because the hash function using a tensor is not implemented.
So we have to "convert" a vertice tensor of 3 points to a tuple to be able to keep track of the triangle.

The second problem is how to remove the duplicated triangles.
I choose to iterate on the index of the unique triangles and to copy them in a new tensor

I did this like that because we don't know the size of the new tensor of triangles until we have looped on all triangles.

I have tried to use :
	tensor.reshape(), but it reshapes the tensor in a tensor with the same number of elements.
	tensor.append(), but we need it doesn't increase the size of the vector where we append elements.

In the worst case where we have only one duplicated triangles we will need to iterate one time on the number of triangles and an other time on number of triangles - 1,
so the complexity is in O(n*(n-1)) and in the best case O(n) due to the test if before the second loop.

I have also treated the case if there are some normales triangles. I suppose that the number of triangles and normales triangles are the same and they are at a corresponding index.

- - -
This change is [Reviewable](https://reviewable.io/reviews/isl-org/Open3D/6409)

### DIFF
--- a/cpp/open3d/t/geometry/TriangleMesh.cpp
+++ b/cpp/open3d/t/geometry/TriangleMesh.cpp
@@ -180,12 +180,12 @@ TriangleMesh &TriangleMesh::RemoveDuplicatedTriangles() {
     if (num_triangles < 2) {
         return *this;
     }
-
     // unordered_map to keep track of existing triangles
     // hash function with code::Tensor not implemented so we use std::tuple
-    typedef std::tuple<int, int, int> Index3;
+    typedef std::tuple<int64_t, int64_t, int64_t> Index3;
     std::unordered_map<Index3, size_t, utility::hash_tuple<Index3>>
             triangle_to_old_index;
+    std::vector<int64_t> triangle_vec;
 
     // Check if mesh has normals triangles and colors
     bool has_tri_normal = HasTriangleNormals();
@@ -199,35 +199,27 @@ TriangleMesh &TriangleMesh::RemoveDuplicatedTriangles() {
     for (int64_t i = 0; i < num_triangles; i++) {
         Index3 triangle_in_tuple;
         // Extract vertices of the triangle
-        core::Tensor triangle_to_check =
-                triangles.GetItem({core::TensorKey::Index(i)});
-        int64_t vertice_0 =
-                triangle_to_check.GetItem({core::TensorKey::Index(0)})
-                        .Item<int64_t>();
-        int64_t vertice_1 =
-                triangle_to_check.GetItem({core::TensorKey::Index(1)})
-                        .Item<int64_t>();
-        int64_t vertice_2 =
-                triangle_to_check.GetItem({core::TensorKey::Index(2)})
-                        .Item<int64_t>();
+        core::Tensor triangle_to_check = triangles[i];
+
+        if (triangles.GetDtype() == core::Int32) {
+            auto vertice_0 = triangle_to_check[0].Item<int32_t>();
+            auto vertice_1 = triangle_to_check[1].Item<int32_t>();
+            auto vertice_2 = triangle_to_check[2].Item<int32_t>();
+            triangle_vec = {vertice_0, vertice_1, vertice_2};
+        } else if (triangles.GetDtype() == core::Int64) {
+            auto vertice_0 = triangle_to_check[0].Item<int64_t>();
+            auto vertice_1 = triangle_to_check[1].Item<int64_t>();
+            auto vertice_2 = triangle_to_check[2].Item<int64_t>();
+            triangle_vec = {vertice_0, vertice_1, vertice_2};
+        }
+
         // We first need to find the minimum index. Because triangle (0-1-2)
         // and triangle (2-0-1) are the same.
-        if (vertice_0 <= vertice_1) {
-            if (vertice_1 <= vertice_2)
-                triangle_in_tuple =
-                        std::make_tuple(vertice_0, vertice_1, vertice_2);
-            else
-                triangle_in_tuple =
-                        std::make_tuple(vertice_0, vertice_2, vertice_1);
-        } else {
-            if (vertice_1 <= vertice_2)
-                triangle_in_tuple =
-                        std::make_tuple(vertice_1, vertice_2, vertice_0);
-            else
-                triangle_in_tuple =
-                        std::make_tuple(vertice_1, vertice_0, vertice_2);
-        }
-        // check if triangle is already present in the mesh
+        std::sort(triangle_vec.begin(), triangle_vec.end());
+        triangle_in_tuple = std::make_tuple(triangle_vec[0], triangle_vec[1],
+                                            triangle_vec[2]);
+
+        // check is triangle is already present in the mesh
         if (triangle_to_old_index.find(triangle_in_tuple) ==
             triangle_to_old_index.end()) {
             mask_of_triangles[i] = true;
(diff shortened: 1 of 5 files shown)


cpp/open3d/t/geometry/TriangleMesh.cpp
```diff
@@ -174,6 +174,98 @@ TriangleMesh &TriangleMesh::Rotate(const core::Tensor &R,
...
+                triangles.GetItem({core::TensorKey::Index(i)});
+        int64_t vertice_0 =
+                triangle_to_check.GetItem({core::TensorKey::Index(0)})
+                        .Item<int64_t>();
+        int64_t vertice_1 =
+                triangle_to_check.GetItem({core::TensorKey::Index(1)})
+                        .Item<int64_t>();
+        int64_t vertice_2 =
+                triangle_to_check.GetItem({core::TensorKey::Index(2)})
+                        .Item<int64_t>();
```
ssheorey, line 208: We need to ensure this works for different dtypes for vertices (specifically int32_t and int64_t). This will output junk values for int32_t vertices.

**minor:**

- `operator[]` is shorter: `triangle_to_check[0].Item<dtype>()`
- Would using `std::rotate` to reorder elements in the tuple be cleaner than the multiple `ifs` ?
- Can you comment on parallelization here? Is it possible and would it provide a useful speedup?

cpp/open3d/t/geometry/TriangleMesh.cpp
```diff
@@ -174,6 +174,98 @@ TriangleMesh &TriangleMesh::Rotate(const core::Tensor &R,
...
+                        std::make_tuple(vertice_1, vertice_2, vertice_0);
+            else
+                triangle_in_tuple =
+                        std::make_tuple(vertice_1, vertice_0, vertice_2);
+        }
+        // check is triangle is already present in the mesh
+        if (triangle_to_old_index.find(triangle_in_tuple) ==
+            triangle_to_old_index.end()) {
+            triangle_to_old_index[triangle_in_tuple] = i;
+            index_to_keep.push_back(i);
```
ssheorey, line 230: Would using a boolean mask of triangles to keep, instead of a vector of indices be faster? Tensors support indexing with a Boolean mask, similar to numpy. See `SelectFacesByMask()`.

cpp/open3d/t/geometry/TriangleMesh.cpp
```diff
@@ -174,6 +174,98 @@ TriangleMesh &TriangleMesh::Rotate(const core::Tensor &R,
...
+
+    // do this only if there are some triangle to remove
+    if (new_size != num_triangles) {
+        core::Tensor triangles_without_dup = core::Tensor::Empty(
+                {new_size, 3}, triangles.GetDtype(), GetDevice());
+
+        core::Tensor triangles_normals, triangles_normals_without_dup;
+
+        // fill only is there are normals triangle to avoid errors
+        if (has_tri_normal) {
```
ssheorey, line 243: Can you support other triangle attributes? (e.g. triangle colors, etc.) An important feature of the Tensor API data structures is support for generic triangle and vertex attributes.

ssheorey (changes requested):

cpp/open3d/t/geometry/TriangleMesh.cpp
PierreIntel: std::rotate can be used but it will need multiple ifs too to know which vertices to rotate. Example {2,1,0} will need 2 rotations to get {0,1,2}. What can be used is instantiate a std::vector<int64_t> of 3 elements, sort it with std::sort and next initialize the tuple for the unordered map.
Also elements of triangle_vec are int64_t like that it can get vertices in int32_t or int64_t.
```
        std::vector<int64_t> triangle_vec;
        if (triangles.GetDtype() == core::Int32)
        {
            auto vertice_0 = triangle_to_check[0].Item<int32_t>();
            auto vertice_1 = triangle_to_check[1].Item<int32_t>();
            auto vertice_2 = triangle_to_check[2].Item<int32_t>();
            triangle_vec = {vertice_0, vertice_1, vertice_2};
        }
        else if (triangles.GetDtype() == core::Int64)
        {
            auto vertice_0 = triangle_to_check[0].Item<int64_t>();
            auto vertice_1 = triangle_to_check[1].Item<int64_t>();
            auto vertice_2 = triangle_to_check[2].Item<int64_t>();
            triangle_vec = {vertice_0, vertice_1, vertice_2};
        }

        // We first need to find the minimum index. Because triangle (0-1-2)
        // and triangle (2-0-1) are the same.
        std::sort(triangle_vec.begin(), triangle_vec.end());
        triangle_in_tuple =
                        std::make_tuple(triangle_vec[0], triangle_vec[1], triangle_vec[2]);
```

The parallelization is possible but we need to be sure that the access to the unordered map is in a critical section to avoid multiple accesses adding the duplicate triangle in the unordered map.

cpp/open3d/t/geometry/TriangleMesh.cpp
PierreIntel: I succeed to manage this in : https://github.com/isl-org/Open3D/pull/6409/commits/98fe0622ffdfc92fa329bb9fcf271a0f79303a5d

PierreIntel: I succeed to manage it in https://github.com/isl-org/Open3D/pull/6409/commits/3aa93dd50c4b97af754e69ce4ff64df9e1a556ac

cpp/open3d/t/geometry/TriangleMesh.cpp
PierreIntel: Yes the triangle colors are manage in this : https://github.com/isl-org/Open3D/pull/6409/commits/3aa93dd50c4b97af754e69ce4ff64df9e1a556ac

cpp/open3d/t/geometry/TriangleMesh.cpp
```diff
@@ -174,6 +174,98 @@ TriangleMesh &TriangleMesh::Rotate(const core::Tensor &R,
     return *this;
 }
 
+TriangleMesh &TriangleMesh::RemoveDuplicatedTriangles() {
+    core::Tensor triangles = GetTriangleIndices();
+    int64_t num_triangles = triangles.GetLength();
+    if (num_triangles < 1) {
+        utility::LogError("Mesh must have at least two triangles.");
+    }
```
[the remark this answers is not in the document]
PierreIntel: This is done in https://github.com/isl-org/Open3D/pull/6409/commits/a6ed51c8eafca3da5eeeca6552051505f2eb14a1

cpp/open3d/t/geometry/TriangleMesh.cpp
PierreIntel: done in https://github.com/isl-org/Open3D/pull/6409/commits/c890dc81096ad395074fad5d02cc813c7cd789bf

ssheorey (changes requested):

PR #6415, opened 2023-10-06
Implement open3d::t::geometry::TriangleMesh::SelectByIndex

nsaiapova:
The method uses tensor API. It takes a list of indices and returns a new mesh built with the selected vertices and triangles formed by these vertices.

## Type

## Motivation and Context

The method is available with the `open3d::geometry::TriangleMesh` but not in `open3d::t::geometry::TriangleMesh`.

## Checklist:

    to my code.
    -   [?] Both C++ (Doxygen) and Python (Sphinx / Google style) **documentation** is
        updated accordingly.   <---- here I am not sure about the Python doc, as I could not generate it locally.
    
## Description

The implementation is inspired by
  open3d::geometry::TriangleMesh::SelectByIndex.
and by
  open3d::t::geometry::TriangleMesh::SelectFacesByMask.

We first compute a mask of vertices to be selected.  If the input index exceeds the maximum number of vertices, we throw an exception. Based on the vertex mask we build a mapping index vector using inclusive prefix sum algorithm. We then update the selected vertex indices in the triangles CPU copy and build the triangles mask. I put that to a templated helper static function to avoid the code duplication for Int32 and Int64 vertex indices types.

With the computed masks we can select the vertices and triangles back on the mesh's device and copy the attributes.

FYI @ssheorey

- - -
This change is [Reviewable](https://reviewable.io/reviews/isl-org/Open3D/6415)

### DIFF
--- a/cpp/open3d/t/geometry/TriangleMesh.cpp
+++ b/cpp/open3d/t/geometry/TriangleMesh.cpp
@@ -993,6 +993,54 @@ int TriangleMesh::PCAPartition(int max_faces) {
     return num_parititions;
 }
 
+/// A helper to compute new vertex indices out of vertex mask.
+/// \param tris_cpu tensor with triangle indices to update.
+/// \param vertex_mask tensor with the mask for vertices.
+template <typename T>
+static void UpdateTriangleIndicesByVertexMask(core::Tensor &tris_cpu,
+                                              const core::Tensor &vertex_mask) {
+    int64_t num_verts = vertex_mask.GetLength();
+    int64_t num_tris = tris_cpu.GetLength();
+    const T *vertex_mask_ptr = vertex_mask.GetDataPtr<T>();
+    std::vector<T> prefix_sum(num_verts + 1, 0);
+    utility::InclusivePrefixSum(vertex_mask_ptr, vertex_mask_ptr + num_verts,
+                                &prefix_sum[1]);
+
+    // update triangle indices
+    T *vert_idx_ptr = tris_cpu.GetDataPtr<T>();
+    for (int64_t i = 0; i < num_tris * 3; ++i) {
+        vert_idx_ptr[i] = prefix_sum[vert_idx_ptr[i]];
+    }
+}
+
+/// A helper to copy mesh attributes.
+/// \param dst destination mesh
+/// \param src source mesh
+/// \param vertex_mask vertex mask of the source mesh
+/// \param tri_mask triangle mask of the source mesh
+static void CopyAttributesByMasks(TriangleMesh &dst,
+                                  const TriangleMesh &src,
+                                  const core::Tensor &vertex_mask,
+                                  const core::Tensor &tri_mask) {
+    if (src.HasVertexPositions() && dst.HasVertexPositions()) {
+        for (auto item : src.GetVertexAttr()) {
+            if (!dst.HasVertexAttr(item.first)) {
+                dst.SetVertexAttr(item.first,
+                                  item.second.IndexGet({vertex_mask}));
+            }
+        }
+    }
+
+    if (src.HasTriangleIndices() && dst.HasTriangleIndices()) {
+        for (auto item : src.GetTriangleAttr()) {
+            if (!dst.HasTriangleAttr(item.first)) {
+                dst.SetTriangleAttr(item.first,
+                                    item.second.IndexGet({tri_mask}));
+            }
+        }
+    }
+}
+
 TriangleMesh TriangleMesh::SelectFacesByMask(const core::Tensor &mask) const {
     core::AssertTensorShape(mask, {GetTriangleIndices().GetLength()});
     core::AssertTensorDtype(mask, core::Bool);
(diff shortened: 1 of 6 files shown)


cpp/open3d/t/geometry/TriangleMesh.cpp
```diff
@@ -1061,6 +1061,106 @@ TriangleMesh TriangleMesh::SelectFacesByMask(const core::Tensor &mask) const {
...
+    if (GetTriangleIndices().GetDtype() == core::Int32) {
+        core::AssertTensorDtype(indices, core::Int32);
+    } else {
+        // we allow both Int32 and Int64 if the mesh indicies are Int64
+        core::AssertTensorDtypes(indices, {core::Int32, core::Int64});
+    }
+
+    // really copy triangles to CPU as we will modify the indices
+    core::Tensor tris_cpu =
+            GetTriangleIndices().To(core::Device(), true).Contiguous();
```
nsaiapova, line 1119: The `true` argument here makes the "true" copy, which the `SelectFacesByMask` does not do. In my case, if I don't have it, the triangles do not match the original mesh if the device is CPU.

cpp/open3d/t/geometry/TriangleMesh.cpp
```diff
@@ -1061,6 +1061,108 @@ TriangleMesh TriangleMesh::SelectFacesByMask(const core::Tensor &mask) const {
...
+                                           core::Tensor &tris_mask,
+                                           core::Tensor &tris_cpu) {
+    const int64_t num_tris = tris_cpu.GetLength();
+    const int64_t num_verts = vertex_mask.GetLength();
+
+    // compute the vertices mask
+    intT *vertex_mask_ptr = vertex_mask.GetDataPtr<intT>();
+    const intT *indices_ptr = indices.GetDataPtr<intT>();
+    for (int64_t i = 0; i < indices.GetLength(); ++i) {
+        if (indices_ptr[i] >= num_verts) {
```
nsaiapova, line 1078: just realised I need to check `< 0` too, and I assume that the vertex indices are consecutive w/o gaps in between.

cpp/tests/t/geometry/TriangleMesh.cpp
```diff
@@ -941,5 +941,92 @@ TEST_P(TriangleMeshPermuteDevices, CreateMobius) {
             triangle_indices_custom));
 }
 
+TEST_P(TriangleMeshPermuteDevices, SelectByIndex_Box) {
```
nsaiapova, line 944: we can also add a test for an empty mesh, empty index list, index lists with negatives and duplicates.

cpp/open3d/t/geometry/TriangleMesh.cpp
```diff
@@ -1061,6 +1061,108 @@ TriangleMesh TriangleMesh::SelectFacesByMask(const core::Tensor &mask) const {
     return result;
 }
 
+// A helper to compute the vertex and triangle masks based on indices.
+// Additionally updates tris_cpu to new indices.
```
nsaiapova, line 1065: I was not sure whether the static helpers should be documented via doxygen.

nsaiapova:
In the last fixup I pushed more tests and a few fixes for issues found by these tests.

cpp/open3d/t/geometry/TriangleMesh.cpp
```diff
@@ -1061,6 +1061,124 @@ TriangleMesh TriangleMesh::SelectFacesByMask(const core::Tensor &mask) const {
...
+}
+
+TriangleMesh TriangleMesh::SelectByIndex(const core::Tensor &indices) const {
+    core::AssertTensorShape(indices, {indices.GetLength()});
+    if (!HasTriangleIndices()) {
+        utility::LogError("[SelectByIndex] mesh has no triangle indices.");
+    }
+    if (!HasVertexPositions()) {
+        utility::LogError("[SelectByIndex] mesh has no vertex positions.");
+    }
```
ssheorey, line 1118: We can return an empty mesh for corner case such as when indices are empty.

cpp/open3d/t/geometry/TriangleMesh.cpp
```diff
@@ -1061,6 +1061,124 @@ TriangleMesh TriangleMesh::SelectFacesByMask(const core::Tensor &mask) const {
...
+        if (vertex_mask_ptr[tris_cpu_ptr[3 * i]] == 1 &&
+            vertex_mask_ptr[tris_cpu_ptr[3 * i + 1]] == 1 &&
+            vertex_mask_ptr[tris_cpu_ptr[3 * i + 2]] == 1) {
+            tris_cpu_ptr[3 * i] = prefix_sum[tris_cpu_ptr[3 * i]];
+            tris_cpu_ptr[3 * i + 1] = prefix_sum[tris_cpu_ptr[3 * i + 1]];
+            tris_cpu_ptr[3 * i + 2] = prefix_sum[tris_cpu_ptr[3 * i + 2]];
+            tris_mask_ptr[i] = true;
+        }
+    }
+}
```
ssheorey, line 1109: In the interest of reducing duplicate code, do you want to factor out the common code from here and `TriangleMesh::SelectFacesByMask()`?

ssheorey (commented):
Looks great! Some minor comments about corner cases and a refactoring suggestion.

cpp/open3d/t/geometry/TriangleMesh.cpp
```diff
@@ -1061,6 +1061,124 @@ TriangleMesh TriangleMesh::SelectFacesByMask(const core::Tensor &mask) const {
...
+        utility::LogError("[SelectByIndex] mesh has no vertex positions.");
+    }
+    GetTriangleAttr().AssertSizeSynchronized();
+    GetVertexAttr().AssertSizeSynchronized();
+    if (GetTriangleIndices().GetDtype() == core::Int32) {
+        core::AssertTensorDtype(indices, GetTriangleIndices().GetDtype());
+    } else {
+        // we allow both Int32 and Int64 if the mesh indicies are Int64
+        core::AssertTensorDtypes(indices, {core::Int32, core::Int64});
+    }
```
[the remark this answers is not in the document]
nsaiapova: I now accept all integral types with the help of 
https://github.com/isl-org/Open3D/pull/6415/files#diff-c8634e2912960fb9c6991e572771973ca0f2c77a9639172ca68b0e788fb358c6R117-R146

cpp/open3d/t/geometry/TriangleMesh.cpp
nsaiapova: I added two static helpers: one to update the triangle indices and one to copy the mesh attributes.

cpp/open3d/t/geometry/TriangleMesh.cpp
```diff
@@ -1045,18 +1093,126 @@ TriangleMesh TriangleMesh::SelectFacesByMask(const core::Tensor &mask) const {
     core::Tensor verts = GetVertexPositions().IndexGet({vertex_mask});
     TriangleMesh result(verts, tris);
 
-    // copy attributes
-    for (auto item : GetVertexAttr()) {
-        if (!result.HasVertexAttr(item.first)) {
-            result.SetVertexAttr(item.first,
-                                 item.second.IndexGet({vertex_mask}));
-        }
+    CopyAttributesByMasks(result, *this, vertex_mask, mask);
```
nsaiapova, line 1089: ops, that went to a wrong commit.

nsaiapova:
@ssheorey I am not exactly sure why the CI fails, this link does not give me the clarity: https://github.com/isl-org/Open3D/actions/runs/6798298171/job/18482179687?pr=6415. Could you please help me to find out whether it is caused by my PR?

cpp/open3d/t/geometry/TriangleMesh.cpp
```diff
@@ -993,7 +993,66 @@ int TriangleMesh::PCAPartition(int max_faces) {
...
+            }
+        }
+    }
+}
+
 TriangleMesh TriangleMesh::SelectFacesByMask(const core::Tensor &mask) const {
+    if (!HasVertexPositions()) {
+        utility::LogWarning(
+                "[SelectFacesByMask] mesh has no vertex positions.");
+        return {};
```
ssheorey, line 1048: minor nit: return TriangleMesh dtype, device not necessarily equal to current TriangleMesh dtype, device. Use 
```return TriangleMesh(Tensor::EmptyLike(GetVertexPositions()), Tensor::EmptyLike(GetTriangleIndices()));```

File: cpp/open3d/t/geometry/TriangleMesh.cpp [hunk 10]
```diff
@@ -993,7 +993,66 @@ int TriangleMesh::PCAPartition(int max_faces) {
...
+    }
+    if (!HasTriangleIndices()) {
+        utility::LogWarning(
+                "[SelectFacesByMask] mesh has no triangle indices.");
+        return {};
```
ssheorey, line 1053: same

ssheorey (approved):

cpp/open3d/t/geometry/TriangleMesh.cpp
nsaiapova: But there is no `GetVertexPositions()` as per condition on line 1045. Btw, if there is no vertex positions, can I rely on that there are no triangles? Or there could be a weirdly constructed meshes?

ssheorey: OK, looks good in that case. Ok to return anything reasonable for erroneously constructed meshes (e.g. triangles but no vertices).

PR #6488, opened 2023-11-13
Add Open3D architecture diagram

ssheorey:
Open3D_layers


errissa (approved):
Looks good. The diagram being a video may be a little bit of overkill but that's not a blocker.

Issue #6498, opened 2023-11-20
Error followed by a crash (segmentation fault) with GUI "selection" tab with LineSet object


EwingKang:
### Checklist

- [X] I have searched for [similar issues](https://github.com/isl-org/Open3D/issues).
- [X] For Python issues, I have tested with the [latest development wheel](http://www.open3d.org/docs/latest/getting_started.html#development-version-pip).
- [X] I have checked the [release documentation](http://www.open3d.org/docs/release/) and the [latest documentation](http://www.open3d.org/docs/latest/) (for `master` branch).

### Describe the issue

I'm using c++ devel
If there exists an `open3d::geometry::LineSet` object, GUI from `open3d::visualization::Draw()` would crash upon selection of  _Actions -> show settings -> Mouse control (drop down) -> Selection (tab)_.

Upon inspections with GDB, there are two part of this issue. The first one is a log error at [cpp/open3d/visualization/gui/PickPointsInteractor.cpp:85](https://github.com/isl-org/Open3D/blob/2a11e0e40d5f9db4c80d9b6543e455d70a368822/cpp/open3d/visualization/gui/PickPointsInteractor.cpp#L85). And the second actual segmentation fault is a memory-related crash during destructor. The traceback of second crash is attached.

### Steps to reproduce the bug
In C++:
```c++
auto linset_aaa_t = std::make_shared<open3d::t::geometry::LineSet>(aaa, aaa_indices);
std::vector<open3d::visualization::DrawObject> drawobject;
drawobject.push_back(open3d::visualization::DrawObject("lineset", linset_aaa_t, true));
open3d::visualization::Draw(drawobject, "path");
```
* Execute the program
* Actions -> show settings -> Mouse control (drop down) -> Selection (tab).

Crash immediately

### Error message

[Open3D Error] (void open3d::visualization::gui::SelectionIndexLookup::Add(const string&, size_t)) /root/Open3D/cpp/open3d/visualization/gui/PickPointsInteractor.cpp:85: start_index 25142 must be larger than all previously added objects 25142.
[Open3D INFO] Memory Statistics: (Device) (#Malloc) (#Free)
[Open3D INFO] ---------------------------------------------
[Open3D WARNING] CPU:0: 62 58 --> 4 with 19768 total bytes
[Open3D WARNING]     0x5653a5dbd750 @ 2288 bytes
[Open3D WARNING]     0x5653a5ffd4a0 @ 2288 bytes
[...]

PR #6499, opened 2023-11-20
Fix the triggering in PickPointsInteractor when there are linesets pr…

EwingKang:
Adds handling of LineSet that is similar to Mesh and PointCloud in SetPickableGeometry()

## Type

## Motivation and Context

This change fixes https://github.com/isl-org/Open3D/issues/6498

## Checklist:

    to my code. **(On U22.04, clang_format 10 is not available, the default install script didn't worked, I eyeballed the format instead)**
        updated accordingly.
    
## Description

GUI no-longer crashes when there's lineset object in the scene when switching to `Actions -> show settings -> Mouse control (drop down) -> Selection (tab).`


godwincharan:
I could help review this PR

EwingKang:
Seems like Codacy caught an error (thanks!)
I fixed it. But not sure if I should just rebase (squash) the to make the commit tree cleaner.
Also, automated style check reported an error, but it only says "Style error found in the following files:... Error: Process completed with exit code 1.". I'm also not sure where to check the error report.

godwincharan (commented):
> *[Reviewable](https://reviewable.io/reviews/isl-org/Open3D/6499)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @EwingKang)

___
*[`cpp/open3d/visualization/gui/PickPointsInteractor.cpp` line 166 at r1](https://reviewable.io/reviews/isl-org/Open3D/6499#-NjxixZU7UJaxUTdABmV:-NjxixZU7UJaxUTdABmW:bbt9nwm) ([raw file](https://github.com/isl-org/Open3D/blob/837a76f6ffd2f3b223b585c47e3fe6f3625c0380/cpp/open3d/visualization/gui/PickPointsInteractor.cpp#L166)):*
> ```cpp
>         auto tmesh =
>                 dynamic_cast<const t::geometry::TriangleMesh *>(pg.tgeometry);
>         auto lineset = dynamic_cast<const geometry::LineSet *>(pg.geometry);
> ```

Can we implement polymorphism here rather than dynamically casting and checking for nullptr?

EwingKang:
Hi @godwincharan ,
I think it would be possible. However the structure `PickableGeometry` defined in [SceneWidget.h#L131](https://github.com/isl-org/Open3D/blob/master/cpp/open3d/visualization/gui/SceneWidget.h#L131) is not a templated class by itself. It actually stores two raw pointers to objects:
```c++
        const geometry::Geometry3D* geometry = nullptr;
        const t::geometry::Geometry* tgeometry = nullptr;
```
So one would have to check which one is the valid pointer essentially. 
Speaking about polymorphism, my first idea of the solution is to make `PickableGeometry` templated by itself, and implement specializations for different objects.

Another way to implement this is to add some sort of base function that returns interest points in their parent class. 
For the newer `open3d::t::geometry::` namespace, [There are 5 chlidrens in `DrawableGeometry`](http://www.open3d.org/docs/release/cpp_api/classopen3d_1_1t_1_1geometry_1_1_drawable_geometry.html#a72987c98386c0a92919cdd2c945af2a2). I think this is doable.
For the old `open3d::geometry::XXXXX` object, [there are many chlidren](http://www.open3d.org/docs/release/cpp_api/classopen3d_1_1geometry_1_1_geometry3_d.html) that all inherits `open3d::geometry`. I don't think it's a good idea to enforce the base function.

The third way is to make pickable objects inherit a new base class. This would require some careful design directly from the owner/maintainers IMHO.

My original intent is to quickly fix this crashing issue for me, so I simply follow what it has been done there. But I would love to implement a more complete solution (at least for the newer `open3d::t::` namespace as described in solution 2). What does everyone think about this issue? I would be happy to have some inputs from you.

ssheorey (approved):
Thanks @EwingKang for the fix!

PS: Here's the last part of the output of `make apply-style` for me:
```
...
Please correct license header *manually* in the following files (see util/check_style.py for the standard header):
/home/ssheorey/Documents/Open3D/Code/Open3D/examples/cpp/TImage.cpp
make[3]: *** [CMakeFiles/apply-style.dir/build.make:71: CMakeFiles/apply-style] Error 1
make[2]: *** [CMakeFiles/Makefile2:2053: CMakeFiles/apply-style.dir/all] Error 2
make[1]: *** [CMakeFiles/Makefile2:2060: CMakeFiles/apply-style.dir/rule] Error 2
make: *** [Makefile:709: apply-style] Error 2
```
I was able to fix it by updating the license header text. Is this the issue you ran into?

ssheorey:
@godwincharan agree with your assessment that polymorphism may help to simplify this code. However, the refactoring will take some careful planning and effort as @EwingKang has noted. Please reach out to me if you would like to help.

EwingKang:
@ssheorey Thanks for merging the code. 
I'm actively using Open3D in my project so I'm definitely going to have some feedbacks. It is probably within the reach for me to contribute the polymorphism part, but I think I'll first take some time to get familiar with the code base and design philosophy.

About the style checker, I just re-ran the checker with `python3 util/check_style.py --apply` and it did worked. But I've done a lot to my system so practically everything is changed. I'll try it again in a few weeks when I'm going to get a new PC, to see which step is missing.
Cheers.

EwingKang:
> 
> About the style checker, I just re-ran the checker with `python3 util/check_style.py --apply` and it did worked. But I've done a lot to my system so practically everything is changed. I'll try it again in a few weeks when I'm going to get a new PC, to see which step is missing. Cheers.

So I've test it again with a fresh Ubuntu 22.04 install. It turns out that the when installing the style checker dependencies [following the instructions](https://www.open3d.org/docs/release/contribute/styleguide.html#style-guide), the install defaults to `~/.local/bin`. It also gave warning like this:
```bash
...
  WARNING: The scripts clang-format, clang-format-diff.py and git-clang-format are installed in '/home/user/.local/bin' which is not on PATH.
  Consider adding this directory to PATH or, if you prefer to suppress this warning, use --no-warn-script-location.
  Attempting uninstall: nbformat
    Found existing installation: nbformat 5.9.2
    Uninstalling nbformat-5.9.2:
      Successfully uninstalled nbformat-5.9.2
  WARNING: The script jupyter-trust is installed in '/home/user/.local/bin' which is not on PATH.
  Consider adding this directory to PATH or, if you prefer to suppress this warning, use --no-warn-script-location.
Successfully installed clang-format-10.0.1.1 nbformat-5.7.0
...
```
I just installed it to the system with `sudo` and the `clang-format` style checker `python util/check_style.py` can be executed correctly
```bash
sudo pip3 install clang-format==10.0.1.1
```

PR #6529, opened 2023-12-09
Fix unused variable error when OpenMP is disabled

### LINKED ISSUE #6514: Build error during build from source on Sonoma M1 Max
### Checklist

#### I first cloned Open3D by:
```
git clone https://github.com/isl-org/Open3D.git
cd Open3D
```

#### Then, I build Open3D (on Ubuntu 20.04, with CUDA 11.5) with:
```
mkdir build && cd build
cmake .. -DBUILD_PYTHON_MODULE=OFF 
make -j$(sysctl -n hw.physicalcpu)
```

### Error message

```shell
[ 84%] Built target core
/Open3Ds/open3dlatest/cpp/open3d/pipelines/registration/Feature.cpp:158:15: error: unused variable 'kOuterThreads' [-Werror,-Wunused-variable]
    const int kOuterThreads = std::min(kMaxThreads, num_searches);
              ^
/Open3Ds/open3dlatest/cpp/open3d/pipelines/registration/Feature.cpp:159:15: error: unused variable 'kInnerThreads' [-Werror,-Wunused-variable]
    const int kInnerThreads = std::max(kMaxThreads / num_searches, 1);
              ^
2 errors generated.
make[2]: *** [cpp/open3d/pipelines/CMakeFiles/pipelines.dir/registration/Feature.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....
/Open3Ds/open3dlatest/cpp/open3d/t/pipelines/registration/Feature.cpp:104:15: error: unused variable 'kOuterThreads' [-Werror,-Wunused-variable]
    const int kOuterThreads = std::min(kMaxThreads, num_searches);
              ^
1 error generated.
```

```markdown
- Operating system: macOS 14.1.1
- Python version: Doesn't matter ( I think :D )
- Open3D version: tried 17.0 and latest from master branch
- System architecture: apple-silicon (M1 Max)
- Is this a remote workstation?: No
- Compiler version (if built from source): Clang 14 (from Xcode)
```

### Additional information

So yeah, firstly I got error during building VTK, but then got this. Now will surpress those warnings and then provide more comments

sitic:
## Type

-   [x] Bug fix (non-breaking change which fixes an issue): Fixes #6514

## Motivation and Context

```
error: unused variable 'kInnerThreads' [-Werror,-Wunused-variable]
```
type of errors occur when building without OpenMP on macOS Sonoma.

## Checklist:

    to my code.
        updated accordingly.
    
## Description

Adds ``(void)kOuterThreads;`` style fixes to suppress the warnings.

### DIFF
--- a/cpp/open3d/pipelines/registration/Feature.cpp
+++ b/cpp/open3d/pipelines/registration/Feature.cpp
@@ -157,6 +157,8 @@ CorrespondenceSet CorrespondencesFromFeatures(const Feature &source_features,
 
     const int kOuterThreads = std::min(kMaxThreads, num_searches);
     const int kInnerThreads = std::max(kMaxThreads / num_searches, 1);
+    (void)kOuterThreads;
+    (void)kInnerThreads;
 #pragma omp parallel for num_threads(kOuterThreads)
     for (int k = 0; k < num_searches; ++k) {
         geometry::KDTreeFlann kdtree(features[1 - k]);
--- a/cpp/open3d/pipelines/registration/Feature.cpp
+++ b/cpp/open3d/pipelines/registration/Feature.cpp
@@ -154,10 +154,9 @@ CorrespondenceSet CorrespondencesFromFeatures(const Feature &source_features,
     std::vector<CorrespondenceSet> corres(num_searches);
 
     const int kMaxThreads = utility::EstimateMaxThreads();
-
     const int kOuterThreads = std::min(kMaxThreads, num_searches);
     const int kInnerThreads = std::max(kMaxThreads / num_searches, 1);
-    (void)kOuterThreads;
+    (void)kOuterThreads;  // Avoids compiler warning if OpenMP is disabled
     (void)kInnerThreads;
 #pragma omp parallel for num_threads(kOuterThreads)
     for (int k = 0; k < num_searches; ++k) {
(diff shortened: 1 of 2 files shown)


saurabheights:
My C++ is very rusty, so just use these as suggestions (I could be wrong). Ideally, if a variable is not used, it should not be even declared. Furthermore, casting to void is too cryptic. There might be a better alternative where openmp related code is defined in `#ifdef` block - https://stackoverflow.com/questions/24488239/use-variable-from-cmake-in-c

sitic:
Thank you for your suggestions, @saurabheights.

``(void)variable;`` is a common pattern used to suppress the unused variable warnings, which are escalated to errors here. While it’s not the most elegant solution (in C++17 we could use ``[[maybe_unused]]``), it's the most suitable approach in this case, in my opinion. I've added a code comment to clarify their purpose.

Indeed ideally unused variables should not be declared. Typically the number of threads will be computed in the OpenMP specific ``#pragma omp parallel`` line, however, in these two cases, they are separated out into new lines to manage line length. I don’t believe enclosing them in an ``#ifdef`` block is appropriate here, as it would make the code harder to read and more complex. They don’t do any meaningful work if OpenMP is disabled and will be optimized away, so it boils down to a preference in style.

ssheorey (approved):
Thanks @sitic we are planning to switch from C++14 to C++17 soon (some MSVC compatibility issues still remain), but this is a good workaround in the meantime.

PR #6288, opened 2023-07-28
Add support for python3.11

OlivierLDff:
## Type

-   [x] Bug fix (non-breaking change which fixes an issue): Fixes #
-   [ ] New feature (non-breaking change which adds functionality). Resolves #
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected) Resolves #

## Motivation and Context

Fix #5620 

## Checklist:

-   [ ] I have run `python util/check_style.py --apply` to apply Open3D **code style**
    to my code.
-   [ ] This PR changes Open3D behavior or adds new functionality.
    -   [ ] Both C++ (Doxygen) and Python (Sphinx / Google style) **documentation** is
        updated accordingly.
    -   [ ] I have added or updated C++ and / or Python **unit tests** OR included **test
        results** (e.g. screenshots or numbers) here.
-   [x] I will follow up and update the code if CI fails.
    
-   [ ] For fork PRs, I have selected **Allow edits from maintainers**.

## Description

- Update the linux/mac/windows CI to handle python 3.11
- Update pybind11

I guess some documentation will also be required to be updated somewhere. I am not familiar with the workspace, I just happened to have python3.11 on my machine so wanted to give a go to the update. Maybe there more stuff to update than what I did. Let's see what the CI have to say.

Companion Open3D-ML PR: https://github.com/isl-org/Open3D-ML/pull/619

- - -
This change is [Reviewable](https://reviewable.io/reviews/isl-org/Open3D/6288)

PR #6536, opened 2023-12-15
Update supported Python versions (PyPI) in readme

lobis:
Just a small update of the readme, I just tried to install from pip on Python 3.11 but wasn't able to. According to the readme this should be possible.

I looked at the list of wheels (https://pypi.org/project/open3d/#files) and found that they are only provided for Python 3.7-3.10 so I updated the readme with this information.

I also had to install `libomp` in my M1 macbook air (macos) but didn't find it mentioned on the readme, not sure if this was supposed to be packaged with the wheel (I installed the 3.10 one).

Thanks for your work!

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ For more, please visit the [Open3D documentation](http://www.open3d.org/docs).
 ## Python quick start
 
 Pre-built pip packages support Ubuntu 18.04+, macOS 10.15+ and Windows 10+
-(64-bit) with Python 3.8-3.11.
+(64-bit) with Python 3.7-3.10.
 
 ```bash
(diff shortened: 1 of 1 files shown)


johnthagen:
This is because the `main` branch does support 3.11

- #6288

We are waiting for the 0.18 release to be made and published to PyPI.

lobis:
> This is because the `main` branch does support 3.11
> 
> * [Add support for python3.11 #6288](https://github.com/isl-org/Open3D/pull/6288)
> 
> But we are waiting for the 0.18 release to be made and published to PyPI.

Makes sense! Thanks.

PR #6541, opened 2023-12-18
Add env var to disable automatic WebVisualizer use in Jupyter environment

gutnar:
Add checking `DISABLE_OPEN3D_WEB_VISUALIZER` environment variable to disable automatically enabling web visualizer when in a Jupyter environment.

## Type

## Motivation and Context

The problem is that when using the Jupyter extension in Visual Studio Code then the web visualizer does not function. This change allows to set an environment variable so that the web visualizer would not automatically be enabled.

## Checklist:

    to my code.
        updated accordingly.

## Description

When using the `gui.Application` functionality and running it using the Jupyter extension in Visual Studio Code then no visualization appears. When setting the env var `DISABLE_OPEN3D_WEB_VISUALIZER=True` then the WebVisualizer is not enabled and an application window opens.


ssheorey (approved):
Thanks @gutnar for this useful update. I added docs, allowed lower case env var and made OPEN3D the prefix for the env var.

PR #6619, opened 2024-01-22
Add Python pathlib support for file IO

### LINKED ISSUE #6344: Add pathlib support to IO methods in Python
### Checklist

According to my tests with `Open3D 0.17` with `Python 3.10`, this is not yet implemented.

It would be nice if the reading/writing methods in Open3D supported `pathlib` as inputs, since this is the preferred way of dealing with system paths for many people. This way, instead of having to transform the path to a string:
```Python
from pathlib import Path
import open3d as o3d
path = Path('./Data/CAD/example.stl')
mesh = o3d.io.read_triangle_mesh(path.as_posix())
```
Or equivalently:
```Python
mesh = o3d.io.read_triangle_mesh(str(path))
```
We could just directly enter the `Path` object:
```Python
from pathlib import Path
import open3d as o3d
path = Path('./Data/CAD/example.stl')
mesh = o3d.io.read_triangle_mesh(path)
```
This is, of course, very minor and would just add a little convenience to the code. But this would also be pretty straightforward to implement, which is why I think it is worth it.

sitic:
Adds support for `pathlib.Path`/`os.PathLike` type arguments to IO functions. E.g:

```python
from pathlib import Path
import open3d as o3d

folder = Path('.')
mesh = o3d.io.read_triangle_mesh(folder / 'test.ply')
o3d.io. write_triangle_mesh(folder / 'test.ply', mesh)
```

is now supported in addition to `str`-type arguments.

## Type

-   [x] New feature (non-breaking change which adds functionality). Resolves #6344

## Motivation and Context

Currently, only string-type filenames are supported in all IO functions. Python's `pathlib` is quite popular and should be supported. This commit adds support for all libraries which are compatible with `os.PathLike` as `pathlib` is.

## Checklist:

    to my code.
        updated accordingly.
    
## Description

Using `std::filesystem::path` as type for filename arguments allows pybind to support both `pathlib.Path` and `string` through [<pybind11/stl/filesystem.h>](https://github.com/pybind/pybind11/blob/master/include/pybind11/stl/filesystem.h) (which I backported to C++14 with minimal changes). Unfortunately this means we have to call `filename.string()` in the wrappers when calling the open3d C++ code.

I've added support everywhere except for `o3d.data` classes and some `o3d.t.io.*Sensor` class functions.

The documentation shows will _os.PathLike_ as type for filenames:

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@
 -   Fix geometry picker Error when LineSet objects are presented (PR #6499)
 -   Fix mis-configured application .desktop link for the Open3D viewer when installing to a custom path (PR #6599)
 -   Fix regression in printing cuda tensor from PR #6444 🐛
+-   Add Python pathlib support for file IO (PR #6619)
 
 ## 0.13
 
--- a/cpp/pybind/io/class_io.cpp
+++ b/cpp/pybind/io/class_io.cpp
@@ -105,10 +105,10 @@ void pybind_class_io(py::module &m_io) {
     // open3d::geometry::Image
     m_io.def(
             "read_image",
-            [](const std::string &filename) {
+            [](const fs::path &filename) {
                 py::gil_scoped_release release;
                 geometry::Image image;
-                ReadImage(filename, image);
+                ReadImage(filename.string(), image);
                 return image;
             },
             "Function to read Image from file", "filename"_a);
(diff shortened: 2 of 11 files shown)


ssheorey (approved):
Thanks very much @sitic . This will PR really makes using Open3D easier!
PS: Regarding C++17 - we are ready to switch to that. I think the only issue was MSVC errors when built in C++17 mode but they go away in c++-latest mode. Let's merge this PR with C++14 in mind, but OK to use C++17 in the future. We can help fixing any build issues.

Kroppeb:
I love you. This is awesome!

PR #6648, opened 2024-02-10
VTK based implementation of extract trianglemesh

intelshashi:
## Type

## Checklist:

    to my code.
        updated accordingly.
    

cpp/open3d/pipelines/integration/UniformTSDFVolume.h
```diff
@@ -52,6 +52,8 @@ class UniformTSDFVolume : public TSDFVolume {
                    const Eigen::Matrix4d &extrinsic) override;
     std::shared_ptr<geometry::PointCloud> ExtractPointCloud() override;
     std::shared_ptr<geometry::TriangleMesh> ExtractTriangleMesh() override;
+    /// Temporary function will be deleted/"merged" with ExtractTriangleMesh
+    std::shared_ptr<geometry::TriangleMesh> ExtractTriangleMesh_v2();
```
theNded, line 56: It would be better to inject the implementation as an option in `ExtractTriangleMesh(bool use_vtk = False)`.

cpp/pybind/pipelines/integration/integration.cpp
```diff
@@ -136,6 +136,8 @@ In SIGGRAPH, 1996)");
                  "Debug function to inject the voxel TSDF data.", "tsdf"_a)
             .def("inject_volume_color", &UniformTSDFVolume::InjectVolumeColor,
                  "Debug function to inject the voxel Color data.", "color"_a)
+            // TODO remove after debugging
+            .def("extract_triangle_mesh_v2", &UniformTSDFVolume::ExtractTriangleMesh_v2)
```
theNded, line 140: Same, do not duplicate API.

cpp/open3d/pipelines/integration/UniformTSDFVolume.cpp
```diff
@@ -138,6 +154,64 @@ std::shared_ptr<geometry::PointCloud> UniformTSDFVolume::ExtractPointCloud() {
...
+#else
+    vtkNew<vtkMarchingCubes> surface;
+#endif
+    surface->SetInputData(imageData);
+    surface->ComputeNormalsOn();
+    const double isoValue = 0.0;
+    surface->SetValue(0, isoValue);
+
+    //// To remain largest region.
+    //// vtkNew<vtkPolyDataConnectivityFilter> confilter;
```
theNded, line 202: Either remove unused code or add a flag to run the code.

cpp/open3d/pipelines/integration/UniformTSDFVolume.cpp
```diff
@@ -138,6 +154,64 @@ std::shared_ptr<geometry::PointCloud> UniformTSDFVolume::ExtractPointCloud() {
     return pointcloud;
 }
 
+std::shared_ptr<geometry::TriangleMesh>
+UniformTSDFVolume::ExtractTriangleMesh_v2() {
+    auto mesh = std::make_shared<geometry::TriangleMesh>();
+    //float *image_array = new float[resolution_  * resolution_ * resolution_];
```
theNded, line 160: Remove unused comments.

cpp/open3d/pipelines/integration/UniformTSDFVolume.cpp
```diff
@@ -138,6 +154,64 @@ std::shared_ptr<geometry::PointCloud> UniformTSDFVolume::ExtractPointCloud() {
...
+    //const long r_sq = resolution_ * resolution_;
+    int dims[] = {resolution_, resolution_, resolution_};
```
theNded, line 162: Use std::array or modern containers, or explicitly declare array size.

cpp/open3d/pipelines/integration/UniformTSDFVolume.cpp
```diff
@@ -138,6 +154,64 @@ std::shared_ptr<geometry::PointCloud> UniformTSDFVolume::ExtractPointCloud() {
...
+    float ***values = new float **[dims[0]];
```
theNded, line 163: All pointers are allocated but not released. It's better to allocate a flattened std::vector and unroll the indices.

cpp/open3d/pipelines/integration/UniformTSDFVolume.cpp
```diff
@@ -138,6 +154,64 @@ std::shared_ptr<geometry::PointCloud> UniformTSDFVolume::ExtractPointCloud() {
...
+    for (int x = 0; x < resolution_ - 1; x++) {
+        for (int y = 0; y < resolution_ - 1; y++) {
+            for (int z = 0; z < resolution_ - 1; z++) {
+                Eigen::Vector3i idx0(x, y, z);
+                const float f = voxels_[IndexOf(idx0)].tsdf_;
+                values[x][y][z] = f;
+            }
+        }
+    }
+    vtkIdType totalSize = (vtkIdType)dims[0] * (vtkIdType) dims[1] * (vtkIdType) dims[2];
```
theNded, line 180: Use static_cast.

cpp/open3d/pipelines/integration/UniformTSDFVolume.cpp
```diff
@@ -138,6 +154,64 @@ std::shared_ptr<geometry::PointCloud> UniformTSDFVolume::ExtractPointCloud() {
...
+            values[i][j] = new float[dims[2]];
+        }
+    }
+
+    for (int x = 0; x < resolution_ - 1; x++) {
+        for (int y = 0; y < resolution_ - 1; y++) {
+            for (int z = 0; z < resolution_ - 1; z++) {
+                Eigen::Vector3i idx0(x, y, z);
+                const float f = voxels_[IndexOf(idx0)].tsdf_;
+                values[x][y][z] = f;
```
theNded, line 176: Only TSDF is used here, weights are not handled, see the [original implementation](https://github.com/isl-org/Open3D/blob/f8c94a1993d08cfa7ec974a76dea8b61c7304946/cpp/open3d/pipelines/integration/UniformTSDFVolume.cpp#L128). This is known to provide problematic mesh especially around boundaries.

cpp/open3d/pipelines/integration/UniformTSDFVolume.cpp
```diff
@@ -138,6 +154,64 @@ std::shared_ptr<geometry::PointCloud> UniformTSDFVolume::ExtractPointCloud() {
...
+    //// To remain largest region.
+    //// vtkNew<vtkPolyDataConnectivityFilter> confilter;
+    //// confilter->SetInputConnection(surface->GetOutputPort());
+    //// confilter->SetExtractionModeToLargestRegion();
+    //// vtkPolyData *pdata = confilter->GetOutput();
+    //// cout << "pdata: " << pdata << std::endl;
+    vtkPolyData *sdata = surface->GetOutput();
+    auto tmesh = open3d::t::geometry::vtkutils::CreateTriangleMeshFromVtkPolyData(sdata);
+    auto lmesh = tmesh.ToLegacy();
+    std::shared_ptr<geometry::TriangleMesh> pmesh(new geometry::TriangleMesh(lmesh.vertices_, lmesh.triangles_));
```
theNded, line 210: Memory is copied twice in the conversion. It is also unsafe to use new to initialize a shared_ptr. Please use std::make_shared and reduce redundant memory copy.

theNded (changes requested):
Thanks for the PR! Please kindly check my comments. The PR also needs unit tests and benchmarking. 
VTK version is more limited (does not allow weights and the perturbation of isosurface). Additional overhead is added through memory copy from TSDFVolume to VTK and tensor mesh to legacy mesh (they can be potentially optimized). 
I think it would be fine to merge the PR if we can justify that the VTK version is faster. Otherwise, it might be good to rethink the necessity.

python/test/t/geometry/test_trianglemesh.py
```diff
@@ -287,6 +287,16 @@ def test_create_text():
     assert mesh.triangle.indices.shape == (936, 3)
 
 
+def test_create_from_volume():
+    coords = np.stack(np.meshgrid(*3 * [np.linspace(-1, 1, num=64)],
```
ssheorey, line 291:
```py
"""Create signed distance field for sphere of radius 0.5 and extract sphere from it."""
```

cpp/open3d/t/geometry/TriangleMesh.h
```diff
@@ -594,6 +594,18 @@ class TriangleMesh : public Geometry, public DrawableGeometry {
...
             const core::Device &device = core::Device("CPU:0"));
 
+    /// Create a mesh from a scalar volume by computing the isocontour.
+    /// This method uses the Flying Edges dual contouring method that similarly
+    /// like Marching Cubes computes the isocontour.
+    /// \param volume 3D tensor with the volume.
+    /// \param contour_values A list of contour values at which isocontours will
+    /// be generated.The values used for contouring. The default \param device
+    /// The device for the returned mesh.
+    static TriangleMesh CreateFromVolume(
```
ssheorey, line 604: Not sure if `CreateFromVolume` is easily understood. How about using `CreateIsoSurfaces` instead? The VTK blog and paper refers to isocontours and 3D image instead of volume, while wikipedia article is about isosurfaces.  Matlab and other sources seem to prefer iso surfaces. 3D scalar field is perhaps more clear than volume.

cpp/pybind/t/geometry/trianglemesh.cpp
```diff
@@ -519,6 +519,45 @@ This example shows how to create a hemisphere from a sphere::
 
         mesh = o3d.t.geometry.TriangleMesh.create_text('Open3D', depth=1)
         o3d.visualization.draw([{'name': 'text', 'geometry': mesh}])
+)");
+
+    triangle_mesh.def_static(
+            "create_from_volume",
```
ssheorey, line 525: Function name (see comment above for C++ function)

cpp/open3d/t/geometry/VtkUtils.cpp
```diff
@@ -182,6 +183,32 @@ static vtkSmartPointer<vtkPoints> CreateVtkPointsFromTensor(
...
+    if (tensor.NumDims() != 2 && tensor.NumDims() != 3) {
+        utility::LogError(
+                "Cannot convert Tensor to vtkImageData. The number of "
+                "dimensions must be 2 or 3 but is {}",
+                tensor.NumDims());
+    }
+
+    // Create a flat tensor that can be converted to a vtkDataArray
+    auto tensor_flat = tensor.Reshape({tensor.NumElements(), 1});
+    copy = copy || tensor.GetDataPtr() != tensor_flat.GetDataPtr();
```
ssheorey, line 199: clarity: `copy ||= tensor.GetDataPtr() != tensor_flat.GetDataPtr();`
or even:  `if (tensor.GetDataPtr() != tensor_flat.GetDataPtr()) {copy = true;}`

cpp/open3d/t/geometry/VtkUtils.cpp
```diff
@@ -213,7 +240,7 @@ static vtkSmartPointer<vtkCellArray> CreateVtkCellArrayFromTensor(
     const int cell_size = tensor.GetShape()[1];
 
     auto tensor_flat = tensor.Reshape({tensor.NumElements(), 1}).Contiguous();
-    copy = copy && tensor.GetDataPtr() == tensor_flat.GetDataPtr();
+    copy = copy || tensor.GetDataPtr() != tensor_flat.GetDataPtr();
```
ssheorey, line 243: clarity: `copy ||= tensor.GetDataPtr() != tensor_flat.GetDataPtr();`
or even:  `if (tensor.GetDataPtr() != tensor_flat.GetDataPtr()) {copy = true;}`

ssheorey (changes requested):
This implementation is not linked to the voxel grid / voxel block grid anymore. An example of how to use it with data in those formats would make it more accessible.

cpp/open3d/t/geometry/VtkUtils.cpp
benjaminum: done

cpp/open3d/t/geometry/VtkUtils.cpp
benjaminum: done

cpp/open3d/t/geometry/TriangleMesh.h
benjaminum: changed function name to CreateIsosurfaces/create_isosurfaces

python/test/t/geometry/test_trianglemesh.py
benjaminum: Only needing a 3D Tensor or numpy array will make using this function straightforward. Maybe we can think of how to include VoxelGrid in another PR?

ssheorey: Sounds good. The intention was that the current workflow creates a voxel grid through RGBD integration. We can add support for using this in RGBD integration later.

PR #6653, opened 2024-02-12
Fix TriangleMesh::SamplePointsUniformly not sampling triangles unifor…

### LINKED ISSUE #6144: TriangleMesh::SamplePointsUniformly doesn't sample triangles uniformly
### Checklist

### Describe the issue

When using TriangleMesh::SamplePointsUniformly(), the function does not actually sample the triangles uniform randomly. It calculates the number of points per triangle based on the surface area and the number of points. And then for each triangle, the corresponding number of points is sampled uniformly from the triangle. So before even doing any random sampling, the number of points per triangle is deterministically calculated based on surface area (and the ordering of triangles in the mesh). This means that with repeated calls to SamplePointsUniformly(), the same number of points will be sampled on each triangle every time, which is inherently non-random. For a larger enough number of sampled points, this is an okay approximation. But for a small number of points it becomes problematic. One major consequence of this is that if the number of points is too small, certain triangles can never be sampled. But the general problem is that triangles do not actually get sampled exactly proportional to their area, and this is amplified significantly for a small number of sample points. 

Is this the intended behavior? 

```c++
auto sphere = open3d::geometry::TriangleMesh::CreateSphere();
    sphere->ComputeTriangleNormals();
    bool use_triangle_normals = true;
    for (int i = 0; i != 10; i++){
        auto sampled_pts = sphere->SamplePointsUniformly(1, use_triangle_normals);
        for (int j = 0; j != sampled_pts->points_.size(); j++){
            std::cout << "Pt: " << sampled_pts->points_[j].transpose() << " normal: " << sampled_pts->normals_[j].transpose() << std::endl;
        }   
    }
```
Sample output:

Pt:  0.881371 -0.464317  0.035942 normal:  0.850028 -0.520898 0.0782187
Pt:  0.825666 -0.558827 0.0119147 normal:  0.850028 -0.520898 0.0782187
Pt:  0.836437 -0.534213 0.0587785 normal:  0.850028 -0.520898 0.0782187
Pt:  0.886223 -0.453663 0.0541698 normal:  0.850028 -0.520898 0.0782187
Pt:  0.887608 -0.454505 0.0335078 normal:  0.850028 -0.520898 0.0782187
Pt:  0.852025 -0.514956 0.0176215 normal:  0.850028 -0.520898 0.0782187
Pt:  0.873439 -0.475443 0.0480524 normal:  0.850028 -0.520898 0.0782187
Pt:  0.877829 -0.462695 0.0852355 normal:  0.850028 -0.520898 0.0782187
Pt:   0.86169 -0.487152 0.0977552 normal:  0.850028 -0.520898 0.0782187
Pt:  0.879574 -0.460373 0.0817351 normal:  0.850028 -0.520898 0.0782187

### Error message

_No response_

### Expected behavior

The above code sample is the extreme example where only 1 point is sampled with each call to SamplePointsUniformly(). I would expect to see different triangles sampled each time SamplePointsUniformly() is called (i.e. the normal direction should change with each call). But what is happening is that the exact same triangle is sampled each time (so the normal direction is the same each time). Basically the cdf for the triangles is calculated deterministically and as a result the only triangle ever sampled will be the first triangle with a cdf > 0.5. And all other triangles will never be sampled. 

```markdown
- Operating system: Ubuntu 22.04
- Python version: N/A (C++)
- Open3D version: v0.17
- System architecture: x86
- How did you install Open3D?: released binary v0.17
- Compiler version (if built from source): N/A
```
### Additional information

If this is not the intended behavior, my proposal would be to sample a triangle for each point according to the cdf instead of calculating the number of points for each triangle deterministically. The downside of doing it this way is that instead of it being O(N+M) like it is now, it would be O(NlogM+M). Where N is the number of points and M is the number of triangles in the mesh. If this is not desirable, another option would be to shuffle the triangle surface areas (keeping track of the shuffled indices) before calculating the cdf. This would not be exactly "uniform" proportional to the triangle surface areas, but it would certainly be closer to uniform random than doing it deterministically. Shuffling the triangle surface areas would keep the complexity O(N+M), which is a benefit over the "correct" way which would be O(NlogM+M).

iyadahmed:
…mly (#6144)

Better uniform sampling of triangle meshes

## Type

-   [x] Bug fix (non-breaking change which fixes an issue): Fixes #6144 

## Motivation and Context

A problem was reported in the issue of uniform sampling not being really random nor uniform, as it iterated triangles deterministically and resulted in same triangle being picked many times in a row.

This change improves the situation by using [`std::discrete_distribution`](https://en.cppreference.com/w/cpp/numeric/random/discrete_distribution)

## Checklist:

    to my code.
        updated accordingly.
        **It seems that the random generators are not exposed to Python? so Python docs were not updated only in source C++ docs (hopefully I did not miss something)**
    -   [ ] I have added or updated C++ and / or Python **unit tests** OR included **This also was not done but no tests exist for other random generators as well right?**
In case I'm unavailable later anyone is welcome to modify the PR, since most of the heavy lifting is done inside `std::discrete_distribution` anyways, any edits will hopefully be easy, also hopefully CI won't fail the changes are very non intrusive

## Description

Sample code based on original issue author
```cpp
#include <iostream>
#include <open3d/Open3D.h>
int main() {
  auto sphere = open3d::geometry::TriangleMesh::CreateSphere();
  sphere->ComputeTriangleNormals();
  bool use_triangle_normals = true;
  for (int i = 0; i != 10; i++) {
    auto sampled_pts = sphere->SamplePointsUniformly(1, use_triangle_normals);
    for (int j = 0; j != sampled_pts->points_.size(); j++) {
      std::cout << "Pt: " << sampled_pts->points_[j].transpose()
                << " normal: " << sampled_pts->normals_[j].transpose()
                << std::endl;
    }
  }
  return 0;
}
```

Original result reported in issue:
```
Pt: 0.881371 -0.464317 0.035942 normal: 0.850028 -0.520898 0.0782187
Pt: 0.825666 -0.558827 0.0119147 normal: 0.850028 -0.520898 0.0782187
Pt: 0.836437 -0.534213 0.0587785 normal: 0.850028 -0.520898 0.0782187
Pt: 0.886223 -0.453663 0.0541698 normal: 0.850028 -0.520898 0.0782187
Pt: 0.887608 -0.454505 0.0335078 normal: 0.850028 -0.520898 0.0782187
Pt: 0.852025 -0.514956 0.0176215 normal: 0.850028 -0.520898 0.0782187
Pt: 0.873439 -0.475443 0.0480524 normal: 0.850028 -0.520898 0.0782187
Pt: 0.877829 -0.462695 0.0852355 normal: 0.850028 -0.520898 0.0782187
Pt: 0.86169 -0.487152 0.0977552 normal: 0.850028 -0.520898 0.0782187
Pt: 0.879574 -0.460373 0.0817351 normal: 0.850028 -0.520898 0.0782187
```

Example result after using `std::discrete_distribution` (note the result is random and will change every time we run, but you can see indeed it is much better now, much better than picking the same triangle every time, truly more random and uniform in terms of unit area on a mesh)
```
Pt: -0.643155   -0.5143    0.5607 normal: -0.648898 -0.554211  0.521326
Pt:  0.613348 -0.783479 0.0359056 normal:  0.647458 -0.758076 0.0782187
Pt:  0.785933  0.454903 -0.406661 normal:  0.788092  0.482943 -0.381676
Pt:   0.759766 -0.0684343  -0.639144 normal:   0.759048 -0.0597384  -0.648288
Pt: -0.706693  -0.20519  0.671707 normal: -0.740358 -0.177744  0.648288
Pt: 0.283229 0.133719 0.949072 normal: 0.354486 0.146833 0.923461
Pt: -0.00406476    0.816619    0.574857 normal: -0.0669537   0.850727   0.521326
Pt: -0.253326 -0.791642 -0.551146 normal: -0.199212 -0.829779 -0.521326
Pt:  0.536542 -0.374171   0.75161 normal:  0.554734 -0.339941  0.759415
Pt: -0.858788  0.159791  0.480014 normal: -0.829779  0.199212  0.521326
````

### DIFF
--- a/cpp/open3d/geometry/TriangleMesh.cpp
+++ b/cpp/open3d/geometry/TriangleMesh.cpp
@@ -433,20 +433,10 @@ std::shared_ptr<TriangleMesh> TriangleMesh::FilterSmoothTaubin(
 
 std::shared_ptr<PointCloud> TriangleMesh::SamplePointsUniformlyImpl(
         size_t number_of_points,
-        std::vector<double> &triangle_areas,
-        double surface_area,
+        const std::vector<double> &triangle_areas,
         bool use_triangle_normal) {
-    if (surface_area <= 0) {
-        utility::LogError("Invalid surface area {}, it must be > 0.",
-                          surface_area);
-    }
-
-    // triangle areas to cdf
-    triangle_areas[0] /= surface_area;
-    for (size_t tidx = 1; tidx < triangles_.size(); ++tidx) {
-        triangle_areas[tidx] =
-                triangle_areas[tidx] / surface_area + triangle_areas[tidx - 1];
-    }
+    utility::random::DiscreteGenerator<size_t> triangle_index_generator(
+            triangle_areas.begin(), triangle_areas.end());
 
     // sample point cloud
     bool has_vert_normal = HasVertexNormals();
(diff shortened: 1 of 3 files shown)


iyadahmed:
I found a potential issue with using `std::discrete_distribution` it might affect other calls to other distributions (the `uniform_real_distribution` used to sampling points in a triangle)

benjaminum:
> I found a potential issue with using `std::discrete_distribution` it might affect other calls to other distributions (the `uniform_real_distribution` used to sampling points in a triangle)
> 
> Edit: I found this, so I guess maybe it is a separate issue to use different engines for different distributions? https://stackoverflow.com/a/18081251/8094047
> 
> or maybe only this bug happens in my program, maybe open3d seeding avoid this
> 
> Edit2: from further investigation in my own program, it seems that it will likely be an issue for Open3D as well (random values from different distribution objects being correlated because they share the random engine), but it is a separate issue than this, so I guess I should not include it in this pr

Yes I think investigating/fixing this is another PR

benjaminum:
Running `mesh.sample_points_uniformly(1)` for 1000 times and collecting the sampled points yields the following.
Old behavior 
image
New behavior
image

benjaminum (approved):

PR #6657, opened 2024-02-15
Implement ComputeTriangleAreas, GetNonManifoldEdges and RemoveNonManifoldEdges in t::geometry::TriangleMesh

nsaiapova:
Implement a few methods which are available with the geometry::TriangleMesh API.

## Type

## Checklist:

    to my code.
        updated accordingly.
    
## Description

In this PR three methods for t::geometry::TriangleMesh are implemented.

**t::geometry::TriangleMesh::ComputeTriangleAreas**
Split the logic from ComputeSurfaceArea into a helper static function
and introduce a new method which computes triangle areas and writes the
resulting tensor into attributes of the mesh.
IIUC, in geometry::TriangleMesh the corresponding function is GetSurfaceArea(std::vector &areas), where the areas are filled with the triangle areas.

**t::geometry::TriangleMesh::GetNonManifoldEdges** mimics the logic of the legacy method.

**t::geometry::TriangleMesh::RemoveNonManifoldEdges** follows the logic of
the legacy method but there are a few differences:
* the main difference is that the outer while-loop is removed. I don't
  see how after the first iteration any edge can have more than 2
  adjucent triangles, which makes the further iterations unnecessary.
* I count triangles with non-negative areas immediately and do not rely
  on the total number of adjacent triangles (which would also include
  triangles marked for removal).
* To choose a triangle with the minimal area out of the existing
  adjacent triangles I use a heap structure.

### DIFF
--- a/cpp/open3d/t/geometry/TriangleMesh.cpp
+++ b/cpp/open3d/t/geometry/TriangleMesh.cpp
@@ -309,6 +309,9 @@ TriangleMesh &TriangleMesh::ComputeTriangleAreas() {
     }
 
     if (!HasTriangleIndices()) {
+        SetTriangleAttr("areas", core::Tensor::Empty(
+                                         {0}, GetVertexPositions().GetDtype(),
+                                         GetDevice()));
         utility::LogWarning("TriangleMesh has no triangle indices.");
         return *this;
     }
--- a/cpp/open3d/t/geometry/TriangleMesh.cpp
+++ b/cpp/open3d/t/geometry/TriangleMesh.cpp
@@ -283,6 +283,42 @@ TriangleMesh &TriangleMesh::ComputeVertexNormals(bool normalized) {
     return *this;
 }
 
+static core::Tensor ComputeTriangleAreasHelper(const TriangleMesh &mesh) {
+    const int64_t triangle_num = mesh.GetTriangleIndices().GetLength();
+    const core::Dtype dtype = mesh.GetVertexPositions().GetDtype();
+    core::Tensor triangle_areas({triangle_num}, dtype, mesh.GetDevice());
+    if (mesh.IsCPU()) {
+        kernel::trianglemesh::ComputeTriangleAreasCPU(
+                mesh.GetVertexPositions().Contiguous(),
+                mesh.GetTriangleIndices().Contiguous(), triangle_areas);
+    } else if (mesh.IsCUDA()) {
+        CUDA_CALL(kernel::trianglemesh::ComputeTriangleAreasCUDA,
+                  mesh.GetVertexPositions().Contiguous(),
+                  mesh.GetTriangleIndices().Contiguous(), triangle_areas);
+    } else {
+        utility::LogError("Unimplemented device");
+    }
+
+    return triangle_areas;
+}
+
+TriangleMesh &TriangleMesh::ComputeTriangleAreas() {
+    if (IsEmpty()) {
+        utility::LogWarning("TriangleMesh is empty.");
+        return *this;
+    }
+
+    if (!HasTriangleIndices()) {
+        utility::LogWarning("TriangleMesh has no triangle indices.");
+        return *this;
+    }
+
+    core::Tensor triangle_areas = ComputeTriangleAreasHelper(*this);
+    SetTriangleAttr("areas", triangle_areas);
+
+    return *this;
+}
+
 double TriangleMesh::GetSurfaceArea() const {
     double surface_area = 0;
     if (IsEmpty()) {
@@ -295,22 +331,7 @@ double TriangleMesh::GetSurfaceArea() const {
         return surface_area;
     }
 
-    const int64_t triangle_num = GetTriangleIndices().GetLength();
-    const core::Dtype dtype = GetVertexPositions().GetDtype();
-    core::Tensor triangle_areas({triangle_num}, dtype, GetDevice());
-
-    if (IsCPU()) {
-        kernel::trianglemesh::ComputeTriangleAreasCPU(
-                GetVertexPositions().Contiguous(),
-                GetTriangleIndices().Contiguous(), triangle_areas);
-    } else if (IsCUDA()) {
-        CUDA_CALL(kernel::trianglemesh::ComputeTriangleAreasCUDA,
-                  GetVertexPositions().Contiguous(),
-                  GetTriangleIndices().Contiguous(), triangle_areas);
-    } else {
-        utility::LogError("Unimplemented device");
-    }
-
+    core::Tensor triangle_areas = ComputeTriangleAreasHelper(*this);
     surface_area = triangle_areas.Sum({0}).To(core::Float64).Item<double>();
 
     return surface_area;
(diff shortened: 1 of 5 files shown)


cpp/pybind/t/geometry/trianglemesh.cpp
```diff
@@ -962,6 +962,18 @@ or has a negative value, it is ignored.
         box = o3d.t.geometry.TriangleMesh.create_box()
         top_face = box.select_by_index([2, 3, 6, 7])
 )");
+    triangle_mesh.def("remove_non_manifold_edges",
```
benjaminum, line 965: Please add docstrings for python in the [google format](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) similar to other functions in this file. Examples are optional but very welcome 😄

benjaminum (changes requested):

cpp/open3d/t/geometry/TriangleMesh.cpp
```diff
@@ -1252,6 +1273,175 @@ TriangleMesh TriangleMesh::SelectByIndex(const core::Tensor &indices) const {
...
+    DISPATCH_INT_DTYPE_PREFIX_TO_TEMPLATE(tri_dtype, tris, [&]() {
+        auto edges = GetEdgeToTrianglesMap<scalar_tris_t>(tris_cpu);
+        std::vector<scalar_tris_t> non_manifold_edges;
+
+        for (auto &kv : edges) {
+            if ((allow_boundary_edges &&
+                 (kv.second.size() < 1 || kv.second.size() > 2)) ||
+                (!allow_boundary_edges && kv.second.size() != 2)) {
+                non_manifold_edges.push_back(std::get<0>(kv.first));
+                non_manifold_edges.push_back(std::get<1>(kv.first));
```
ssheorey, line 1433: nit: can do a single operation: `non_manifold_edges.insert(non_manifold_edges.end(), std::begin(kv.first), std::end(kv.first));`

cpp/open3d/t/geometry/TriangleMesh.cpp
```diff
@@ -1252,6 +1273,175 @@ TriangleMesh TriangleMesh::SelectByIndex(const core::Tensor &indices) const {
...
+    core::Tensor tri_areas_cpu =
+            GetTriangleAttr("areas").To(core::Device()).Contiguous();
+
+    DISPATCH_FLOAT_INT_DTYPE_TO_TEMPLATE(
+            GetVertexPositions().GetDtype(), tris_cpu.GetDtype(), [&]() {
+                scalar_t *tri_areas_ptr = tri_areas_cpu.GetDataPtr<scalar_t>();
+                auto edges_to_tris = GetEdgeToTrianglesMap<int_t>(tris_cpu);
+
+                // lambda to compare triangles areas by index
+                auto area_greater_compare = [=](size_t lhs, size_t rhs) {
```
ssheorey, line 1346: We only need to capture `tri_areas_ptr`:
`[=tri_areas_ptr](...) {...}`

cpp/open3d/t/geometry/TriangleMesh.cpp
```diff
@@ -1252,6 +1273,175 @@ TriangleMesh TriangleMesh::SelectByIndex(const core::Tensor &indices) const {
...
+                    return tri_areas_ptr[lhs] > tri_areas_ptr[rhs];
+                };
+
+                // go through all edges and for those who has more than 2
```
ssheorey, line 1350: typo: `...for those that have more...`

cpp/open3d/t/geometry/TriangleMesh.cpp
```diff
@@ -1252,6 +1273,175 @@ TriangleMesh TriangleMesh::SelectByIndex(const core::Tensor &indices) const {
...
+
+                    // now erase all triangle indices already marked for removal
+                    kv.second.erase(tris_end, kv.second.end());
+
+                    // the number of triangles we need to delete to make the
+                    // edge manifold
+                    int n_tris_to_delete = n_tris - 2;
+
+                    // create min-area heap
+                    std::make_heap(kv.second.begin(), kv.second.end(),
```
ssheorey, line 1380: `std::nth_element` implements selection algorithm (i.e. select largest 2 elements in range) and we can use it directly.

ssheorey: We should be able to directly use `std::nt_element` after `std::remove_if` and simplify the code.

cpp/open3d/t/geometry/TriangleMesh.cpp
```diff
@@ -1252,6 +1273,175 @@ TriangleMesh TriangleMesh::SelectByIndex(const core::Tensor &indices) const {
...
+    }
+    return tris_per_edge;
+}
+
+TriangleMesh TriangleMesh::RemoveNonManifoldEdges() {
+    if (HasTriangleAttr("texture_uvs")) {
+        utility::LogWarning(
+                "[RemoveNonManifoldEdges] This mesh contains triangle uvs that "
+                "are not handled in this function");
+    }
```
ssheorey, line 1316: I think we should be able to support triangle uvs here. Perhaps just copy triangle uvs for the triangles that are retained?

ssheorey (commented):
Thanks @nsaiapova! Some minor comments. Feel free to ignore comments labeled with nit. We can chat about the algorithm....

cpp/pybind/t/geometry/trianglemesh.cpp
```diff
@@ -962,10 +962,23 @@ or has a negative value, it is ignored.
...
+                      "allow_boundary_edges"_a = true,
+                      "Return the list consisting of non-manifold edges.");
 
     triangle_mesh.def("remove_unreferenced_vertices",
                       &TriangleMesh::RemoveUnreferencedVertices,
                       "Removes unreferenced vertices from the mesh in-place.");
+
+    triangle_mesh.def("compute_triangle_areas",
+                      &TriangleMesh::ComputeTriangleAreas,
+                      "Compute triangle areas and save it as \"areas\" "
```
benjaminum, line 980: Just a suggestion. An example cold show how to access the new attribute and sum it to get the total surface area.

python/test/t/geometry/test_trianglemesh.py
```diff
@@ -709,3 +709,63 @@ def check_remove_unreferenced_vertices(device, int_t, float_t):
 def test_remove_unreferenced_vertices(device, int_t, float_t):
     check_no_unreferenced_vertices(device, int_t, float_t)
     check_remove_unreferenced_vertices(device, int_t, float_t)
+
+
+@pytest.mark.parametrize("device", list_devices())
+@pytest.mark.parametrize("int_t", (o3c.int32, o3c.int64))
+@pytest.mark.parametrize("float_t", (o3c.float32, o3c.float64))
+def test_compute_triangle_areas(device, int_t, float_t):
+    torus = o3d.t.geometry.TriangleMesh.create_torus(2, 1, 6, 3, float_t, int_t,
```
benjaminum, line 718: Thanks for adding more unit tests

ssheorey:
Hi @nsaiapova can you take a look at the failing CI checks?
Looks like we are almost ready to merge after that.

ssheorey (approved):
Thanks @nsaiapova ! Looks good!

PR #6672, opened 2024-02-29
Update behavior of FromLegacy to make more flexible with materials

errissa:
## Type

## Motivation and Context

Current behavior of `FromLegacy` on a legacy `TriangleMesh` with more than one material is to warn the user that Tensor-based meshes don't support multiple materials and not convert _any_ materials. Annoyingly, sometimes our importers (mostly ASSIMP) import or create dummy materials so converting those to Tensor-based `TriangleMesh` result in meshes with no materials. With this PR those cases will convert to a Tensor-based `TriangleMesh` with the first material converted as well. In most cases this should produce a mesh with a usable material. In any case, a message is output to the user warning that only the first material was converted.

## Checklist:

    to my code.
        updated accordingly.
    
### DIFF
--- a/cpp/open3d/t/geometry/TriangleMesh.cpp
+++ b/cpp/open3d/t/geometry/TriangleMesh.cpp
@@ -365,7 +365,7 @@ geometry::TriangleMesh TriangleMesh::FromLegacy(
                         .Reshape({-1, 3, 2}));
     }
 
-    // Convert material first material if legacy mesh has one
+    // Convert first material only if one or more are present
     if (mesh_legacy.materials_.size() > 0) {
         const auto &mat = mesh_legacy.materials_.begin()->second;
         auto &tmat = mesh.GetMaterial();
@@ -397,7 +397,8 @@ geometry::TriangleMesh TriangleMesh::FromLegacy(
     if (mesh_legacy.materials_.size() > 1) {
         utility::LogWarning(
                 "Legacy mesh has more than 1 material which is not supported "
-                "by Tensor-based mesh. Only first material was converted.");
+                "by Tensor-based mesh. Only material {} was converted.",
+                mesh_legacy.materials_.begin()->first);
     }
     return mesh;
 }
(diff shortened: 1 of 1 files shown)


cpp/open3d/t/geometry/TriangleMesh.cpp
```diff
@@ -365,8 +365,8 @@ geometry::TriangleMesh TriangleMesh::FromLegacy(
                         .Reshape({-1, 3, 2}));
     }
 
-    // Convert material if legacy mesh only has one
-    if (mesh_legacy.materials_.size() == 1) {
+    // Convert material first material if legacy mesh has one
```
ssheorey, line 368: typo in comment

cpp/open3d/t/geometry/TriangleMesh.cpp
```diff
@@ -393,10 +393,11 @@ geometry::TriangleMesh TriangleMesh::FromLegacy(
                     Image::FromLegacy(*mat.clearCoatRoughness));
         if (mat.anisotropy)
             tmat.SetAnisotropyMap(Image::FromLegacy(*mat.anisotropy));
-    } else if (mesh_legacy.materials_.size() > 1) {
+    }
+    if (mesh_legacy.materials_.size() > 1) {
         utility::LogWarning(
                 "Legacy mesh has more than 1 material which is not supported "
-                "by Tensor-based meshes.");
+                "by Tensor-based mesh. Only first material was converted.");
```
ssheorey, line 400: If available, print a description of material that was converted.

ssheorey (approved):
Thanks @errissa ! Looks good with minor nits.

cpp/open3d/t/geometry/TriangleMesh.cpp
errissa: Done

cpp/open3d/t/geometry/TriangleMesh.cpp
errissa: Done. Now outputting name of material converted.

PR #6676, opened 2024-03-03
Improved performance of function geometry::PointCloud::RemoveRadiusOutliers, function geometry::ClusterDBSCAN and feature counting

PiatrouskiIM:
## Type

## Motivation and Context

Performance improvements.

## Checklist:

    to my code.
        updated accordingly.
    
## Description

1. Determining the presence of a sufficient number of neighbors in a given radius in the geometry::PointCloud::RemoveRadiusOutliers function can be carried out by the SearchHybrid function instead of the SearchRadius function.
2.0 When clustering in PointCloud::ClusterDBSCAN, a set of visited points is created for each new cluster. Elements are added to this set based on the presence of a certain label. This set is subsequently used to add elements to the set of unvisited points. The same thing can be done much more efficiently based only on labels. 
2.1 And there is also no need to represent the collection of unvisited points as an unordered set, given that the only difference is the additional protection against duplication when adding a new element.
3. When counting features in ComputePairFeatures, angles are compared, but given that the arc cosine is a monotonically decreasing function, its influence on the comparison result is limited to a change in sign.

### DIFF
--- a/cpp/open3d/geometry/PointCloud.cpp
+++ b/cpp/open3d/geometry/PointCloud.cpp
@@ -589,7 +589,8 @@ PointCloud::RemoveRadiusOutliers(size_t nb_points,
     for (int i = 0; i < int(points_.size()); i++) {
         std::vector<int> tmp_indices;
         std::vector<double> dist;
-        size_t nb_neighbors = kdtree.SearchRadius(points_[i], search_radius,
+        size_t nb_neighbors = kdtree.SearchHybrid(points_[i], search_radius,
+                                                  int(nb_points + 1),
                                                   tmp_indices, dist);
         mask[i] = (nb_neighbors > nb_points);
         ++progress_bar;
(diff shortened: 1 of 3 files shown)


ssheorey:
Hi @PiatrouskiIM thanks for this optimization! Can you provide some benchmark results of speed comparisons with the code in the `main` branch? You can use any reasonably sized data (eg something from the data module should be useful).

PiatrouskiIM:
I was mistaken about the presence of performance improvements when switching to SearchHybrid (my tests sets were regenerated when running benchmark of the another implementation).

PiatrouskiIM:
@benjaminum, SearchHybrid is implemented on top of SearchKNN, which is usually implemented differently (I won’t say for nanoflann, but I saw the implementation from scipy written in c). I was wrong about improving performance; it’s more likely a trade-off in favor of searching by radius. There are improvements only if there are many more points in the radius than the set threshold. I can prepare a more detailed comparison to find a visual tradeoff, but it seems that most often searching by radius is almost twice as fast.

benjaminum:
Sounds good. Looking forward to the comparison!

ssheorey:
Reverted switch from radius -> hybrid search:

> @benjaminum, SearchHybrid is implemented on top of SearchKNN, which is usually implemented differently (I won’t say for nanoflann, but I saw the implementation from scipy written in c). I was wrong about improving performance; it’s more likely a trade-off in favor of searching by radius. There are improvements only if there are many more points in the radius than the set threshold. I can prepare a more detailed comparison to find a visual tradeoff, but it seems that most often searching by radius is almost twice as fast.

ssheorey (approved):

PR #6719, opened 2024-03-25
:sparkles: Create the function ComputeClosestPoint to speedup the que…

### LINKED ISSUE #6718: Speedup the ComputeClosestPoint query for a single 3D Point
### Checklist

Adding a function ComputeClosestPoint(const Eigen::Vector3d& query_point) which is used to computes the closest point on the surface of the scene to a single query point.

I created a PR for this feature request.

### References

_No response_

### Additional information

_No response_

sebastienmascha:
Computes the closest point on the surface of the scene to a single query point.

This function calculates the closest point on any surface within the scene to the provided query point.
It's optimized for single-point queries, making it suitable for applications that require precise, real-time calculations of nearest surface points, such as collision detection or interactive simulations.

## Type
-   [x] New feature (non-breaking change which adds functionality). Resolves #6718 

## Motivation and Context
- It speeds up the query when you just have a point cloud and you want to do registration for instance.

## Checklist:
    to my code.
        updated accordingly.
    
### DIFF
--- a/cpp/open3d/t/geometry/RaycastingScene.cpp
+++ b/cpp/open3d/t/geometry/RaycastingScene.cpp
@@ -688,6 +688,33 @@ struct RaycastingScene::Impl {
                     LoopFn);
         }
     }
+
+    Eigen::Vector3d ComputeClosestPoint(const Eigen::Vector3d& query_point){
+        if (!scene_committed_) {
+            rtcCommitScene(scene_);
+            scene_committed_ = true;
+        }
+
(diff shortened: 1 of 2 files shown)


benjaminum (changes requested):
@sebastienmascha Thanks for this PR! 
Do you have a runtime comparison for the new function compared to the function processing multiple queries?
About the application using this function, do you use this from C++ or from Python?
Would there be a significant performance drop with using the O3D Tensor?

Issue #4794, opened 2022-02-22
TriangleMesh's operator "+=" gets rid of UV coordinates when no texture is attached to the mesh.


sebastienchappuis:
### Checklist

- [X] I have searched for [similar issues](https://github.com/isl-org/Open3D/issues).
- [X] For Python issues, I have tested with the [latest development wheel](http://www.open3d.org/docs/latest/getting_started.html#development-version-pip).
- [X] I have checked the [release documentation](http://www.open3d.org/docs/release/) and the [latest documentation](http://www.open3d.org/docs/latest/) (for `master` branch).

### My Question

Hello,

I noticed that the `+=` operator overload for the `TriangleMesh` class does not append UV coordinates in the case of a mesh that has UVs but no texture attached to it. I guess this is determined by this boolean [here](https://github.com/isl-org/Open3D/blob/fbf549b23c079ae255d1b8a20db33fd35f942585/cpp/open3d/geometry/TriangleMesh.cpp#L74-L76). I think it would be more intuitive if UVs were also copied over when appending meshes together, regardless of if they have a texture attached or not.

Is there any particular reason for the current behavior ? If not I could make a small PR to modify this.

thanks in advance and thanks for the great lib 👍

PR #6728, opened 2024-03-27
[Mesh] TriangleMesh's "+=" operator applies regardless of the presenc…

cdbharath:
…e of existing features.

Allows appending of UV coordinates, textures and material IDs with `+=` operator overload in `TriangleMesh` class irrespective of the presence of these attributes in the existing mesh (mesh on LHS to += operator)

## Type

-   [x ] Bug fix (non-breaking change which fixes an issue): Fixes # #4794

## Motivation and Context

The `+=` operator overload of `TriangleMesh` class does not append UV coordinates, textures and material IDs when the existing mesh (mesh on LHS to += operator) already contains any of these attributes. There is no reason to limit `+=` for UVs to `TriangleMesh` with the attributes already assigned as it currently happens [here](https://github.com/isl-org/Open3D/blob/fbf549b23c079ae255d1b8a20db33fd35f942585/cpp/open3d/geometry/TriangleMesh.cpp#L74-L76).  

## Checklist:

    to my code.
        updated accordingly.
    
## Description

The current change doesn't check if the presence of UV coordinates, textures and material IDs in the existing mesh and appends only based on their availability in the incoming mesh. 

I am attaching a screenshot that shows the addition of UV coordinates based on `+=` operator. The code is as follows
```
    auto box = geometry::TriangleMesh::CreateBox(1.0, 1.0, 1.0, true, true);    
    auto sphere = geometry::TriangleMesh::CreateSphere(1.0, 6, true);   
    rendering::MaterialRecord material;
    material.albedo_img = texture_image;
    material.name = "material";
            
    auto scene = std::make_shared<geometry::TriangleMesh>();
        
    // Append new meshes using +=
    *scene += *box;
    *scene += *sphere;
            
    gui::Application::GetInstance().Initialize(argc, argv);
    auto vis1 = std::make_shared<visualizer::O3DVisualizer>("Textures", 1024, 768);
    vis1->AddGeometry("scene", scene, &material);
    vis1->ResetCameraToDefault();
    gui::Application::GetInstance().AddWindow(vis1);
    gui::Application::GetInstance().Run();
```
The following is the resulting mesh with its texture UVs.
Screenshot from 2024-03-27 15-41-48

### DIFF
--- a/cpp/open3d/geometry/TriangleMesh.cpp
+++ b/cpp/open3d/geometry/TriangleMesh.cpp
@@ -52,9 +52,6 @@ TriangleMesh &TriangleMesh::Rotate(const Eigen::Matrix3d &R,
 
 TriangleMesh &TriangleMesh::operator+=(const TriangleMesh &mesh) {
     if (mesh.IsEmpty()) return (*this);
-    bool add_textures = HasTriangleUvs() && HasTextures() &&
-                        HasTriangleMaterialIds() && mesh.HasTriangleUvs() &&
-                        mesh.HasTextures() && mesh.HasTriangleMaterialIds();
     size_t old_vert_num = vertices_.size();
     MeshBase::operator+=(mesh);
     size_t old_tri_num = triangles_.size();
@@ -77,19 +74,22 @@ TriangleMesh &TriangleMesh::operator+=(const TriangleMesh &mesh) {
     if (HasAdjacencyList()) {
         ComputeAdjacencyList();
     }
-    if (add_textures) {
+    if (mesh.HasTriangleUvs()) {
         size_t old_tri_uv_num = triangle_uvs_.size();
         triangle_uvs_.resize(old_tri_uv_num + mesh.triangle_uvs_.size());
         for (size_t i = 0; i < mesh.triangle_uvs_.size(); i++) {
             triangle_uvs_[old_tri_uv_num + i] = mesh.triangle_uvs_[i];
         }
-
+    }
+    if (mesh.HasTextures()) {
         size_t old_tex_num = textures_.size();
         textures_.resize(old_tex_num + mesh.textures_.size());
         for (size_t i = 0; i < mesh.textures_.size(); i++) {
             textures_[old_tex_num + i] = mesh.textures_[i];
         }
-
+    }
+    if (mesh.HasTriangleMaterialIds()) {
+        size_t old_tex_num = textures_.size();
         size_t old_mat_id_num = triangle_material_ids_.size();
         triangle_material_ids_.resize(old_mat_id_num +
                                       mesh.triangle_material_ids_.size());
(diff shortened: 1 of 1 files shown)


cpp/open3d/geometry/TriangleMesh.cpp
```diff
@@ -77,30 +74,29 @@ TriangleMesh &TriangleMesh::operator+=(const TriangleMesh &mesh) {
     if (HasAdjacencyList()) {
         ComputeAdjacencyList();
     }
-    if (add_textures) {
+    if (mesh.HasTriangleUvs()) {
         size_t old_tri_uv_num = triangle_uvs_.size();
         triangle_uvs_.resize(old_tri_uv_num + mesh.triangle_uvs_.size());
```
benjaminum, line 79: If `this` does not have any UV coordinates then `triangle_uvs_.size()` will be different from the total number of triangles in the resulting mesh. This would break with the assumption that `triangle_uvs_.size() == 3*number of triangles`

benjaminum:
Hi @cdbharath , we think mixing meshes with and without texture uvs can be problematic. I added a special case for empty meshes that allows your workflow with the += operator.

cpp/open3d/geometry/TriangleMesh.cpp
```diff
@@ -77,30 +75,30 @@ TriangleMesh &TriangleMesh::operator+=(const TriangleMesh &mesh) {
...
                                       mesh.triangle_material_ids_.size());
         for (size_t i = 0; i < mesh.triangle_material_ids_.size(); i++) {
             triangle_material_ids_[old_mat_id_num + i] =
                     mesh.triangle_material_ids_[i] + (int)old_tex_num;
         }
-    } else {
-        triangle_uvs_.clear();
-        textures_.clear();
-        triangle_material_ids_.clear();
     }
```
ssheorey, line 102: We need to keep this portion - `clear()` properties if only one mesh has them (unless the other mesh is empty), else we will get meshes with only half the triangles with material ids, etc. (e.g. if `this` has material_ids, but `mesh` doesn't, output should not have material ids.) Half populated properties will cause seg faults.

Also, textures and triangle_material_ids go together, so we should have a single `if()` for them. We want `n_textures == max(triangle_material_ids_)+1` and either both or none should be present always. 

@cdbharath this PR may break other mesh functions if there is a bug -- please do the following tests and make sure there are no errors. Check for the case where:

- a has texture_uvs but no textures / triangle_material_ids
- b has all of texture_uvs, textures and triangle_material_ids 
- c has none of the 3 (but has triangles)

With these 3 meshes, perform this check:

- `a+=b` and `b+=a` should produce the same results in a and b respectively (test mesh equality). Also for the pair `a+=c`, `c+=a`, and separately for the pair `b+=c` and `c+=b`. [Add this to C++ unit tests if not present already] Visualize the resulting mesh correctly with draw_geometries() and draw() [Just test visualization on your computer - no unit tests required]

cdbharath:
@benjaminum @ssheorey I have addressed the posted issues. I also added an unit test for `+=` operator along with visual examples attached to the PR description. I have requested for review again. Thank you.

benjaminum (approved):
lgtm

PR #6815, opened 2024-06-04
Adapted find_dependencies.cmake to build with CUDA >= 12.0 with dynamic libraries as well

davconde:
## Type

## Motivation and Context

Currently, CUDA support when building searches explicitly for static dependencies. However, in absence of static lib binaries, Open3D fails to build. I think this is important since some CUDA installations may lack the required static dependencies.

For instance, the official NVIDIA DeepStream 7.0 Docker images don't include the static libraries, while the framework include multiple new features oriented toward LiDAR data processing, camera calibration and 3D tracking, so being able to develop custom GStreamer plugins taking advantage of Open3D would be very helpful.

## Checklist:

    to my code.
        updated accordingly.
    
## Description

When importing CUDA dependencies for cuBLAS if version is >= 12.0 it now checks first if dynamic libraries are available to build as usual. Otherwise, it will look for static libraries. This change has been tested in these both NVIDIA Docker images, being the DeepStream image the one without static libraries:

- `nvcr.io/nvidia/cuda:12.2.0-devel-ubuntu22.04`
- `nvcr.io/nvidia/deepstream:7.0-gc-triton-devel`

After basic containers setups, building was performed by running :
```bash
git clone https://github.com/davconde/Open3D.git
cd Open3D
apt install -y libfmt-dev libomp-dev
SUDO=command util/install_deps_ubuntu.sh assume-yes
mkdir build && cd build
export OPEN3D_INSTALL_PATH=/usr/local
cmake -DBUILD_SHARED_LIBS=ON -DBUILD_PYTHON_MODULE=OFF -DBUILD_CUDA_MODULE=ON -DCMAKE_INSTALL_PREFIX=${OPEN3D_INSTALL_PATH} ..
make -j$(nproc)
make install
echo 'export LD_PRELOAD=/usr/local/lib/libOpen3D.so:${LD_PRELOAD}' >> ~/.bashrc
export LD_PRELOAD=/usr/local/lib/libOpen3D.so:${LD_PRELOAD}
```

And tested with the following snippet:
```cpp
#include <iostream>
#include <open3d/Open3D.h>
#include <open3d/core/CUDAUtils.h>  // Include this for CUDA utilities

int main() {
    using namespace open3d;

    // Print Open3D version
    std::cout << "Open3D " << OPEN3D_VERSION << std::endl;

    // Check if CUDA is available
    if (core::cuda::IsAvailable()) {
        std::cout << "CUDA is available" << std::endl;
    } else {
        std::cout << "CUDA is NOT available" << std::endl;
    }

    // Create a simple point cloud
    core::Tensor points = core::Tensor::Init<float>({{0, 0, 0}, {1, 0, 0}, {0, 1, 0}, {0, 0, 1}});
    t::geometry::PointCloud pcd(points);
    std::cout << "PointCloud:\n" << pcd.GetPointPositions().ToString() << std::endl;

    return 0;
}

```

Which yielded coherent results in both cases.


ssheorey:
Hi @davconde, thanks for submitting this PR. Building with cuda shared libraries is definitely useful. Some comments:
- Please add an option BUILD_WITH_CUDA_STATIC (default ON) to CMakeLists. 
- We should not guess the build environment - if BUILD_WITH_CUDA_STATIC is ON and any static lib is not found, it's a FATAL_ERROR. Same if it's OFF and shared libs are not found.
- Test both the Open3D shared lib and static lib (using BUILD_SHARED_LIBS = ON/OFF options ) and BUILD_UNIT_TESTS=ON `tests` executable.
- Testing must be done in an empty new dcker container. Not the build container. Otherwise the binary is not relocatable / portable. 
- With dynamic linking, the cuda libraries are now part of the Open3D ABI. This interface info needs to be present in the cmake and package config scripts created when we do `make package`. To test that this is setup correctly, build a test Open3D app using each of the `find_package(Open3D)` and `open3d.pc` methods and ensure that the build files look for the correct CUDA libraries. See https://www.open3d.org/docs/latest/getting_started.html#id6 for a simple test app.

Knighthood2001:
I write CMakeList.txt in https://github.com/Knighthood2001/open3d_cuda12.4_test

PR #6816, opened 2024-06-05
Consistent orientation of mesh hole-filled triangles.

stormboy:
## Type

## Checklist:

    to my code.
        updated accordingly.
    

benjaminum:
@stormboy Thanks for this PR!
I think it would be very good to get consistent orientation for the newly created triangles. However, the proposed solution affects all triangles. This could change the orientation of the existing triangles which is an unexpected side effect.
Do you want to give a solution a try that only affect the hole triangles?

stormboy:
> @stormboy Thanks for this PR! I think it would be very good to get consistent orientation for the newly created triangles. However, the proposed solution affects all triangles. This could change the orientation of the existing triangles which is an unexpected side effect. Do you want to give a solution a try that only affect the hole triangles?

I agree that this change adds a dirty way to try solve the problem.  I have in fact found a few cases where this solution results in opposing normals in much of an original *almost* closed volume mesh.

I am in fact planning to have another look at this problem in the coming month.  This is the first time I have looked at VTK, which is what is being used, so it might take me a while to understand approaches that can be taken here.

benjaminum:
This may require to directly manipulate the mesh arrays. Something like this
1. Identify the new triangles
2. Identify all edges connecting existing and new triangles
3. Count number of edges for which the triangle orientations disagree between existing and new triangles
4. Change orientation of new triangles if the majority in step 3 disagrees

ssheorey:
Closing for now. Please re-open this or a new PR once it is ready.

PR #6854, opened 2024-07-04
Nix build

netogallo:
I am building a derivation for NixOS 24.05 of open3d-cpu so it can be easily used in nix. To guarantee reproducibility, Nix cannot fetch resources during build and best practices recommend using packages already present in the nix store. For that reason, I removed the use of deprecated functions from ImGui.

Additionally, Nix allows users to easily modify the build parameters of packages, to that end I included a flag to disable the "ComputeUVAtlas" function as it is a single function that leads to the inclusion of significant dependencies.

Finally, Nix does not install packages in systemwide directories. Instead, each package and version gets installed in an isolated directory and nix uses symlinks and env variables to bring the packages into scope. Therefore, a hardcoded DESKTOP_INSTALL_DIR is not suitable for this environment and must be supplied through a compile time variable. If this variable is not supplied (default), the behavior remains unchanged.

-   [X] New feature (non-breaking change which adds functionality). Resolves #

## Motivation and Context

These changes are required to eventually update ImGui to version 1.9.6. Furthermore, it adds a CMake flag that allows users to build a smaller package by removing the TriangleMesh::ComptueUVAtlas function which adds many additional dependencies to the project. The function is included by default in order to preserve existing behavior.

## Checklist:

    to my code.
        updated accordingly.
    
## Description

Allow compiling open3d-cpu python package on NixOs 24.05
* Remove the use of deprecated functions in ImGui
* CMake option to exclude uvatlas as dependency as a single function leads to the inclusion of many libraries.
* CMake option to override the desktop install path.

The derivation for packaging v0.18.0 is available at: [https://github.com/netogallo/nix-packages/blob/main/pkgs/python-packages/open3d/open3d-common.nix](https://github.com/netogallo/nix-packages/blob/main/pkgs/python-packages/open3d/open3d-common.nix)

It currently applies a similar patch before building.


ssheorey:
@netogallo thanks for submitting this PR. Removing deprecated imgui functions is very useful.

About making UVAtlas optional, @benjaminum what do you think? I think this adds extra complexity to the build system without being widely useful, so I would recommend to maintain that in a fork of Open3D.

ssheorey:
@netogallo if there is a constraint on fetching resources during build, you can build Open3D once on an internet connected system. All required 3rdparty downloads are stored in the 3rdparty_downloads folder. If you include this in the Open3D source, Open3D can be built without an internet connection. cmake looks in the 3rdparty_downloads folder first before downloading from the internet.

benjaminum:
@netogallo Thank you for these contributions!
I think being able to override the DESKTOP_INSTALL_DIR is a good idea. Please see the change which allows you to directly set the variable from the cmdline.
Thanks for updating ListView!

About UVAtlas, I removed the build option as we have already quite a large number of build options and likely will expand UV map functionality in the future. Happy to hear from you how we can make it easier to deal with dependencies for Nix.

benjaminum (approved):

PR #6865, opened 2024-07-11
Add O3DVisualizer API to enable collapse control of verts in the side panel

EwingKang:
Add API SetCollapsableVertOpen to control the toggling of the verts in the side panel

## Type

## Motivation and Context
When we have start the viewer with side panel ("show settings"), the default toggle vert (pane) in the side panel defaults to ON for the "Mouse Controls" and "Scene", but no "Lighting". These default on panels took a lot of space and I find myself always collapsing the settings since I mostly want to control the visibility of objects in "Geometries". This becomes quite annoying very soon.

## Checklist:

    to my code.
        updated accordingly.
    
## Description

**Before**: Side vert "Mouse Controls" and "Scene" always defaults to ON

Screenshot from 2024-07-10 11-44-12

**After** The displayed result after calling the new API
```c++
O3DVisualizer::SetCollapsableVertOpen("Mouse Controls", false);
```
Screenshot from 2024-07-11 09-59-02

### DIFF
--- a/cpp/open3d/visualization/visualizer/O3DVisualizer.cpp
+++ b/cpp/open3d/visualization/visualizer/O3DVisualizer.cpp
@@ -1706,7 +1706,7 @@ Ctrl-alt-click to polygon select)";
         }
     }
 
-    void SetCollapsableVertOpen(const std::string &name, bool open) {
+    void SetPanelOpen(const std::string &name, bool open) {
         if (name == settings.mouse_panel->GetText()) {
             settings.mouse_panel->SetIsOpen(open);
         } else if (name == settings.scene_panel->GetText()) {
@@ -2458,8 +2458,8 @@ void O3DVisualizer::SetMouseMode(SceneWidget::Controls mode) {
     impl_->SetMouseMode(mode);
 }
 
-void O3DVisualizer::SetCollapsableVertOpen(const std::string &name, bool open) {
-    impl_->SetCollapsableVertOpen(name, open);
+void O3DVisualizer::SetPanelOpen(const std::string &name, bool open) {
+    impl_->SetPanelOpen(name, open);
 }
 
 void O3DVisualizer::EnableGroup(const std::string &group, bool enable) {
--- a/cpp/open3d/visualization/visualizer/O3DVisualizer.cpp
+++ b/cpp/open3d/visualization/visualizer/O3DVisualizer.cpp
@@ -291,6 +291,7 @@ struct O3DVisualizer::Impl {
     std::function<bool()> on_animation_tick_;
     std::shared_ptr<io::rpc::ZMQReceiver> receiver_;
     std::shared_ptr<MessageProcessor> message_processor_;
+    std::map<std::string, CollapsableVert *> side_panel_verts_;
 
     UIState ui_state_;
     bool can_auto_show_settings_ = true;
@@ -420,6 +421,7 @@ struct O3DVisualizer::Impl {
         settings.mouse_panel =
                 new CollapsableVert("Mouse Controls", v_spacing, margins);
         settings.panel->AddChild(GiveOwnership(settings.mouse_panel));
+        side_panel_verts_.emplace("Mouse Controls", settings.mouse_panel);
 
         settings.mouse_tab = new TabControl();
         settings.mouse_panel->AddChild(GiveOwnership(settings.mouse_tab));
@@ -541,6 +543,7 @@ Ctrl-alt-click to polygon select)";
         // Scene controls
         settings.scene_panel = new CollapsableVert("Scene", v_spacing, margins);
         settings.panel->AddChild(GiveOwnership(settings.scene_panel));
+        side_panel_verts_.emplace("Scene", settings.scene_panel);
 
         settings.show_skybox = new Checkbox("Show Skybox");
         settings.show_skybox->SetOnChecked(
@@ -649,6 +652,7 @@ Ctrl-alt-click to polygon select)";
         settings.light_panel = new CollapsableVert("Lighting", 0, margins);
         settings.light_panel->SetIsOpen(false);
         settings.panel->AddChild(GiveOwnership(settings.light_panel));
+        side_panel_verts_.emplace("Lighting", settings.light_panel);
 
         h = new Horiz(v_spacing);
         settings.use_ibl = new Checkbox("HDR map");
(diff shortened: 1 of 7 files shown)


ssheorey (changes requested):
Hi @EwingKang thanks for this PR.

One suggestion to simplify this is to add`CollapsableVert::GetText()` in `Layout.cpp`. This way, there is no need to store the text -> pointer map. You can instead iterate through all the children of the vert panel and just match the text.
 
Also, please add python bindings.

EwingKang:
Hi @ssheorey 
Thanks for your reply. I'll add the Python bindings (not sure how to do that exactly, but I'll try)
Also, adding `CollapsableVert::GetText()` in `Layout.cpp` is not a problem.   

However, for the pointers suggestion, do you suggest having a `std::vector<CollapsableVert *>` instead of a map here? Does that mean the removal of `CollapsableVert *mouse_panel,` `CollapsableVert *scene_panel,` `CollapsableVert *light_panel,` and `CollapsableVert *geometries_panel`?
The upside is the simplification of members. We'll only have one pointer per object. 
The downside is the increasing difficulties of using the object and slower speed if there are a lot of panels. For example, if someone wants to use `geometries_panel` within the `O3DVisualizer.cpp`, they would now have to find the pointer to that `CollapsableVert * through the same look API with a key string. The API would have O(N) instead of O(logN). Of course, the former might still be faster, considering we have only 4 panels now.

EwingKang:
Add `GetText()` and `SetText()` to CollapsableVert
Here are some test screenshots of setting the text (the "hahahaha#" was originally "Advanced lighting")
Screenshot from 2024-07-18 10-28-43
Screenshot from 2024-07-18 10-28-47

cpp/open3d/visualization/visualizer/O3DVisualizer.cpp
```diff
@@ -1706,6 +1711,14 @@ Ctrl-alt-click to polygon select)";
         }
     }
 
+    void SetCollapsableVertOpen(const std::string &name, bool open) {
+        auto it = side_panel_verts_.find(name);
+        if (it != side_panel_verts_.end() && it->second) {
+            it->second->SetIsOpen(open);
+            window_->SetNeedsLayout();
+        }
```
ssheorey, line 1719: I think we don't need `side_panel_verts_` and can remove it. Instead an if-else works:
```cpp
if (name=='Geometries') {
settings.geometries_panel.SetIsOpen(open);
...
} else if (name==...) {
...
} else if ...
```
Code wise it's about the same, all the logic is collected in this one function (no updates to `side_panel_verts_` needed.

cpp/open3d/visualization/visualizer/O3DVisualizer.cpp
```diff
@@ -1706,6 +1711,14 @@ Ctrl-alt-click to polygon select)";
         }
     }
 
+    void SetCollapsableVertOpen(const std::string &name, bool open) {
```
ssheorey, line 1714: nit: `SetPanelOpen` is clearer for users. O3DVisualizer is meant to be more high level, so referring to the widgets as panels is easier for a new user to find than CollapsableVert.

cpp/open3d/visualization/visualizer/O3DVisualizer.h
```diff
@@ -175,6 +175,7 @@ class O3DVisualizer : public gui::Window {
     void SetLineWidth(int line_width);
     void EnableGroup(const std::string& group, bool enable);
     void SetMouseMode(gui::SceneWidget::Controls mode);
+    void SetCollapsableVertOpen(const std::string& name, bool open);
```
ssheorey, line 178: nit: consider naming `SetPanelOpen` instead.

cpp/pybind/visualization/o3dvisualizer.cpp
```diff
@@ -344,6 +344,10 @@ void pybind_o3dvisualizer(py::module& m) {
                  "enable"_a)
             .def("show_skybox", &O3DVisualizer::ShowSkybox,
                  "Show/Hide the skybox", "show"_a)
+            .def("set_collapsable_vert_open",
```
ssheorey, line 347: nit: consider naming `set_panel_open` instead for clarity.

ssheorey (commented):
Looks good after a few minor updates. Thanks @EwingKang

cpp/open3d/visualization/visualizer/O3DVisualizer.h
EwingKang: Interestingly, `SetPanelOpen` is exactly what I originally named the function. I did change it because the source code seems to refer the "Panel" as the setting panel, and "Verts" as the elements within the panel. This is not absolute though, different places uses different names. I don't know if we have any definitions to the UI elements but I strongly think that it should probably exists.
For the mean time, I'll happily change the name if you think it's clearer for the user/ coders.

cpp/open3d/visualization/visualizer/O3DVisualizer.cpp
EwingKang: Updated in commit https://github.com/isl-org/Open3D/pull/6865/commits/a19cdf12bc6329a88047f2c602175005f7267eaf

cpp/pybind/visualization/o3dvisualizer.cpp
EwingKang: Updated in commit https://github.com/isl-org/Open3D/pull/6865/commits/a19cdf12bc6329a88047f2c602175005f7267eaf

cpp/open3d/visualization/visualizer/O3DVisualizer.h
EwingKang: Updated in https://github.com/isl-org/Open3D/pull/6865/commits/459c7ab7e9b369a9daba4491e3e2edb2f85c82b9

cpp/open3d/visualization/visualizer/O3DVisualizer.cpp
EwingKang: Updated in commit https://github.com/isl-org/Open3D/pull/6865/commits/a19cdf12bc6329a88047f2c602175005f7267eaf

ssheorey (approved):
Thanks @EwingKang Looks good!

PR #6874, opened 2024-07-16
Support numpy v2.

ssheorey:
Pybind11 2.12 added numpy 2 support. We update to the latest 2.13.1.
Other dependent packages (e.g. tensorboard) also need update, so keeping the numpy<2 in requirements.txt for now.

## Type

-   [x] Bug fix (non-breaking change which fixes an issue): Fixes #6845, #6856, #6839, #6840
-   [ ] New feature (non-breaking change which adds functionality). Resolves #
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected) Resolves #

## Checklist:

-   [ ] I have run `python util/check_style.py --apply` to apply Open3D **code style**
    to my code.
-   [ ] This PR changes Open3D behavior or adds new functionality.
    -   [ ] Both C++ (Doxygen) and Python (Sphinx / Google style) **documentation** is
        updated accordingly.
    -   [ ] I have added or updated C++ and / or Python **unit tests** OR included **test
        results** (e.g. screenshots or numbers) here.
-   [ ] I will follow up and update the code if CI fails.
    
-   [ ] For fork PRs, I have selected **Allow edits from maintainers**.

PR #6879, opened 2024-07-17
add numpy version warning to documentation

rxba:
## Motivation and Context
Release of `numpy==2.0.0` causes many Open3D features to segfault, which I saw will be fixed with the `pybind` update in #6874 .
Due to only requiring `numpy>=1.18.0` in the `Open3D<=0.18.0` releases, new Open3D installs will also install `numpy==2.0.0`, and thus break a large part of the functionality.
I think it's important to communicate this to users, although I'm unsure if this is the right place and way to do so, especially since this documentation will also only update for the next release of Open3D.
I hope this feedback is helpful, thank you!

## Description

Add warning about `numpy` versions  to the "Getting Started" documentation.

### DIFF
--- a/docs/getting_started.in.rst
+++ b/docs/getting_started.in.rst
@@ -52,6 +52,15 @@ Pip (PyPI)
 
 .. warning::
 
+   Versions of ``numpy>=2.0.0`` are not fully supported by ``Open3D<=0.18.0`` and may lead to errors.
+   Consider downgrading ``numpy``, e.g. with
+
+   .. code-block:: bash
+
+        pip install "numpy<2.0.0"
+
+.. warning::
+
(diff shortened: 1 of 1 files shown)


ssheorey:
Hi @rxba the current requirements.txt lists (merged a few days ago):

https://github.com/isl-org/Open3D/blob/2c8bb608f960301401ff33867c6b5831231ea462/python/requirements.txt#L1

If you install the latest devel wheels (from github releases), pip will install numpy 1.x. If numpy 2 is installed, pip will complain.

ssheorey (approved):

ssheorey:
Makes sense. Thanks for the PR!

PR #6896, opened 2024-07-26
Fixes for pybind11-stubgen errors

timohl:
## Type

-   [x] Bug fix (non-breaking change which fixes an issue): Fixes #
-   [ ] New feature (non-breaking change which adds functionality). Resolves #
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected) Resolves #

## Motivation and Context

This pull request is a follow up for/based on #6869 and contains more fixes for documentation and typing generation.
Related issues are #3052 and #6867.

## Checklist:

-   [x] I have run `python util/check_style.py --apply` to apply Open3D **code style**
    to my code.
-   [ ] This PR changes Open3D behavior or adds new functionality.
    -   [ ] Both C++ (Doxygen) and Python (Sphinx / Google style) **documentation** is
        updated accordingly.
    -   [ ] I have added or updated C++ and / or Python **unit tests** OR included **test
        results** (e.g. screenshots or numbers) here.
-   [x] I will follow up and update the code if CI fails.
    
-   [x] For fork PRs, I have selected **Allow edits from maintainers**.

## Description

Typings are generated using:
```bash
# Create venv
python -m venv .venv
. .venv/bin/activate
pip install python/requirements.txt
pip install python/requirements_build.txt

# Build open3d and install in .venv
mkdir build && cd build
cmake ..
make -j$(nproc) && make install-pip-package

# Generate typings
pip install pybind11-stubgen
pybind11-stubgen open3d -o typings
```

Issue #5450, opened 2022-08-18
From NUMPY RGBD To o3d.RGBD


wangmiaowei:
### Checklist

- [X] I have searched for [similar issues](https://github.com/isl-org/Open3D/issues).
- [X] For Python issues, I have tested with the [latest development wheel](http://www.open3d.org/docs/latest/getting_started.html#development-version-pip).
- [X] I have checked the [release documentation](http://www.open3d.org/docs/release/) and the [latest documentation](http://www.open3d.org/docs/latest/) (for `master` branch).

### My Question

RuntimeError: [Open3D Error] (static std::shared_ptr<open3d::geometry::RGBDImage> open3d::geometry::RGBDImage::CreateFromColorAndDepth(const open3d::geometry::Image&, const open3d::geometry::Image&, double, double, bool)) /root/Open3D/cpp/open3d/geometry/RGBDImageFactory.cpp:42: [CreateFromColorAndDepth] Unsupported image format.

        radius_normal = voxel_size * 2
        color_raw = Image.open(path+"/color/0000"+str(i)+".jpg")
        depth_raw = Image.open(path+"/depth/0000"+str(i)+".png")
        color  =255*rgb_transform(color_raw)
        depth = dep_transform(np.asarray(depth_raw),dims)
        color_o3d = o3d.geometry.Image(color.detach().numpy().astype(np.uint8))
        depth_o3d = o3d.geometry.Image((depth.detach().numpy()).astype(np.uint16))
        print('color_o3d.shape: ',(np.asarray(color_o3d)).max())
        rgbd_image = o3d.geometry.RGBDImage.create_from_color_and_depth(color_o3d,depth_o3d)
        pcd = o3d.geometry.PointCloud.create_from_rgbd_image(rgbd_image, pinhole_camera_intrinsic)

wangmiaowei:
OK,this problem is solved by myself. I'd want to paste the solution! All the problems come from unsupported formats.The map type and shaped should be converted to the following format:
![Screenshot from 2022-08-19 11-15-13](https://user-images.githubusercontent.com/27915265/185535149-d0f0a6b0-75f4-4dfe-b877-16c893fc07d8.png)

PR #6917, opened 2024-08-14
Include generated stubs in python package

### LINKED ISSUE #3052: Open3D type checking bug/issue
**Describe the bug**
Since the Open3D update I get this kind of a warning (see screenshots) in PyCharm when I try to add a point cloud as a geometry in a visualizer.

I've also seen issues like "Expected type 'PointCloud.py', got 'PointCloud' instead", where the only difference is that the expected type is the type of the object I am already giving but followed by a ".py"

**To Reproduce**
1. Start PyCharm
2. Write a test program that uses the geometry, or pointcloud objects (even some others)
3. Create an object of some of those types, but explicitly define the type, i.e.:
    some_pointcloud: open3d.geometry.PointCloud = open3d.geometry.PointCloud() (added another screenshot for this example)
4. Give it as an input to the method that uses such object 

**Expected behavior**
The method should accept the input object and PyCharm shouldn't complain about it.

**Screenshots**
image
image

- Python version: Python 3.8 (but it happens in other Python versions I believe)
- Open3D version: Open3D 12 (but I think it also happened in 11, not 100% sure though)
- How did you install Open3D?: pip (through PyCharm)
- Compiler version (if built from source): n/a

**Additional context**
I opened this discussion on Discord, but there were no responses, so I assume that it is a bug. Also, in the geometry issue, I've tried casting it to a geometry object type, but there were some issues with the geometry object constructor. I searched the documentation to try and see if I'm doing something wrong, but I couldn't find anything specific to this.

By the way, nothing crashes, everything works as intended, it's just these silly warnings that the static code analyzer complains about.

### LINKED ISSUE #4347: VS Code IntelliSense doesn't work with Open3D when using the PyLance server
**Describe the bug**
The [IntelliSense](https://code.visualstudio.com/docs/editor/intellisense) feature of VS Code does not work with the Open3D library when using the PyLance language server in VS Code. 

**To Reproduce**

1. Get VS Code. 
2. Install the Python and PyLance extension. 
3. Create a new python file, and import open3d.
4. Write something from the library. Auto-completion and function docstring should be presented. 

**Expected behavior**
Here's an example of expected behavior using numpy. 
numpywork

**Workaround**
It kinda works using the Jedi language server. Although the docstrings are quite bad.
I mentioned this in #751 but never got a response. 
open3djedi

- Operating system:Description: Ubuntu 20.04.3 LTS
- Python version: 3.8.10
- Open3D version: 0.13.0
- How did you install Open3D?: pip

### LINKED ISSUE #5802: Request to add type hint for Open3D package?
### Checklist

just like numpy, it's mainly implemented by c/c++
But it still has type hint, it is convinient for coders.

When I type "numpy." and "Tab", there're many functions to be selected.
When I type "open3d.geometry", there's nothing

### References

_No response_

### Additional information

_No response_

timohl:
## Type

-   [x] New feature (non-breaking change which adds functionality). fixes #3052, fixes #4347, fixes #5802

## Motivation and Context

Currently, the Open3D python package does not come with typing stubs.
Stubs can be generated manually using pybind11-stubgen (with currently wip fixes in #6896).
This pull request tries to integrate the generated typing stubs into the python package.
In this way, Open3D can ship with typing support out of the box.

## Checklist:

    to my code.
        updated accordingly.
    
## Description

This adds the following steps to the build process:
* Add pybind11-stubgen to requirements_build.txt
* Generate stubs using pybind11-stubgen
* Write stubs into python package
* Add `py.typed` file to mark typing support (see [PEP 561](https://peps.python.org/pep-0561/#packaging-type-information)

TODO:
* [x] In vscode imports do not work correctly when integrating stubs. Usage of, e.g., `open3d.camera` instead of `open3d.cpu.pybind.camera` in stubs are not resolved correctly. Further investigation requiered...
* [x] Maybe add build option IGNORE_STUBGEN_ERRORS to toggle whether any errors in pybind11-stubgen call should stop the build. This might be usefull to add a CI check so typing stubs stay clean of errors.


timohl:
This is ready for review now.

In [85dd074](https://github.com/isl-org/Open3D/pull/6917/commits/85dd0748553dcb7eeffb299b4eeb0ab637b4090b) I moved the namespace fixing (`open3d.[cpu|cuda].pybind.*` -> `open3d.*`) to sphinx autodoc callbacks, since the imports in the generated stubs were not working correctly.

I added a cmake flag `IGNORE_STUBGEN_ERRORS` which might be useful when a CI check for stub generation is added.

timohl:
The open3d python package installs the following now:
```
open3d:
__init__.py
__init__.pyi
_build_config.py
app.py
cpu
examples
libc++.so.1
libc++abi.so.1
ml
py.typed
resources
tools
visualization
web_visualizer.py
```
Note the following new files:
* `*.pyi` files: Contain the typing stub information (more can be found in subdirectories)
* `py.typed` file: Marks that this package supports typing (see [PEP 561](https://peps.python.org/pep-0561/#packaging-type-information))

timohl:
I fixed some build errors from CI checks.
For that, I changed the workflows to always use requirements_build.txt instead of locally defining the build dependencies.
I applied this to all locally defined python dependencies, so the requirements*.txt become the single source of truth for dependencies.

There is still an issue when open3d-ml is included.
pybind11-stubgen requires runtime dependencies to be available, so the open3d-ml requirements have to be installed in docker/Dockerfile.wheel.
See [Ubuntu Wheel / Build wheel (3.11, false)](https://github.com/isl-org/Open3D/actions/runs/10404892703/job/28814386616?pr=6917) with error from pybind11-stubgen:
`ModuleNotFoundError: No module named 'sklearn'`

I will fix that later today.

timohl:
I fixed some CI stuff, but there are still two issues left in the Windows checks:
* With `BUILD_CUDA_MODULE` enabled pybind11-stubgen throws: ` [Open3D Error] (void __cdecl open3d::core::__OPEN3D_CUDA_CHECK(enum cudaError,const char *,const int)) D:\a\Open3D\Open3D\cpp\open3d\core\CUDAUtils.cpp:310: D:\a\Open3D\Open3D\cpp\open3d\core\CUDAUtils.cpp:25 CUDA runtime error: CUDA driver version is insufficient for CUDA runtime version`
* With `BUILD_SHARED_LIBS` pybind11-stubgen throws: `ImportError: DLL load failed while importing pybind: The specified module could not be found.`

Not sure right now how to fix those.

timohl:
I disabled stubgen for windows shared libs and windows cuda workflows by introducing the `WITH_STUBGEN` cmake flag.

Currently, I cannot really check if stubgen works for CUDA since I do not have access to a GPU supporting CUDA at home right now.
The `__init__.py` seems to only import `open3d.cuda.pybind` if there is any CUDA device available:
https://github.com/isl-org/Open3D/blob/6cb32f6d7b4d3a39555abccf2e20a56d5f852ec4/python/open3d/__init__.py#L65-L69
So `pybind11-stubgen` does not work if no device is available.

At this point, I would love to get some feedback.
Most importantly, those commits which add the stub generation/inclusion:
[564af88](https://github.com/isl-org/Open3D/pull/6917/commits/564af883103061232f695ca51b132aa012e61a5d)
[9a9cc95](https://github.com/isl-org/Open3D/pull/6917/commits/9a9cc9547f1b760699af5736591e04113ac31ac3)
[85dd074](https://github.com/isl-org/Open3D/pull/6917/commits/85dd0748553dcb7eeffb299b4eeb0ab637b4090b)
[7966fc0](https://github.com/isl-org/Open3D/pull/6917/commits/7966fc0d560f712f55438cb70f6403a7cdf601ad)
But also about the integration into the build process and CI.

ManuCorrea:
Looking forward to this to get merged! This will improve a lot the safety and use of Open3D in large applications.

ssheorey:
Hi @timohl thanks for this very useful contribution!

The goal of this PR is to improve type checking support when using Open3D. One way to test this is by running mypy over the Python examples and looking at the output with and without this PR (i.e. with open3d v0.18 installed vs the Python wheel from this PR installed.) When I tried that, I ran into this error with the wheel from this PR:

```sh
$ mypy examples/python/geometry/triangle_mesh_with_numpy.py 
/Users/ssheorey/.pyenv/versions/3.11.4/envs/o3d-311/lib/python3.11/site-packages/open3d/__init__.pyi:51: error: invalid syntax  [syntax]
Found 1 error in 1 file (errors prevented further checking)
```

Here's what it's complaining about (last line of `__init__.pyi` looks incomplete):
```py
open3d = 
```

Can you take a look?

timohl:
Honestly, I am not sure why there is this `open3d = ` at the end.
I am no expert in `pybind11-stubgen` internals and the `__init__.py` from which the `__init__.pyi` is generated, is quite complex.
When I add `del open3d` in [288d51d](https://github.com/isl-org/Open3D/pull/6917/commits/288d51d77abc864cca263df2f73495c90008c4b5) it seems to be resolved and everything else is working.

Now mypy has a new errror:
```bash
mypy examples/python/geomet
ry/triangle_mesh_with_numpy.py 
/home/timohl/projects/Open3D/.venv/lib/python3.10/site-packages/open3d/cpu/pybind/core/__init__.pyi:732: error: non-default argument follows default argument  [syntax]
Found 1 error in 1 file (errors prevented further checking)
```

refering to:
```python
@staticmethod
@typing.overload
def arange(start: int | None = None, stop: int, step: int | None = None, dtype: Dtype | None = None, device: Device | None = None) -> Tensor:
        """
        Create a 1D tensor with evenly spaced values in the given interval.
        """
```

I will try to fix all mypy errors, but soonest I can find time for that might be next weekend.

timohl:
I fixed some mypy errors:
### Changed Tensor.arange signature to be like in numpy (230a9a6).

This is how numpy defines arange:
```python
@overload
def arange(  # type: ignore[misc]
    stop: _IntLike_co,
    /, *,
    dtype: None = ...,
    device: None | L["cpu"] = ...,
    like: None | _SupportsArrayFunc = ...,
) -> _1DArray[int, signedinteger[Any]]: ...
@overload
def arange(  # type: ignore[misc]
    start: _IntLike_co,
    stop: _IntLike_co,
    step: _IntLike_co = ...,
    dtype: None = ...,
    *,
    device: None | L["cpu"] = ...,
    like: None | _SupportsArrayFunc = ...,
) -> _1DArray[int, signedinteger[Any]]: ...
```
(see https://github.com/numpy/numpy/blob/b3ddf2fd33232b8939f48c7c68a61c10257cd0c5/numpy/_core/multiarray.pyi#L769-L786)

Note how `start` is not optional in the second signature.
Here, optional is not needed, because the first overload handles that case.

Also, it adds `/, *,` in the first signature and `*,` in the second.
`/` forces all args before to be positional-only and `*` forces all arguments after to be keyword-only.
In pybind this can be achieved using `py::pos_only()` and `py::kw_only()`.



### Changed `lambda` to `lambda_penalty` in `orient_normals_consistent_tangent_plane` (6df9d85).
This is a straight forward argument rename, since `lambda` is a reserved keyword in python.
I also checked for other parameters named `lambda` and this seems to be the only one.



### Changed lookat, up, front, and zoom to optional args (54eab24).
mypy comlained that there are arg without default value after args with default value.
I changed those to be optional (passing none/nullptr).
This is in line with the cpp implementation.
Note that I used utility::optional instead of a simple pointer, which generates correct types in stubgen (`<type> | None = None` instead of `<type> = None` which mypy also complains about).

timohl:
mypy now does not complain about anything in the stub files, but has two errors in the example code:
```bash
mypy examples/python/geometry/triangle_mesh_with_numpy.py 
examples/python/geometry/triangle_mesh_with_numpy.py:14: error: Argument 1 to "read_triangle_mesh" has incompatible type "str"; expected "PathLike[Any]"  [arg-type]
examples/python/geometry/triangle_mesh_with_numpy.py:15: error: Argument 1 to "paint_uniform_color" of "TriangleMesh" has incompatible type "list[float]"; expected "ndarray[Any, Any]"  [arg-type]
Found 2 errors in 1 file (checked 1 source file)
```

The first error comes from pybind11 generating `os.PathLike` from `fs::path`, however it should be `os.PathLike | str` since C++ supports implicit conversion from `std::string` to `fs::path`.
So I think this is a bug in pybind11.

We could change `cpp/pybind/pybind_filesystem.h` (not sure how easy it is to play with template/macro magic there).
Or I think the easier route would be to change `cpp/pybind/docstring.cpp` to replace any `os.PathLike` in argument types to `os.PathLike | str`.

The second error involves this signature in the stubs:
```python
def paint_uniform_color(self, arg0: numpy.ndarray[numpy.float64[3, 1]]) -> MeshBase:
...
```
pybind11 uses `numpy.ndarray[numpy.float64[3, 1]]` for `Eigen::Vector3d`.
Apparently there is also some implicit conversion going on, but I cannot easily find it in the pybind11/Eigen code.
A solution using [numpy.typing.ArrayLike](https://numpy.org/devdocs/reference/typing.html#numpy.typing.ArrayLike) would probably be nice, however shape info would be lost.
Not sure how to proceed here.

timohl:
288d51d also seems to cause an error in CI:
```bash
  File "<string>", line 1, in <module>
  File "/home/runner/work/Open3D/Open3D/open3d_test.venv/lib/python3.11/site-packages/open3d/__init__.py", line 61, in <module>
    load_cdll(str(next((Path(__file__).parent / "cpu").glob("pybind*"))))
  File "/home/runner/work/Open3D/Open3D/open3d_test.venv/lib/python3.11/site-packages/open3d/__init__.py", line 36, in load_cdll
    raise FileNotFoundError(f"Shared library file not found: {path}.")
FileNotFoundError: Shared library file not found: /home/runner/work/Open3D/Open3D/open3d_test.venv/lib/python3.11/site-packages/open3d/cpu/pybind.
```
https://github.com/isl-org/Open3D/actions/runs/10647126362/job/29514998533?pr=6917

timohl:
[70b3f81](https://github.com/isl-org/Open3D/pull/6917/commits/70b3f81767df968d3b7e644efadbb515b3757817) fixes `fs::path` python typing by changing it from `os.PathLike` to `Union[os.PathLike, str]` which is translated to `os.PathLike | str` by pybind11-stubgen and sphinx-docs. It was actually easier to change than I previously thought.

The major amount of remaining mypy errors are implicit conversions from `list` to `np.ndarray`.
Having dimensions in docs is very nice (e.g., `numpy.ndarray[numpy.float64[3, 1]]`) but since shapes are not checked by mypy.
So, maybe [numpy.typing.ArrayLike](https://numpy.org/devdocs/reference/typing.html#numpy.typing.ArrayLike) could be used just in stubs and not docs.

Another common error is mypy not resolving imports correctly:
```bash
mypy examples/python/io/realsense_io.py 
examples/python/io/realsense_io.py:28: error: Cannot find implementation or library stub for module named "open3d.t.io"  [import-not-found]
examples/python/io/realsense_io.py:28: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
examples/python/io/realsense_io.py:28: error: Cannot find implementation or library stub for module named "open3d.t"  [import-not-found]
examples/python/io/realsense_io.py:29: error: Cannot find implementation or library stub for module named "open3d.t.geometry"  [import-not-found]
examples/python/io/realsense_io.py:30: error: Cannot find implementation or library stub for module named "open3d.visualization.gui"  [import-not-found]
examples/python/io/realsense_io.py:31: error: Cannot find implementation or library stub for module named "open3d.visualization.rendering"  [import-not-found]
examples/python/io/realsense_io.py:69: error: "Callable[[Any, Any, Any, Any, Any], Any]" has no attribute "fid"  [attr-defined]
Found 6 errors in 1 file (checked 1 source file)
```
mypy is able to `import open3d as o3d`, but any import like  `from open3d.t.geometry import PointCloud` is not getting resolved.
Not sure what the source of those import errors are, but maybe I can find some info at https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports

timohl:
Some notes (mostly for myself, but happy for any help/advice) regarding implicit list -> np.ndarray conversions causing mypy errors like this:
```python
mypy examples/python/geometry/triangle_mesh_with_numpy.py 
examples/python/geometry/triangle_mesh_with_numpy.py:15: error: Argument 1 to "paint_uniform_color" of "TriangleMesh" has incompatible type "list[float]"; expected "ndarray[Any, Any]"  [arg-type]
Found 1 errors in 1 file (checked 1 source file)
```

### list is not always convertible to np.ndarray, but only when using Eigen types
Example:
```python
import numpy as np
import open3d as o3d

#[1] Uses Eigen::Vector3d
bb = o3d.geometry.AxisAlignedBoundingBox([0, 0, 0], np.array([1, 1, 1]))
print(bb)
#[2] Uses explicit overloads for py::list
t1 = o3d.core.Tensor([0, 1.5, 0], dtype=o3d.core.float32)
print(t1)
#[3] Same as [2]. shape is a ambigous here, since this will not set the shape but initilize the Tensor with those values
t2 = o3d.core.Tensor(shape=[0, 1, 0], dtype=o3d.core.float32)
print(t2)
#[4] Uses py::array -> Error when passing list
t3 = o3d.core.Tensor.from_numpy([0, 5, 0])
print(t3)
```
Output:
```bash
AxisAlignedBoundingBox: min: (0, 0, 0), max: (1, 1, 1)
[0 1.5 0]
Tensor[shape={3}, stride={1}, Float32, CPU:0, 0x5623bc51ff80]
[0 1 0]
Tensor[shape={3}, stride={1}, Float32, CPU:0, 0x5623bc562150]
  File "/home/timohl/projects/pybind11-stubgen/o3d_test.py", line 10, in <module>
    t3 = o3d.core.Tensor.from_numpy([0, 5, 0])
TypeError: from_numpy(): incompatible function arguments. The following argument types are supported:
    1. (arg0: numpy.ndarray) -> open3d.cpu.pybind.core.Tensor

Invoked with: [0, 5, 0]
```

### How could annotation using ArrayLike for Eigen types look like?
Using imports this is a more compact view and could easily be displayed in sphinx docs like this:
```python
import numpy as np
import numpy.typing as npt
from typing import Annotated

def f1(arg0: np.ndarray[np.float64[3, 1]] | npt.ArrayLike) -> None: ...
def f2(arg0: Annotated[npt.ArrayLike, np.float64[3, 1]]) -> None: ...
```
Right now I would tend to `f2`.

### How to insert ArrayLike annotations?
I am not sure, however, how to get there, so that only Eigen Arguments and not Return values are using ArrayLike.
* Wrapping all occurences of Eigen types in the o3d binding code seems like a lot of work.
* Customizing `pybind11/eigen/matrix.h` could be an option, but it is pretty complicated template magic code and diverging from pybind11 could be bad for updates.
* Another option would be to allow lists for every np.ndarray argument. Then changing to ArrayLike would be easily possible as a post-processing step after pybind11 compilation. But then all functions taking arrays must be made compatible with lists.

timohl:
I have opened a pull request at pybind/pybind11#5358 to add a feature for adding different type hints for args and return types.
That PR would be the cleanest solution in my opinion to resolve a lot of mypy issues.
Hope it gets merged soon.

timohl:
Just a quick update:
I do not have so much time this month but I hopefully can start working on this again next month.

Some notes on what to do next:
The base idea of [pybind/pybind11#5358](https://github.com/pybind/pybind11/pull/5358) was good but it was hard to get all edge cases within the template magic used in pybind11.
I think a simpler scope would be easier to implement and test:
Just support specialized arg+return type hints for types that are not template parameters.
Something like std::vector<std::filesystem::path> or std::vector<std::vector<std::filesystem::path>> is hard to get right in a generalized manner.
So, using specialized arg+return type hints for just std::filesystem::path and falling back to the default type hints for (nested) templates like std::vector<std::filesystem::path> would be a safe way to add this without breaking a lot of other stuff.

With specialized arg+return type hints most mypy errors should be easy to fix and generates more precise documentation on what types are actually valid for function parameters.

timohl:
My pybind11 pull request #5450 to fix the type hint for `fs::path` is making good progress and lays the foundation to improve typing of Eigen/Numpy classes.
That part might take more time, though.

Meanwhile, I could create some smaller pull requests to already merge some fixes, that also improve the documentation.
The following fixes could already improve the documentation:
* A lot of default arguments are rendered as `(with default value)` and have kind of ugly values in the text, e.g.:
  * [o3d.t.geometry.PointCloud.create_from_depth_image](https://www.open3d.org/docs/release/python_api/open3d.t.geometry.PointCloud.html#open3d.t.geometry.PointCloud.create_from_depth_image) where maybe `Tensor.eye()` could be used
  * [o3d.geometry.TriangleMesh.filter_sharpen](https://www.open3d.org/docs/release/python_api/open3d.geometry.TriangleMesh.html#open3d.geometry.TriangleMesh.filter_sharpen) where the parameters in the doc text are not displayed correctly
* `robust` param in [o3d.t.geometry.OrientedBoundingBox.create_from_points](https://www.open3d.org/docs/0.18.0/python_api/open3d.t.geometry.OrientedBoundingBox.html#open3d.t.geometry.OrientedBoundingBox.create_from_points) is weird because of a bug in `cpp/pybind/docstring.cpp`
* [o3d.geometry.PointCloud.orient_normals_consistent_tangent_plane](https://www.open3d.org/docs/release/python_api/open3d.geometry.PointCloud.html#open3d.geometry.PointCloud.orient_normals_consistent_tangent_plane) uses lambda as argument name, but lambda is a reserved keyword in Python

Maybe it would be easier to merge multiple smaller PRs than just one big PR.
There is also #6896 including some fixes.
@ssheorey what do you think about that?

timohl:
My pull request pybind/pybind11#5450 fixing `fs::path` type hints was successfully merged into pybind11. :)
Next, I will try to work correct type hints for np.array/Eigen into pybind11.

I also created some smaller pull requests for commits that change the Python API.
That way, the scope of this PR would only be to integrate stubgen into the packaging and CI.
For now, until the smaller PRs are merged and the correct np.array/Eigen typing is in pybind11, I will convert this PR to a draft.

timohl:
### Latest progress update:
I managed to get rid of all mypy errors from the examples in my fork [timohl/Open3D:pybind11-test](https://github.com/timohl/Open3D/tree/pybind11-test) 🎉 
```bash
> mypy --ignore-missing-imports examples/python/
Success: no issues found in 143 source files
```

For that, I worked on pybind11 to improve typing.
My currently open PR [pybind11/pybind11#5486](https://github.com/pybind/pybind11/pull/5486) lays the foundation for that.
After this, I am planning a final PR to actually change numpy typing in pybind11.

Then this PR can be reopened.

pablovela5620:
Other than waiting for this PR to land, is the only way to get typehints to work with an LSP like vscode is to manually build Open3D and generate stubs onself?

timohl:
You can already get partial stubs working without compiling:
```bash
pip install open3d pybind11-stubgen # Make sure to first create/activate your venv
pybind11-stubgen open3d -o typings
```
The default path for local stubs in vscode is the `typings` directory inside your project.
You can change that in the vscode settings here: `python.analysis.stubPath`.

Make sure to install latest open3d (v0.19) since it includes a lot of fixes for stubgen.

Some annotations are still not correct though.
For example the type checker will have a problem when passing lists to args expecting `np.ndarray` (although it works and should be annotated with `npt.ArrayLike`).
Another problem is passing strings where paths are expected (also works but type checker complains).

I have a branch in my fork which includes more fixes, but uses a modified pybind11 (with the fixes from my in-progress PRs).
[timohl/Open3D:pybind11-test](https://github.com/timohl/Open3D/tree/pybind11-test)
You could clone, compile and run stubgen.
If you are interested feel free to ask and I can help.

timohl:
Latest progress on pybind11 can be followed in pybind/pybind11#5212 and pybind/pybind11#5502
This should be the last required fix in pybind11, but is still in development.

timohl:
The final pybind11 PR pybind/pybind11#5212 got merged. 🎉 
I will start working on updating this PR on the weekend, so typing support can finally be included.

ssheorey:
Hi @timohl how are things? Please let me know if you need any help.

timohl:
Sorry, I was and still am pretty busy with writing my master thesis (deadline in under a month).
I am planning on finishing #7249 this weekend (already have some fixes almost ready).
Pybind11 released [v3.0.0rc3](https://github.com/pybind/pybind11/releases/tag/v3.0.0rc3) two weeks ago.
Should we wait for the full release of v3.0.0 before merging #7249 (assuming everything works)?

After #7249, I can continue here, but probably not before mid next month.

timohl:
Hi, I'm picking this up again. I've rebased onto current `main` and simplified the approach:

- **Stub generation is now disabled by default** (`WITH_STUBGEN=OFF`), so this has zero impact on existing builds and CI.
- **No CI workflow changes** — this PR only adds the build system plumbing. CI changes can follow in a separate PR once the feature is validated more broadly.
- When opted in via `-DWITH_STUBGEN=ON`, `pybind11-stubgen` runs after the Python package is assembled and generates `.pyi` stubs + a `py.typed` marker, both included in the wheel.
- `IGNORE_STUBGEN_ERRORS=ON` (default) tolerates stubgen warnings for complex pybind modules; set to `OFF` to fail the build on any stubgen error.
- **requirements_stubgen.txt** contains the pybind11-stubgen dependency separate from the other requirements
- **Fixed small export issue** in `__init__.py`. It was exporting open3d which confuses pybind11 (should have no impact code importing open3d)

I've verified on Windows with VS 2022 / Python 3.12 that:
1. Default build (`WITH_STUBGEN=OFF`) — wheel builds normally, no stubs present.
2. `WITH_STUBGEN=ON` — wheel contains 40+ `.pyi` files covering all pybind modules and a `py.typed` marker.

The idea is to land this as a minimal, low-risk PR that lets others test typing support in their local builds, then follow up with CI integration separately.

ssheorey (approved):

ssheorey:
Thanks @timohl !

PR #6898, opened 2024-07-28
Fix minimal oriented bounding box of MeshBase derived classes and add new unit tests

nicolaloi:
## Type

-   [x] Bug fix (non-breaking change which fixes an issue): Fixes #6866
-   [ ] New feature (non-breaking change which adds functionality). Resolves #
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected) Resolves #

## Motivation and Context

Fixing the [`MeshBase::GetMinimalOrientedBoundingBox`](https://github.com/isl-org/Open3D/blob/v0.18.0/cpp/open3d/geometry/MeshBase.cpp#L53-L55) method, which did not return the minimal oriented bounding box. Also adding new unit tests for the derived classes `TriangleMesh` and `TetraMesh` for the different bounding box generation algorithms, since they were not covered.

## Checklist:

-   [x] I have run `python util/check_style.py --apply` to apply Open3D **code style**
    to my code.
-   [x] This PR changes Open3D behavior or adds new functionality.
    -   [ ] Both C++ (Doxygen) and Python (Sphinx / Google style) **documentation** is
        updated accordingly.
    -   [x] I have added or updated C++ and / or Python **unit tests** OR included **test
        results** (e.g. screenshots or numbers) here.
-   [x] I will follow up and update the code if CI fails.
    
-   [x] For fork PRs, I have selected **Allow edits from maintainers**.

## Description

The [`MeshBase::GetMinimalOrientedBoundingBox`](https://github.com/isl-org/Open3D/blob/v0.18.0/cpp/open3d/geometry/MeshBase.cpp#L53-L55) method was calling `OrientedBoundingBox::CreateFromPoints` (so same functionality as [MeshBase::GetOrientedBoundingBox](https://github.com/isl-org/Open3D/blob/v0.18.0/cpp/open3d/geometry/MeshBase.cpp#L49-L51)) instead of `OrientedBoundingBox::CreateFromPointsMinimal` (as done in [PointCloud::GetMinimalOrientedBoundingBox](https://github.com/isl-org/Open3D/blob/v0.18.0/cpp/open3d/geometry/PointCloud.cpp#L55-L58)). I have also added new unit tests for the _minimal oriented bounding box_, _oriented bounding box_, _axis-aligned bounding box_, and _get center_ for both `TriangleMesh` and `TetraMesh` objects.

#### Testing

Testing code (in the visualization the bounding boxes of the meshes are slightly enlarged for better visibility)

```python
import numpy as np
[...]

PR #6946, opened 2024-09-03
Expose minimal oriented bounding box to python api

sharon-br:
This PR exposes the minimal volume bounding box function to the python API.

-   [X] New feature (non-breaking change which adds functionality). Resolves #

## Motivation and Context

Only the static function `create_from_points` is currently exposed to the Python API.
However, in the CPP API, another option exists which is to create a bounding box with a minimal volume that is not exposed.

## Checklist:

    to my code.
    -   [X] Both C++ (Doxygen) and Python (Sphinx / Google style) **documentation** is
        updated accordingly.
    
### DIFF
--- a/cpp/pybind/geometry/boundingvolume.cpp
+++ b/cpp/pybind/geometry/boundingvolume.cpp
@@ -84,7 +84,7 @@ Returns:
      bounding box is oriented such that the axes are ordered with respect to
      the principal components.
 )doc")
-            .def_static("create_minimal_from_points",
+            .def_static("create_from_points_minimal",
                         &OrientedBoundingBox::CreateFromPointsMinimal,
                         "points"_a, "robust"_a = false,
                         R"doc(
--- a/cpp/pybind/geometry/boundingvolume.cpp
+++ b/cpp/pybind/geometry/boundingvolume.cpp
@@ -84,6 +84,27 @@ Returns:
      bounding box is oriented such that the axes are ordered with respect to
      the principal components.
 )doc")
+            .def_static("create_from_points_minimal",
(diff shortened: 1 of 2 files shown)


sharon-br:
I was unsure how to name the function - another option was `create_from_points_minimal` to match the CPP function name but it felt odd so I went with this. Happy to change if you have any other preference.

Also, any plans on 0.19 to include this and #6898 ?

cpp/pybind/geometry/boundingvolume.cpp
```diff
@@ -83,6 +83,27 @@ The returned bounding box is an approximation to the minimal bounding box.
      open3d.geometry.OrientedBoundingBox: The oriented bounding box. The
      bounding box is oriented such that the axes are ordered with respect to
      the principal components.
+)doc")
+            .def_static("create_minimal_from_points",
```
ssheorey, line 87: Can we name `create_from_points_minimal` instead so that it matches the C++ function?

ssheorey (changes requested):
Thanks @sharon-br ! Looks good after a rename.

cpp/pybind/geometry/boundingvolume.cpp
sharon-br: Done. Is there any timeline regarding the 0.19 release?

ssheorey (approved):

cpp/pybind/geometry/boundingvolume.cpp
ssheorey: As soon as we can get the Python 3.12 support PR merged. There are still some test failures blocking that.

PR #6948, opened 2024-09-04
Add Tensor-based implementation of FarthestPointDownSample

manuelvogel12:
This PR adds a FarthestPointDownSample implementation for tensor-based point clouds. This enables GPU-acceleration for a performance increase of up to 8x. Before, point clouds were copied to the legacy format and then being downsampled on the CPU before being copied back.

## Motivation and Context

FarthestPointDownSample is an important downsampling method, however it's performance render this implementation inapplicable for certain scenarios. This PR is trying to increase the performance

## Checklist:

    to my code.
        updated accordingly.
    
### DIFF
--- a/cpp/open3d/t/geometry/PointCloud.cpp
+++ b/cpp/open3d/t/geometry/PointCloud.cpp
@@ -384,11 +384,47 @@ PointCloud PointCloud::RandomDownSample(double sampling_ratio) const {
 }
 
 PointCloud PointCloud::FarthestPointDownSample(size_t num_samples) const {
-    // We want the sampled points has the attributes of the original point
-    // cloud, so full copy is needed.
-    const open3d::geometry::PointCloud lpcd = ToLegacy();
-    return FromLegacy(*lpcd.FarthestPointDownSample(num_samples),
-                      GetPointPositions().GetDtype(), GetDevice());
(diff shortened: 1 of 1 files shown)


manuelvogel12:
I use the legacy method when dealing with CPU-Pointclouds, because my tensor-based implementation is roughly 10x slower on the CPU than the legacy version. 

However, with PR https://github.com/isl-org/Open3D/pull/6989, the performance is only 4 times slower (compared to the legacy implementation).

It can be further sped up to 2.5 times slower (compared to the legacy implementation) by replacing

```
core::Tensor diff = GetPointPositions() - selected;
core::Tensor distances_to_selected = (diff * diff).Sum({1});
```
with
```
core::Tensor diff = GetPointPositions() - selected;
diff *= diff;
core::Tensor distances_to_selected = diff.IndexExtract(1,0) + diff.IndexExtract(1,1) + diff.IndexExtract(1,2);
```

Therefore, I think it makes sense to remove the `if (IsCUDA())` condition and also run this version for CPU-Pointclouds, because with the legacy version, properties such as the intensity are lost, and the dtype of other properties might be changed.

benjaminum:
Thanks @manuelvogel12 for the update!

benjaminum (approved):

PR #6964, opened 2024-09-12
MSVC Ninja fixes

dbs4261:
## Type

## Motivation and Context

Not all windows builds use the MSVC generator, for example the CLion can be used as an IDE and dispatch to the MSVC compiler. MSVC creates different names for some libraries eg. `libpng-d.lib`. This does not happen with the Ninja generator, so these debug libraries cannot be found when using a different generator.

## Checklist:

    to my code.
        updated accordingly.
    
## Description

This PR has cmake check the generator instead of the compiler to determine if these different names should be added. Also updated the tbb target name that is copied for the `test.exe` target to its namespaced version TBB::tbb.


dbs4261:
I don't think these changes are worth updating the changelog, but I'm open to other opinions.

dbs4261:
Hmmm, the rabbit hole goes deeper than I expected. This is not ready to merge yet.

ssheorey:
Hi @dbs4261 do you plan to continue working on this? This can be quite useful since it will allow us to use ccache compiler caching for Windows builds as well.
Also please see the recent updates in Windows CI - we added support for Intel (LLVM) compiler + Ninja on Windows.

PR #6966, opened 2024-09-13
Replace conda with pyenv to fix incorrect libstdc++ use in jammy CI.

ssheorey:
## Type

## Motivation and Context

CUDA jammy CI (python tests) fail since pytest causes the conda libstdc++ library to be loaded instead of the system libstdc++ library. This is older and prevents Open3D cuda pybind lib from loading.

## Checklist:

    to my code.
        updated accordingly.
    
## Description

- Replace python install from conda with pyenv python install. This uses the system libraries.
- Use RPATH instead of RUNPATH to load libc++abi.so directly in Python. No need to find and load explicitly for filament.
- Use libc++ v11 to build in Ubuntu 22.04. Warn if newer version is used. libc++ v12 and later use LLVM libunwind, which is incompatible with system libunwind and causes the Python crash.

**TODO**: Solution for Ubuntu 24.04 (newer libc++ in general).

### DIFF
--- a/docker/Dockerfile.ci
+++ b/docker/Dockerfile.ci
@@ -68,6 +68,8 @@ RUN if [ "${BUILD_SYCL_MODULE}" = "ON" ]; then \
     fi
 
 # Dependencies: basic and python-build
+# gcc-11 causes a seg fault in tensormap when built in Release mode. Upgrade to
+# gcc-13 instead.
 RUN apt-get update && apt-get install -y \
     git  \
     wget \
@@ -87,11 +89,16 @@ RUN apt-get update && apt-get install -y \
     libxmlsec1-dev \
     libffi-dev \
     liblzma-dev \
+ && if c++ --version | grep -F ' 11.' ; then \
+        apt-get install -y g++-13 \
+        && update-alternatives --install /usr/bin/c++ c++ /usr/bin/g++-13 100 \
+        && c++ --version; \
+    fi \
  && rm -rf /var/lib/apt/lists/*
 
 # pyenv or Intel Python
 # The pyenv python paths are used during docker run, in this way docker run
-# does not need to activate the environment again. 
+# does not need to activate the environment again.
 # The soft link from the python patch level version to the python mino version
 # ensures python wheel commands (i.e. open3d) are in PATH, since we don't know
 # which patch level pyenv will install (latest).
--- a/docker/Dockerfile.ci
+++ b/docker/Dockerfile.ci
@@ -68,8 +68,6 @@ RUN if [ "${BUILD_SYCL_MODULE}" = "ON" ]; then \
     fi
 
 # Dependencies: basic and python-build
-# gcc-11 causes a seg fault in tensormap when built in Release mode. Upgrade to
-# gcc-12 instead.
 RUN apt-get update && apt-get install -y \
     git  \
     wget \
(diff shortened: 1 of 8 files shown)


ssheorey:
Seg fault in python `test_tensormap.py` here when built with gcc 11 (Ubuntu 20.04 jammy) in release config. No seg fault in debug or relwithdebinfo configs.

```py
...
def test_tensormap(device):
...
    # __delitem__ operator. SEGFAULT
    with pytest.raises(RuntimeError) as excinfo:
        del tm.positions
        assert 'cannot be deleted' in str(excinfo.value)
...
   # Set primary key. SEGFAULT
    with pytest.raises(KeyError) as e:
        tm.primary_key = o3c.Tensor.ones((2, 3), dtype, device)
...
    # Get unknown attributes. SEGFAULT
    with pytest.raises(KeyError) as e:
        normals = tm.normals
```

Fix: switch to gcc 13.

Update nanoflann to fix this warning as error:
```sh
    inlined from ‘open3d::core::nns::impl::{anonymous}::_KnnSearchCPU<float, int, open3d::core::nns::NeighborSearchAllocator<float, int>, 1>(open3d::core::nns::NanoFlannIndexHolderBase*, int64_t*, size_t, const float*, size_t, const float*, size_t, int, bool, bool, open3d::core::nns::NeighborSearchAllocator<float, int>&)::<lambda(const float*, const float*, size_t)>’ at /home/ssheorey/Documents/Open3D/Code/Open3D/cpp/open3d/core/nns/NanoFlannImpl.h:126:5,
    inlined from ‘open3d::core::nns::impl::{anonymous}::_RadiusSearchCPU<float, int, open3d::core::nns::NeighborSearchAllocator<float, int>, 1>(open3d::core::nns::NanoFlannIndexHolderBase*, int64_t*, size_t, const float*, size_t, const float*, size_t, const float*, bool, bool, bool, bool, open3d::core::nns::NeighborSearchAllocator<float, int>&)::<lambda(const tbb::detail::d1::blocked_range<long unsigned int>&)>’ at /home/ssheorey/Documents/Open3D/Code/Open3D/cpp/open3d/core/nns/NanoFlannImpl.h:258:41:
/usr/include/c++/13/bits/new_allocator.h:168:33: error: ‘void operator delete(void*, std::size_t)’ called on pointer ‘<unknown>’ with nonzero offset [4, 9223372036854775804] [-Werror=free-nonheap-object]
```

benjaminum (approved):
lgtm

ssheorey:
Seg fault with Python, CUDA 11.8, 12.1 and Ubuntu 22.04. No issue with C++, the open3d-cpu package, or if CUDA_VISIBLE_DEVICES is empty. No issue on Ubuntu 20.04 or Windows.

docker/Dockerfile.ci
```diff
@@ -67,34 +67,55 @@ RUN if [ "${BUILD_SYCL_MODULE}" = "ON" ]; then \
         rm -rf /etc/apt/sources.list.d/oneAPI.list; \
     fi
 
-# Dependencies: basic
+# Dependencies: basic and python-build
+# gcc-11 causes a seg fault in tensormap when built in Release mode. Upgrade to
```
benjaminum, line 71: Have we found the root cause for that or is it really a bug only due to gcc11?

benjaminum (approved):
approved. See question comment

docker/Dockerfile.ci
ssheorey: Thanks for catching that. I think it was actually the libunwind issue. I've been working with gcc-11 on jammy (22.04) without issues. Reverted.

PR #6969, opened 2024-09-15
fix build with fmt-v11

daizhirui:
`fmt v11` requires that `format` function of user-defined formatter should be `const`.


ssheorey (approved):
Thanks @daizhirui . Looks good!

PR #6989, opened 2024-09-26
Faster CPU (Arg-)Reductions

manuelvogel12:
This pull request aims to speed up Reductions and ArgReductions with single outputs on the CPU.

## Type

## Motivation and Context

For Reductions (e.g. sum), the problem of the current code is the access of multiple threads to a shared array in each loop iteration, which can be fixed using a single local variable.

Argreductions for single outputs (e.g. points.argmax()) run with the current code on a single CPU core, although the operation can be sped up in the same way as the reduction.

This PR fixes both problems and has a 400% speedup for these reductions (4x faster) and 2000% speedup for these argreductions (20x faster) on my PC.  

Note that this does only apply for single-output operations performed on the CPU.
## Checklist:

    to my code.
        updated accordingly.
    
## Description
The benchmarking was done on a 24core cpu, using the sum() and argmax() functions of arrays with 150M elements.

### DIFF
--- a/cpp/open3d/core/kernel/ReductionCPU.cpp
+++ b/cpp/open3d/core/kernel/ReductionCPU.cpp
@@ -122,13 +122,14 @@ private:
         for (int64_t thread_idx = 0; thread_idx < num_threads; ++thread_idx) {
             int64_t start = thread_idx * workload_per_thread;
             int64_t end = std::min(start + workload_per_thread, num_workloads);
+            scalar_t local_result = identity;
             for (int64_t workload_idx = start; workload_idx < end;
                  ++workload_idx) {
                 scalar_t* src = reinterpret_cast<scalar_t*>(
                         indexer.GetInputPtr(0, workload_idx));
-                thread_results[thread_idx] =
-                        element_kernel(*src, thread_results[thread_idx]);
+                local_result = element_kernel(*src, local_result);
             }
+            thread_results[thread_idx] = local_result;
         }
         scalar_t* dst = reinterpret_cast<scalar_t*>(indexer.GetOutputPtr(0));
         for (int64_t thread_idx = 0; thread_idx < num_threads; ++thread_idx) {
--- a/cpp/open3d/core/kernel/ReductionCPU.cpp
+++ b/cpp/open3d/core/kernel/ReductionCPU.cpp
@@ -191,14 +191,25 @@ public:
         // elements. We need to keep track of the indices within each
         // sub-iteration.
         int64_t num_output_elements = indexer_.NumOutputElements();
+        if (num_output_elements <= 1) {
+            LaunchArgReductionKernelTwoPass(indexer_, reduce_func, identity);
+        } else {
+            LaunchArgReductionParallelDim(indexer_, reduce_func, identity);
+        }
+    }
 
+    template <typename scalar_t, typename func_t>
+    static void LaunchArgReductionParallelDim(const Indexer& indexer,
+                                              func_t reduce_func,
+                                              scalar_t identity) {
+        int64_t num_output_elements = indexer.NumOutputElements();
 #pragma omp parallel for schedule(static) \
         num_threads(utility::EstimateMaxThreads())
         for (int64_t output_idx = 0; output_idx < num_output_elements;
              output_idx++) {
             // sub_indexer.NumWorkloads() == ipo.
-            // sub_indexer's workload_idx is indexer_'s ipo_idx.
-            Indexer sub_indexer = indexer_.GetPerOutputIndexer(output_idx);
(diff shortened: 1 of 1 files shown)


cpp/open3d/core/kernel/ReductionCPU.cpp
```diff
@@ -190,14 +191,25 @@ class CPUArgReductionEngine {
...
+    }
 
+    template <typename scalar_t, typename func_t>
+    static void LaunchArgReductionParallelDim(const Indexer& indexer,
+                                              func_t reduce_func,
+                                              scalar_t identity) {
+        int64_t num_output_elements = indexer.NumOutputElements();
 #pragma omp parallel for schedule(static) \
         num_threads(utility::EstimateMaxThreads())
         for (int64_t output_idx = 0; output_idx < num_output_elements;
```
ssheorey, line 208: Since usually `num_output_elements << num_threads`, we should collapse the two for loops to improve thread utilization with`#pragma omp for collapse(2)` I believe Visual Studio now supports this (OpenMP v3) but good to check Windows CI.

cpp/open3d/core/kernel/ReductionCPU.cpp
```diff
@@ -212,6 +224,52 @@ class CPUArgReductionEngine {
...
+                    "single-output arg reduction ops.");
+        }
+        int64_t num_workloads = indexer.NumWorkloads();
+        int64_t num_threads = utility::EstimateMaxThreads();
+        int64_t workload_per_thread =
+                (num_workloads + num_threads - 1) / num_threads;
+        std::vector<int64_t> thread_results_idx(num_threads, 0);
+        std::vector<scalar_t> thread_results_val(num_threads, identity);
+
+#pragma omp parallel for schedule(static) \
```
ssheorey, line 246: Can we simplify this with  for loop collapse in OpenMP?

ssheorey (commented):
Hi @manuelvogel12 thanks for this PR! Speeding up basic ops is really useful. Can I request some additional benchmarking? Essentially, we want to ensure there are no speed regressions for different use cases. An example of multiple scenarios:

number of points: 10K, 100K, 1M, 10M, 100M
number of CPU cores:  4, 8, 16, 24
[while counting cores, please ignore hyperthreading]

You can use (and update if needed) `cpp/benchmarks/core/Reduction.cpp` for benchmarking this. [Build with cmake option `-D BUILD_BENCHMARKS=ON`]

You can use `taskset --cpu-list 0-7 open3dbenchmark...` to restrict the benchmark to only CPU cores 0-7 inclusive. [This is 4 cores for hyperthreading CPUs].

cpp/open3d/core/kernel/ReductionCPU.cpp
benjaminum: I just want to remind about #6626 as this will add new OpenMP code

manuelvogel12:
Here are the benchmarking results. I always reduced along dim 0, with varying shapes.
Since this PR aims only at single-output reductions, only the tests with single shape inputs are relevant here (e.g. (100) or (10000))

This is for all cores (24cores, 48 threads) 
Screenshot from 2024-09-30 14-39-16

This is 4 cores with SMT (taskset --cpu-list 0-3,24-27)
Screenshot from 2024-09-30 14-39-28

This is 8 cores without SMT (taskset --cpu-list 0-7)
Screenshot from 2024-09-30 14-39-41

To interpret the results, I would say my claims earlier made (400% speedup for these reductions and 2000% speedup for these argreductions) are valid

cpp/open3d/core/kernel/ReductionCPU.cpp
manuelvogel12: When simply launching `LaunchArgReductionKernelTwoPass` inside the outer for loop, it improves the performance when reducing over the large_dim, but worsens the performance of reducing over the small dim. Therefore, this would require a more sophisticated way to determine how many threads we should launch in the inner loop.

The image shows the time, so `+4.5` means this approach would be 4.5 times slower in that benchmark, 
`-0.78` means roughly 4.5 times faster 
image

ssheorey: Thanks @manuelvogel12 for the detailed benchmarking. This looks good to me.

To address @benjaminum's point, we will be switching from OpenMP to TBB soon (PR #6626). This helps quite a bit with lack of good support for OpenMP on Windows and Apple, simplifies the code and also comes with the benefit of composable parallelism. We also expect (some) speedup from TBB's work stealing scheduler model.

I think we can merge this PR due to its speed boost for now and then switch this code to TBB as well along with the rest of the code. @manuelvogel12 I would encourage you to check out PR #6626 and help us test that.

PR #7022, opened 2024-10-22
Update ZLib 3rdparty library to use ZLib-ng and Minizip-ng so patching is not necessary

dbs4261:
## Type

## Motivation and Context

Open3D has used a patched version of ZLib that was patched to also build minizip. This causes problems when including Open3D in a larger project if some other dependency provides a version of ZLib that isn't patched before Open3D's provider is called. The current approach of WITH_MINIZIP is not descriptive enough, firstly because it can only be used with a system installed minizip package, and secondly the option is not connected to using a system version of ZLib. With this change, external dependency providers like the system package manager or VCPKG could provide one or both of these dependencies. If only a system zlib is provided, minizip will be built from source. This patch uses the NG versions of these libraries in comparability mode purely because the updated cmake build system is easier to integrate with. Existing builds using the WITH_MINIZIP option will not break as minizip will be provided by default, however the unused option will produce a warning from cmake.

## Checklist:

    to my code.
        updated accordingly.
    
## Description

This patch changes the ZLib provider to provide the ZLib-ng library in compatibility mode without any patch. Additionally it adds a Minizip-ng provider also in compatibility mode. Both of these providers are now optional, and can instead use `find_package(...)` if USE_SYSTEM_ZLIB and USE_SYSTEM_MINIZIP are turned on respectively.


3rdparty/find_dependencies.cmake
```diff
@@ -853,6 +853,25 @@ if (BUILD_LIBREALSENSE)
     endif()
 endif()
 
+if(USE_SYSTEM_ZLIB)
+    open3d_find_package_3rdparty_library(3rdparty_zlib
+        PACKAGE ZLIB
+        TARGETS ZLIB::ZLIB
+    )
+    list(APPEND Open3D_3RDPARTY_PRIVATE_TARGETS_FROM_SYSTEM Open3D::3rdparty_zlib)
```
meetgandhi-dev, line 861: I think You should set `USE_SYSTEM_ZLIB` to `OFF` if the package can't be found.

dbs4261:
Good call, I'll make that update

ssheorey (changes requested):
Thanks @dbs4261. Looks good in general. We can merge after:
- addressing @meetgandhi-dev 's comment
- CI passes. Please check the Ubuntu (OFF, OFF) failing workflow - libpng seems to be looking for some zlib functions.
https://github.com/isl-org/Open3D/actions/runs/11466322539/job/31906749596?pr=7022

sr425:
@dbs4261 I like the effort to make build properly working on most recent MacOS. Do you need some assistance getting this MR running? Not too deep into the lib, but would be willing to try to contribute to get this merged if needed.

PR #7032, opened 2024-10-26
fix infinite loop in segment_plane if num_points < ransac_n

rxba:
## Type

-   [X] Bug fix (non-breaking change which fixes an issue): Fixes #

## Motivation and Context

Just a small fix, but if the number of points in the pointcloud is smaller than `ransac_n`, `segment_plane` is caught in an infinite loop and does not terminate.
It's a bit of an unlikely scenario, but I did stumble across it because I was fitting planes on clustering results, and one of those point clusters only had two points.

The reason for the infinite loop is because `RandomSampler` is called with `num_points`
https://github.com/isl-org/Open3D/blob/fcc396e494dd3f1204e53571f37400cac50976ef/cpp/open3d/geometry/PointCloudSegmentation.cpp#L164-L165

, so in the case of `num_points < ransac_n` the condition can never become true in `RandomSampler`:
https://github.com/isl-org/Open3D/blob/fcc396e494dd3f1204e53571f37400cac50976ef/cpp/open3d/geometry/PointCloudSegmentation.cpp#L37-L45

Code to reproduce:
```
import open3d as o3d
import numpy as np

num_points = 2
points = np.random.rand(num_points, 3)
pcd = o3d.geometry.PointCloud()
pcd.points = o3d.utility.Vector3dVector(points)
plane_model, inliers = pcd.segment_plane(distance_threshold=0.01,
                                         ransac_n=3,
                                         num_iterations=1000)
```

## Checklist:

    to my code.
        updated accordingly.
    
## Description

* I just moved the already existing checks for this to before the RandomSampler is called so this error can actually be caught
* I also removed some of the comments around this part of the code, as they are redundant in my opinion

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,7 @@
 -   Fix tensor EstimatePointWiseNormalsWithFastEigen3x3 (PR #6980)
 -   Fix alpha shape reconstruction if alpha too small for point scale (PR #6998)
 -   Fix render to depth image on Apple Retina displays (PR #7001)
+-   Fix infinite loop in segment_plane if num_points < ransac_n (PR #7032)
 
 ## 0.13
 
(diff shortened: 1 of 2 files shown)


benjaminum:
Thanks for this bugfix @rxba !

PR #7056, opened 2024-11-15
SamplePointsUniformly, Chamfer Distance, Hausdorff Distance and F-Score

ssheorey:
## Type

## Checklist:

    to my code.
        updated accordingly.
    
### Tensor API SamplePointsUniformly for TriangleMesh: 
   - Only CPU, not parallel.
   - Fairly fast: ~0.25s for 1 million points.
   - Use normals, albedo texture or colors from the mesh.
### F-Score:
 - Tanks and Temples: Benchmarking Large-Scale Scene Reconstruction.
 - TODO: Check correctness in Python unit test.
### Chamfer Distance, Hausdorff distance

Screenshot 2024-11-15 at 9 14 13 PM
Screenshot 2024-11-15 at 9 14 07 PM


cpp/open3d/t/geometry/kernel/Metrics.cpp
```diff
@@ -0,0 +1,58 @@
...
+                                         core::Tensor distance21,
+                                         std::vector<Metric> metrics,
+                                         MetricParameters params) {
+    std::vector<float> metric_values;
+    float metric_val;
+
+    for (Metric metric : metrics) {
+        switch (metric) {
+            case Metric::ChamferDistance:
+                metric_val = 0.5 *
```
benjaminum, line 26: I am still checking this but it looks like we should drop the 0.5 to be compatible with other libs

cpp/open3d/t/geometry/TriangleMesh.h
```diff
@@ -1007,6 +1008,31 @@ class TriangleMesh : public Geometry, public DrawableGeometry {
...
+    /// Compute various distances / metrics between two triangle meshes. This
+    /// uses ray casting for distance computations between a triangle mesh and a
+    /// sampled point cloud.
+    /// \param mesh2 Other point cloud to compare with.
+    /// \param metrics List of Metric s to compute. Multiple metrics can be
+    /// computed at once for efficiency.
+    /// \param params MetricParameters struct holds parameters required by
+    /// different metrics.
+    std::vector<float> ComputeDistance(
+            const TriangleMesh &mesh2,
```
benjaminum, line 1032: Should we warn that this function adds triangle.areas as a side effect?

benjaminum (changes requested):

cpp/open3d/t/geometry/kernel/Metrics.cpp
ssheorey: https://fwilliams.info/point-cloud-utils/sections/shape_metrics/
This uses 0.5.
Kaolin is customizable, but default is 1.
https://kaolin.readthedocs.io/en/latest/modules/kaolin.metrics.pointcloud.html

I think PDAL is a good standard to follow here, and that doesn't have the 0.5, so I took it off. Let me know what you think:

https://pdal.io/en/2.4.3/apps/chamfer.html

Also PDAL has Hausodorff, which seems common in point cloud analysis (perhaps not for ML). Added that as well.

cpp/open3d/t/geometry/TriangleMesh.h
ssheorey: Done.

cpp/open3d/t/geometry/PointCloud.cpp
```diff
@@ -1331,6 +1332,34 @@ int PointCloud::PCAPartition(int max_points) {
...
+
+    if (!tree2.KnnIndex()) {
+        utility::LogError("[ComputeDistance] Building KNN-Index failed!");
+    }
+    if (!tree1.KnnIndex()) {
+        utility::LogError("[ComputeDistance] Building KNN-Index failed!");
+    }
+
+    std::tie(indices12, distance12) = tree2.KnnSearch(points1, 1);
+    std::tie(indices21, distance21) = tree2.KnnSearch(points2, 1);
```
benjaminum, line 1358: I think these are the squared distances

cpp/open3d/t/geometry/kernel/Metrics.h
```diff
@@ -0,0 +1,22 @@
...
+
+#include <vector>
+
+#include "open3d/core/Tensor.h"
+#include "open3d/t/geometry/Geometry.h"
+namespace open3d {
+namespace t {
+namespace geometry {
+
+core::Tensor ComputeMetricsCommon(core::Tensor distance12,
```
benjaminum, line 16: Need to define here that distancexx is the Euclidean distance

benjaminum (changes requested):

cpp/open3d/t/geometry/kernel/Metrics.h
ssheorey: Comment added.

python/test/t/geometry/test_pointcloud.py
```diff
@@ -193,3 +193,26 @@ def test_pickle(device):
...
+    from open3d.t.geometry import TriangleMesh, PointCloud, Metric, MetricParameters
+    # box is a cube with one vertex at the origin and a side length 1
+    pos = TriangleMesh.create_box().vertex.positions
+    pcd1 = PointCloud(pos.clone())
+    pcd2 = PointCloud(pos * 1.1)
+
+    # (1, 3, 3, 1) vertices are shifted by (0, 0.1, 0.1*sqrt(2), 0.1*sqrt(3))
+    # respectively
+    metric_params = MetricParameters(
+        fscore_radius=o3d.utility.FloatVector((0.01, 0.11, 0.15, 0.18)))
```
benjaminum, line 209: Do we need o3d.utility.FloatVector or can we pass a python list/tuple?

ssheorey: I think it's supposed to work automatically for STL containers and pybind11, but it doesn't work in Open3D for `std::vector<T>` due to some customizations. Going through a `deque` enabled it. :-(

benjaminum: Is the problem with std::vector related to Eigen?

ssheorey: Probably. We do some custom bindings for Eigen matrices / vectors as well as std::vector<int> and std::vector<double> That's likely disabling pybind11 default bindings. We should switch to pybind11's default bindings. [I don't know the original reason for customization, or if it's still needed.]

PR #7069, opened 2024-11-22
fix bug that cannot call torch/tf under open3d.ml

USTCKAY:
## Type

## Checklist:

    to my code.
        updated accordingly.
    
## Description

At present, we can only directly import ml module via `import open3d.ml.torch`, but cannot call torch/tf under `open3d.ml` module. 
e4aa23b249bf5fab17e84014c3997b16
This PR solve this problem by import torch if `BUILD_PYTORCH_OPS` is set as well as import tf if  `BUILD_TENSORFLOW_OPS` is set in `python\open3d\ml\__init__.py`.
After apply this change , torch can be find under `open3d.ml` when I built from source with `BUILD_PYTORCH_OPS` on.
image

### DIFF
--- a/python/open3d/ml/__init__.py
+++ b/python/open3d/ml/__init__.py
@@ -7,6 +7,7 @@
 
 import os as _os
 import open3d as _open3d
+from open3d import _build_config
 if _open3d.__DEVICE_API__ == 'cuda':
     from open3d.cuda.pybind.ml import *
 else:
@@ -16,3 +17,7 @@ from . import configs
 from . import datasets
 from . import vis
 from . import utils
+if _build_config["BUILD_TENSORFLOW_OPS"]:
+    from . import tf
+if _build_config["BUILD_PYTORCH_OPS"]:
+    from . import torch


USTCKAY:
@benjaminum @ssheorey Hi, could you please help to take a review?

ssheorey:
Hi @USTCKAY not importing torch / tf is intentional. Otherwise you will get an import error if torch is not installed. torch is an optional dependency.

USTCKAY:
> Hi @USTCKAY not importing torch / tf is intentional. Otherwise you will get an import error if torch is not installed. torch is an optional dependency.

@ssheorey  Yes, I know that, so I only import torch when BUILD_PYTORCH_OPS is set because I suppose torch is installed under this condition. However I found the release whl also set BUILD_PYTORCH_OPS true at the same time torch is not in requirments.txt lol.
image
Maybe we should add torch/tf to requirements.txt when BUILD_PYTORCH_OPS/BUILD_TENSORFLOW_OPS is on.

PR #7072, opened 2024-11-23
Copy SampleCount when making a copy of FilamentView

patrikhuber:
If that function doesn't copy the sample count, the copied view will not be an exact copy - it'll contain the default sample count (=4), and not the sample count of the view it's copying from.

## Type

-   [x] Bug fix (non-breaking change which fixes an issue): Fixes - separate issue not created, just see above problem & fix

## Motivation and Context

See above.

## Checklist:

    to my code.
        updated accordingly.

## Description

See above.


ssheorey (approved):
Looks good @patrikhuber  Thanks!

PR #6714, opened 2024-03-24
Update librealsense to v2.54.2

sitic:
## Type

-   [ ] Bug fix (non-breaking change which fixes an issue): Fixes #
-   [ ] New feature (non-breaking change which adds functionality). Resolves #
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected) Resolves #

## Motivation and Context

Update to latest librealsense version for PR
- https://github.com/isl-org/Open3D/pull/6695

## Checklist:

-   [x] I have run `python util/check_style.py --apply` to apply Open3D **code style**
    to my code.
-   [ ] This PR changes Open3D behavior or adds new functionality.
    -   [ ] Both C++ (Doxygen) and Python (Sphinx / Google style) **documentation** is
        updated accordingly.
    -   [ ] I have added or updated C++ and / or Python **unit tests** OR included **test
        results** (e.g. screenshots or numbers) here.
-   [x] I will follow up and update the code if CI fails.
    
-   [x] For fork PRs, I have selected **Allow edits from maintainers**.

## Description

I've browed through the librealsense [Release Notes](https://github.com/IntelRealSense/librealsense/wiki/Release-Notes#release-2542) and didn't see any relevant breaking changes listed.

PR #7074, opened 2024-11-24
Patch librealsense v2.44.0 to include `<chrono>` for `system_clock`

StephanTLavavej:
I work on Microsoft Visual C++, where we regularly build popular open-source projects, including yours, with development builds of our compiler and libraries to detect and prevent shipping regressions that would affect you. This also allows us to provide advance notice of breaking changes, which is the case here.

I just merged https://github.com/microsoft/STL/pull/5105, which revealed a conformance issue in librealsense used by Open3D.

Compiler error with this STL change:
```
C:\gitP\isl-org\Open3D\build_amd64\librealsense\src\ext_librealsense\src\backend.h(108,79): error C2039: 'system_clock': is not a member of 'std::chrono'
```

librealsense was assuming that including `<thread>` makes the `chrono::system_clock` type available, which is not guaranteed by the C++ Standard.

I'm adding a patch to properly include `<chrono>` and updating librealsense.cmake to apply it.

(Note that upstream librealsense needs a different patch, for which I've created https://github.com/IntelRealSense/librealsense/pull/13537. This would be relevant if you merged an update like #6714.)

I've verified that building Open3D with microsoft/STL `main` fails without this patch and succeeds with it.


ssheorey (approved):
Thanks @StephanTLavavej !

PR #7094, opened 2024-12-12
ComputeAdjacencyList Assignment Open3D

adityaranigaon:
* Implement C++ function
* Add gtests
* Add pybind
* Add pytests

### DIFF
--- a/cpp/open3d/t/geometry/TriangleMesh.cpp
+++ b/cpp/open3d/t/geometry/TriangleMesh.cpp
@@ -18,6 +18,7 @@
 #include <Eigen/Core>
 #include <string>
 #include <unordered_map>
+#include <set>
 
 #include "open3d/core/CUDAUtils.h"
 #include "open3d/core/Device.h"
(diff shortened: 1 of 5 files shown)


cpp/open3d/t/geometry/TriangleMesh.cpp
```diff
@@ -65,6 +66,54 @@ TriangleMesh::TriangleMesh(const core::Tensor &vertex_positions,
...
+        return std::make_pair(core::Tensor(), core::Tensor());
+    }
+
+    if(!HasTriangleIndices()) {
+        utility::LogWarning("TriangleMesh has no Indices. No attributes computed!");
+        return std::make_pair(core::Tensor(), core::Tensor());
+    }
+
+    core::Tensor tris_cpu =
+            GetTriangleIndices().To(core::Device()).Contiguous();
```
ssheorey, line 83: TODO: Also enable on cuda and sycl.

cpp/open3d/t/geometry/TriangleMesh.cpp
```diff
@@ -65,6 +66,54 @@ TriangleMesh::TriangleMesh(const core::Tensor &vertex_positions,
...
+
+    std::unordered_map< size_t, std::set<size_t> > adjacencyList;
+    auto insertEdge = [&adjacencyList](size_t s, size_t t){
+        adjacencyList[s].insert(t);
+    };
```
ssheorey, line 88: Edges are undirected. For a triangle mesh, 2 neighboring faces will share an edge. If they are oriented in-consistently, the directions will be opposite, i.e. if one face is [0, 1, 2], another may be [2, 1, 3]. In this case we will get duplicate entries in the adjacency list. Essentially, we just need the lower triangle of the adjacency matrix - the upper triangle is redundant.

ssheorey: Also, some tri meshes are not manifold (degenerate) - this can lead to situations where an edge is shared by 3 faces. Using a set as the value type does address this.

PS: Why do we need an `unordered_map<size_t, set>` instead of a simpler `vector<set>`?

ssheorey (changes requested):
Thanks @adityaranigaon , this looks good. Some thoughts:
- Can we change the name to `ComputeAdjacencyMatrix() / compute_adjacency_matrix()`. The adjacency list uses an equivalent but slightly different format. We have that as part of the Eigen API, so having the same name may confuse users.
- Do add these as triangle attributes as well. They are needed for other algorithms to be added later (e.g. mesh filtering).

cpp/pybind/t/geometry/trianglemesh.cpp
```diff
@@ -734,11 +734,14 @@ This function always uses the CPU device.
 
         o3d.visualization.draw([{'name': 'difference', 'geometry': ans}])
 )");
-
+    triangle_mesh.def("compute_adjacency_list", &TriangleMesh::ComputeAdjacencyList,
+    "Return Mesh Adjacency Matrix in CSR format using Triangle indices attribute.");
+
```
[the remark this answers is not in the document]
asrathore-ai: I am guessing that adjst() would be added to vertex attribute and adjv() to triangle attributes ? 

However the check HasVertexAttribute will fail because it checks the length of attribute is equal to the number of vertices, however as per CSR format the length of adjv is "num_vertices + 1"

cpp/open3d/t/geometry/TriangleMesh.cpp
asrathore-ai: yeah, vector would work. I will update

PR #7095, opened 2024-12-15
Changed o3c.Tensor.arange signature to be like np.arange.

timohl:
## Type

-   [x] Breaking change (fix or feature that would cause existing functionality to not work as expected) Resolves #

## Motivation and Context

The current signatures of `open3d.core.Tensor.arange` are not valid:
```python
arange(stop, dtype=None, device=None)
arange(start=None, stop, step=None, dtype=None, device=None)
```
in the second signature, `start` has a default value but is before `stop`.
See [The Python Language Reference](https://docs.python.org/3/reference/compound_stmts.html#function-definitions):

> If a parameter has a default value, all following parameters up until the “*” must also have a default value — this is a syntactic restriction that is not expressed by the grammar.

This is how numpy defines arange:
```python
@overload
def arange(  # type: ignore[misc]
    stop: _IntLike_co,
    /, *,
    dtype: None = ...,
    device: None | L["cpu"] = ...,
    like: None | _SupportsArrayFunc = ...,
) -> _1DArray[int, signedinteger[Any]]: ...
@overload
def arange(  # type: ignore[misc]
    start: _IntLike_co,
    stop: _IntLike_co,
    step: _IntLike_co = ...,
    dtype: None = ...,
    *,
    device: None | L["cpu"] = ...,
    like: None | _SupportsArrayFunc = ...,
) -> _1DArray[int, signedinteger[Any]]: ...
```
(see https://github.com/numpy/numpy/blob/b3ddf2fd33232b8939f48c7c68a61c10257cd0c5/numpy/_core/multiarray.pyi#L769-L786)

Note how `start` is not optional in the second signature.
Here, optional is not needed, because the first overload handles that case.

Also, it adds `/, *,` in the first signature and `*,` in the second.
`/` forces all args before to be positional-only and `*` forces all arguments after to be keyword-only.
In pybind this can be achieved using `py::pos_only()` and `py::kw_only()`.

## Checklist:

    to my code.
        updated accordingly.
    
## Description

This PR changes the signatures to (similar to numpy.arange):
```python
arange(stop, /, *, dtype=None, device=None)
arange(start, stop, step=None, dtype=None, *, device=None)
```
There is already an unittest for `o3c.Tensor.arange`: `test_arange` in `python/test/core/test_core.py`
The test still succeeds with this change.

However, existing code might break if it does not use `dtype` or `device` as a keyword argument, e.g., like this:
`o3c.Tensor.arange(2, o3c.int64)`
This must now be written as:
`o3c.Tensor.arange(2, dtype=o3c.int64)`
(which the unit test already does)

This is a fix used in #6917 since `mypy` would complain about a syntax error in stubs generated by `pybind11-stubgen`:
```bash
mypy examples/python/geometry/triangle_mesh_with_numpy.py 
/home/timohl/projects/Open3D/.venv/lib/python3.10/site-packages/open3d/cpu/pybind/core/__init__.pyi:732: error: non-default argument follows default argument  [syntax]
Found 1 error in 1 file (errors prevented further checking)
```


ssheorey (approved):
Looks good. Thanks @timohl

PR #7097, opened 2024-12-16
Changed draw_geometries to use default args instead of overload

timohl:
## Type

## Motivation and Context

Currently, `o3d.visualization.draw_geometries` uses overloading to provide default values for `lookat`, `up`, `front`, and `zoom`. 
This causes two problems:
* It defines a function with default arguments before non-default arguments (forbidden by [The Python Language Reference](https://docs.python.org/3/reference/compound_stmts.html#function-definitions))
* It clutters the documentation with redundancy (see [docs/0.18.0/python_api/open3d.visualization.draw_geometries](https://www.open3d.org/docs/0.18.0/python_api/open3d.visualization.draw_geometries.html))
## Checklist:

    to my code.
        updated accordingly.
    
## Description

This PR uses `utility::optional` to provide a single binding function (no overload) with a clean argument list.
The arguments are in line with the cpp implementation (which also uses default args).
This solution avoids named args before positional args.

This is a fix used in #6917 since `mypy` would complain about `non-default argument follows default argument  [syntax]`.


ssheorey (approved):
Looks good. Thanks @timohl !

PR #7081, opened 2024-11-29
Docstring updates for NearestNeighborSearch and compute_metrics

benjaminum:
## Type

-   [x] Bug fix (non-breaking change which fixes an issue): Fixes #
-   [ ] New feature (non-breaking change which adds functionality). Resolves #
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected) Resolves #

## Checklist:

-   [x] I have run `python util/check_style.py --apply` to apply Open3D **code style**
    to my code.
-   [ ] This PR changes Open3D behavior or adds new functionality.
    -   [ ] Both C++ (Doxygen) and Python (Sphinx / Google style) **documentation** is
        updated accordingly.
    -   [ ] I have added or updated C++ and / or Python **unit tests** OR included **test
        results** (e.g. screenshots or numbers) here.
-   [ ] I will follow up and update the code if CI fails.
    
-   [ ] For fork PRs, I have selected **Allow edits from maintainers**.

## Description

- Adding missing documentation and improving format issues for some docstrings for PointCloud and TriangleMesh
- Fix unknown escape sequence warning in make_docs.py
- **Proposal of a template docstring to point out pitfalls and make writing docstrings easier** [template](https://github.com/isl-org/Open3D/blob/bu/docstring_updates/docs/template_python_docstring.txt)

The rendered template looks like this

image

PR #7103, opened 2024-12-18
Changed namespace fix to use Sphinx autodoc-process-* events

timohl:
## Type

## Motivation and Context

Open3D uses namespaces `open3d.cpu.pybind.` and `open3d.cuda.pybind.` to differentiate between both device builds.
This adds a lot of clutter to the documentation (see screenshots below).
There is already a partial fix for functions documented using `docstring::ClassMethodDocInject` and `docstring::FunctionDocInject`.
However, a big part of Open3D does not use those functions for documentation and #7081 removed those from the contribute docs (so they should probably be avoided in the future).

## Checklist:

    to my code.
        updated accordingly.
    
## Description

This PR moves the namespace fix away from `docstring::ClassMethodDocInject` and `docstring::FunctionDocInject` to the sphinx documentation generation.
It uses the autodoc events `autodoc-process-signature` and `autodoc-process-docstring` to change namespaces to `open3d.` during doc generation.
With this PR all namespaces are fixed in the documentation (checked using the search function).

This is a fix used in https://github.com/isl-org/Open3D/pull/6917 since mypy would complain about missing imports for functions with fixed namespaces.

Notice in the following before/after screenshots that the namespaces in the `clear` function were already fixed due to using `docstring::ClassMethodDocInject`. This PR improves the namespaces in `__init__` and `as_tensor` though, resulting in better readability.

Here before:
before
and after:
after

### DIFF
--- a/cpp/pybind/docstring.cpp
+++ b/cpp/pybind/docstring.cpp
@@ -289,17 +289,9 @@ std::string FunctionDoc::ToGoogleDocString() const {
     return rc.str();
 }
 
-std::string FunctionDoc::NamespaceFix(const std::string& s) {
-    std::string rc = std::regex_replace(s, std::regex("::(\\S)"), ".$1");
-    rc = std::regex_replace(rc, std::regex("open3d\\.(cpu|cuda)\\.pybind\\."),
-                            "open3d.");
-    return rc;
-}
-
 std::string FunctionDoc::StringCleanAll(std::string& s,
                                         const std::string& white_space) {
     std::string rc = utility::StripString(s, white_space);
-    rc = NamespaceFix(rc);
     return rc;
 }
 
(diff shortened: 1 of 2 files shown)


timohl:
Documentation CI was already failing in previous commits.
With the following error:
```bash
The following packages have unmet dependencies:
 cmake : Depends: libssl1.1 (>= 1.1.1) but it is not installable
```

The following is the function installing docs dependencies.
Maybe an issue in the kitware repository?
Not sure why this is even required.
Why not use the cmake from ubuntu repositories?
https://github.com/isl-org/Open3D/blob/0ae7b1a74325f4083ec2f1806a998cf1f67522c5/util/ci_utils.sh#L338-L375

benjaminum (approved):
@timohl This is great! Thanks!

We are moving away from the `*Inject` functions to simplify writing docstrings.

PR #7108, opened 2024-12-20
Fix packaged dlls not found

### LINKED ISSUE #7104: Open3D Python cannot find DLLs in add_dll_directory-locations
### Checklist

### Describe the issue

I try to import custom-build Open3D with CUDA on Windows. In `__init__.py` [when `open3d/cuda/pybind.cp310-win_amd64.pyd` is loaded](https://github.com/isl-org/Open3D/blob/0ae7b1a74325f4083ec2f1806a998cf1f67522c5/python/open3d/__init__.py#L64), `open3d/tbb12.dll` is not found, although its location has been added to the DLL-search-directories [with `os.add_dll_directory` before](https://github.com/isl-org/Open3D/blob/0ae7b1a74325f4083ec2f1806a998cf1f67522c5/python/open3d/__init__.py#L46).

Workarounds:

1. Copying `tbb12.dll` into `Windows/system32` resolves the problem.
2. Loading `tbb12.dll` explicitly with `CDLL(str(Path(__file__).parent / "tbb12.dll"))` in `__init__.py` before loading `cuda/pybind.cp310-win_amd64.pyd` resolves the problem.
3. Adding `tbb12.dll`'s location to `PATH` resolves the problem. I.e. doing `os.environ['PATH'] = str(Path(__file__).parent) + ';' + os.environ.get('PATH', '')` in `__init__.py`.

**Main question: can anyone reproduce the problem or has any insights into why `add_dll_directory` seems not to be respected here? If this is a general problem, I'd open a PR that applies workaround 3.**

```python
Build Open3D with the software versions stated below and try to import it with `import open3d`.
```

### Error message

I get the warning below. Note that the hint about the CUDA libraries is wrong in my case and only the "Reported error" is relevant.

```
"Open3D was built with CUDA support, but an error ocurred while loading the Open3D CUDA Python bindings. This is usually because the CUDA libraries could not be found. Check your CUDA installation. Falling back to the CPU pybind library. Reported error: Could not find module 'c:\workspace\visionkit\venv310\lib\site-packages\open3d\cuda\pybind.cp310-win_amd64.pyd' (or one of its dependencies). Try using the full path with constructor syntax."
```

### Expected behavior

`pybind.cp310-win_amd64.pyd` is loaded without error. I.e. `tbb12.dll` can be located and loaded during the call to `CDLL()`.

```markdown
- Operating system: Windows 11
- Python version: Python 3.10.11
- Open3D version: 0.18.0+8e434558a
- System architecture: x86
- Compiler version (if built from source): msvc 14.3
```

### Additional information

~If I understand [this](https://github.com/uxlfoundation/oneTBB/issues/1465) correctly, the problem could perhaps also be that `tbb12.dll` is not finding its dependencies and not that `tbb12.dll` is not found. I did not follow this particular rabbit hole, further, though. If this is the problem, Open3D should perhaps implement a workaround or switch to a different tbb-version.~

[edit: with some further research I found that this problem also exists for CUDA-DLLs, so I'm pretty sure the issue above has nothing to do with this issue.]

fegorsch:
## Type

-   [x] Bug fix (non-breaking change which fixes an issue): Fixes #7104 
-   [x] Breaking change (fix or feature that would cause existing functionality to not work as expected) -> see below for description

## Motivation and Context

[This PR](https://github.com/isl-org/Open3D/pull/5259) introduced `winmode=0` for loading DLLs during import. This however leads to directories added via `os.add_dll_directory` not being considered, when loading DLLs (see [here](https://github.com/isl-org/Open3D/issues/7104#issuecomment-2556557535) or [here](https://stackoverflow.com/questions/59330863/cant-import-dll-module-in-python/59333711#59333711)). In particular this leads to `open3d/tbb12.dll` not being found.

Fix this by going back to the default DLL-search mode (`winmode=None`).

In order to still be able to find CUDA-DLLs, that are not part of the open3d-package, search their directory in PATH and explicitly add it to the DLL-search-locations for CUDA-builds.

Note that this PR results in DLLs generally not being searched in `PATH` anymore, when importing `open3d`, as is the default behaviour of Python-packages since 3.8. The exception are CUDA-libraries in the heuristically determined directory. If anyone relies on `PATH` being searched, this PR breaks functionality.

## Checklist:

    to my code.
        updated accordingly.
    
### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ## Main
+-   Fix DLLs not being found in Python-package. Also prevent PATH from being searched for DLLs, except CUDA (PR #7108)
 -   Fix MSAA sample count not being copied when FilamentView is copied
 -   Fix TriangleMesh::SamplePointsUniformly and TriangleMesh::SamplePointsPoissonDisk now sampling colors from mesh if available (PR #6842)
 -   Fix TriangleMesh::SamplePointsUniformly not sampling triangle meshes uniformly (PR #6653)
--- a/python/open3d/__init__.py
+++ b/python/open3d/__init__.py
@@ -15,6 +15,7 @@
 # https://github.com/dmlc/xgboost/issues/1715
 import os
 import sys
+import re
 
 os.environ["KMP_DUPLICATE_LIB_OK"] = "True"
 # Enable thread composability manager to coordinate Intel OpenMP and TBB threads. Only works with Intel OpenMP.
@@ -26,22 +27,6 @@ from pathlib import Path
 import warnings
 from open3d._build_config import _build_config
 
-
-def load_cdll(path):
-    """
-    Wrapper around ctypes.CDLL to take care of Windows compatibility.
-    """
-    path = Path(path)
-    if not path.is_file():
-        raise FileNotFoundError(f"Shared library file not found: {path}.")
-
-    if sys.platform == "win32" and sys.version_info >= (3, 8):
-        # https://stackoverflow.com/a/64472088/1255535
-        return CDLL(str(path), winmode=0)
-    else:
-        return CDLL(str(path))
-
-
 if sys.platform == "win32":  # Unix: Use rpath to find libraries
     _win32_dll_dir = os.add_dll_directory(str(Path(__file__).parent))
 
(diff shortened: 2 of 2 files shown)


ssheorey (approved):
Thanks @fegorsch !

PR #7118, opened 2024-12-30
Add optional indices arg for fast computation of a small subset of FPFH features

### LINKED ISSUE #7089: Calculate FPFH only in some points
### Checklist

One problem with Open3D is that it calculates the FPFH for all points in the cloud, which is not efficient. Ideally, there should be a way of calculating the FPFH only for certain indices, such as the ISS detector keypoints. This has the potential to speed up global registration by several orders of magnitude. 

For example, in the registration of clouds from topographic/terrestrial laser scanners, even after reduction by 10 cm voxealization, it is common for the cloud to still have more than 1 million points. But using only ISS keypoints, which are calculated much faster than FPFH, this is reduced to 10,000 points. 

So what I'm asking for is just a function that calculates the FPFH on a list of indices of the points in the cloud, not on the whole cloud 

Translated with DeepL.com (free version)

### References

_No response_

### Additional information

_No response_

nicolaloi:
## Type

-   [x] New feature (non-breaking change which adds functionality). Resolves #7089

## Motivation and Context

Allows optional selection of specific points when computing FPFH features for a point cloud. Previously, if you only needed FPFH features from a small subset of points (e.g., ISS keypoints), you had to compute the features for the entire point cloud (which was slow and unnecessary) and then select the relevant features. Now, you can directly compute the FPFH features for just the selected points, resulting in faster performance.

Unrelated, also fixed a bug in #7039 when calling SelectByIndex with `invert=true`.

## Checklist:

    to my code.
        updated accordingly.
    
## Description

Made significant modifications to multiple files. Extra processing for a correct dataflow was needed when the user provides the required point indices. 

An optional argument `indices` is added to [`ComputeFPFHFeature`](https://github.com/isl-org/Open3D/blob/583eaa3351f9411f2d2f1faf7ac17479cc11975b/cpp/open3d/pipelines/registration/Feature.h#L69) and [`t\ComputeFPFHFeature`](https://github.com/isl-org/Open3D/blob/583eaa3351f9411f2d2f1faf7ac17479cc11975b/cpp/open3d/t/pipelines/registration/Feature.h#L41). If `indices` is provided, the function computes and returns the FPFH features for only the selected points. Some additional processing is performed to prepare the data. The catch is that even if the user selects _n_ points, to compute the corresponding _n_ FPFH features, the SPFH features of all the neighbors of each selected point must first be computed. And to compute the SPFH feature for a point, you need the data of its neighbors.

This means that selecting 1% or 10% of the points will not result in a 100x or 10x speed-up, because the many SPFH features of all the neighbors still need to be computed. Moreover, the new inner pipelines within the FPFH functions are optimized for selecting a small fraction of indices (around 0-10%). If the fraction of selected indices is large, the speed-up will be minimal. In fact, if the fraction of selected points approaches 100%, calling the function without specifying indices (i.e., computing FPFH features for the entire point cloud) will actually be faster. In the next days I can add benchmark tests too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,7 @@
 -   Fix render to depth image on Apple Retina displays (PR #7001)
 -   Fix infinite loop in segment_plane if num_points < ransac_n (PR #7032)
 -   Add select_by_index method to Feature class (PR #7039)
+-   Add optional indices arg for fast computation of a small subset of FPFH features (PR #7118).
 
 
 ## 0.13
--- a/cpp/benchmarks/t/pipelines/registration/Feature.cpp
+++ b/cpp/benchmarks/t/pipelines/registration/Feature.cpp
@@ -5,8 +5,6 @@
 // SPDX-License-Identifier: MIT
 // ----------------------------------------------------------------------------
 
-#include <map>
-
 #include "open3d/t/pipelines/registration/Feature.h"
 
 #include <benchmark/benchmark.h>
--- a/cpp/benchmarks/t/pipelines/registration/Feature.cpp
+++ b/cpp/benchmarks/t/pipelines/registration/Feature.cpp
@@ -5,6 +5,8 @@
 // SPDX-License-Identifier: MIT
 // ----------------------------------------------------------------------------
 
+#include <map>
+
 #include "open3d/t/pipelines/registration/Feature.h"
 
 #include <benchmark/benchmark.h>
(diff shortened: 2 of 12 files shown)


nicolaloi:
@ssheorey @benjaminum Before adding new benchmarks, I tried running the existing benchmarks in the main branch running `./build/bin/benchmarks`, but I immediately get a `Segmentation fault (core dumped)`, even with commits from a few months ago. Is there anything else I need to do apart from building with the `-DBUILD_BENCHMARKS=ON` flag?

nicolaloi:
Here are the benchmarks I have added for the legacy FPFH and CPU/CUDA tensor FPFH using the Hybrid NN search.
`null`: no indices selected (compute FPFH for the whole point cloud as original behavior)
`0.0001`/`0.001`/.../`1.0`: select random indices with a specified ratio to the total number of points.

benchmark

With the new commit 24ce40a3a9bdb6ed6b89d0cacb1e81dcbd26ddc3, I have added a speedup when the ratio of selected indices is high (>=0.01).

nicolaloi:
@ssheorey @benjaminum friendly ping if you have time

benjaminum:
@nicolaloi Thanks a lot for the PR! It looks very good!
I have a high level question about the point selection. Would permuting the point cloud to movie the selected points to the beginning simplify the index computations? Just want to know what you think

nicolaloi:
@benjaminum It makes sense so the number of _FPFH-idx to point-idx_ and _SPFH-idx to point-idx_ mappings in the code can be reduced. I am thinking though if there would be an important slowdown for very large point clouds and small sets of selected FPFH points: the input point cloud is a reference to the user point cloud, so you would need to copy the whole point cloud just to move the required points, which could be just <1‰ of the total points.

Anyway, I see it like this: **1)** Copy the original point cloud,   **2)** Move the selected FPFH points to the beginning of the copied point cloud,   **3)** Find neighbors of these selected points,   **4)** Move the neighboring points too to the beginning of the copied point cloud (since they are needed to compute the SPFH features needed to compute the FPFH features).

benjaminum:
I see. Having a full copy can be quite costly and doing and undoing the permutation in place may become a source of strange side effects and potential errors. I think the index structures are the best solution to avoid all this

benjaminum (approved):
Looks good! Thanks a lot @nicolaloi !

benjaminum:
CI errors are not related to this PR.

Link error on Windows with CUDA
```
error LNK2019: unresolved external symbol cudaMemcpy
```

IO test
```
[  FAILED  ] TPointCloudIO.ReadWrite3DGSPLYToSPLAT
[  FAILED  ] TPointCloudIO.ReadWrite3DGSSPLAT
```

Documentation
```
#17 259.4 ModuleNotFoundError: No module named 'open3d'
```

PR #7128, opened 2025-01-06
Build fixes for v0.19

ssheorey:
Prevent libomp linking when libomp5-dev is installed, due to cmake adding the libc++ dir as a linker search path. In sycl mode, libiomp is the default OMP runtime.

Do not install clang to prevent libomp-dev being installed. DEVELOPER_BUILD option for SYCL CI
ccache gz -> xz
Use cxx11 abi pytorch to allow building both torch and tf docs at once.

glew updated to latest (v2.2)

GLFW detects platform at runtime (except NULL for headless rendering). Only glew needs separate build for HEADLESS rendering.

Issues in current build:

- [x] SYCL CI does not allow turning off the git hash name (DEVELOPER_MODE env var)
- [x] SYCL wheel libOpen3D.so links both libiomp5 and libomp5 (with BUILD)SHARED_LIBS only).
- [x] ENABLE_HEADLESS_RENDERING does not work with OSMesa with the new GLFW v3.4 with `draw_geometries()`
```
[Open3D WARNING] GLFW Error: Failed to detect any supported platform
[Open3D WARNING] Failed to initialize GLFW
```

GLFW 

## Type

## Checklist:

    to my code.
        updated accordingly.
    

johnthagen:
@ssheorey I noticed 0.19 was published to PyPI but there are no Linux ARM64 wheels (`-manylinux_2_27_aarch64.whl`). I saw that last release it took a day to get ARM64 wheels out, but just wanted to report this in case it got overlooked

- https://pypi.org/project/open3d/0.19.0/#files

ssheorey:
There's a dynamic linker error on Linux ARM64 that's preventing releasing the wheels. Please see the release notes and the CI run with the error here: 

> ARM64 Linux builds are not available due to a runtime dynamic linker error: "cannot allocate memory in static TLS block". e.g. see [here](https://github.com/isl-org/Open3D/actions/runs/12604362802/job/35131136413).

Happy to release with a fix if someone can contribute a fix for the error.

johnthagen:
@ssheorey Thanks for the reference. I opened an issue to track this

- #7130

nanotuxi:
@ssheorey  said "BUILD_HEADLESS_RENDERING option (OSMesa) gives a build error in Ubuntu 24.04"

This workaround seems to fix headless rendering for Open3D on Ubuntu 24.04
sudo sed -i 's/APIENTRY//' /usr/include/GL/osmesa.h

It can be undone with:
sudo apt reinstall libosmesa6-dev

But it's not for the faint of heart.

see https://github.com/nigels-com/glew/issues/411#issuecomment-2271361805

works in current version, but you should set the parameter allow_arbitrary=True, i.e.,
view_control.convert_from_pinhole_camera_parameters(cam_params, allow_arbitrary=True)
in the python example

emersonknapp:
@ssheorey or @johnnynunez would you be able to do another release of 0.19 to PyPI now with the Linux ARM artifacts, given the fix is now in for #7130 ?

ssheorey:
You can use the main-devel release from Github release for now. We plan to do a v0.20 release soon.

PR #7136, opened 2025-01-13
Fix for #7135 (error with named argument "width")

### LINKED ISSUE #7135: Named argument "width" not accepted by `create_box` in tensor API
### Checklist

### Describe the issue

The equivalent of the `o3d.geometry.TriangleMesh.create_box()` method in the tensor API is `o3d.geometry.TriangleMesh.create_box()`. Contrary to the legacy API, the tensor API does not accept the named argument "width". 

This is due to a MISSING COMMA between the end of the method documentation and the first argument:

https://github.com/isl-org/Open3D/blob/1e7b17438687a0b0c1e5a7187321ac7044afe275/cpp/pybind/t/geometry/trianglemesh.cpp#L346-L354

(Notice also that spaces should be present on each line of the documentation before the closing quotation marks.)

```python
import open3d as o3d

m = o3d.t.geometry.TriangleMesh.create_box(5, 4, 3)  # works
m = o3d.t.geometry.TriangleMesh.create_box(width=5, depth=4, height=3)  # doesn't work
```

### Error message

```
  File "<string>", line 1, in <module>
TypeError: create_box(): incompatible function arguments. The following argument types are supported:
    1. (Create a box triangle mesh. One vertex of the boxwill be placed at the origin and the box alignswith the positive x, y, and z axes.width: float = 1.0, height: float = 1.0, depth: float = 1.0, float_dtype: open3d.cpu.pybind.core.Dtype = Float32, int_dtype: open3d.cpu.pybind.core.Dtype = Int64, device: open3d.cpu.pybind.core.Device = CPU:0) -> open3d.cpu.pybind.t.geometry.TriangleMesh

Invoked with: kwargs: width=5, height=4, depth=3
``` 

### Expected behavior

The `o3d.t.geometry.TriangleMesh.create_box` should accept the `width` named argument.

```markdown
- Python version: Python 3.11, Python 3.12
- Open3D version: 0.18.0, 0.19.0
```

### Additional information

_No response_

joseph-sch:
## Type

-   [x] Bug fix (non-breaking change which fixes an issue): Fixes #7135

## Checklist:

    to my code.
        updated accordingly.
    
## Description

* Add the missing comma before the first argument's name
 * Add the missing spaces in the function description

### DIFF
--- a/cpp/pybind/t/geometry/trianglemesh.cpp
+++ b/cpp/pybind/t/geometry/trianglemesh.cpp
@@ -344,9 +344,9 @@ This example shows how to create a hemisphere from a sphere::
     // Triangle Mesh's creation APIs.
     triangle_mesh
             .def_static("create_box", &TriangleMesh::CreateBox,
-                        "Create a box triangle mesh. One vertex of the box"
-                        "will be placed at the origin and the box aligns"
-                        "with the positive x, y, and z axes."
+                        "Create a box triangle mesh. One vertex of the box "
+                        "will be placed at the origin and the box aligns "
+                        "with the positive x, y, and z axes.",
                         "width"_a = 1.0,
                         "height"_a = 1.0, "depth"_a = 1.0,
                         "float_dtype"_a = core::Float32,
--- a/cpp/pybind/t/geometry/trianglemesh.cpp
+++ b/cpp/pybind/t/geometry/trianglemesh.cpp
@@ -347,8 +347,7 @@ This example shows how to create a hemisphere from a sphere::
                         "Create a box triangle mesh. One vertex of the box "
                         "will be placed at the origin and the box aligns "
                         "with the positive x, y, and z axes.",
-                        "width"_a = 1.0,
-                        "height"_a = 1.0, "depth"_a = 1.0,
+                        "width"_a = 1.0, "height"_a = 1.0, "depth"_a = 1.0,
                         "float_dtype"_a = core::Float32,
                         "int_dtype"_a = core::Int64,
                         "device"_a = core::Device("CPU:0"))


joseph-sch:
> Thanks for submitting this pull request! The maintainers of this repository would appreciate if you could update the **CHANGELOG.md** based on your changes.

I'd happily update the **CHANGELOG.md** file, but I don't think it's being maintained... so I wouldn't know where to add the change description, nor what good it would do...

timohl:
Good catch and fix. :)

The last three lines in the docs are also corrupted:
```
(open3d.core.Device (device) – 0): Device of the create mesh.
optional – 0): Device of the create mesh.
default=CPU – 0): Device of the create mesh.
```
https://www.open3d.org/docs/latest/python_api/open3d.t.geometry.TriangleMesh.html#open3d.t.geometry.TriangleMesh.create_box

If it is no problem, can you check if your fix also resolves this?
Just a quick [docs build](https://www.open3d.org/docs/latest/builddocs.html) and compare if it looks fine or not.
Otherwise, I can help looking into why it gets corrupted and we can fix that as well.

ssheorey (approved):
Thanks @joseph-sch ! Looks good.

joseph-sch:
> Good catch and fix. :)
> 
> The last three lines in the docs are also corrupted:
> 
> ```
> (open3d.core.Device (device) – 0): Device of the create mesh.
> optional – 0): Device of the create mesh.
> default=CPU – 0): Device of the create mesh.
> ```
> 
> https://www.open3d.org/docs/latest/python_api/open3d.t.geometry.TriangleMesh.html#open3d.t.geometry.TriangleMesh.create_box
> 
> If it is no problem, can you check if your fix also resolves this? Just a quick [docs build](https://www.open3d.org/docs/latest/builddocs.html) and compare if it looks fine or not. Otherwise, I can help looking into why it gets corrupted and we can fix that as well.

Regarding the mangled documentation, apparently it's a **sphinx parsing issue** caused by the string in the default argument (i.e. by encountering quotes in the default argument expression `core::Device("CPU:0")`).

The issue does _not_ appear when a _custom docstring is specified_, so that could be a workaround.

Indeed, compare the following two (`create_sphere` and `create_text`, the latter having a custom docstring):

image

image

In the second case, the word "Optional" and the default value are not present alongside each argument, but the `device` argument is correctly displayed without mangling.

According to ChatGPT, another solution would be to "Create a static default `Device` instance outside the binding to prevent inline string construction" with

```cpp
static const core::Device default_device("CPU:0");
```

and use that instead of the expression `core::Device("CPU:0")`.

There could also be a configuration of sphinx that avoids the issue, and that would be ideal.

I'm afraid I didn't manage to build the C++ code yet (I tried only on WSL so far) and I don't manage to rebuild the sphinx documentation after modifying the sources, so I can't check by myself yet. I suggest opening a separate issue and a separate PR.

joseph-sch:
> Thanks @joseph-sch ! Looks good.

Thank you! I've now applied the style check and pushed the change. Do I need to do anything else?

timohl:
Interesting.
I can look into it on the weekend in a separate issue/PR.

joseph-sch:
> Interesting. I can look into it on the weekend in a separate issue/PR.

OK, I've opened #7147 for the separate issue of the corrupted documentation.

joseph-sch:
I managed to build the package and documentation to check the success of the patch:

image

PR #7177, opened 2025-02-22
fixes for using system packages

christian-rauch:
I was building Open3D with most system packages enabled and some features deactivated. This uncovered a couple of compilation and linking issues, such as:
1. using`ExtractZIP` when `WITH_MINIZIP` is off
2. using GLFW APIs that are only available from version 3.4 onwards
3. linking OpenSSL when it has not been found before when `USE_SYSTEM_CURL` and `USE_SYSTEM_OPENSSL` are set

## Type

## Motivation and Context

Fixing the build

## Checklist:

    to my code.
        updated accordingly.
    

3rdparty/find_dependencies.cmake
```diff
@@ -867,29 +869,31 @@ if(USE_SYSTEM_CURL)
...
+        INCLUDE_DIRS ${BORINGSSL_INCLUDE_DIRS}
+        INCLUDE_ALL
+        INCLUDE_DIRS ${BORINGSSL_INCLUDE_DIRS}
+        LIB_DIR      ${BORINGSSL_LIB_DIR}
+        LIBRARIES    ${BORINGSSL_LIBRARIES}
+        DEPENDS      ext_zlib ext_boringssl
+    )
+endif()
 
+if(NOT USE_SYSTEM_CURL)
```
ssheorey, line 896: Hi @christian-rauch we need SSL as an indirect dependency of cURL. So with USE_SYSTEM_CURL, we do not expliclty need to find and link system SSL. Can you update this so that SSL is ignored with USE_SYSTEM_CURL?

ssheorey (changes requested):
Thanks @christian-rauch for these fixes! Please see one comment below.

3rdparty/find_dependencies.cmake
christian-rauch: But isn't this exactly what the commit is fixing? Previously, with `USE_SYSTEM_CURL` set to `ON`, we would not link SSL. Hence, the linking issues. So with `USE_SYSTEM_CURL=ON`, we still need to link SSL explicitly.

christian-rauch:
Btw, `python util/check_style.py --apply` requires an old clang-format version. Also with the oldest possible setting `required_clang_format_major = 14` on Ubuntu 24.04, this failed in the regex. Ubuntu returns something like `Ubuntu clang-format version 18.1.3 (1ubuntu1)`, i.e. it prepends `Ubuntu ` to the string.

3rdparty/find_dependencies.cmake
christian-rauch: @ssheorey

christian-rauch:
The CI is showing errors seemingly unrelated to this PR, such as
```
2025-03-12T22:53:54.4354260Z [1;33m[Open3D WARNING] Read LOG failed: unable to open file: /Users/runner/open3d_data/extract/SampleRedwoodRGBDImages/trajectory.log[0;m
2025-03-12T22:53:54.4381880Z ./util/ci_utils.sh: line 302: 99802 Segmentation fault: 11  ./bin/tests "$unitTestFlags"
```
and the main branch also does not pass the CI.

christian-rauch:
@ssheorey Can you have a look at this again?

ssheorey (approved):
Thanks @christian-rauch looks good. 

I'll update with `main` to check if we can pass CI.

christian-rauch:
@ssheorey I remove the "WITH_MINIZIP" commit to see if this reduces test failures. I am still getting test failures without this commit, but those are the same as on the `main` branch, so I guess this is ok?

PR #7180, opened 2025-02-26
Fix missing dependencies in Open3D-viewer post-install script

meetgandhi-dev:
# Fix missing dependencies in Open3D-viewer post-install script

## Type

## Motivation and Context

The script `cpp/apps/Open3DViewer/postinstall-linux.sh` was generating the following error messages:
```bash
#24 904.4 /root/Open3D/cpp/apps/Open3DViewer/postinstall-linux.sh: 5: update-desktop-database: not found
#24 904.4 /root/Open3D/cpp/apps/Open3DViewer/postinstall-linux.sh: 6: gtk-update-icon-cache: not found
```

This PR adds the necessary dependencies for installation:

`update-desktop-database` is provided by [desktop-file-utils](https://manpages.ubuntu.com/manpages/focal/en/man1/update-desktop-database.1.html)
`gtk-update-icon-cache` is provided by [gtk-update-icon-cache](https://manpages.ubuntu.com/manpages/focal/en/man1/gtk-update-icon-cache.1.html)

## Checklist:

    to my code.
        updated accordingly.
    
### Log snippet while generating TXZ package

```text
#24 688.0 Run CPack packaging tool...
#24 688.0 /cmake-3.29.2-linux-x86_64/bin/cpack --config ./CPackConfig.cmake
#24 688.0 CPack: Create package using TXZ
#24 688.0 CPack: Install projects
#24 688.0 CPack: - Run preinstall target for: Open3D
#24 900.1 CPack: - Install project: Open3D []
#24 904.4 /root/Open3D/cpp/apps/Open3DViewer/postinstall-linux.sh: 5: update-desktop-database: not found
#24 904.4 /root/Open3D/cpp/apps/Open3DViewer/postinstall-linux.sh: 6: gtk-update-icon-cache: not found
#24 904.4 CPack: Create package
#24 981.6 CPack: - package: /root/Open3D/build/package/open3d-devel-linux-x86_64-cxx11-abi-0.19.0.tar.xz generated.
```


ssheorey:
Hi @meetgandhi-dev these packages are installed in the standard Ubuntu desktop install. Which OS / variant were they missing from?

meetgandhi-dev:
Hi @ssheorey 
I found the errors while building packages for cpu-xxx variants. For eg: https://github.com/isl-org/Open3D/blob/main/docker%2Fdocker_build.sh#L371
I believe the base docker image doesn't contain desktop application(s).

ssheorey:
Ah I see. This is a post install script, so meant to be run on the users system after installation. I'm not sure if running it inside docker is useful. The issue with adding these 2 packages to docker is that they will pull in the entire Ubuntu desktop, which will make the docker image significantly larger. 

I would recommend instead modifying the post install script so that it does not run these commands if they are not found in the system.

meetgandhi-dev:
@ssheorey Understood.
Should I list them as dependencies for the created Debian package [here](https://github.com/isl-org/Open3D/blob/main/cpp/apps/Open3DViewer/Debian/CMakeLists.in.txt#L36)?
Also, do you think the [postinstall](https://github.com/isl-org/Open3D/blob/main/cpp/apps/Open3DViewer/postinstall-linux.sh) script needs to be included in the Debian package for proper installation?

ssheorey:
I don't think these should be dependencies - dependencies are supposed to be mandatory, while these are optional convenience utilities. We run them if present, else don't do anything. 

I think the post install script is included in the Debian package for the viewer at the moment (see the open3d-viewer deb package in github releases.)

meetgandhi-dev:
> I don't think these should be dependencies - dependencies are supposed to be mandatory, while these are optional convenience utilities. We run them if present, else don't do anything.

Got it. I will update the script accordingly.

> I think the post install script is included in the Debian package for the viewer at the moment (see the open3d-viewer deb package in github releases.)

I think the post-install script is executed during the Cmake installation process ([ref](https://github.com/isl-org/Open3D/blob/main/cpp/apps/CMakeLists.txt#L73C13-L73C25)) and is not part of the debian package. However, the required dependencies are not available in the Docker image, so they aren't being installed. I propose using the [postinst feature](https://cmake.org/cmake/help/latest/cpack_gen/deb.html#variable:CPACK_DEBIAN_PACKAGE_CONTROL_EXTRA) of the CPack DEB to trigger the installation script during the Debian package installation. (This will only execute commands if the dependencies are available, as mentioned above.)

ssheorey:
Sounds good. Can you update the PR ? Thanks!

Issue #7154, opened 2025-01-30
Visualize 3D Gaussian splats


ssheorey:
3D Gaussian splatting is a novel 3D representation that enables fast scene reconstruction and real time photorealistic rendering. We are adding support for rendering 3DGS point clouds in Open3D.

**Steps [filament renderer]:**

Study the Open3D filament based visualizer `cpp/open3d/visualization/rendering/filament` and the [filament material format](https://google.github.io/filament/Materials.html).

1.  Study the [3DGS paper](https://repo-sam.inria.fr/fungraph/3d-gaussian-splatting/), reference implementation of the 3DGS viewer [gsplat/rendering.py](https://github.com/nerfstudio-project/gsplat/blob/0880d2b471e6650d458aa09fe2b2834531f6e93b/gsplat/rendering.py#L28C16-L28C16) and read its docs.
2.  We want to port the Python code to C++17 and the CUDA kernels to GLSL vertex and fragment shaders. At the moment, we will only port the Python code and use mock-ups for the GLSL shaders. To enable the mock-ups to be functional, we will provide sample returned data. We will load this data to the GPU as buffer objects and the mock-up shaders will simply return them. The result should be a correct rendering of the 3DGS scene, but from a fixed viewpoint.
3.  Create a new filament material "3dgs" with mock vertex and fragment shaders in ESSL 3.0 (OpenGL ES variant of GLSL) to render 3DGS point clouds. ESSL 3.0 is a filament limitation.
4.  Add the new material to Open3D, update `O3DVisualizer` to use the 3dgs material. Provide all required point cloud attributes to the GPU when 3DGS attributes are detected. See the folder `cpp/open3d/visualization/rendering/` for the rendering implementation. For now, also create buffers for the intermediate data returned by the shaders. 
6. Use it to render 3DGS point clouds. 

**Tests:**

- Load these Numpy files and display scene from a few different viewpoints. Compare with provided reference images.

**Open3D Tensor point cloud representation:**

- `pcd.point[‘positions’]` (N,3) – (x,y,z) for each splat.
- `pcd.point[‘opacity’]` – (N,) opacity
- `pcd.point[‘rot’]` (N, 4) - quaternion rotation of Gaussian.
- `pcd.point[‘scale']` (N, 3) - x, y, z scales for Gaussian.
- `pcd.point[‘f_dc’]` (N, 3) – DC components for RGB colors.
[...]

PR #7193, opened 2025-03-08
add gaussianSplat application

WeiguangHan:
## Type

-   [ ] Bug fix (non-breaking change which fixes an issue): Fixes #
-   [x] New feature (non-breaking change which adds functionality). Resolves #
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected) Resolves #

## Checklist:

-   [ ] I have run `python util/check_style.py --apply` to apply Open3D **code style**
    to my code.
-   [ ] This PR changes Open3D behavior or adds new functionality.
    -   [ ] Both C++ (Doxygen) and Python (Sphinx / Google style) **documentation** is
        updated accordingly.
    -   [ ] I have added or updated C++ and / or Python **unit tests** OR included **test
        results** (e.g. screenshots or numbers) here.
-   [ ] I will follow up and update the code if CI fails.
    
-   [ ] For fork PRs, I have selected **Allow edits from maintainers**.

PR #7192, opened 2025-03-08
add gaussianSplat mat file for shader

WeiguangHan:
## Type

-   [ ] Bug fix (non-breaking change which fixes an issue): Fixes #
-   [x] New feature (non-breaking change which adds functionality). Resolves #
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected) Resolves #

## Checklist:

-   [ ] I have run `python util/check_style.py --apply` to apply Open3D **code style**
    to my code.
-   [ ] This PR changes Open3D behavior or adds new functionality.
    -   [ ] Both C++ (Doxygen) and Python (Sphinx / Google style) **documentation** is
        updated accordingly.
    -   [ ] I have added or updated C++ and / or Python **unit tests** OR included **test
        results** (e.g. screenshots or numbers) here.
-   [ ] I will follow up and update the code if CI fails.
    
-   [ ] For fork PRs, I have selected **Allow edits from maintainers**.

Issue #7153, opened 2025-01-30
3D Gaussian splatting IO


ssheorey:
3D Gaussian splatting is a novel 3D representation that enables fast scene reconstruction and real time photorealistic rendering. We are adding support for  3DGS in Open3D. As part of that, we want to add the ability to read and write 3DGS to files. 3DGS files are commonly stored as [ply format](https://paulbourke.net/dataformats/ply/) files. [original spec](https://gamma.cs.unc.edu/POWERPLANT/papers/ply.pdf). An alternate more compact representation (without support for view dependent effects) is the [.splat](https://github.com/antimatter15/splat) file format.

**Sub-tasks:**

Study Open3D point cloud IO code for reading / writing ply files. 3DGS is a point cloud with additional properties.

1. Add code to support reading / writing ply files with additional vertex properties. Add C++ unit test with very small toy data.
2. Add code to support reading / writing `.splat` files. Add C++ unit test with very small toy data.
3. Add documentation - tutorial Python notebook showing reading a 3DGS, converting it to a point cloud, visualizing it, converting to a mesh, adding albedo texture using the calibrated images and saving the point cloud and mesh. Also crop the original 3DGS point cloud and save it to disk.
4. [Optional] Convert a 3DGS point cloud to a normal dense point cloud by sampling multiple points with colors from each Gaussian. See [this](https://github.com/Lewis-Stuart-11/3DGS-to-PC/blob/main/gauss_to_pc.py) reference implementation. [Arxiv report](https://arxiv.org/pdf/2501.07478). Add C++ unit test with very small toy data.

**Test:** Open each of these files and save them back. Open the saved files in the reference viewer and ensure that the original and saved files look indistinguishable.

**Reference:**

- 3DGS files save to .ply format: https://github.com/graphdeco-inria/gaussian-splatting/blob/main/scene/gaussian_model.py
- 3DGS files saved to .splat format: https://github.com/antimatter15/splat/blob/main/convert.py
- Reference viewer: https://projects.markkellogg.org/threejs/demo_gaussian_splats_3d.php
- 3DGS to point cloud reference: https://github.com/Lewis-Stuart-11/3DGS-to-PC/blob/main/gauss_to_pc.py

**3DGS tensor point cloud representation:**

- `pcd.point[‘positions’]` (N,3) – (x,y,z) for each splat.
- `pcd.point[‘opacity’]` – (N,) opacity
[...]

PR #7194, opened 2025-03-08
Visualizing gaussian splats

billamiable:
## Type

## Motivation and Context

This PR, along with two other PRs (https://github.com/isl-org/Open3D/pull/7192, https://github.com/isl-org/Open3D/pull/7193), will fix https://github.com/isl-org/Open3D/issues/7154

## Checklist:

    to my code.
        updated accordingly.
    

cpp/open3d/visualization/rendering/filament/GaussianSplatBuffers.cpp
```diff
@@ -0,0 +1,187 @@
...
+        utility::LogWarning(
+            "TGaussianSplatBuffers is constructed for a geometry that is not GaussianSplat."
+        );
+    }
+
+    std::vector<std::string> check_list = {"f_dc", "opacity", "rot", "scale", "f_rest"};
+    for (const auto& check_item : check_list) {
+        if (geometry_.HasPointAttr(check_item) &&
+            geometry_.GetPointAttr(check_item).GetDtype() != core::Float32) {
+            auto check_item_instance = geometry_.GetPointAttr(check_item);
```
benjaminum, line 53: I think we need to check the attribute shapes here or in Pointcloud::IsGaussianSplat(). In ConstructBuffers() assumes that the attributes have specific sizes. f_rest can have different shapes (num_verts, num_coeffs, 3) with num_coeffs = 3, 8, or 15

cpp/open3d/visualization/rendering/filament/GaussianSplatBuffers.cpp
```diff
@@ -0,0 +1,187 @@
...
+    vbuf->setBufferAt(engine, 3, std::move(scale_descriptor));
+
+    const size_t f_rest_array_size = n_vertices * 4 * sizeof(float);
+    const size_t custom_buffer_num = 6;
+    const size_t custom_buffer_start_index = 4;
+    for(size_t i = 0; i < custom_buffer_num; i++) {
+        float* f_rest_array = static_cast<float*>(malloc(f_rest_array_size));
+        if (geometry_.HasPointAttr("f_rest")) {
+            float* f_rest_src = static_cast<float*>(geometry_.GetPointAttr("f_rest").GetDataPtr());
+            std::memcpy(f_rest_array, f_rest_src + i * n_vertices * 4, f_rest_array_size);
```
benjaminum, line 166: f_rest source may only have 9 elements per vertex if the shape is (N, num_coeffs=3, 3)

benjaminum (changes requested):
Thanks for the PR! Please have a look at the comments

cpp/open3d/visualization/rendering/filament/GaussianSplatBuffers.cpp
```diff
@@ -0,0 +1,187 @@
...
+    const size_t color_array_size = n_vertices * 4 * sizeof(float);
+    float* color_array = static_cast<float*>(malloc(color_array_size));
+    if (geometry_.HasPointAttr("f_dc") && geometry_.HasPointAttr("opacity")) {
+        float* f_dc_ptr = static_cast<float*>(geometry_.GetPointAttr("f_dc").GetDataPtr());
+        float* opacity_ptr = static_cast<float*>(geometry_.GetPointAttr("opacity").GetDataPtr());
+        for (size_t i = 0; i < n_vertices; i++) {
+            std::memcpy(color_array + i * 4, f_dc_ptr + i * 3, 3 * sizeof(float));
+            std::memcpy(color_array + i * 4 + 3, opacity_ptr + i, sizeof(float));
+        }
+    } else {
```
ssheorey, line 121: This if-else should be unnecessary, right? We are guaranteed that the mandatory attributes are present (f_dc. opacity, scale and rot).

cpp/open3d/visualization/rendering/filament/GaussianSplatBuffers.cpp
```diff
@@ -0,0 +1,187 @@
...
+                color_array, color_array_size,
+                GeometryBuffersBuilder::DeallocateBuffer);
+    vbuf->setBufferAt(engine, 1, std::move(color_descriptor));
+
+    const size_t rot_array_size = n_vertices * 4 * sizeof(float);
+    float* rot_array = static_cast<float*>(malloc(rot_array_size));
+    if (geometry_.HasPointAttr("rot")) {
+        std::memcpy(rot_array, geometry_.GetPointAttr("rot").GetDataPtr(),
+               rot_array_size);
+    } else {
```
ssheorey, line 136: This if-else should be unnecessary, right? We are guaranteed that the mandatory attributes are present (f_dc, opacity, scale and rot).

cpp/open3d/visualization/rendering/filament/GaussianSplatBuffers.cpp
```diff
@@ -0,0 +1,187 @@
...
+        }
+    }
+    VertexBuffer::BufferDescriptor rot_descriptor(
+                rot_array, rot_array_size,
+                GeometryBuffersBuilder::DeallocateBuffer);
+    vbuf->setBufferAt(engine, 2, std::move(rot_descriptor));
+
+    const size_t scale_array_size = n_vertices * 4 * sizeof(float);
+    float* scale_array = static_cast<float*>(malloc(scale_array_size));
+    std::memset(scale_array, 0, scale_array_size);
```
ssheorey, line 148: if is unnecessary.

cpp/open3d/visualization/rendering/filament/GaussianSplatBuffers.cpp
ssheorey: With 0,1,2 orders, we can have 0, 9, 24 floats per vertex (we cannot support order 3 / 45 floats per vertex with filament CUSTOM vertex attributes). The number of custom buffers should be different based on the SH order. If we use all the CUSTOM attributes every time, loading will be slow for splat format (quite common, does not contain f_rest for faster rendering).

ssheorey (changes requested):

cpp/open3d/visualization/rendering/filament/GaussianSplatBuffers.cpp
billamiable: We are adopting similar logic inside `PointCloudBuffers.cpp`, see implementation [here](https://github.com/isl-org/Open3D/blob/71798783b1c052293b7778c75126cf2d82f0959f/cpp/open3d/visualization/rendering/filament/PointCloudBuffers.cpp#L302).

cpp/open3d/visualization/rendering/filament/GaussianSplatBuffers.cpp
billamiable: So do you think we need to add additional function to check how many orders are provided inside raw data, as well as how many orders do we want to use during rendering?

cpp/open3d/t/geometry/PointCloud.cpp
```diff
@@ -1369,6 +1369,16 @@ core::Tensor PointCloud::ComputeMetrics(const PointCloud &pcd2,
                                 metrics, params);
 }
 
+bool PointCloud::IsGaussianSplat() const {
```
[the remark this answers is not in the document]
billamiable: Sorry for the confusion. This PR is based on PR https://github.com/isl-org/Open3D/issues/7153 because documentation needs to add above the newly added functions.

cpp/open3d/visualization/rendering/filament/FilamentResourceManager.cpp
```diff
@@ -1017,6 +1019,25 @@ void FilamentResourceManager::LoadDefaults() {
     lit_mat->setDefaultParameter("anisotropyMap", texture, default_sampler);
     materials_[kDefaultLit] = BoxResource(lit_mat, engine_);
 
+    const auto gaussian_path = resource_root + "/gaussianSplat.filamat";
+    auto gaussian_mat = LoadMaterialFromFile(gaussian_path, engine_);
```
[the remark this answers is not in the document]
billamiable: Not sure what this means. Is it supposed to be the same as https://github.com/isl-org/Open3D/pull/7192#discussion_r1988156555?

cpp/open3d/visualization/rendering/filament/GaussianSplatBuffers.cpp
billamiable: I wonder if this is also the case for point cloud, because inside `PointCloudBuffers.cpp` https://github.com/isl-org/Open3D/blob/71798783b1c052293b7778c75126cf2d82f0959f/cpp/open3d/visualization/rendering/filament/PointCloudBuffers.cpp#L389, default values are set.

cpp/open3d/visualization/rendering/filament/GaussianSplatBuffers.cpp
billamiable: same as above

cpp/open3d/visualization/rendering/filament/GaussianSplatBuffers.cpp
benjaminum: > Agree. PR #7153 adds this checking support for `IsGassianSplat()`. We can skip the checks here.

We should use the same checks as for IO here but I am not sure if we can skip checks here since the user can create a GS point cloud for visualization in memory.

ssheorey: `IsGaussianSplat()` is already called before this ctor is run (see line 255 in FilamentGeometryBuffersBuilder.cpp). So we don't need to repeat the check.
We should add a commment about preconditions to this function:
```cpp
///IsGaussianSplat() has verified that this point cloud contains the necessary data for a Gaussian Splat.
```

cpp/open3d/t/geometry/PointCloud.cpp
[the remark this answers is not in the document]
billamiable: I see, I thought this PR is also one of ours.

cpp/open3d/visualization/rendering/filament/GaussianSplatBuffers.cpp
zhangyang-intel: In the latest commit, i add a member variable `sh_degree` in class `MaterialRecord`.
Then we can determine how many custom buffers we to create according to `sh_degree`.
So @benjaminum @ssheorey , which options would you prefer, add function `PointCloud::GaussianSplatGetSHOrder()` or use param `sh_degree` .

cpp/open3d/visualization/rendering/filament/GaussianSplatBuffers.cpp
```diff
@@ -0,0 +1,187 @@
...
+    float* vertex_array = static_cast<float*>(malloc(vertex_array_size));
+    memcpy(vertex_array, points.GetDataPtr(), vertex_array_size);
+    VertexBuffer::BufferDescriptor pts_descriptor(
+            vertex_array, vertex_array_size,
+            GeometryBuffersBuilder::DeallocateBuffer);
+    vbuf->setBufferAt(engine, 0, std::move(pts_descriptor));
+
+    const size_t color_array_size = n_vertices * 4 * sizeof(float);
+    float* color_array = static_cast<float*>(malloc(color_array_size));
+    if (geometry_.HasPointAttr("f_dc") && geometry_.HasPointAttr("opacity")) {
```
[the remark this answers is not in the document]
zhangyang-intel: Yes, i fixed this.

cpp/open3d/visualization/rendering/filament/GaussianSplatBuffers.cpp
zhangyang-intel: I remove unnecessary if-else for positions, scales, rot, f_dc, opacity.

cpp/open3d/visualization/rendering/filament/GaussianSplatBuffers.cpp
ssheorey: I would prefer the function, since it's a quick calculation and the function is generally useful for users, beyond this purpose.

zhangyang-intel: I have added function `PointCloud::GaussianSplatGetSHOrder()` and allocate buffer according to sh degree, please review it in latest commit.

cpp/open3d/visualization/rendering/filament/GaussianSplatBuffers.cpp
zhangyang-intel: I will update this function according to PR #7153 .

cpp/open3d/visualization/rendering/filament/GaussianSplatBuffers.cpp
zhangyang-intel: I have removed unnecessary if-else.

cpp/open3d/visualization/rendering/filament/GaussianSplatBuffers.cpp
```diff
@@ -112,61 +123,55 @@ GeometryBuffersBuilder::Buffers TGaussianSplatBuffersBuilder::ConstructBuffers()
...
     float* rot_array = static_cast<float*>(malloc(rot_array_size));
-    if (geometry_.HasPointAttr("rot")) {
-        std::memcpy(rot_array, geometry_.GetPointAttr("rot").GetDataPtr(),
-               rot_array_size);
-    } else {
-        for (size_t i = 0; i < n_vertices * 4; ++i) {
-            rot_array[i] = 1.f;
-        }
-    }
+    std::memset(rot_array, 0, rot_array_size);
```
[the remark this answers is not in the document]
zhangyang-intel: I removed this, but why we can skip `memset` just before `memcpy`?
For sh_degree = 1, there are 9 floats. We need create 3 buffers so 12 floats.
In this situation, the last 3 floats in buffer have no default value.

cpp/open3d/t/geometry/PointCloud.cpp
```diff
@@ -1369,6 +1369,78 @@ core::Tensor PointCloud::ComputeMetrics(const PointCloud &pcd2,
...
+            if (degree.value() > 2) {
+                utility::LogWarning("The SH coeffs degree higher than 2 is not supported for now.");
+            }
+            return degree.value();
+        } else {
+            utility::LogError("The number of SH coeffs is not a valid value.");
+        }
+        return -1;
+    }
+}
```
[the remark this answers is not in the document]
zhangyang-intel: Many thanks for your advice! This is simpler!
May be the degree is `sqrt(Nc+1) - 1`. For example, degree = 1, `Nc = 3`, `sqrt(Nc+1) = 2`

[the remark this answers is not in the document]
ssheorey: You are right!

cpp/open3d/visualization/rendering/filament/GaussianSplatBuffers.cpp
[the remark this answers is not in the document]
ssheorey: I just meant the mandatory buffers: "rot", "color", "scale" - (line 151, 161, ). Here all values are always present, so a `memset` is not required. For scale, the last value is unused, so we can set it to 0 for safety in the for loop.
For `f_rest`, `memset` makes sense for the 3rd buffer (only 1 value present) for sh_degree=1. Not needed in other cases.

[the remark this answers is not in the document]
zhangyang-intel: I understand. I only use `memset` for scale buffer and the last f_rest buffer.

cpp/open3d/visualization/rendering/filament/FilamentResourceManager.cpp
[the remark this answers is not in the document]
zhangyang-intel: After some tests, i think color space is linear RGB based on the following:
1. the tpye of `f_dc` and `f_rest` is float32 in `.npz` file.
2. some of data in `f_dc` is negative.
3. I try to use `linear_to_sRGB` and `sRGB_to_linear` in `.mat` file
**raw picture:**
raw
**linear_to_sRGB:**
l_to_srgb
**sRGB_to_linear**
srgb_to_l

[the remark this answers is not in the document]
ssheorey: About the negative f_dc values: this is due to "ringing" artifacts ([wikipedia](https://en.wikipedia.org/wiki/Ringing_artifacts)) that you see when trying to approximate a high frequency signal with low frequency approximation. Spherical harmonics are 2D analog of Fourier series and exhibit the same issue. To mitigate this issue, we should clamp the final output color values to the [0, 1] range. Apply any color processing (e.g. linear_to_sRGB) after the clamping.

cpp/open3d/t/geometry/PointCloud.cpp
```diff
@@ -1369,6 +1369,55 @@ core::Tensor PointCloud::ComputeMetrics(const PointCloud &pcd2,
...
+
+        deg = deg - 1;
+        if (deg > 2) {
+            utility::LogWarning(
+                    "SH degree higher than 2 is not supported. The SH degree "
+                    "of in input data is {}, we set it as 2.",
+                    deg);
+        }
+
+        return deg - 1;
```
ssheorey, line 1418: Shouldn't this be `return deg;`?

zhangyang-intel: Sure, my fault. I have fixed this.

ssheorey (approved):

PR #7205, opened 2025-03-18
Update metadata

### LINKED ISSUE #7163: Please remove deprecated metadata
Hello, 

Please remove [`obsoletes` ](https://packaging.python.org/en/latest/specifications/core-metadata/#obsoletes) and [`provides`](https://packaging.python.org/en/latest/specifications/core-metadata/#provides) which have been deprecated for long time.
https://github.com/isl-org/Open3D/blob/c6d474b3fa0b47adbcff51219f5928855c3bb806/python/setup.py#L189

Why? Because those metadata are deprecated but also because some internal tools that manipulate wheel format and metadata breaks on it.

One can instead use [`obsoletes-dist`](https://packaging.python.org/en/latest/specifications/core-metadata/#obsoletes-dist-multiple-use) and [`provides-dist`](https://packaging.python.org/en/latest/specifications/core-metadata/#provides-dist-multiple-use)

Thanks,

ccoulombe:
Replace deprecated `provides` with `prodives-dist`. 
Replace deprecated `obsoletes` with `obsoletes-dist`.

## Type

-   [X] Bug fix (non-breaking change which fixes an issue): Fixes #7163 

## Motivation and Context

Remove deprecated metadata

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -190,9 +190,8 @@ setup_args = dict(
     description="@PROJECT_DESCRIPTION@",
     long_description=long_description,
     long_description_content_type="text/x-rst",
-    # Metadata below is valid but currently ignored by pip (<=v23)
-    obsoletes=["open3d_python"],
-    provides=["open3d", "open3d_cpu", "open3d_xpu"],  # For open3d-cpu
+    obsoletes_dist=["open3d_python"],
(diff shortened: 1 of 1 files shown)


ssheorey (approved):
Thanks @ccoulombe !

PR #7230, opened 2025-04-25
fix(cmake): Correctly report system BLAS in summary

sincethestudy:
# fix(cmake): Correctly report system BLAS in configuration summary

## Type

## Motivation and Context

When configuring Open3D with the `USE_SYSTEM_BLAS=ON` option, the final configuration summary table incorrectly reported `BLAS ..... no`. This occurred even though CMake successfully found the system BLAS/LAPACK libraries and configured the build to use them (verified via `CMakeCache.txt` and the addition of the `USE_BLAS` preprocessor definition).

This misleading summary output could cause confusion for users building with system BLAS. This change corrects the summary output to accurately reflect the configuration.

## Checklist:

-   [ ] I have run `python util/check_style.py --apply` to apply Open3D **code style** to my code. 
    -   [ ] I have added or updated C++ and / or Python **unit tests** OR included **test results** (e.g. screenshots or numbers) here.
    
## Description

This PR modifies the `cmake/Open3DPrintConfigurationSummary.cmake` script.

The original logic for printing the status of third-party dependencies checked for the existence of an internal CMake target (`Open3D::3rdparty_${dep_lower}`). However, when `USE_SYSTEM_BLAS=ON`, this target (`Open3D::3rdparty_blas`) is apparently not created, causing the script to fall into the `else` block and print `"no"`.

The fix adds a specific condition within this `else` block: if the dependency being checked is `"BLAS"` and the `USE_SYSTEM_BLAS` variable is ON, it now correctly prints `"yes (system)"` instead of `"no"`. This ensures the summary accurately reflects the use of a system-provided BLAS library.

**Before:**
-- Third-Party Dependencies: -- BLAS .................................... no
**After:**
-- Third-Party Dependencies: -- BLAS .................................... yes (system)


ssheorey (approved):
Thanks @sincethestudy !

PR #7234, opened 2025-04-29
Add error handling for insufficient correspondences in AdvancedMatching

sutatoruta:
## Type

## Motivation and Context

The tuple test within AdvancedMatching requires at least 3 correspondences, but this requirement wasn't previously checked. This lack of validation led to an error during random number generation specifically when there were no correspondences. While having 1 or 2 correspondences did not cause this specific error, processing them is also ineffective for this test.

This PR implements the check for the minimum requirement of 3 correspondences. This prevents the error observed with 0 correspondences and also avoids unproductive processing for the ineffective cases (1 or 2 correspondences), without otherwise altering the behavior.

## Checklist:

    to my code.
        updated accordingly.
    

cpp/open3d/pipelines/registration/FastGlobalRegistration.cpp
```diff
@@ -69,6 +69,14 @@ static std::vector<std::pair<int, int>> AdvancedMatching(
     int ncorr = static_cast<int>(corres_cross.size());
     int number_of_trial = ncorr * 100;
 
+    if (ncorr <= 2) {
+        utility::LogError(
```
ssheorey, line 73: Hi @sutatoruta thanks for catching this! Can you update the Error to a Warning? LogError will throw an exception and disrupt calling code. Use it only when there is no valid result that we can return. Here `{}` is a valid result.

sutatoruta: Thank you for the review! You're absolutely right - I've fixed it in https://github.com/isl-org/Open3D/pull/7234/commits/e195077b53f8d2050d954d77dcc4f762174580fc.

ssheorey (approved):
Thanks!

PR #7252, opened 2025-05-27
only compile ExtractZIP when WITH_MINIZIP is enabled

christian-rauch:
Fix compilation with `WITH_MINIZIP` set to `OFF`.

## Type

## Motivation and Context

Fix compiling `ExtractZIP.cpp` when `WITH_MINIZIP` is `OFF`. See https://github.com/isl-org/Open3D/pull/7177#issuecomment-2911193232.

## Checklist:

    to my code.
        updated accordingly.
    

christian-rauch:
@ssheorey Many tests fail because the log file `/root/open3d_data/extract/SampleRedwoodRGBDImages/trajectory.log` cannot be read. I guess this log file is compressed and cannot be read because of the missing `ExtractZIP.cpp` support. Should the test be disabled in this case?

ssheorey:
Hi @christian-rauch I'd prefer to just keep minizip ON all the time, to simplify the config and build process. We can just remove that option and fail the build if minizip is not available.

christian-rauch:
@ssheorey Can you check this again? `WITH_MINIZIP` is now `ON` again.

PR #7258, opened 2025-05-31
Expose the `get_plotly_fig` function, return fig from `draw_plotly` and `draw_plotly_server`

gspencergoog:
## Type

-   [X] New feature (non-breaking change which adds functionality).

## Motivation and Context

When trying to use `draw_plotly` in Colab, it's impossible to modify things like the aspect ratio, and other aspects of the layout, since they aren't parameters to `draw_plotly`. It's also not possible to modify them *before* it is shown.

## Checklist:

    to my code.
-   [X] This PR changes Open3D behavior or adds new functionality.
    -   [X] Both C++ (Doxygen) and Python (Sphinx / Google style) **documentation** is
        updated accordingly.
    -   [] I have added or updated C++ and / or Python **unit tests** OR included **test

## Description

This change exports the `get_plotly_fig` function so that you can generate a `Figure`, set its layout parameters, and then show it.  It also modifies `draw_plotly` to return the `fig` object when called, so that if you capture the figure, you can modify its layout after it is shown.  Similarly, I modified `draw_plotly_server` to return the `app` and the `fig`, since those also allow configuration after the fact.

I also added missing documentation for the three functions.

## Tests

This PR doesn't change any functionality other than publicly exposing an existing function and returning values from functions that used to return nothing, so no unit tests were added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,7 @@
 -   Add optional indices arg for fast computation of a small subset of FPFH features (PR #7118).
 -   Fix CMake configuration summary incorrectly reporting `no` for system BLAS. (PR #7230)
 -   Add error handling for insufficient correspondences in AdvancedMatching (PR #7234)
+-   Exposed `get_plotly_fig` and modified `draw_plotly` to return the `Figure` it creates. (PR #7258)
 
 ## 0.13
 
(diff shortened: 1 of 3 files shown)


gspencergoog:
@ssheorey Would you be the right person to review this?

ssheorey (approved):
Thanks @gspencergoog this is quite useful. Especially appreciate the documentation!

Issue #4581, opened 2022-01-11
Summarize the question. (e.g., "How to run reconstruction system with RealSense D435i? ")


singhw22:
### Checklist

- [X] I have searched for [similar issues](https://github.com/isl-org/Open3D/issues).
- [X] For Python issues, I have tested with the [latest development wheel](http://www.open3d.org/docs/latest/getting_started.html#development-version-pip).
- [X] I have checked the [release documentation](http://www.open3d.org/docs/release/) and the [latest documentation](http://www.open3d.org/docs/latest/) (for `master` branch).

### My Question

Hello guys, is Open3D possible to through OCTree to find neighbours points within a given radius?
Like, I can query the points with distances to the anchor point less than a given radius after building the KDTree in Open3D right now. Any help would be appreciated!

roman-4erkasov:
@shw-how As i understand properlly, I've faced the same problem. This article helped me: https://www.open3d.org/docs/release/python_api/open3d.ml.torch.ops.radius_search.html

ssheorey:
You can use nearest neighbor search:

This uses a KDTree internally on CPU and direct search / hash grid on a GPU.

https://www.open3d.org/docs/release/python_api/open3d.core.nns.NearestNeighborSearch.html

PR #7259, opened 2025-06-01
[Feature] Find neighbours points within a given radius through Octree

chunibyo-wly:
## Type

## Checklist:

    to my code.
        updated accordingly.
    
## Description

This feature implements the issue #4581 , which could return indices of the points used to construct the octree.

However, I have two concerns:

1. The first issue is that, the points are not stored in either the Octree class or the OctreeNode class, causing the new function needing the points to be passed in as input. This feels a bit weird and creates challenges for further development. What do you think about changing the behavior of the octree, for example, saving Octree to a JSON file, the point cloud could be saved separately, with a relative path referenced in the JSON file?
2. The second is the variable member `origin` in Octree means minimum corner of the Octree node. Would it cause misleading in the future?


chunibyo-wly:
Hello, may I ask for any updates on this?

Issue #7130, opened 2025-01-08
Dynamic linker build error on Linux ARM64 prevents Linux ARM64 wheels for 0.19 release


johnthagen:
### Checklist

- [X] I have searched for [similar issues](https://github.com/isl-org/Open3D/issues).
- [X] For Python issues, I have tested with the [latest development wheel](https://www.open3d.org/docs/latest/getting_started.html#development-version-pip).
- [X] I have checked the [release documentation](https://www.open3d.org/docs/release/) and the [latest documentation](https://www.open3d.org/docs/latest/) (for `main` branch).

### Steps to reproduce the issue

> ARM64 Linux builds are not available due to a runtime dynamic linker error: "cannot allocate memory in static TLS block"

See: https://github.com/isl-org/Open3D/actions/runs/12604362802/job/35131136413

### Error message

```shell
____ ERROR collecting test/t/registration/test_transformation_estimation.py ____
ImportError while importing test module '/root/Open3D/python/test/t/registration/test_transformation_estimation.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
../miniconda3/envs/open3d/lib/python3.10/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
python/test/t/registration/test_transformation_estimation.py:8: in <module>
    import open3d as o3d
../miniconda3/envs/open3d/lib/python3.10/site-packages/open3d/__init__.py:100: in <module>
    from open3d.cpu.pybind import (
E   ImportError: /root/miniconda3/envs/open3d/lib/python3.10/site-packages/open3d/cpu/pybind.cpython-310-aarch64-linux-gnu.so: cannot allocate memory in static TLS block
```

### Open3D, Python and System information

```markdown
- Linux
- ARM64
```

### Additional information

This prevents building Linux ARM wheels for the 0.19 release

- #7121 
- #7128

Linux ARM wheels are especially important for macOS developers who run Docker containers locally. Since macOS devices have migrated to ARM64 architecture, Linux containers running on macOS ARM64 require ARM64 Linux wheels.

This prevents us for upgrading to Open3D 0.19, and thus transitively also prevents us from having Python 3.12 support.

fabiannagel:
+1, still an issue. is there any update here?

johnthagen:
Not totally sure, but @ssheorey may be working on this in 

- https://github.com/isl-org/Open3D/pull/7134

It's mentioned in the "CI errors left" section

[...]

PR #7268, opened 2025-06-09
[BUG] Fixed ARM Builds

johnnynunez:
FIXES: #7130 
ARM64 Linux builds are not available due to a runtime dynamic linker error: "cannot allocate memory in static TLS block"

Context:
```
Error migrate from 18.04 to 20.04 with glibc bugged.
```

Fixed:
```
Arm builds come back to build again without LTS Block problem.
```

Includes CUDA ARM Working again with >=12.1 cuda. Tested on Jetson and GH200
Necessary changes:

Upgrade dockers to 22.04
Upgrade filament
Upgrade openblas to support Grace GH200 and newer instructions and fixes for Parallel
Fix Some cpp to do compatible filament
Fix Cmake for correct path lib
Upgrade cutlass to support GB200 blackwell


johnnynunez:
@ssheorey the tests results are good, the problem is that it has some differences with the new compiles between aarc64 and x86. But it is like 1e-5 vs 1e-6

PR #7275, opened 2025-06-15
upgrade pytorch

johnnynunez:
related to:
https://github.com/isl-org/Open3D-ML/pull/677


ssheorey:
Hi @johnnynunez thanks for your efforts in upgrading various Open3D dependencies. These are highly inter-dependent, so we have to be quite careful with versions to ensure everything works on all platforms. Here is the upgrade plan that should work:

```md
### Update plan

- PyTorch 2.7.1
- CUDA 12.6
- Tensorflow 2.19.0 (drop CUDA support, since this is built with CUDA 12.3)
- Ubuntu 20.04 -> Ubuntu 22.04
- Windows 2019, MSVC 16 -> Windows 2022, MSVC 17
- Python 3.8-3.12 -> Python 3.10-3.13 (drop 3.8 and 3.9)
- Linux cxx11_abi only in CI. Keep old ABI option in cmake.
- Update style checkers to latest: clang-format 20.1.6, yapf 0.43.0, nbformat 5.10.4.
```

Can you update your PRs to use these versions everywhere? Please let me know if some of these versions don't work, or we need to update other dependencies, and I'll update this list.

I've opened a draft PR #7286 to take care of build errors, especially since the CUDA CI does not run on PRs from forks. As long as your PRs use the same versions, I'll merge yours first and then merge #7286 for any remaining CI cleanup.

johnnynunez:
> Hi @johnnynunez thanks for your efforts in upgrading various Open3D dependencies. These are highly inter-dependent, so we have to be quite careful with versions to ensure everything works on all platforms. Here is the upgrade plan that should work:
> 
> ```
> ### Update plan
> 
> - PyTorch 2.7.1
> - CUDA 12.6
> - Tensorflow 2.19.0 (drop CUDA support, since this is built with CUDA 12.3)
> - Ubuntu 20.04 -> Ubuntu 22.04
> - Windows 2019, MSVC 16 -> Windows 2022, MSVC 17
> - Python 3.8-3.12 -> Python 3.10-3.13 (drop 3.8 and 3.9)
> - Linux cxx11_abi only in CI. Keep old ABI option in cmake.
> - Update style checkers to latest: clang-format 20.1.6, yapf 0.43.0, nbformat 5.10.4.
> ```
> 
> Can you update your PRs to use these versions everywhere? Please let me know if some of these versions don't work, or we need to update other dependencies, and I'll update this list.
> 
> I've opened a draft PR #7286 to take care of build errors, especially since the CUDA CI does not run on PRs from forks. As long as your PRs use the same versions, I'll merge yours first and then merge #7286 for any remaining CI cleanup.

Perfect. Thank you. I’m going to do

johnnynunez:
> 12.6

it is better to upgrade to 12.8 to have blackwell support

PR #7329, opened 2025-08-19
Add serialization and deserialization methods to Octree

lumurillo:
Add serialization and deserialization methods to the Octree.

## Type

## Checklist:

    to my code.
        updated accordingly.
    
### DIFF
--- a/cpp/open3d/geometry/Octree.cpp
+++ b/cpp/open3d/geometry/Octree.cpp
@@ -49,6 +49,32 @@ std::shared_ptr<OctreeNode> OctreeNode::ConstructFromJsonValue(
     return node;
 }
 
+std::shared_ptr<OctreeNode> OctreeNode::ConstructFromBinaryStream(const std::string& in, size_t& offset) {
+    std::string class_id(&in[offset], 24);
+
+    std::shared_ptr<OctreeNode> node = nullptr;
+    if (std::string(&in[offset], 18) == "OctreeInternalNode") {
+        node = std::make_shared<OctreeInternalNode>();
+    } else if (std::string(&in[offset], 23) == "OctreeInternalPointNode") {
+        node = std::make_shared<OctreeInternalPointNode>();
+    } else if (std::string(&in[offset], 19) == "OctreeColorLeafNode") {
+        node = std::make_shared<OctreeColorLeafNode>();
+    } else if (std::string(&in[offset], 24) == "OctreePointColorLeafNode") {
+        node = std::make_shared<OctreePointColorLeafNode>();
+    } else {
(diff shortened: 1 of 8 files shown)


ssheorey (commented):
Hi @lumurillo thanks for this nice addition. Some thoughts:
- Can we use msgpack, instead of raw binary format for serialization? This is already used elsewhere in Open3D, is simple to use, should provide good speed and smaller size and will also be cross-platform. Raw binary format sometimes has issues if you write on Windows and read in Linux or vice versa. It will also enable reading / writing from plain Python through the Python msgpack library. 
- msgpack in Open3D: https://github.com/isl-org/Open3D/blob/fb7088ceebef38d54c575b47935e568979b50954/cpp/open3d/io/rpc/Messages.h#L250 (and other places - search for msgpack). We use `*.msgpack` files to store tensor geometries in the Tensorboard plugin, for example.
- Tutorial: https://github.com/msgpack/msgpack-c/wiki/v2_0_cpp_tutorial

ssheorey:
From Github Copilot, about the macOS errors:

The build is failing due to missing override specifiers for functions that override virtual methods in the Octree classes. The compiler treats this as an error because of the -Werror,-Winconsistent-missing-override flag. For example:

```
error: 'SerializeToBinaryStream' overrides a member function but is not marked 'override' [-Werror,-Winconsistent-missing-override]
```

To fix this, add the override keyword to all overriding methods in the derived classes. Specifically, in cpp/open3d/geometry/Octree.h (ref: 1fada55e1c46344d4b0bd63d3668240a0ff2f9dc), update the relevant function declarations as follows:

For OctreeInternalNode:
```cpp
bool SerializeToBinaryStream(std::string& out) const override;
bool DeserializeFromBinaryStream(const std::string& in, size_t& offset) override;
```

For OctreeInternalPointNode:
```cpp
bool SerializeToBinaryStream(std::string& out) const override;
bool DeserializeFromBinaryStream(const std::string& in, size_t& offset) override;
```

For OctreeColorLeafNode:
```cpp
bool SerializeToBinaryStream(std::string& out) const override;
bool DeserializeFromBinaryStream(const std::string& in, size_t& offset) override;
```

For OctreePointColorLeafNode:
```cpp
bool SerializeToBinaryStream(std::string& out) const override;
bool DeserializeFromBinaryStream(const std::string& in, size_t& offset) override;
```

Add override to all other similar functions that are overriding a base class method but are missing it. This will resolve the compilation errors and allow the build to succeed.

You can see the problematic lines in Octree.h here:  
https://github.com/isl-org/Open3D/blob/1fada55e1c46344d4b0bd63d3668240a0ff2f9dc/cpp/open3d/geometry/Octree.h

PR #7337, opened 2025-09-17
Fix logic for adding -allow-unsupported-compiler to nvcc

patrikhuber:
The previous logic would always add `-allow-unsupported-compiler` and `-D_ALLOW_COMPILER_AND_STL_VERSION_MISMATCH`, even when it was not needed (e.g. when a recent CUDA version is used), and also printing a (confusing) warning which is not applicable in that case.

The revised logic only adds the flags if MSVC>=1940 and CUDA<=12.4. For that, we need to use `find_package(CUDAToolkit REQUIRED)`, like it's also done a few more lines below in the `if(BUILD_COMMON_CUDA_ARCHS)` code - and then we append to `CMAKE_CUDA_FLAGS` only when needed.

The new code also uses CMAKE_CUDA_FLAGS for the `_ALLOW_COMPILER_AND_STL_VERSION_MISMATCH` define, which makes CMake's try_compile (which is used when `enable_language(CUDA)` is called) work reliably if a CMake version >=3.29.4 is used. It doesn't work when `add_compile_definitions(_ALLOW_COMPILER_AND_STL_VERSION_MISMATCH)` is used.

## Type

-   [x] Bug fix (non-breaking change which fixes an issue): Fixes the issue described above.

## Motivation and Context

See above.

## Checklist:

    to my code.
        updated accordingly.
    
## Description

See above.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,7 @@
 -   Fix build with librealsense v2.44.0 and upcoming VS 2022 17.13 (PR #7074)
 -   Fix `deprecated-declarations` warnings when compiling code with C++20 standard (PR #7303)
 -   Fix logic for adding -allow-unsupported-compiler to nvcc (PR #7337)
+-   Fix linker error "library limit of 65535 objects exceeded" with Ninja generator on MSVC (PR #7335)
 
 ## 0.13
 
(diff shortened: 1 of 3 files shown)


patrikhuber:
@ssheorey could we get this merged?

patrikhuber:
@ssheorey Anything I can do to get this merged?

patrikhuber:
> Thanks @patrikhuber I simplified the update by removing the duplicate flag check. Duplicate flags should be harmless.

Thanks for merging! I personally think the version with the check is better (and only marginally longer), as the version you merged will add those two flags _every time_ CMake configure is run (which can be very often!). However, maybe one argument for your stripped down version is that medium-term everyone is going to be at CUDA above version 12.5, so this will not be needed anymore anyway (and the flags aren't added for anyone already on CUDA > 12.5 anyway).

PR #7360, opened 2025-11-04
Added assertion to guard against memory corruption

TwentyPast4:
## Type

## Motivation and Context

This is not strictly a bug fix, it is just a low-cost method of making sure advanced indexing operations don't go out of bounds. The consequence of this happening could be either memory corruption, illegal memory access, or just undefined behavior. It makes sense to guard against these cases, as it is probably one of the more opaque ways a tensor operation might result in such behavior. Other comparable invalid operations might be caught by the user, but in my experience this one is the most opaque one. An example that would cause this issue:
```cpp
Tensor a = Tensor::Init<int64_t>({0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10});
Tensor b = Tensor::Zeros({10}, open3d::core::Float32);
b.IndexAdd_(0, a, Tensor::Ones({a.GetLength()}, open3d::core::Float32));
// one of the operations would end up being (pseudocode) b[10] += 1;
```

The use cases of IndexSet, IndexGet, IndexAdd that I've come across have been of a computational nature, meaning the index was computed based on some arbitrary input data. Meaning that without proper due diligence, the validity of the indices is dependent on input data.

## Checklist:

    to my code.
        updated accordingly.
    
## Description

Added an assertion check to Indexer, in a function which all index operations call to get input & output pointers.


ssheorey:
Hi @TwentyPast4 thanks for this useful PR! How does this compare to the behavior of PyTorch and Numpy? The Open3D tensors are designed to be consistent with them.

Also, `assert` will be ignored in Release mode, where most users of Open3D work. Debug mode is only for Open3D developers.

ssheorey (approved):

ssheorey:
Thanks for the questions — addressing both here:

**PyTorch / NumPy parity:** Out-of-range advanced indexing is undefined in NumPy/PyTorch (silent memory corruption or garbage), not a catchable Python exception. This PR adds **developer-facing** guards (similar to PyTorch’s CUDA device-side `assert` / `TORCH_CHECK` on device) so invalid indexing fails loudly during development instead of corrupting memory. It is intentionally stricter than end-user NumPy semantics, aligned with how PyTorch treats device indexing bugs.

**Release vs `assert`:** Host code uses `OPEN3D_ASSERT` → `Logger::LogError_` (not libc `assert`, so not stripped in Release). CUDA/SYCL device paths use device-side reporting + trap, comparable to PyTorch’s device asserts, so invalid indexing is still caught when running GPU/XPU kernels even in Release builds.

Happy to adjust messaging or scope if we want host-only checks in a follow-up.

PR #7374, opened 2025-11-16
[Draft] support PointCloud pickle

chunibyo-wly:
## Type

## Motivation and Context

Currently, the Non-tensor version of Geometry Class can not work with multiprocessing very well, as these classes miss the implementation of `py::pickle`.

May I ask if it is the proper implementation to support pickle? If so, I will continue to finish the remaining classes (lineset, triangle mesh, ...).

### DIFF
--- a/cpp/pybind/geometry/pointcloud.cpp
+++ b/cpp/pybind/geometry/pointcloud.cpp
@@ -45,6 +45,21 @@ void pybind_pointcloud_definitions(py::module &m) {
                  })
             .def(py::self + py::self)
             .def(py::self += py::self)
+            .def(py::pickle(
+                    [](const PointCloud &pcd) {
+                        return py::make_tuple(pcd.points_, pcd.normals_,
+                                              pcd.colors_, pcd.covariances_);
+                    },
+                    [](py::tuple t) {
+                        PointCloud pcd;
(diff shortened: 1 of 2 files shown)


chunibyo-wly:
The minimal test case is

```python
import pickle

import numpy as np
import open3d as o3d

pcd = o3d.io.read_point_cloud(o3d.data.KnotMesh().path)
pcd.estimate_normals()
pcd.estimate_covariances()
print(pcd.has_covariances(), pcd.has_normals())
data = pickle.dumps(pcd)
pcd_loaded = pickle.loads(data)


print(np.all(np.asarray(pcd.points) == np.asarray(pcd_loaded.points)))
print(np.all(np.asarray(pcd.normals) == np.asarray(pcd_loaded.normals)))
print(np.all(np.asarray(pcd.colors) == np.asarray(pcd_loaded.colors)))
print(np.all(np.asarray(pcd.covariances) == np.asarray(pcd_loaded.covariances)))
```

And the stdout is
```
True True
True
True
True
True
```

ssheorey:
Hi @chunibyo-wly yes, you are correct - pickling support would enable multiprocessing with eigen based geometry types. Happy to merge this PR if you want to add support for all of the eigen based / legacy geometry types. This is not a priority for us since we are migrating to the new tensor based types that already support pickling.

PR #7382, opened 2025-12-09
Add value-based equality for TriangleMesh

sean85914:
## Type

-   [x] New feature (non-breaking change which adds functionality).

## Motivation and Context

Currently, `TriangleMesh` instances in Python compare by identity due to the lack of a C++ `operator==`.
This leads to unintuitive behavior:

```python
m1 = o3d.geometry.TriangleMesh.create_sphere(0.1)
m1.compute_vertex_normals()
m2 = o3d.geometry.TriangleMesh(m1)

m1 == m2   # False
```
Although `m2` is a copy of `m1`, equality returns False.
This PR adds a value-based equality operator for TriangleMesh to fix this.

## Checklist:

    to my code.
        updated accordingly.

## Description
* Added `bool TriangleMesh::operator==(const TriangleMesh &mesh) const`
* Equality is defined as vertices and triangles being identical.
* `vertices_` comparison uses `isApprox` for floating-point precision.
* `triangles_` comparison uses exact equality for integer indices.
* Bound Python equality using `py::self == py::self` for TriangleMesh.
* Updated Python behavior:
```python
m1 = o3d.geometry.TriangleMesh.create_sphere(0.1)
m2 = o3d.geometry.TriangleMesh(m1)
assert m1 == m2   # True
```
* Copy constructor remains consistent with value-based equality semantics.
* Normals, colors, and other auxiliary attributes are not considered in equality, as they can be recomputed.

### DIFF
--- a/cpp/open3d/geometry/TriangleMesh.cpp
+++ b/cpp/open3d/geometry/TriangleMesh.cpp
@@ -114,15 +114,15 @@ TriangleMesh TriangleMesh::operator+(const TriangleMesh &mesh) const {
 }
 
 bool TriangleMesh::operator==(const TriangleMesh &mesh) const {
-    if(vertices_.size() != mesh.vertices_.size()) return false;
-    if(triangles_.size() != mesh.triangles_.size()) return false;
+    if (vertices_.size() != mesh.vertices_.size()) return false;
+    if (triangles_.size() != mesh.triangles_.size()) return false;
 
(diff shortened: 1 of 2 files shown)


ssheorey:
A feature like this needs motivation (Why?) and use cases (When would you use this?). The implementation provided here is incorrect, since it ignores a reordering of the vertices.

Issue #7397, opened 2026-01-13
"Cannot compile on Ubuntu 24.04 on Jetson Thor with Cuda 13"


preet-dhir-azalea:
### Checklist

- [x] I have searched for [similar issues](https://github.com/isl-org/Open3D/issues).
- [x] For Python issues, I have tested with the [latest development wheel](https://www.open3d.org/docs/latest/getting_started.html#development-version-pip).
- [x] I have checked the [release documentation](https://www.open3d.org/docs/release/) and the [latest documentation](https://www.open3d.org/docs/latest/) (for `main` branch).

#### I first cloned Open3D by:
```
git clone https://github.com/isl-org/Open3D.git
cd Open3D
```

#### Then, I build Open3D (on Ubuntu 24.04, with CUDA 13.0 on Jetson Thor) with:
```
mkdir build

cd build

cmake -DBUILD_CUDA_MODULE=ON -DBUILD_GUI=ON -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Release -DPYTHON_EXECUTABLE=$(which python3) -DUSE_SYSTEM_OPENSSL=ON -DUSE_SYSTEM_CURL=ON ..

make -j$(nproc)
```

### Error message

```shell
-- 
-- Configuring done (0.1s)
-- Generating done (0.0s)
-- Build files have been written to: /home/azalea-robotics-arc-001/Open3D/build/stdgpu/src/ext_stdgpu-build
[ 33%] Performing build step for 'ext_stdgpu'
[ 16%] Building CXX object src/stdgpu/CMakeFiles/stdgpu.dir/impl/device.cpp.o
[ 33%] Building CXX object src/stdgpu/CMakeFiles/stdgpu.dir/impl/iterator.cpp.o
[ 50%] Building CXX object src/stdgpu/CMakeFiles/stdgpu.dir/impl/memory.cpp.o
[ 66%] Building CXX object src/stdgpu/CMakeFiles/stdgpu.dir/cuda/impl/device.cpp.o
[ 83%] Building CXX object src/stdgpu/CMakeFiles/stdgpu.dir/cuda/impl/memory.cpp.o
/home/azalea-robotics-arc-001/Open3D/build/stdgpu/src/ext_stdgpu/src/stdgpu/cuda/impl/device.cpp: In function ‘void stdgpu::cuda::print_device_information()’:
[...]

PR #7398, opened 2026-01-13
Upgrade stdgpu to fix compilation errors on Jetson Thor with Ubuntu 24.04 and Cuda 13.0

preet-dhir-azalea:


preet-dhir-azalea:
Fix for https://github.com/isl-org/Open3D/issues/7397

Issue #6778, opened 2024-04-29
Integrate DNN runtime for running inference in C++


ssheorey:
Enable 3DML processing in C++

This allows an increasing number of 3D processing methods to be available to the Open3D C++ API.

Options:
- [ ] libtorch
- [ ] onnx

adityamwagh:
I've worked on similar stuff in one of my projects. Can definitely help! Can you point me to relevant material so I can get started?

ssheorey:
Hi @adityamwagh thanks for volunteering! I've added more details and a high level plan in the issue description. Let me know if you have any questions.

adityamwagh:
Thanks for the heads up! I'll have a look into it.

adityamwagh:
Hey! @ssheorey  Apologies for the delay. I have started working on this issue, should have a PR ready in a few days. 

I see that this issue is mentioned in the v0.20.0 milestone. When do you plan to relase v0.20.0?

xiaoya27:
Hi @ssheorey,
I’m very interested in this feature and have started working on the implementation for a DNN inference runtime in C++. I noticed this issue has been open for a while, so I’ve put together a roadmap to contribute this incrementally.
I have just opened a Draft PR #7414 which covers the first step of this plan.
My proposed implementation plan:

PR #7414 (Draft): Add open3d::ml::Model class skeleton.
PR2: Implement load_model with LibTorch support.
PR3: Add DLPack tensor conversion for zero-copy data exchange.
PR4: Implement forward() pass.
PR5: Integration example (e.g. GeDI model).
I’d love to get your thoughts on it, looking forward to your feedback!

adityamwagh:
Hey @xiaoya27 

There's also a basic implementation for this in my fork [here](https://github.com/isl-org/Open3D/compare/main...adityamwagh:Open3D:main) that you can refer to. BTW it seems torch inductor is not stable yet and needs PyTorch installed.

Please feel free take over this issue 😄 as I haven't had time to work on this since a long time. @ssheorey Can you assign @xiaoya27 to this PR?

PR #7414, opened 2026-01-25
Feature/ml model skeleton

xiaoya27:
## Type

- [ ] Bug fix (non-breaking change which fixes an issue): Fixes #
- [x] New feature (non-breaking change which adds functionality). Resolves #
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected) Resolves #

## Motivation and Context

This is part of the #6778 initiative and part of the roadmap in the discussion . It adds the foundational `open3d::ml::Model` class that will allow loading PyTorch models exported via AOTInductor for C++ inference.

This PR establishes the API skeleton. Follow-up PRs will implement:
- LibtorchLoader with dlopen (PR2)
- DLPack tensor conversion to enable zero-copy data exchange between Open3D and PyTorch (PR3)
-  Implementation of the forward() pass with a basic linear layer unit test (PR4)
- End-to-end integration example using a model like GeDI(PR5) 

## Checklist:

- [x] I have run `python util/check_style.py --apply` to apply Open3D **code style** to my code.
- [x] This PR changes Open3D behavior or adds new functionality.
    - [ ] Both C++ (Doxygen) and Python (Sphinx / Google style) **documentation** is updated accordingly.
    - [x] I have added or updated C++ and / or Python **unit tests** OR included **test results** (e.g. screenshots or numbers) here.
- [x] I will follow up and update the code if CI fails.
- [x] For fork PRs, I have selected **Allow edits from maintainers**.

## Description

Adds `open3d::ml::Model` class with:
- `LoadModel(path)` - stub for loading model artifacts
- `Forward(inputs)` - stub for running inference  
- `IsPyTorchRuntimeEnabled()` - compile-time check for `BUILD_PYTORCH_OPS`
- `OPEN3D_BUILD_PYTORCH_OPS` compile definition

Unit tests included for basic API validation.


ssheorey:
Hi @xiaoya27 thanks for picking this up! This is awesome!

We already have DLPack support, so you should be able to skip this.

Also, since the roadmap discussion, PyTorch has released **executorch** which is specifically designed for deployment (inference). This is quite a bit lighter than full PyTorch, while being compatible with PyTorch checkpoints. We should switch to executorch for Open3D.

https://pytorch.org/blog/introducing-executorch-1-0/

xiaoya27:
Thanks @ssheorey! ExecuTorch makes a lot of sense here given its focus on lean inference.  I'd like to clarify the integration strategy:

- Integration Method: Given ExecuTorch's modular nature, would you prefer it as a CMake build option (e.g., -DBUILD_EXECUTORCH=ON) that links the runtime at compile-time, or should I stick to the dlopen approach to keep the core Open3D shared library completely independent? (I'm leaning towards CMake for better stability and simpler code).
- DLPack: Noted! I see Open3D has `Tensor::ToDLPack()` / `Tensor::FromDLPack()` already. I'll use these for tensor conversion. 
- Revised PR: I will update PR #7414 to reflect this shift toward ExecuTorch

ssheorey:
Either approach - dlopen() or static / shared link at build time is fine. We will accept either. The cmake code required will likely be very similar. dlopen() has the advantage of keeping this heavy feature optional for users of the Open3D binaries / python package.

Issue #6612, opened 2024-01-18
Summarize the bug (e.g., "Segmentation Fault for Colored Point Cloud Registration")


dsurovik-bdai:
### Checklist

- [X] I have searched for [similar issues](https://github.com/isl-org/Open3D/issues).
- [X] For Python issues, I have tested with the [latest development wheel](https://www.open3d.org/docs/latest/getting_started.html#development-version-pip).
- [X] I have checked the [release documentation](https://www.open3d.org/docs/release/) and the [latest documentation](https://www.open3d.org/docs/latest/) (for `main` branch).

### Describe the issue

I am using `o3d.t.geometry.PointCloud.project_to_rgbd_image()` and have found strange and problematic black-pixel artifacts in the color result _only_ when using CUDA - they do not appear on CPU.  The depth image appears normal at the corresponding pixels.

I first observed this running on a live feed of a static scene -- the artifacts were fairly constant but did change somewhat with the changing sensor noise.  Below I'm including code for reproducing the effect on a fake point cloud.

The point cloud is generated on an xy grid and the projection comes from an overhead (z-direction) camera.  I offset the point z coordinates either randomly or as a function of radius.
- Random --> artifact locations also appear random
- Radial --> artifacts curiously occupy one diagonal half of the view.

![o3d17_rand](https://github.com/isl-org/Open3D/assets/129541401/9b289072-1d87-40d0-807e-858a10e6702e)
![o3d18dev](https://github.com/isl-org/Open3D/assets/129541401/697b4cef-43d5-48ca-8750-1b7839a80134)

### Steps to reproduce the bug

```python
import matplotlib.pyplot as plt
import numpy as np
import open3d as o3d

# Parameters
size_m = 0.25  # Size of square image in meters
size_px = 64  # Size of square image in pixels
size_pts = 300  # Number of points in created pointcloud (gets squared)
camera_height_m = 10
device_names = ["CPU:0", "CUDA:0"]
height_mode = "random" # in ["random", "radial"]

# Derived values
half_size_px = size_px // 2
m_per_px = size_m / size_px
focal_length_px = camera_height_m / m_per_px
projection_kwargs = {
        'width': size_px,
        'height': size_px,
        'extrinsics': o3d.core.Tensor([  # 10m above origin, pointing straight down
            [1, 0, 0, 0],
            [0, -1, 0, 0],
            [0, 0, -1, camera_height_m],
            [0, 0, 0, 1],
        ]),
[...]

PR #7424, opened 2026-01-31
Fix color artifacts in PointCloud projection due to CUDA race condition

ManuCorrea:
## Type

## Motivation and Context

It solves the artifacts described in https://github.com/isl-org/Open3D/issues/6612 which happen to be because a race condition on the CUDA code.

The existing ProjectCUDA implementation suffered from a race condition in the color assignment pass. Multiple threads with similar depths in the color assignment pass (within the precision_bound) could simultaneously pass the depth test and write to the RGB color buffer, leading to the artifacts shown in the mentioned issue.

## Checklist:

    to my code.
        updated accordingly.
    
## Description
To fix it I implemented a buffer to keep track in the pass 1 of which worker should write to the color tensor in the pass 2.

### Pass 1 (Depth & Index Acquisition):
  - Packs the depth (32-bit float bits) into the High Bits and the Point Index (32-bit workload_idx) into the Low Bits.
  - Uses atomicMin on the 64-bit value. This ensures an **unique worker for every pixel** based on the minimum depth.

### Pass 2 (Deterministic Color Assignment):
- Each thread reads the 64-bit winner from the index buffer.
- By applying a 32-bit mask (& 0xFFFFFFFF), we extract the unique ID.
- Only the thread whose workload_idx matches the ID writes to the color tensor.

Old graph from the issue:
297903988-9b289072-1d87-40d0-807e-858a10e6702e

My results:
CPUvsCUDA_correct


ManuCorrea:
I will add testing and the changes in CHANGELOG.md in the following days.

ssheorey (approved):

PR #7431, opened 2026-02-08
Fix remove to legacy in docs

h812755591:
just one file docs/jupyter/t_geometry/pointcloud.ipynb 
1,  
remove  pointcloud.to_legacy method  
2,
Conversion between tensor and legacy point cloud  ===>Conversion between tensor and point cloud 
PointCloud can be converted to/from legacy open3d.geometry.PointCloud.====>PointCloud can be converted to Tensor

### DIFF
--- a/docs/jupyter/t_geometry/pointcloud.ipynb
+++ b/docs/jupyter/t_geometry/pointcloud.ipynb
@@ -157,8 +157,8 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "## Conversion between tensor and legacy point cloud\n",
-    "PointCloud can be converted to/from legacy `open3d.geometry.PointCloud`."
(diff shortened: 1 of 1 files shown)


h812755591:
This fixes outdated documentation: to_legacy() is no longer part of the Tensor PointCloud API.
The updated wording reflects the current conversion behavior

h812755591:
This is a documentation-only change with no user-facing behavior change,
so I did not update the CHANGELOG.

ssheorey (commented):
Hi @h812755591 thanks for submitting this PR. However, we don't have support for tensor point clouds in `draw_geometries()`. You have to use the new `draw()` function, which supports both legacy and tensor point clouds.

ssheorey:
Hi @h812755591 , Happy to reopen PR after you address comment above.

PR #7447, opened 2026-02-25
fix windows CI

johnnynunez:
Fix windows CI related with MSVC

Commit eb47dea downgraded Open3D's own fmt to 9.1.0 on MSVC and removed the global /utf-8 flag from the root CMakeLists.txt...
Then commit bf8a5f9 (the tensor slicing PR) landed on top, and the CI ran, but the failure isn't from the tensor slicing change at all. It's from the removal of /utf-8 in eb47dea.

The global /utf-8 was removed because Open3D's bundled fmt no longer needs it (pinned to 9.1.0 on MSVC). But PyTorch's bundled fmt (v11+, in torch/include/fmt/base.h) still requires it, and the open3d_torch_ops target includes those PyTorch headers....

The fix I already applied is the right one,.....mmmm 
 scoped /utf-8 to just the open3d_torch_ops target rather than restoring it globally (which would re-introduce the CUDA+nvcc issue that eb47dea was trying to solve).
 
 @ssheorey

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,7 +72,7 @@
 -   Download tarballs instead of Git repos for "3rdparty/uvatlas" (PR #7371)
 -   macOS x86_64 not longer supported, only macOS arm64 is supported.
 -   Python 3.13+3.14 support
--   Fix Windows build failure for PyTorch ops due to PyTorch's bundled fmt (v11+) requiring `/utf-8` with MSVC
+-   Fix Windows build failure for PyTorch ops due to PyTorch's bundled fmt (v11+) requiring `/utf-8` with MSVC (PR #7447)
 
 
 ## 0.13


johnnynunez:
fyi @ssheorey https://docs.python.org/3.15/whatsnew/3.15.html#whatsnew315-utf8-default

investigating pytorch, it using it from 2.9.0

johnnynunez:
@ssheorey ready to merge
image

I will add now the comments on changelog

johnnynunez:
If all is green we can tag a release 0.20.0 and then upgrade to cuda 13.2 @ssheorey with 0.20.1

What do you think? There is no new wheels since 08/01/25

ssheorey (approved):

ssheorey:
Thanks @johnnynunez . Yes we are planning to release v0.20 soon.

PR #7451, opened 2026-03-02
Add expand/collapse support to TreeView

### LINKED ISSUE #7396: Add the "collapse" and "open" functions to TreeView in python
### Checklist

When adding many items to the TreeView (with different hierarchies) the widget get crowded very quickly. It's possible to collapse and open a single parent item from the interface with the mouse, but I could not find a way to do it from code.

I would like to have the option to collapse and open them via code.

### References

_No response_

### Additional information

_No response_

MiriShulman:
Adds expand/collapse support to TreeView.

This introduces:
- SetItemExpanded
- IsItemExpanded
- Python bindings exposure

Fixes #7396


ssheorey (commented):
Hi @MiriShulman thanks for reworking the PR! Could you close the old PR, please?

AI Review comments for this one (please verify yourself if they are correct):

1: `SetItemExpanded` won't actually expand or collapse already-rendered nodes because `ImGuiTreeNodeFlags_DefaultOpen` only affects the *initial* state of a tree node. After the first frame, ImGui stores its own internal state and ignores this flag entirely. The codebase already has the correct pattern in `Layout.cpp`, which uses `ImGui::SetNextItemOpen()` before the tree node call to override ImGui's stored state. Without that, calling `SetItemExpanded` sets `item.expanded`, but on the next draw frame ImGui returns its own stored state, immediately overwriting the value back.

2: The `TreeViewTest` executable is defined outside the `if (BUILD_GUI)` guard, so it's always added to the build regardless of whether GUI support is enabled. Since it includes GUI headers (`Application.h`, `Window.h`, `TreeView.h`) and links to `Open3D::Open3D` expecting GUI symbols, it will fail to compile or link when `BUILD_GUI` is `OFF`. The existing test content in the same file is properly guarded.

3. git and treeview-clean are accidentally added to the commit.

Let me know if I can help.

PR #7518, opened 2026-07-11
Replace OSMesa with EGL for legacy visualizer offscreen rendering

ssheorey:
## Summary

- Replace OSMesa-based offscreen rendering with a GPU-accelerated EGL offscreen context for the legacy visualizer (`cpp/open3d/visualization/{shader,visualizer}`) in the standard Open3D binary.
- Runtime auto-selection on Linux: `Visualizer::CreateVisualizerWindow` first tries GLFW (windowed); if no display is available, it falls back to an `EGLOffscreenContext` (pbuffer, GL 3.3 core, GPU-backed). Windows/macOS behavior is unchanged (GLFW only).
- All direct `glfwMakeContextCurrent`/`glfwSwapBuffers`/poll/wait/clipboard calls in `Visualizer`, `VisualizerRender`, `VisualizerWithEditing`, and `VisualizerWithVertexSelection` are routed through a new `Visualizer::MakeContextCurrent()` helper and guarded for headless mode.
- Removed `ENABLE_HEADLESS_RENDERING` option, `HEADLESS_RENDERING` define, `GLEW_OSMESA`, `FindOSMesa.cmake`, and OSMesa system deps; `OpenGL::EGL` is now a required, unconditional dependency on non-Apple/non-Windows builds.
- `GLInfo` now reports EGL GPU info when no display is available (instead of requiring a separate headless build).
- Documentation build simplified: `ubuntu.yml` now uploads the Python wheel from the standard build and uses it (plus the built C++ binaries) to run notebooks and generate Sphinx/Doxygen docs in a new `build-docs` job (`needs: ubuntu`), producing the same `open3d-<sha>-docs.tar.gz` artifact and `main-devel` release update. The separate `documentation.yml` workflow and `Dockerfile.docs` (which required a second, headless-only Open3D build) are removed.
- Public Python/C++ `Visualizer` APIs are unchanged; `create_window(visible=False)` + `add_geometry` + `capture_screen_float_buffer`/`capture_screen_image` now work without `DISPLAY` using the GPU.

## Test plan

- [x] CMake reconfigure: `Found OpenGL: ... found components: EGL`; `ENABLE_HEADLESS_RENDERING` no longer appears in the configuration summary.
- [x] Built `GLInfo`, `Open3D` shared lib, Python wheel (`pip-package`), and C++ `tests` target successfully.
- [x] Ran `bin/GLInfo` with no `DISPLAY` on a machine with an NVIDIA GPU: falls back to EGL 1.5 offscreen context and correctly reports `GL_RENDERER: NVIDIA GeForce RTX 4090/...`.
- [x] Added `python/test/visualization/test_legacy_headless_rendering.py`: spawns a subprocess with no `DISPLAY`/`WAYLAND_DISPLAY`, creates a legacy `Visualizer`, adds geometry, and captures a real (non-blank) GPU-rendered frame via EGL. Passes.
- [x] Ran `examples/python/visualization/headless_rendering.py` end-to-end with no `DISPLAY`: captured 31 real (non-blank) color+depth frames along a camera trajectory via EGL.
- [x] Ran C++ `bin/tests --gtest_filter="*Visualiz*:*Render*"` headlessly: passes.
- [x] Ran full `python/test/visualization/` + `python/test/test_flags.py` suite: passes (unrelated tests skipped due to missing optional deps).

Made with [Cursor](https://cursor.com)


ssheorey:
Reviewed the Copilot inline comments — all 5 were valid and are fixed in c2f4496:

1. `cpp/tools/GLInfo.cpp` `ReportEGLGLInfo()`: consolidated all early-return paths through a single `cleanup()` lambda that always terminates the EGL display/destroys surface+context, and now checks `eglBindAPI()`'s return value.
2. `cpp/tools/GLInfo.cpp` `TryGLVersion()`: clarified the doc comment — returning `true` when `glfwCreateWindow()` fails (but `glfwInit()` succeeded) is intentional, since that means the *specific GL version* is unsupported, not that there's no display. Only `glfwInit()` failing means "no display" (verified this is the only path `main()`'s `has_display` check relies on).
3. `Visualizer::WaitEvents()`: added a short `sleep_for(10ms)` in headless mode. Without it, `Run()`'s `while (WaitEvents())` loop (used when no animation callback is registered) busy-spun at ~100% CPU. Verified via `/proc/<pid>/stat` CPU-time sampling: busy-loop was consuming ~100% CPU, sleep-based version uses ~6%.
4. `util/ci_utils.sh` `build_docs()`: now `cd`s via `"${OPEN3D_SOURCE_ROOT}/docs"` instead of a relative path, consistent with the rest of the file.
5. `test_legacy_headless_rendering.py`: added `proc.join()` after `proc.kill()` to reap the process, and switched to `pytest.fail(...)` for clearer failure reporting.

Rebuilt and re-ran the full validation suite (GLInfo, wheel, pytest, C++ tests) on real GPU hardware after each fix — all pass.

PR #7524, opened 2026-07-18
Fix reading OBJ meshes from UTF-8 paths on Windows

### LINKED ISSUE #7336: Bug report: Cannot load .obj file when path contains Korean characters
### Checklist

## Description

When attempting to load an ASCII .obj file using open3d.io.read_triangle_mesh() in a Windows environment, the function fails with a UTF-8 decoding error if the file path contains Korean (Hangul) characters. However, the same file saved as ASCII .ply loads successfully under identical conditions.

## Steps to Reproduce
1. Save an ASCII `.obj` file with UTF-8 encoding.
2. Place the file in a directory with Korean characters in the path, e.g., `C:/Users/사용자/샘플/sample.obj`
3. Run the following code:

```python
import open3d as o3d
mesh = o3d.io.read_triangle_mesh("C:/Users/사용자/샘플/sample.obj")
```

4. Observe the error:
```
'utf-8' codec can't decode byte 0xbb in position XX: invalid start byte
```

5. Save the same mesh as `.ply` in ASCII format and load it from the same path:

```python
mesh = o3d.io.read_triangle_mesh("C:/Users/사용자/샘플/sample.ply")
```

6. This works without error.

```python
open3d as o3d
### this does not work
mesh = o3d.io.read_triangle_mesh("C:/Users/사용자/샘플/sample.obj")
### but this works
mesh = o3d.io.read_triangle_mesh("C:/Users/사용자/샘플/sample.ply")
```

### Error message

`'utf-8' codec can't decode byte 0xbb in position XX: invalid start byte`

## Expected Behavior
Both `.obj` and `.ply` files should load successfully from paths containing non-ASCII characters, assuming the files themselves are properly encoded.

## Actual Behavior
Only `.ply` files load successfully. `.obj` files fail with a UTF-8 decoding error.

```markdown
OS: Windows 10/11 (Korean locale)
Open3D version: 0.19.0
Python version: 3.10
File encoding: UTF-8 (confirmed)
File format: ASCII .obj and .ply
```

## Additional Notes
- The `.obj` file has been verified to be UTF-8 encoded.
- The issue seems to be related to how Open3D parses `.obj` files internally, possibly involving texture or material references.
- This issue does not occur when the path contains only ASCII characters.

## Suggested Fix
- Ensure that Open3D’s `.obj` parser properly handles Unicode file paths and internal references.
- Consider using wide-character APIs or Python’s `Path` objects with proper encoding handling.

Fstarnb:
## Description

`ReadTriangleMeshFromOBJ` used `tinyobj::LoadObj(filename, ...)`, which opens the path with narrow `fopen`/`ifstream`. On Windows that breaks when the path contains non-ASCII characters (e.g. Korean), while ASCII PLY paths work because other loaders use `utility::filesystem::FOpen` / `FReadToBuffer` (UTF-8 → wide open).

This change:
1. Loads the `.obj` bytes through `FReadToBuffer`
2. Parses via `tinyobj::LoadObj` stream overload
3. Resolves `.mtl` files with a small `MaterialReader` that also uses `FReadToBuffer` + `MaterialStreamReader`

## Related Issue

Fixes #7336

## Checklist

- [x] Focused change
- [x] No open competing PR for #7336 at publish time

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ## Main
+-   Fix reading OBJ/MTL meshes from UTF-8 paths on Windows by loading through `utility::filesystem::FReadToBuffer` and tinyobj stream APIs instead of narrow `fopen`/`ifstream` (issue #7336)(PR #7524).
 -   Add SYCL tensor backends for HashMap, nearest-neighbor search (KNN, fixed-radius, hybrid), geometry transforms, and registration / odometry / feature pipelines (parity with CUDA paths where applicable).
 -   Add compressed SPZ file I/O for tensor-based Gaussian splats, with zstd dependency integration, round-trip tests, and notebook samples.
 -   Add Windows shared-library CUDA and SYCL Python wheels (`open3d-cuda`, `open3d-xpu`) built against the installed devel package; ship NVIDIA CUDA 12.6 runtime pip dependencies (`python/requirements_win_cuda.txt`) since CUDA is linked dynamically on Windows
--- a/cpp/open3d/io/file_format/FileOBJ.cpp
+++ b/cpp/open3d/io/file_format/FileOBJ.cpp
@@ -8,7 +8,9 @@
 #include <tiny_obj_loader.h>
 
 #include <fstream>
+#include <map>
 #include <numeric>
+#include <sstream>
 #include <vector>
 
 #include "open3d/io/FileFormatIO.h"
@@ -20,6 +22,51 @@
 
 namespace open3d {
 namespace io {
+namespace {
+
+// tinyobj::MaterialFileReader / LoadObj(filename) open paths with narrow
+// fopen/ifstream, which breaks UTF-8 paths on Windows (issue #7336). Route file
+// IO through utility::filesystem::FReadToBuffer, which uses UTF-8-aware FOpen.
+class UTF8MaterialFileReader : public tinyobj::MaterialReader {
+public:
+    explicit UTF8MaterialFileReader(const std::string& mtl_basedir)
+        : mtl_basedir_(mtl_basedir) {}
+
+    bool operator()(const std::string& matId,
+                    std::vector<tinyobj::material_t>* materials,
+                    std::map<std::string, int>* matMap,
+                    std::string* warn,
+                    std::string* err) override {
+        std::string filepath = matId;
+        if (!mtl_basedir_.empty()) {
+            filepath = mtl_basedir_ + matId;
+        }
+        std::vector<char> buffer;
+        std::string io_err;
+        if (!utility::filesystem::FReadToBuffer(filepath, buffer, &io_err)) {
+            // Fall back to matId alone (tinyobj MaterialFileReader behavior).
+            if (!mtl_basedir_.empty() &&
+                utility::filesystem::FReadToBuffer(matId, buffer, &io_err)) {
+                // ok
+            } else {
+                if (err) {
+                    (*err) += "WARN: Material file [ " + filepath +
+                              " ] not found: " + io_err + "\n";
+                }
+                return false;
+            }
+        }
+        std::string content(buffer.begin(), buffer.end());
+        std::istringstream iss(content);
+        tinyobj::MaterialStreamReader stream_reader(iss);
+        return stream_reader(matId, materials, matMap, warn, err);
+    }
+
+private:
+    std::string mtl_basedir_;
+};
+
+}  // namespace
 
 FileGeometry ReadFileGeometryTypeOBJ(const std::string& path) {
     return FileGeometry(CONTAINS_TRIANGLES | CONTAINS_POINTS);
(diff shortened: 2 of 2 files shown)


ssheorey:
Hi @Fstarnb, thanks for submitting this fix. Have you checked if the latest (main branch) of tinyobjloader fixes this? Its easier to update versions than have special workarounds in Open3D.

Alternately, will adding "/utf-8" to MSVC compile options fix this?

Fstarnb:
@ssheorey Thanks for the review questions — I checked both options carefully.

### 1. Does latest tinyobjloader main already fix this?

**Yes on main, no on the pin Open3D uses.**

Open3D currently vendors **tinyobjloader v2.0.0rc8** (`3rdparty/tinyobjloader/tinyobjloader.cmake`). That release opens paths with a narrow `std::ifstream ifs(filename)` / equivalent — no `MultiByteToWideChar` / wide path handling — so UTF-8 paths on Windows fail exactly as in #7336.

On **tinyobjloader main** (post-rc8), the filename `LoadObj` path was updated to open via `LongPathW(UTF8ToWchar(filename))` (and the same for material file readers). So upstream *has* fixed the classic filename API.

I did **not** bump the third-party pin in this PR because:

- There is still **no newer tagged release** after `v2.0.0rc8` (only that tag).
- Main has since grown a large pure-C11 loader and other surface area; pinning an untagged main commit is a wider dependency risk than this one IO path.
- Open3D already has the right primitive for this: `utility::filesystem::FOpen` / `FReadToBuffer` (UTF-8 → `_wfopen` on Windows), which other mesh loaders already use. Routing OBJ/MTL through that keeps path semantics consistent with the rest of Open3D IO even if tinyobj is upgraded later.

If you prefer a dedicated follow-up to bump tinyobjloader once there is a tagged release (or a pinned commit you trust), happy to do that separately. This PR stays a minimal, pin-safe fix for #7336.

### 2. Would MSVC `/utf-8` fix it?

**No.** `/utf-8` only sets the source/execution charset for the compiler (how string *literals* and source files are decoded). It does **not** change the CRT/MSVC behavior of narrow `fopen` / `std::ifstream` with a runtime path string: on a typical Windows ANSI/ACP code page those still go through the narrow API and break on characters outside that code page (e.g. Korean in the reporter’s path).

Open3D already learned this elsewhere: other loaders go through `FOpen` → `MultiByteToWideChar(CP_UTF8, …)` → `_wfopen`. That is the same mechanism this PR uses, just via `FReadToBuffer` + tinyobj’s **stream** overloads (`LoadObj(istream*)` + `MaterialStreamReader`), which exist on the current v2.0.0rc8 pin.

### 3. CHANGELOG

Also pushed a `CHANGELOG.md` entry for this fix (bot request).

Happy to adjust further if you’d rather see a tinyobj pin bump instead of the stream wrapper — just say which direction you prefer.

Fstarnb:
Follow-up: rebased onto current `main` and resolved a `CHANGELOG.md` conflict with the recent SYCL/SPZ entries so the PR is mergeable again. Code change in `FileOBJ.cpp` is unchanged.

ssheorey:
Hi @Fstarnb we would prefer to switch to the upstream `tinyobjloader` with the utf-8 fix, rather than maintain our own workaround code for this. rc version is OK. We can also wait for a new release with the fix if you prefer.

Fstarnb:
Thanks for the direction. While preparing the dependency-only update, I checked the current main call path and found an important detail: both `cpp/open3d/io/TriangleMeshIO.cpp` and `cpp/open3d/t/io/TriangleMeshIO.cpp` map `.obj` to `ReadTriangleMeshUsingASSIMP`, which calls `Assimp::Importer::ReadFile(filename.c_str(), ...)`. `ReadTriangleMeshFromOBJ` is not used by either public read dispatch.

That means updating tinyobjloader alone would not change the Python `open3d.io.read_triangle_mesh()` path from issue #7336. The two complete options appear to be:

1. Route legacy OBJ reads through `ReadTriangleMeshFromOBJ` (and decide how to handle the tensor API and its current ASSIMP/PBR behavior), or
2. Keep ASSIMP for OBJ and add an Open3D UTF-8-aware ASSIMP `IOSystem`/`IOStream`.

I have not pushed the dependency-only update because it would leave the reported path unchanged. Which direction would you prefer for this PR? I can prepare the focused implementation and regression test accordingly.

Fstarnb:
Additional verification: the bundled Assimp version on current `main` is v6.0.5, whose `DefaultIOSystem::Open` already converts UTF-8 paths to UTF-16 and calls `_wfopen` on Windows. Since both public OBJ read dispatches use `ReadTriangleMeshUsingASSIMP`, the Python path from issue #7336 is already handled by Assimp in this baseline.

This makes a tinyobjloader-only update (and the unused `ReadTriangleMeshFromOBJ` path) redundant for the reported API. Unless you want a separate fix for callers of the direct `ReadTriangleMeshFromOBJ` helper, I recommend closing this PR rather than merging a dependency change that does not affect the public path.

ssheorey:
Thanks for researching this - glad that the issue is resolved. 

Future TODO: Remove tinyobjloader and fully use ASSIMP for loading obj files.

PR #7531, opened 2026-08-07
Add USE_SYSTEM_ZSTD option to use pre-installed zstd

eclipse0922:
## Summary

Add a `USE_SYSTEM_ZSTD` CMake option so Open3D can link against a pre-installed zstd (e.g. from vcpkg) instead of always building the bundled `ext_zstd` ExternalProject.

## Motivation

When building Open3D with `OPEN3D_USE_VCPKG=ON` on Windows (dynamic linkage), the `ext_spz` ExternalProject's `find_package(zstd)` call picks up vcpkg's dynamic zstd, which causes `spz` to compile with `__imp_ZSTD_*` (dllimport) references. At link time, Open3D links `spz` against the bundled *static* `ext_zstd` (which exports only plain `ZSTD_*` symbols), producing:

```
spz.lib(load-spz.cc.obj) : error LNK2019: unresolved external symbol __imp_ZSTD_decompress
spz.lib(load-spz.cc.obj) : error LNK2019: unresolved external symbol __imp_ZSTD_compressBound
...
Open3D.dll : fatal error LNK1120: 7 unresolved externals
```

With `USE_SYSTEM_ZSTD=ON`, the bundled `ext_zstd` is not built at all. Both `spz` and Open3D use the same zstd from `find_package(zstd CONFIG)`, so the symbols are consistent.

## Changes

| File | Change |
|------|--------|
| `CMakeLists.txt` | Add `option(USE_SYSTEM_ZSTD ...)` and include it in the `OPEN3D_USE_VCPKG` defaults |
| `3rdparty/zstd/zstd.cmake` | When `USE_SYSTEM_ZSTD`: use `find_package(zstd CONFIG REQUIRED)` instead of `ExternalProject_Add` |
| `3rdparty/spz/spz.cmake` | When `USE_SYSTEM_ZSTD`: skip `ext_zstd` build dependency and don't pass `-Dzstd_DIR` (spz finds zstd through the inherited toolchain) |
| `3rdparty/find_dependencies.cmake` | When `USE_SYSTEM_ZSTD`: create `3rdparty_zstd` as a simple INTERFACE target linking to `zstd::libzstd`; link `zstd::libzstd` directly in `3rdparty_spz`. When bundled: keep the existing archive-ordering workaround for GNU ld |

## Backward compatibility

`USE_SYSTEM_ZSTD` defaults to **OFF**, so builds without vcpkg are unchanged — zstd is still built as a bundled static ExternalProject with the same link-ordering workaround.


ssheorey (approved):
Thanks @eclipse0922 !

PR #7559, opened 2026-09-10
Implement Transform for OrientedBoundingBox and OrientedBoundingEllipsoid

### LINKED ISSUE #7558: geometry::OrientedBoundingBox::Transform() always throws, even for a rigid transform
### Checklist

- [x] I have searched for [similar issues](https://github.com/isl-org/Open3D/issues). The closest one is #4875, which reports the same limitation as a question and is still open.
- [x] For Python issues, I have tested with the [latest development wheel](https://www.open3d.org/docs/latest/getting_started.html#development-version-pip). That page currently lists no wheel links, so I reproduced with 0.19.0 from PyPI and confirmed the code path is unchanged on `main` at 1a9eb990f9, see the permalink below.

### Describe the issue

`geometry::OrientedBoundingBox::Transform()` always raises, even for a plain rigid transform:

https://github.com/isl-org/Open3D/blob/1a9eb990f9a20936c30c428568c602bdef760744/cpp/open3d/geometry/BoundingVolume.cpp#L154-L160

An oriented bounding box is stored as a center, a rotation matrix and an extent, so it is closed under any similarity transform (rotation, uniform scale, translation): only the center and the rotation move, and the extent scales. That is exactly the case that matters in practice, where a box is carried between coordinate frames next to the point cloud it bounds:

```cpp
cloud.Transform(target_T_source);
box.Transform(target_T_source);  // throws
```

Three things make the current behavior surprising:

1. Every other `Geometry3D` accepts `Transform()`, so a box cannot be moved through generic code that transforms a geometry.
2. The tensor API already implements it, as `Rotate()` followed by `Translate()`, so the legacy and tensor APIs disagree on the same operation:
   https://github.com/isl-org/Open3D/blob/1a9eb990f9a20936c30c428568c602bdef760744/cpp/open3d/t/geometry/BoundingVolume.cpp#L464-L478
3. The suggested workaround ("Call Translate, Scale, and Rotate") is easy to get wrong, because `Rotate(R)` rotates about the box center by default while `Transform()` rotates about the origin. #4875 is exactly that mistake, and the usual fix people land on is to rebuild the box by hand:
   ```cpp
   OrientedBoundingBox transformed(transform * box.center_,
                                   transform.rotation() * box.R_, box.extent_);
   ```

`geometry::OrientedBoundingEllipsoid::Transform()` on `main` has the same body and the same limitation.

A general affine transform genuinely cannot be supported, since shear or non-uniform scale turns the box into a parallelepiped and the ellipsoid into a general quadric. Those cases should keep raising. The similarity case should not.

```python
import numpy as np
import open3d as o3d

box = o3d.geometry.OrientedBoundingBox([1, 2, 3], np.eye(3), [2, 4, 6])

transformation = np.eye(4)
transformation[:3, 3] = [1, 0, 0]  # translate by 1 along x

box.transform(transformation)
```

### Error message

```
  File "<string>", line 5, in <module>
RuntimeError: [Open3D Error] (virtual open3d::geometry::OrientedBoundingBox& open3d::geometry::OrientedBoundingBox::Transform(const Matrix4d&)) /root/Open3D/cpp/open3d/geometry/BoundingVolume.cpp:58: A general transform of an OrientedBoundingBox is not implemented. Call Translate, Scale, and Rotate.
```

### Expected behavior

A similarity transform is applied to the box: `center_` is mapped by the transform, `R_` is premultiplied by its rotation part and `extent_` is multiplied by its uniform scale. The corners of the result are then the transformed corners of the original box, which is what transforming a point cloud of those corners with the same matrix gives.

A transform with shear, non-uniform scale, mirroring or a projective part keeps raising, with a message that names which part is not supported.

```markdown
- Operating system: Ubuntu 24.04 64-bit
- Python version: 3.12.3
- Open3D version: 0.19.0, and the source of `main` at 1a9eb990f9a20936c30c428568c602bdef760744
- System architecture: x86_64
- How did you install Open3D?: pip, and separately built from source for the C++ side
- Compiler version (if built from source): gcc 13.3
```

### Additional information

I have a fix ready and will open a PR that implements `Transform()` for similarity transforms on both `OrientedBoundingBox` and `OrientedBoundingEllipsoid`, reusing the existing `Rotate()`, `Scale()` and `Translate()` primitives, with C++ and Python unit tests for the supported and the rejected cases.

simonschlaepfer:
## Type

-   [x] Bug fix (non-breaking change which fixes an issue): Fixes #7558

## Motivation and Context

`geometry::OrientedBoundingBox::Transform()` and `geometry::OrientedBoundingEllipsoid::Transform()` raise for every input, including a plain rigid transform, so a box cannot be carried between coordinate frames next to the point cloud it bounds and cannot pass through generic code that transforms a `Geometry3D`.

Both are parametrized by a center, a rotation matrix and an extent (or radii), which is closed under any similarity transform: the center is mapped, the rotation is premultiplied and the extent scales. The tensor API already relies on this and implements `Transform()` as `Rotate()` followed by `Translate()`, so the legacy and tensor APIs disagreed on the same operation.

## Plan

- Goal: `Transform()` applies a similarity transform (rotation, uniform scale, translation) to `OrientedBoundingBox` and `OrientedBoundingEllipsoid`, and keeps raising for transforms whose result those types cannot represent.
- Requirements: no public API change, no new dependency, reuse the existing `Rotate()` / `Scale()` / `Translate()` primitives, and raise with a message naming the unsupported part for shear, non-uniform scale, mirroring and projective transforms.
- Implementation: one file-local helper decomposes the 4x4 matrix into a rotation, a uniform scale and a translation, validating that the last row is affine, that the determinant is positive and that the scaled linear part is orthogonal to 1e-5. Each `Transform()` then calls `Rotate()` and `Scale()` about the origin followed by `Translate()`, which composes into the transform and keeps the center moving with the geometry.
- Test method: the corners of the transformed box must equal the corners of the original box transformed by the same matrix, which is what transforming a point cloud of those corners produces. Plus identity, composition of two transforms, and the four rejected transform classes.
- Risks: silently accepting a transform that deforms the box. Mitigated by the orthogonality and determinant checks and by tests for each rejected class. The 1e-5 tolerance matches the one the tensor `Rotate()` already uses, so a rotation accumulated in single precision is still accepted.

A general affine transform is intentionally still rejected: shear or non-uniform scale turns a box into a parallelepiped and an ellipsoid into a general quadric, neither of which these types can express. Mirroring is rejected as well, since it would leave `R_` improper.

## Checklist:

-   [x] I have run `python util/check_style.py --apply` to apply Open3D **code style** to my code.
    -   [x] Both C++ (Doxygen) and Python (Sphinx / Google style) **documentation** is updated accordingly.
    -   [x] I have added or updated C++ and / or Python **unit tests** OR included **test results** (e.g. screenshots or numbers) here.

## Description

`cpp/open3d/geometry/BoundingVolume.cpp`
- Adds the file-local `DecomposeSimilarityTransform()` helper, shared by the box and the ellipsoid.
- Implements both `Transform()` overrides in terms of `Rotate()`, `Scale()` and `Translate()`.

`cpp/open3d/geometry/BoundingVolume.h`
- Documents what both `Transform()` overrides accept and why the rest raises. Neither declaration was documented before.

`cpp/tests/geometry/BoundingVolume.cpp` (new, registered in the tests CMakeLists)
- `OrientedBoundingBox`: `TransformIdentity`, `TransformRigid`, `TransformUniformScale`, `TransformComposes`, and one test per rejected class (`TransformNonUniformScale`, `TransformShear`, `TransformMirror`, `TransformProjective`).
- `OrientedBoundingEllipsoid`: `TransformUniformScale` and `TransformNonUniformScale`.

`python/test/geometry/test_bounding_volume.py` (new)
- The box corners follow the transform the way point coordinates do, a box and a point cloud of its corners transform consistently, the rejected classes raise, and the ellipsoid case.

### Test evidence

Built on Ubuntu 24.04, gcc 13.3, `BUILD_GUI=OFF`, `BUILD_SHARED_LIBS=ON`.

```
$ ./bin/tests --gtest_filter='OrientedBounding*.Transform*'
[==========] Running 11 tests from 3 test suites.
[       OK ] OrientedBoundingBox.TransformIdentity (0 ms)
[       OK ] OrientedBoundingBox.TransformRigid (0 ms)
[       OK ] OrientedBoundingBox.TransformUniformScale (0 ms)
[       OK ] OrientedBoundingBox.TransformComposes (0 ms)
[       OK ] OrientedBoundingBox.TransformNonUniformScale (0 ms)
[       OK ] OrientedBoundingBox.TransformShear (0 ms)
[       OK ] OrientedBoundingBox.TransformMirror (0 ms)
[       OK ] OrientedBoundingBox.TransformProjective (0 ms)
[       OK ] OrientedBoundingEllipsoid.TransformUniformScale (0 ms)
[       OK ] OrientedBoundingEllipsoid.TransformNonUniformScale (0 ms)
[       OK ] OrientedBoundingBox/OrientedBoundingBoxPermuteDevices.Transform/0 (4 ms)
[==========] 11 tests from 3 test suites ran. (4 ms total)
[  PASSED  ] 11 tests.
```

No regressions in the surrounding suites:

```
$ ./bin/tests --gtest_filter='*Bounding*'
[==========] 77 tests from 12 test suites ran. (47 ms total)
[  PASSED  ] 77 tests.
```

```
$ pytest python/test/geometry/test_bounding_volume.py
......                                                                   [100%]
6 passed in 0.12s
```

The values the new tests expect were also cross-checked against the released 0.19.0 wheel, by building the transformed box with the manual reconstruction this PR replaces. It confirms that the corner ordering of `GetBoxPoints()` is preserved by a similarity transform, so comparing corner by corner is valid, and that the result agrees with transforming a point cloud of those corners:

```
rigid                    corner-order preserved: max|err| = 1.776e-15   vs point cloud: 8.882e-16
uniform scale 2x         corner-order preserved: max|err| = 3.553e-15   vs point cloud: 1.776e-15
identity                 corner-order preserved: max|err| = 0.000e+00   vs point cloud: 0.000e+00
```


kvnloo:
Thanks @simonschlaepfer — keeping the box consistent with a point cloud of its corners is a useful invariant here.

One edge worth adding to the projective-rejection test: a small nonzero coefficient in the last row passes the `1e-5` approximate comparison, but is then discarded by the rotation / scale / translation path.

For example, start with an identity matrix, set `T(3, 0) = 1e-6`, and use a box centered at `(1000, 0, 0)` with extent `(2, 2, 2)`. The decomposition accepts the matrix and leaves the box unchanged. Transforming the same corners with homogeneous division instead differs by about `1.001` world units.

I reproduced this in a compiled standalone Eigen 3.4 probe using the relevant arithmetic from `acea77b`; this was not a full Open3D build.

Would it make sense to enforce the affine last-row requirement separately, while retaining the tolerance for the rotation check? A small-coefficient case alongside the existing `0.1` projective test would protect the stated rejection behavior without expanding scope.

AI-assisted source review and standalone numerical validation.

simonschlaepfer:
Good catch @kvnloo. `isApprox` is relative, so against `[0, 0, 0, 1]` the norm is dominated by the `1` and `1e-6` passes at a `1e-5` tolerance, after which the decomposition drops it. The last row is now checked coefficient by coefficient against `[0, 0, 0, 1]` with an absolute tolerance of `1e-12`, while the orthogonality check keeps `1e-5`. That does not reject valid input: an exact zero row survives matrix products, `Matrix4d::inverse()` with `1e6` scale translations, and a `float32` round trip. Added `TransformSmallProjectiveCoefficient` with your case.

kvnloo:
Thanks @simonschlaepfer for the quick follow-up and for adding the regression.

I reviewed `4968fcf` and rechecked the revised arithmetic independently: the original `1e-6` case is now rejected, while the valid-transform controls still pass, including composed/inverted transforms and a float32 round trip. That numerical check was separate from Open3D's own tests, not a full Open3D build.

Keeping the tighter last-row check separate from the rotation tolerance addresses the concrete concern I raised without tightening unrelated behavior. No further changes requested from me on that point, appreciate your collaboration on this!

AI-assisted source review and independent numerical checks.

PR #7573, opened 2026-09-19
Fix USE_SYSTEM_FILAMENT

Tobias-Fischer:
## Type

## Motivation and Context

`USE_SYSTEM_FILAMENT=ON` could not be used as it stood. I hit four separate problems getting Open3D to build against a packaged Filament (conda-forge's), each of which stops the build on its own.

**1. The package and target names don't match anything.** The code looks for a lowercase `filament` package exporting `filament::filament`. Filament installs no CMake package configuration upstream — its build uses `DIST_DIR` and the only `export(TARGETS ...)` is host tools into a build-tree file — so the naming comes from whoever packaged it. conda-forge ships `lib/cmake/Filament/FilamentConfig.cmake` exporting the `Filament::` namespace. On a case-sensitive filesystem `find_package(filament)` won't even find it.

Both spellings are now accepted rather than one replaced by the other, so any setup relying on the current names keeps working. The namespace is chosen from the targets that actually got defined rather than from which `find_package` call answered: on a case-insensitive filesystem `find_package(Filament)` can succeed against a lowercase config directory, so the two are not reliably correlated.

**2. No ktxreader target is requested,** although `FilamentResourceManager` constructs `image::Ktx1Bundle` and calls `ktxreader::Ktx1Reader::createTexture()` when loading IBL and skybox textures (`FilamentResourceManager.cpp:36,666,669,735,738`).

**3. `FILAMENT_MATC` is hardcoded to `/usr/bin/matc`** rather than using the matc from the package that was just located.

**4. macOS frameworks are never linked.** This is the interesting one, and it's independent of packaging. Open3D's own Objective-C++ sources — `NativeMacOS.mm`, `MenuMacOS.mm`, `ComputeGPUMetal.mm`, `GaussianSplatMetalBackend.mm` — reference AppKit, Metal and QuartzCore directly, but the `if(APPLE)` block that links them lives inside the `if(NOT USE_SYSTEM_FILAMENT)` branch. Open3D was effectively free-riding on the bundled-Filament target's link interface, so with a system Filament `libOpen3D` fails to link:

```
Undefined symbols for architecture arm64:
  "_NSApp", referenced from: ... in NativeMacOS.mm.o
  "_OBJC_CLASS_$_CAMetalLayer", referenced from: objc-class-ref in NativeMacOS.mm.o
  "_OBJC_CLASS_$_MTLTextureDescriptor", referenced from: objc-class-ref in ComputeGPUMetal.mm.o
  "_OBJC_CLASS_$_NSOpenPanel", ... "_OBJC_CLASS_$_NSScreen", ...
```

## Checklist:

-   [x] I have run `python util/check_style.py --apply` to apply Open3D **code style** to my code. — n/a, CMake only; no C++ or Python changed.
    -   [x] I have added or updated C++ and / or Python **unit tests** OR included **test results** here. — test results below.

## Description

Only the `USE_SYSTEM_FILAMENT` branch is touched; the bundled-Filament path is unchanged.

### Test results

Built Open3D 0.20.0 on osx-arm64 against conda-forge's filament 1.77.0 with `USE_SYSTEM_FILAMENT=ON`. `libOpen3D.dylib` links the packaged libraries:

```
$ otool -L lib/libOpen3D.0.20.0.dylib | grep -E 'filament|ktxreader|image|backend|utils|geometry'
	@rpath/libfilament.dylib
	@rpath/libbackend.dylib
	@rpath/libgeometry.dylib
	@rpath/libutils.dylib
	@rpath/libktxreader.dylib
	@rpath/libimage.dylib
```

and the packaging tests pass, including importing `open3d.visualization`:

```
✔ python imports test passed!
✔ pip check passed!
✔ all tests passed!
```

Note that `ktxreader`/`image` had to be added to the conda-forge package first (conda-forge/filament-feedstock#15); they were not previously shipped, which is what made this worth chasing.

### Points for reviewers

**I kept the change strictly additive.** The five `find_library` calls are duplicated into the system branch rather than hoisting the existing `if(APPLE)` block to cover both paths. Hoisting would be cleaner and remove the duplication, but it modifies the bundled-Filament path, which currently works and which I have not re-tested. Happy to do it that way if you prefer.

**Scope of testing:** osx-arm64 only, and only against conda-forge's packaging — I have no other packaged Filament to test against. I have not verified the bundled path still builds, since I did not modify it. Items 1–3 are packaging-dependent; item 4 is a genuine bug on any macOS system-Filament build.

**Version skew:** conda-forge ships Filament 1.77.0 while Open3D 0.20.0 targets 1.76.0. It compiled and linked cleanly, but I have not exercised rendering at runtime, so I can't speak to behavioural differences between those releases.

---
🤖 This pull request was written by [Claude Code](https://claude.com/claude-code) (Claude Opus 5), acting on behalf of @Tobias-Fischer, who has reviewed it.

https://claude.ai/code/session_017R4DS6whKCY8xGcA61kni2


Tobias-Fischer:
Related: #7574 fixes two fmt call sites that break when Open3D is compiled as C++20.

That matters here because conda-forge's Filament config declares `INTERFACE_COMPILE_FEATURES cxx_std_20`, and since compile features propagate as a floor, linking these targets raises every Open3D translation unit to C++20 — so in practice a system-Filament build on that packaging needs both PRs. They are independent changes though, and can be reviewed in either order.

---
🤖 Comment written by [Claude Code](https://claude.com/claude-code) (Claude Opus 5), acting on behalf of @Tobias-Fischer.

Tobias-Fischer:
Updated: both the existing and the packaged spelling are now accepted, rather than replacing one with the other. Any setup relying on the current names keeps working.

The package name is probed in turn and whichever is installed is used. For Filament the namespace is then chosen from the targets that actually got defined, not from which `find_package` call answered — on a case-insensitive filesystem `find_package(Filament)` can succeed against a lowercase config directory, so the two aren't reliably correlated.

Rebuilt Open3D 0.20.0 on osx-arm64 against conda-forge's filament with the updated logic: still resolves the package and passes the packaging tests.

---
🤖 Comment written by [Claude Code](https://claude.com/claude-code) (Claude Opus 5), acting on behalf of @Tobias-Fischer.

PR #7578, opened 2026-09-19
Use the imgui package name that packagings actually export

Tobias-Fischer:
## Type

## Motivation and Context

`USE_SYSTEM_IMGUI` looks for a package called `ImGui` exporting `ImGui::ImGui`. Dear ImGui ships no CMake package configuration of its own, so that naming comes entirely from whoever packaged it — and the capitalised spelling does not appear to match any packaging in common use:

- **conda-forge** exports `imgui::imgui` (its recipe aliases `${PROJECT_NAME}::${PROJECT_NAME}` with `PROJECT_NAME` lowercase and installs the export set with `NAMESPACE imgui::`)
- **vcpkg** likewise names the port `imgui` and exports `imgui::imgui`

There is a second, quieter consequence: `find_package(ImGui)` cannot find a config directory named `imgui` on a case-sensitive filesystem, so `USE_SYSTEM_IMGUI=ON` silently falls back to the bundled copy on Linux even when a system Dear ImGui is installed. The option appears to work while having no effect.

## Checklist:

-   [x] I have run `python util/check_style.py --apply` to apply Open3D **code style** to my code. — n/a, CMake only.
    -   [x] I have added or updated C++ and / or Python **unit tests** OR included **test results** here. — test results below.

## Description

Both spellings are accepted rather than one replaced by the other, so setups relying on the current names keep working. The package name is probed (`imgui`, then `ImGui`) and the target is resolved with the helper's existing `TARGET_ALTERNATIVES`, which is built for exactly this.

Only the `USE_SYSTEM_IMGUI` path is affected, and it retains its existing fallback: if neither spelling is found, `USE_SYSTEM_IMGUI` is still set `OFF` and the bundled copy is used, so no configuration that builds today changes behaviour.

### Test results

Carried as a patch in the conda-forge [open3d-feedstock](https://github.com/conda-forge/open3d-feedstock) since the recipe was created. Note that the feedstock currently builds with `USE_SYSTEM_IMGUI=OFF` for an unrelated reason (an `ImGuiIO::KeyMap` API mismatch), so this specific change is verified as "finds the package and configures", not as a full build against a system Dear ImGui.

### Credit

The patch has been carried in the conda-forge recipe since it was created, by its maintainers (@goanpeca, @Tobias-Fischer, @traversaro).

---
🤖 This pull request was written by [Claude Code](https://claude.com/claude-code) (Claude Opus 5), acting on behalf of @Tobias-Fischer, who has reviewed it.

https://claude.ai/code/session_017R4DS6whKCY8xGcA61kni2


Tobias-Fischer:
Updated: both the existing and the packaged spelling are now accepted, rather than replacing one with the other. Any setup relying on the current names keeps working.

The package name is probed in turn and whichever is installed is used. For Filament the namespace is then chosen from the targets that actually got defined, not from which `find_package` call answered — on a case-insensitive filesystem `find_package(Filament)` can succeed against a lowercase config directory, so the two aren't reliably correlated.

Rebuilt Open3D 0.20.0 on osx-arm64 against conda-forge's filament with the updated logic: still resolves the package and passes the packaging tests.

---
🤖 Comment written by [Claude Code](https://claude.com/claude-code) (Claude Opus 5), acting on behalf of @Tobias-Fischer.

